PR #16, opened 2021-02-19
DEM differencing, coreg.py changes and added Examples

eriksholmlund:
Added DEM differencing function.

Changed the default coregistration method to "nuth_kaab" and improved code.

Added Examples section in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,3 +30,38 @@ In the interest of keeping the documentation simple, please write all docstring 
 ### Testing - again please read!
 These tools are only valuable if we can rely on them to perform exactly as we expect. So, we need testing. Please create tests for every function that you make, as much as you are able. Guidance/examples here for the moment: https://github.com/GeoUtils/georaster/blob/master/test/test_georaster.py
 https://github.com/corteva/rioxarray/blob/master/test/integration/test_integration__io.py
+
+
+
+### Examples
+
+**Coregister a DEM to another DEM**
+```python
+import GeoUtils as gu
+
+reference_dem = "path/to/reference.tif"
+dem_to_be_aligned = "path/to/dem.tif"
+mask = "path/to/mask.shp"  # This is optional. Could for example be glacier outlines.
+
+aligned_dem, error = gu.coreg.coregister(reference_dem, dem_to_be_aligned, mask=mask)
+
+aligned_dem.save("path/to/coreg.tif")
+```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -30,3 +30,38 @@ In the interest of keeping the documentation simple, please write all docstring
...
+
+**Coregister a DEM to another DEM**
+```python
+import GeoUtils as gu
+
+reference_dem = "path/to/reference.tif"
+dem_to_be_aligned = "path/to/dem.tif"
+mask = "path/to/mask.shp"  # This is optional. Could for example be glacier outlines.
+
+aligned_dem, error = gu.coreg.coregister(reference_dem, dem_to_be_aligned, mask=mask)
```
adehecq, line 46: If the argument that is expected for mask is a shapefile, should the variable be named shp or shpfile instead? I would use 'mask' for a raster type. This could also be an option to take a raster mask as input.

README.md
```diff
@@ -30,3 +30,38 @@ In the interest of keeping the documentation simple, please write all docstring
...
+
+aligned_dem.save("path/to/coreg.tif")
+```
+The default coregistration method is a [Nuth and Kääb (2011)](https://doi.org/10.5194/tc-5-271-2011) implementation, but this can be changed with the keyword argument `method=...`, e.g. to `"icp"`.
+The currently supported methods are: `"nuth_kaab"`, `"icp"` and `"deramp"`.
```
adehecq, line 51: Is deramp seperated from nuth_kaab? It's not really a coregistration algorithm as such. I would rather put it as an option to nuth_kaab, no?

eriksholmlund: deramp is part of nuth_kaab but can also be used separately! Once we have proper documentation,  I can describe each method in depth

eriksholmlund: And I would still call it a coregistration method, as one is using a reference DEM to correct another

eriksholmlund: Also, if `deramp_degree=None` is supplied to the `coregister()` function, this will disable deramping.

adehecq: Alright, always good to have it as an additional option. Good to know it's also an option in the nuth_kaab algorithm!

README.md
eriksholmlund: Good suggestion! I'll make it accept either a raster or a vector. I'll make another pull request once I'm finished.

eriksholmlund: @adehecq, this inclusion is now in PR #16

adehecq (approved):
I did not test it, but looks reasonable to me!

PR #17, opened 2021-02-19
Coreg mask can now be Raster or Vector

eriksholmlund:
... or a path to either of the two.

Tests are included to verify that it works properly.


adehecq (approved):
Looks great!

PR #21, opened 2021-03-02
DEM class with vertical referencing + changing package name to xdem

rhugonnet:
Renamed DemUtils into xdem; fixed imports and setup

Added dem.py with DEM class (subclass of SatImg: necessary for parsing of metadata)
Added method to parse vertical referencing from DEM product name
Added class method **set_vref** to set a compound CRS (horizontal + vertical) based on name or any vertical grids available in PROJ-DATA (https://github.com/OSGeo/PROJ-data); added proj-data dependency through conda-forge in environment file; by default the **set_vref** method does not try to compute the CCRS (in case all DEMs are the same vertical referencing; or the user doesn't care to avoid using pyproj-data which represents ~500Mo)
Added class method **to_vref** to transform between vertical references and update DEM data.

Tests are quite basic for now and independent from the Class. I tried several grids (e.g., Alaska GEOID2006 or Iceland geoids) and all is working fine for test points. We need to change the structure of datasets to match that of geoutils.


erikmannerfelt:
Looks nice and I support the name change!

How come a DEM class is needed in xdem? Should the package not use geoutils Raster classes? Or at least have inheritance from geoutils Rasters?

xdem/dem.py
```diff
@@ -0,0 +1,163 @@
...
+    elif product in ['COPDEM']:
+        vref = 'EGM08'
+    else:
+        vref = None
+
+    return vref
+
+class DEM(SatelliteImage):
+
+    def __init__(self, filename, vref=None, vref_grid=None, ccrs = None, read_vref_from_prod=True, silent=False, **kwargs):
```
adehecq, line 35: Need to add a docstring here

rhugonnet:
Thanks ! :)

The problem with Raster classes is that they don't need/have a vertical referencing. Most of the time, that info is not reported in the GTiffs. DEMs have and need one. 
Right now, DEM is a subclass of SatImg which is a subclass of Raster. The reasoning:
A SatImg is a raster with specific metadata associated to it (satellite, sensor, tile, date of acq, etc..). The methods there are quite useful to get the spatial/temporal dimension solely from the filename (tile name, date). For the time dimension, it is usually the only way to get it. For the spatial dimension, it can be useful if the raster metadata can't be read directly (e.g. you have 5 TB of zipped rasters, don't want to unzip everything).
A DEM is a specific kind of Raster, but also of SatImg in the sense that it also has this metadata (satellite, sensor, tile, date, etc...). It needs its own subclass because it also has the specificity of the vertical dimension, for which we need all associated methods on the vertical referencing.

xdem/dem.py
```diff
@@ -0,0 +1,163 @@
...
+
+        super().__init__(filename, **kwargs)
+
+        if self.nbands > 1:
+            raise ValueError('DEM rasters should be composed of only one band only')
```
adehecq, line 40: repetition of "only"

xdem/dem.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # priority to user input
+        self.vref = vref
```
adehecq, line 43: Like the Raster._data attribute, this should be set to a hidden attribute self._vref with a getter/setter method to prevent users from setting it to anything. See geoutils.Raster for an example.

xdem/dem.py
```diff
@@ -0,0 +1,163 @@
...
+    # COPERNICUS DEM: https://spacedata.copernicus.eu/web/cscda/dataset-details?articleId=394198
+
+    if product in ['ArcticDEM/REMA','TDM1','NASADEM-HGTS']:
+        vref = 'WGS84'
+    elif product in ['AW3D30','SRTMv4.1','SRTMGL1','ASTGTM2','NASADEM-HGT']:
+        vref = 'EGM96'
+    elif product in ['COPDEM']:
+        vref = 'EGM08'
+    else:
+        vref = None
```
adehecq, line 29: Maybe raise a warning that this is not know?

xdem/dem.py
```diff
@@ -0,0 +1,163 @@
...
+            self.__parse_vref_from_fn(silent=silent)
+
+
+    def __parse_vref_from_fn(self,silent=False):
+
+        """
+        Attempts to pull vertical reference from product name identified by SatImg
+        """
+
+        if self.product is not None:
```
adehecq, line 58: self.product is never set anywhere.

xdem/dem.py
```diff
@@ -0,0 +1,163 @@
...
+                self.vref = 'EGM96'
+            else:
+                raise ValueError(
+                    'Vertical reference name must be either "WGS84", "EGM96" or "EGM08". Otherwise, provide only'
+                    'a geoid grid from PROJ DATA: https://github.com/OSGeo/PROJ-data')
+        elif not isinstance(vref_grid, str):
+            raise ValueError('Vertical reference grid name must be PROJ DATA file name, '
+                             'such as: "us_noaa_geoid06_ak.tif" for the Alaska GEOID2006')
+        else:
+            self.vref = 'Unknown vertical reference name'
```
adehecq, line 108: Maybe in that case, if both arguments are given, I would use grid only, and set self.vref to whatever is provided by the user?
Unless it doesn't make sense for how vref is used later...

adehecq: Actually, here self.vref should be set to None, otherwise self.to_vref for example will fail.

xdem/spatial_tools.py
```diff
@@ -66,7 +66,7 @@ def subtract_rasters(first_raster: Union[str, gu.georaster.Raster], second_raste
             )
 
     # Reproject the non-reference and subtract the two rasters.
-    difference = first_raster.data - second_raster.reproject(second_raster, resampling=resampling_method).data if \
+    difference = first_raster.data - second_raster.reproject(first_raster, resampling=resampling_method).data if \
```
adehecq, line 69: since this is fixing a bug, ideally this should be a different PR

erikmannerfelt:
> Thanks ! :)
> 
> The problem with Raster classes is that they don't need/have a vertical referencing. Most of the time, that info is not reported in the GTiffs. DEMs have and need one.
> Right now, DEM is a subclass of SatImg which is a subclass of Raster. The reasoning:
> A SatImg is a raster with specific metadata associated to it (satellite, sensor, tile, date of acq, etc..). The methods there are quite useful to get the spatial/temporal dimension solely from the filename (tile name, date). For the time dimension, it is usually the only way to get it. For the spatial dimension, it can be useful if the raster metadata can't be read directly (e.g. you have 5 TB of zipped rasters, don't want to unzip everything).
> A DEM is a specific kind of Raster, but also of SatImg in the sense that it also has this metadata (satellite, sensor, tile, date, etc...). It needs its own subclass because it also has the specificity of the vertical dimension, for which we need all associated methods on the vertical referencing.

I didn't read where the SatelliteImage class was imported from... Sounds good with your additions!

Now perhaps the `coreg.py` functions should accept a `DEM` instance, or perhaps it automatically will since it's a `Raster` subclass..

xdem/spatial_tools.py
erikmannerfelt: This is a merge with a small correction that I did earlier.

adehecq:
As you can see, I made a bunch of comments. Everything that you feel is too long to handle now, just ignore and add an issue for it.

xdem/dem.py
rhugonnet: it is set in the SatImg class, so through:
`        super().__init__(filename, attrs=attrs, load_data=load_data, bands=bands, as_memfile=as_memfile)
`

xdem/spatial_tools.py
rhugonnet: ;)

rhugonnet:
> As you can see, I made a bunch of comments. Everything that you feel is too long to handle now, just ignore and add an issue for it.

Thanks! I will add the docstring and do proper setter/getter like in geoutils on this PR. I think the rest mostly comes down to default behaviour which we can adjust later if desired.

By the way, do either of you know how to fix a merge conflict for a deleted file? (I pulled Erik's change into the commit, but because the file has changed location, the merging issue is still unresolved)

adehecq:
Did you try to fix the conflict on your machine or online? (I wrote you on Slack)

tests/test_spatial_tools.py
```diff
@@ -7,8 +7,8 @@
 import numpy as np
 import pytest
 
-import DemUtils as du
-from test_coreg import EXAMPLE_PATHS
+import xdem as du
+# from test_coreg import EXAMPLE_PATHS
```
erikmannerfelt, line 11: The example paths need to be imported. Otherwise this test won't work.

I guess a better example approach should be made. I think @adehecq mentioned that GeoPandas has a good structure?

erikmannerfelt (approved):
As mentioned, one coreg test now doesn't work. Since the dem tests are not finished, I suggest a new PR after  this to make all tests work again.

xdem/dem.py
rhugonnet: In the end I didn't add it because I don't think the usage of vref is really comparable to data. It would not be used much more than an attribute such as nbands or crs in Raster for instance, for which there is no need for a getter/setter method

xdem/dem.py
rhugonnet: this is raised in the class __init__ function: __parse_vref_from_fn(self,silent=False)

xdem/dem.py
rhugonnet: Changed default behaviour: grid has priority; also added a check that the grid exists (based on pyproj.datadir)

PR #25, opened 2021-03-03
Added example data fetching and removed local data

erikmannerfelt:
A new script (`xdem/examples.py`) contains functions to fetch the Longyearbyen datasets.

I don't claim that the script should have that name and be there. Additionally, absolute paths have not yet been implemented. Therefore, I'm trying the draft pull request mode.

Where should the data be fetched to? Right now, it's just to `examples/Longyearbyen/data` relative to the current directory. Is this good enough for testing purposes?


xdem/examples.py
```diff
@@ -0,0 +1,78 @@
...
+        "east": 529450,
+        "south": 8654330,
+        "north": 8674030
+    }
+
+    async def crop_dem(input_path, output_path, bounds):
+        dem = rio.open(input_path)
+
+        upper, left = dem.index(bounds["west"], bounds["north"])
+        lower, right = dem.index(bounds["east"], bounds["south"])
```
rhugonnet, line 36: would this somehow work directly with the Raster.crop() method (and thus DEM.crop() method)? 
maybe it's too specific with the window reading

xdem/examples.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+
+def load_longyearbyen_examples(exist_ok: bool = True):
+    """
+    Fetch the Longyearbyen example files.
+
+    :param exist_ok: Do not download the files again if they already exist.
+    """
+    if exist_ok and os.path.isfile("examples/Longyearbyen/data/DEM_2009_ref.tif"):
```
rhugonnet, line 72: here you could loop through all files just in case (minor); by the way what is the difference between os.path.exists and os.path.isfile?

xdem/examples.py
```diff
@@ -0,0 +1,78 @@
...
+        print("Datasets exist")
+        return
+    print("Downloading datasets from Longyearbyen")
+    os.makedirs("examples/Longyearbyen/data/glacier_mask", exist_ok=True)
+    loop = asyncio.get_event_loop()
```
rhugonnet, line 77: I'm not familiar with asyncio that much, but it really looks nice

rhugonnet (approved):

xdem/examples.py
erikmannerfelt: Thanks! It's just for fetching the data concurrently. Not that it goes that fast anyway.. Takes ages to run it still. I haven't checked if it's actually faster than doing each at a time!

xdem/examples.py
erikmannerfelt: `os.path.isfile` is only true if the path points toward a file (and not a directory).

I definitely should go though each file. I just didn't want to repeat it once again without first clearing how paths should be handled. Maybe just in an uppercase dictionary in the start of the file?

xdem/examples.py
erikmannerfelt: I wanted to avoid loading the entire raster into memory. Would `Raster.crop()` read a window appropriately?

adehecq: I wanted to test this actually. I can have a look.

adehecq:
I'm also not very familar with asyncio so it is difficult to read the code... You can always merge when ready and I'll test it.

erikmannerfelt:
I've now changed the example paths to be absolute paths within the xdem directory. Tests have been changed accordingly.

PR #26, opened 2021-03-03
Docstring for DEM class + Default priority to grid + Checking proj-data files on disk

rhugonnet:
End of commits from the previous PR (Amaury's comments)

Added docstring for all functions in DEM class
Changed set_vref() priority to vref_grid argument
Checking for existence/correct naming of proj-data grid files through pyproj.datapath()

I was a bit busy for the early afternoon to start back on the coreg, will have more time tonight :) !

### DIFF
--- a/xdem/dem.py
+++ b/xdem/dem.py
@@ -1,6 +1,7 @@
 """
 dem.py provides a class for working with digital elevation models (DEMs)
 """
+import os
 import pyproj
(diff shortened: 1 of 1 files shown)


rhugonnet:
Merging because it's stuff you've already seen in the previous PR (except the docs, but that's pretty straightforward)

PR #27, opened 2021-03-03
Test for DEM class

rhugonnet:
Fixed elif loop typo in set_vref(), pointing to CRS in to_vref()
Created better tests for DEM class and checked those worked properly: load, set_vref, to_vref
Added xdem/version.py file to gitignore

**Still ongoing: DEM.to_vref() test fails when CRS cannot be read by Pyproj**
Using the DEMs from Erik, I can't seem to be able to transfer the CRS read by rasterio into a Pyproj CRS:
```python
img.crs.to_wkt()
Out[10]: 'LOCAL_CS["ETRS89 / UTM zone 33N",GEOGCS["ETRS89",DATUM["unknown",SPHEROID["unretrievable - using WGS84",6378137,298.257223563]],PRIMEM["Greenwich",0],UNIT["degree",0.0174532925199433]],AUTHORITY["EPSG","25833"],UNIT["metre",1,AUTHORITY["EPSG","9001"]]]'
pyproj.CRS.from_wkt(img.crs.to_wkt())
Out[11]: Traceback (most recent call last):
  File "/home/atom/miniconda3/envs/pyddem/lib/python3.8/site-packages/IPython/core/formatters.py", line 224, in catch_format_error
    r = method(self, *args, **kwargs)
  File "/home/atom/miniconda3/envs/pyddem/lib/python3.8/site-packages/IPython/core/formatters.py", line 702, in __call__
    printer.pretty(obj)
  File "/home/atom/miniconda3/envs/pyddem/lib/python3.8/site-packages/IPython/lib/pretty.py", line 394, in pretty
    return _repr_pprint(obj, self, cycle)
  File "/home/atom/miniconda3/envs/pyddem/lib/python3.8/site-packages/IPython/lib/pretty.py", line 700, in _repr_pprint
    output = repr(obj)
  File "/home/atom/miniconda3/envs/pyddem/lib/python3.8/site-packages/pyproj/crs/crs.py", line 983, in __repr__
    datum=self.datum,
  File "pyproj/_crs.pyx", line 2464, in pyproj._crs._CRS.datum.__get__
  File "pyproj/_crs.pyx", line 1331, in pyproj._crs.Datum.create
KeyError: 0
```
This is necessary for creating the compound CRS for vertical reprojection...
Any ideas?


rhugonnet:
Oh and the other tests of test_dem.py are only passing with the fix to SatelliteImage which is PRed to GeoUtils

adehecq:
Is the CRS issue similar to this? https://github.com/GlacioHack/GeoUtils/issues/61
Erik was only able to fix this by updating proj... Or you can put a quick fix by changing the EPSG like Erik did initially.

tests/test_dem.py
```diff
@@ -18,39 +23,75 @@
...
             warnings.filterwarnings("ignore", module="pyproj")
             #init is deprecated by
             ellipsoid=pyproj.Proj(init="EPSG:4326") #WGS84 datum ellipsoid height
             geoid=pyproj.Proj(init="EPSG:4326", geoidgrids='egm96_15.gtx') #EGM96 geoid in Chile, we expect ~30 m difference
-
         transformer = pyproj.Transformer.from_proj(ellipsoid, geoid)
-        print(transformer.transform(lng,lat,z))
+        z_out = transformer.transform(lng,lat,z)[2]
+
+        assert np.logical_and.reduce((np.isfinite(z_out),np.greater(z_out,z),np.less(np.abs(z_out-z),40)))
```
adehecq, line 82: what's the meaning of this? #CommentsAreGreat ;-)

rhugonnet: Added :D

rhugonnet:
> Is the CRS issue similar to this? [GlacioHack/GeoUtils#61](https://github.com/GlacioHack/GeoUtils/issues/61)
> Erik was only able to fix this by updating proj... Or you can put a quick fix by changing the EPSG like Erik did initially.

Looks similar! I haven't been able to update to `proj 7.2.2` through, environment fails

erikmannerfelt:
A workaround for the time being could be:

```python
import json
import subprocess

import pyproj


def get_crs(filepath: str) -> pyproj.CRS:
    """Get the CRS of a raster with the given filepath."""
    info = subprocess.run(
        ["gdalinfo", "-json", filepath],
        stdout=subprocess.PIPE,
        check=True,
        encoding="utf-8"
    ).stdout

    wkt_string = json.loads(info)["coordinateSystem"]["wkt"]

    return pyproj.CRS.from_wkt(wkt_string)

```
Does this work for you, @rhugonnet ? It works for me, but I have an updated proj installation so maybe it won't work for you still.

tests/test_dem.py
adehecq: So why don't you test that it is exactly the value that you currently get?

tests/test_dem.py
```diff
@@ -5,8 +5,13 @@
...
+import inspect
+import xdem
+import numpy as np
 from xdem.dem import DEM
 
+path_module = os.path.dirname(os.path.dirname(os.path.abspath(inspect.getsourcefile(xdem))))
+
 #TODO: move datasets to a "dataset" folder like in geoutils
 EXAMPLE_PATHS = {
     "dem1": "examples/Longyearbyen/data/DEM_2009_ref.tif",
```
erikmannerfelt, line 17: Could this be adapted to use the new path system?
```python
xdem.examples.download_longyearbyen_data(overwrite=False)

xdem.exampled.FILEPATHS["longyearbyen_ref_dem"]  # This is the same as EXAMPLE_PATHS["dem1"] before.
```

rhugonnet: > A workaround for the time being could be:
> 
> ```python
> import json
> import subprocess
> 
> import pyproj
> 
> 
> def get_crs(filepath: str) -> pyproj.CRS:
>     """Get the CRS of a raster with the given filepath."""
>     info = subprocess.run(
>         ["gdalinfo", "-json", filepath],
>         stdout=subprocess.PIPE,
>         check=True,
>         encoding="utf-8"
>     ).stdout
> 
>     wkt_string = json.loads(info)["coordinateSystem"]["wkt"]
> 
>     return pyproj.CRS.from_wkt(wkt_string)
> ```
> 
> Does this work for you, @rhugonnet ? It works for me, but I have an updated proj installation so maybe it won't work for you still.

This worked! I've implemented as temporary fix.
Didn't expect PROJ/pyproj to not support some projections at such advanced versions (7.1)... will open an issue. Maybe we can simply force proj>7.2 in the requirements at a later stage.

xdem/dem.py
```diff
@@ -105,6 +107,20 @@ def set_vref(self,vref_name=None,vref_grid=None,compute_ccrs=False):
         :return:
         """
 
+        # temporary fix for some CRS with proj < 7.2
```
erikmannerfelt, line 110: You can check the proj version using:
```python
if pyproj.proj_version_str >= "7.2.0":
    # do stuff
```

rhugonnet: just pushed the change!

tests/test_dem.py
rhugonnet: changed all paths from test_dem.py

erikmannerfelt (approved):

tests/test_dem.py
```diff
@@ -10,14 +10,8 @@
...
-path_module = os.path.dirname(os.path.dirname(os.path.abspath(inspect.getsourcefile(xdem))))
-
-#TODO: move datasets to a "dataset" folder like in geoutils
-EXAMPLE_PATHS = {
-    "dem1": "examples/Longyearbyen/data/DEM_2009_ref.tif",
-    "dem2": "examples/Longyearbyen/data/DEM_1995.tif",
-    "glacier_mask": "examples/Longyearbyen/data/glacier_mask/CryoClim_GAO_SJ_1990.shp"
-}
+xdem.examples.download_longyearbyen_data(overwrite=False)
+xdem.examples.FILEPATHS["longyearbyen_ref_dem"]  # This is the same as EXAMPLE_PATHS["dem1"] before
```
erikmannerfelt, line 14: This line is not necessary! I just wanted to show how one would use it!

rhugonnet: Ups! went a bit fast :D

xdem/dem.py
adehecq: I didn't know it was possibly to just compare the strings! I thought the comparison would be based on the number of characters.

PR #34, opened 2021-03-09
Move examples to one place: a separate examples directory + ICP coreg improvements

erikmannerfelt:
We had two options for where to put the examples directory:
`xdem/examples/`
or
`examples/`
currently, both are used!

This PR makes only the latter be used.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,6 +6,7 @@ dependencies:
   - ipython
   - matplotlib
   - pdal
+  - opencv
   - python-pdal
   - pyproj
   - rasterio
--- a/xdem/coreg.py
+++ b/xdem/coreg.py
@@ -34,6 +34,12 @@ import scipy.optimize
 from rasterio import Affine
 from tqdm import trange
 
+try:
+    import cv2
+    _has_cv2 = True
+except ImportError:
+    _has_cv2 = False
+
 
 def filter_by_range(ds: rio.DatasetReader, rangelim: tuple[float, float]):
     """
@@ -239,9 +245,9 @@ def check_for_pdal(min_version="2.2.0") -> None:
         raise AssertionError(f"Installed PDAL has version: {version}, min required version is {min_version}")
 
 
-def icp_coregistration(reference_dem: np.ndarray, dem_to_be_aligned: np.ndarray, bounds: rio.coords.BoundingBox,
-                       mask: Optional[np.ndarray] = None, max_assumed_offset: Optional[float] = None,
-                       verbose=False, metadata: Optional[dict[str, Any]] = None, **_) -> tuple[np.ndarray, float]:
+def icp_coregistration_pdal(reference_dem: np.ndarray, dem_to_be_aligned: np.ndarray, bounds: rio.coords.BoundingBox,
+                            mask: Optional[np.ndarray] = None, max_assumed_offset: Optional[float] = None,
+                            verbose=False, metadata: Optional[dict[str, Any]] = None, **_) -> tuple[np.ndarray, float]:
     """
     Perform an ICP coregistration in areas where two DEMs overlap.
 
@@ -260,6 +266,10 @@ def icp_coregistration(reference_dem: np.ndarray, dem_to_be_aligned: np.ndarray,
     dem_to_be_aligned_unmasked = dem_to_be_aligned.copy()
     # Check that PDAL is installed.
     check_for_pdal()
+
+    nodata_value = -9999
+    reference_dem[reference_dem <= nodata_value] = np.nan
+    dem_to_be_aligned[reference_dem <= nodata_value] = np.nan
 
     resolution = (bounds.right - bounds.left) / reference_dem.shape[1]
     # Make sure that the DEMs have the same shape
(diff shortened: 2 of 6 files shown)


xdem/examples.py
```diff
@@ -5,7 +5,7 @@
 import geopandas as gpd
 import rasterio as rio
 
-EXAMPLES_DIRECTORY = os.path.abspath(os.path.join(os.path.dirname(__file__), "examples/"))
+EXAMPLES_DIRECTORY = os.path.abspath(os.path.join(os.path.dirname(__file__), "../", "examples/"))
```
adehecq, line 8: Ok, so the unique one will be directly in the base directory?
Sounds good to me.

erikmannerfelt: @adehecq Yep!

erikmannerfelt:
Crap, I accidentally put another commit in this PR. Does anyone know how to remove it??

environment.yml
```diff
@@ -6,6 +6,7 @@ dependencies:
   - ipython
   - matplotlib
   - pdal
+  - opencv
```
adehecq, line 9: This makes me think. The list of dependencies is starting to grow a bit. Sometimes it can be difficult to resolve conflicts (I've had a lot of issues of conflicts between gdal and opencv, the latter only compatible with old versions of gdal).
Maybe we should make some of these dependencies soft and add a test at the import. If opencv is not available, set opencv_avail=False. Then in any of the function where needed, if this is False, simply raise an error.

erikmannerfelt: @adehecq  I have done exactly that. But as I mentioned, this is in the wrong PR. Sorry!!

adehecq: Never mind, I just saw it's what you did! Then opencv should be listed as soft dependency (maybe same for pdal?).

xdem/coreg.py
```diff
@@ -261,6 +267,10 @@ def icp_coregistration(reference_dem: np.ndarray, dem_to_be_aligned: np.ndarray,
     # Check that PDAL is installed.
     check_for_pdal()
 
+    nodata_value = -9999
+    reference_dem[reference_dem <= nodata_value] = np.nan
```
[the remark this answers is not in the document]
adehecq: This makes me think. I don't think any of the sample data you have so far has no data? This would be good to include for testing these cases.

xdem/coreg.py
```diff
@@ -280,26 +290,35 @@ def icp_coregistration(reference_dem: np.ndarray, dem_to_be_aligned: np.ndarray,
         reference_dem[mask] = np.nan
         dem_to_be_aligned[mask] = np.nan
 
+    # Generate center point coordinates for each pixel
```
adehecq, line 293: You know there's a function that does just this in Raster? self.coords

erikmannerfelt:
I changed the name of the PR since I accidentally added two commits, and I don't know how to undo it!

xdem/coreg.py
```diff
@@ -380,12 +399,100 @@ def icp_coregistration(reference_dem: np.ndarray, dem_to_be_aligned: np.ndarray,
...
+
+    # noqa: DAR101 **_
+    """
+    if not _has_cv2:
+        raise AssertionError("opencv (cv2) is not available and needs to be installed.")
+    dem_to_be_aligned_unmasked = dem_to_be_aligned.copy()
+
+    resolution = reference_dem.shape[1] / (bounds.right - bounds.left)
+    points: dict[str, np.ndarray] = {}
+    x_coords, y_coords = np.meshgrid(
```
adehecq, line 438: Is it worth making a small function for it taht would be used by both pdal and opencv's coreg?

adehecq:
> I changed the name of the PR since I accidentally added two commits, and I don't know how to undo it!

If you've done a commit and push, it's probably too late... What you should have done is to create a different branch for the second feature, then it would not add it to the existing PR.

adehecq:
You can always check if it's possible with git to link a commit to a new branch a posteriori...

xdem/coreg.py
erikmannerfelt: The PDAL and opencv conversion functions are not exactly similar. In that case, the code would have to be reformatted a bit. Maybe a task for a "simplification PR"?

xdem/coreg.py
erikmannerfelt: I did not know that! Good to know!

Right now, the coreg methods work completely independently of the Raster class (Rasters are only used in the main `coregister()` function).

This could be changed in the future, but I propose that it should be a separate "simplification PR"

erikmannerfelt:
Oh and I haven't described the new/improved ICP approaches:

The PDAL ICP approach has been modified so PDAL takes an XYZ point cloud without any NaNs instead of a raster with NaN values (which are later filtered). The running time is almost exactly the same for the Longyearbyen example (31.8s vs 31.6s measured with pytest), but it improves drastically with many NaNs. On a DEM with more than 80% NaNs, the previous implementation failed after >200 Hrs, while this finished in 10 minutes. 

On that note, we should add tests to try the methods on large datasets!

The new approach using opencv is much cleaner than PDAL, and also much faster. pytest measures it to 9.21s, vs 31.63s by PDAL! It also uses only variables in memory instead of writing temporary files to the disk. I had some trouble on servers with small `/tmp` partitions, so this is nice I guess. Finally, opencv is much more widely available than PDAL in terms of platforms (Windows most specifically).

I have set the new opencv approach as the default `"icp"` method.

xdem/coreg.py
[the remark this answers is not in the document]
erikmannerfelt: Removed the bad -9999 conversion and added conversion so that all input rasters will be either `float32` or `float64` in 9852054

xdem/coreg.py
rhugonnet: I agree that georeferenced-based functions such as coords() should be kept to a minimum :+1:

rhugonnet:
> Oh and I haven't described the new/improved ICP approaches:
> 
> The PDAL ICP approach has been modified so PDAL takes an XYZ point cloud without any NaNs instead of a raster with NaN values (which are later filtered). The running time is almost exactly the same for the Longyearbyen example (31.8s vs 31.6s measured with pytest), but it improves drastically with many NaNs. On a DEM with more than 80% NaNs, the previous implementation failed after >200 Hrs, while this finished in 10 minutes.
> 
> On that note, we should add tests to try the methods on large datasets!
> 
> The new approach using opencv is much cleaner than PDAL, and also much faster. pytest measures it to 9.21s, vs 31.63s by PDAL! It also uses only variables in memory instead of writing temporary files to the disk. I had some trouble on servers with small `/tmp` partitions, so this is nice I guess. Finally, opencv is much more widely available than PDAL in terms of platforms (Windows most specifically).
> 
> I have set the new opencv approach as the default `"icp"` method.

This is great, in particular the in-mem operations for applying to large datasets/parallel processing that require a lot of I/O.

erikmannerfelt:
> > Oh and I haven't described the new/improved ICP approaches:
> > The PDAL ICP approach has been modified so PDAL takes an XYZ point cloud without any NaNs instead of a raster with NaN values (which are later filtered). The running time is almost exactly the same for the Longyearbyen example (31.8s vs 31.6s measured with pytest), but it improves drastically with many NaNs. On a DEM with more than 80% NaNs, the previous implementation failed after >200 Hrs, while this finished in 10 minutes.
> > On that note, we should add tests to try the methods on large datasets!
> > The new approach using opencv is much cleaner than PDAL, and also much faster. pytest measures it to 9.21s, vs 31.63s by PDAL! It also uses only variables in memory instead of writing temporary files to the disk. I had some trouble on servers with small `/tmp` partitions, so this is nice I guess. Finally, opencv is much more widely available than PDAL in terms of platforms (Windows most specifically).
> > I have set the new opencv approach as the default `"icp"` method.
> 
> This is great, in particular the in-mem operations for applying to large datasets/parallel processing that require a lot of I/O.

Yeah I forgot to say that the opencv approach is multithreaded! Plus it contains native outlier filtering.

PR #35, opened 2021-03-10
Function to merge rasters using any callable (mean, median, sum etc)

erikmannerfelt:
In GDAL, only the operators "first" and "last" can be applied when merging (mosaicing) rasters. This often creates quite ugly seams, and may hide good data (a NaN in one raster can replace a value in another). Additionally, when dealing with multiple DEMs, one might want to calculate statistics such as the median, standard deviation (or anything else) within each cell.

An example:
```python
dem_west = gu.georaster.Raster("path/to/dem1")
dem_east = gu.georaster.Raster("path/to/dem2")

# This will merge the two DEMs and take the average (np.nanmean) values if they overlap.
# It returns a new Raster with the combined bounds of the two inputs.
merged_dem = xdem.spatial_tools.merge_rasters([dem_west, dem_east])
```

or:
```python
list_of_many_dems = [...]

# Calculate the variation between the different rasters
standard_deviation = xdem.spatial_tools.merge_rasters(list_of_many_dems, merge_algorithm=np.nanstd)
nmad = xdem.spatial_tools.merge_rasters(list_of_many_dems, merge_algorithm=xdem.spatial_tools.nmad)
```

### DIFF
--- a/xdem/spatial_tools.py
+++ b/xdem/spatial_tools.py
@@ -1,7 +1,9 @@
 """
 
 """
-from typing import Union
+from __future__ import annotations
+
+from typing import Callable, Union
 
 import geoutils as gu
 import numpy as np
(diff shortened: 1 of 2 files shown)


xdem/spatial_tools.py
```diff
@@ -78,3 +86,77 @@ def subtract_rasters(first_raster: Union[str, gu.georaster.Raster], second_raste
...
+        modulo = max_bounds[key] % resolution
+        max_bounds[key] -= modulo
+
+        if key in ["right", "top"] and modulo > 0:
+            max_bounds[key] += resolution
+
+    return rio.coords.BoundingBox(**max_bounds)
+
+
+def merge_rasters(rasters: list[gu.georaster.Raster], reference: int = 0, merge_algorithm: Callable = np.nanmean,
```
rhugonnet, line 109: Amaury will argue here that this type of method belongs in geoutils ;) (and I do too)

rhugonnet (approved):
Looks good. I'm thinking of how much we should link Dask to numpy/xarray for those calculations (speed/chunking). In any case, this should be easy enough to implement at a later stage "improvement" if we feel the need to

xdem/spatial_tools.py
erikmannerfelt: > Looks good. I'm thinking of how much we should link Dask to numpy/xarray for those calculations (speed/chunking). In any case, this should be easy enough to implement at a later stage "improvement" if we feel the need to

I totally agree that this would be a great addition! This will work for small to medium sized rasters, but not necessarily for huge ones.

I feel like we should have a test framework for huge DEMs, which would highlight where performance improvements should be made. I can make an associated issue!

erikmannerfelt: @rhugonnet  I agree that it would be a great addition to geoutils. Maybe we can remove it later if it's added there?

adehecq: Please add to the list of geoutils issues and then I can add it there!

PR #36, opened 2021-03-10
Local hypsometric interpolation

erikmannerfelt:
I've added a function `xdem.volume.hypsometric_binning()` which takes two 1D arrays and returns the following dataframe (on a glacier on Svalbard):

```bash
                                            median       mean        std   count
(227.21670532226562, 277.2167053222656] -16.151154 -16.702609   9.718108   597.0
(277.2167053222656, 327.2167053222656]  -33.600098 -29.986920  12.277790   736.0
(327.2167053222656, 377.2167053222656]  -34.610580 -33.213806   7.857803   904.0
(377.2167053222656, 427.2167053222656]  -24.568985 -24.089276   4.625535   772.0
(427.2167053222656, 477.2167053222656]  -19.961365 -19.116361   3.405295  1219.0
(477.2167053222656, 527.2167053222656]  -16.065674 -15.813945   4.325823  1415.0
(527.2167053222656, 577.2167053222656]  -10.828613 -10.168528   4.474668  1065.0
(577.2167053222656, 627.2167053222656]   -3.929382  -4.687018   4.310905  1072.0
(627.2167053222656, 677.2167053222656]    2.258606   1.073841   4.334557   821.0
(677.2167053222656, 727.2167053222656]    0.829834   0.944558   3.454747   495.0
(727.2167053222656, 777.2167053222656]    0.249207   0.142597   3.973010   295.0
(777.2167053222656, 827.2167053222656]   -1.081299  -0.164995   5.557001   298.0
(827.2167053222656, 877.2167053222656]    1.049011  -0.785810   7.675866   273.0
(877.2167053222656, 927.2167053222656]    0.102966  -1.146541   7.116261   242.0
(927.2167053222656, 977.2167053222656]   -3.863617  -5.664418   8.469879   166.0
```
The index is a `pd.IntervalIndex` instance, with nice convenience properties like `df.index.mid` for the midpoint of the intervals.

What do you think of this structure?


erikmannerfelt:
Aand I screwed up the double PRs again... I'm too tired to do git properly!

xdem/volume.py
```diff
@@ -0,0 +1,49 @@
...
+
+def hypsometric_binning(ddem: np.ndarray, dem: np.ndarray):
+
+    assert ddem.shape == dem.shape
+    nan_mask = np.isnan(ddem) | np.isnan(dem)
+    ddem = ddem[~nan_mask]
+    dem = dem[~nan_mask]
+
+    old_dem = dem - ddem
+    elevation_range = np.percentile(old_dem, 99) - np.percentile(old_dem, 1)
```
rhugonnet, line 16: we need to find a way to be robust with the elevation range without removing any glacier area

xdem/volume.py
```diff
@@ -0,0 +1,49 @@
...
+
+    bin_size = 50 if elevation_range > 500 else elevation_range / 10
```
rhugonnet, line 18: maybe we can move this into an argument, with a default setting

xdem/volume.py
```diff
@@ -0,0 +1,49 @@
...
+
+    bins = np.arange(old_dem.min(), old_dem.max() + bin_size, step=bin_size)
+    indices = np.digitize(old_dem, bins=bins)
+
+    medians = np.zeros(shape=bins.shape[0] - 1, dtype=ddem.dtype) + np.nan
```
rhugonnet, line 23: to initialize those in one line, I usually write:
`medians, means, stds = ( np.zeros(shape=,dtype=) * np.nan for i in range(3) )`

xdem/volume.py
erikmannerfelt: What do you mean? The elevation range is only used for calculating 10% steps. The actual bins are generated from `old_dem.min()` to `old_dem.max() + `bin_size`

xdem/volume.py
erikmannerfelt: Added this in 786b385. It can either be in georeferenced units or as a percentage (e.g. 10). Please let me know what you think of the argument names and the structure!

tests/test_volume.py
```diff
@@ -0,0 +1,20 @@
...
+
+    outlines = gu.geovector.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines"])
+    # Filter to only look at the Scott Turnerbreen glacier
+    outlines.ds = outlines.ds.loc[outlines.ds["NAME"] == "Scott Turnerbreen"]
+
+    mask = outlines.create_mask(dem_2009) == 255
+
+    ddem = dem_2009.data - dem_1990.data
+
+    ddem_bins = xdem.volume.hypsometric_binning(ddem.squeeze()[mask], dem_2009.data.squeeze()[mask])
```
adehecq, line 20: The tests should have an assert statement. In the case of sample data, you can check that the output array is equal to what you get today. Or at least a summary of it like mean, sum etc.

xdem/volume.py
```diff
@@ -0,0 +1,67 @@
...
+    :param dem: The reference DEM as a 2D or 1D array.
+    :param bin_size: The bin interval size in georeferenced units (or percent; 0-100, if normalized_bin_size=True)
+    :param normalized_bin_size: If the given bin size should be parsed as a percentage of the glacier's elevation range.
+
+    :returns: A Pandas DataFrame with elevation bins and dDEM statistics.
+    """
+
+    assert ddem.shape == dem.shape
+    # Remove all nans, and flatten the inputs.
+    nan_mask = np.isnan(ddem) | np.isnan(dem)
```
adehecq, line 22: Note that now Raster's nodata are treated using masked arrays. So maybe it's worth mentioning in the docstrings that nodata values should be set to Nan prior to calling this function.
But maybe it's worth testing if the input array is of type np.masked_array and to use array.mask instead.

xdem/volume.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Calculate the mean representative elevations between the two DEMs
+    mean_dem = dem - (ddem / 2)
+
+    # If the bin size should be seen as a percentage.
+    if normalized_bin_size:
+        assert bin_size > 0 and bin_size < 100
+
+        # Get the statistical elevation range to normalize the bin size with
+        elevation_range = np.percentile(mean_dem, 99) - np.percentile(mean_dem, 1)
```
adehecq, line 34: Should the 1/99 be hard-coded?

xdem/volume.py
adehecq: Or simply:
`medians, means, stds = ( np.zeros(shape=,dtype=) * np.nan) * 3`

adehecq: Missed a comma:
`medians, means, stds = ( np.zeros(shape=,dtype=) * np.nan, ) * 3`

xdem/volume.py
```diff
@@ -0,0 +1,49 @@
...
+    stds = medians.copy()
+    counts = np.zeros_like(medians, dtype=int)
+    for i in np.unique(indices):
+        values_in_bin = ddem[indices == i]
+        if values_in_bin.shape[0] == 0:
+            continue
+
+        medians[i - 1] = np.median(values_in_bin)
+        means[i - 1] = np.mean(values_in_bin)
+        stds[i - 1] = np.std(values_in_bin)
```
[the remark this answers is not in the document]
adehecq: Once more, it's all in old geoutils! ;-) You can just use any operator, and have a list of defaults you want to output:
https://github.com/GeoUtils/geoutils/blob/5c41b2a343d1f0fc161eed4426154656c06841a4/geoutils/geovector.py#L660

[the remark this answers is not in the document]
adehecq: Actually, the solution is a few blocks below. The line I gave is for initializing the arrays to Nan.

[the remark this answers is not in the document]
erikmannerfelt: I added an nmad column in 5853875. Would you both prefer to have it as a dynamic callable instead of all the columns right now?

tests/test_volume.py
erikmannerfelt: Added some sanity checks in 115fdf6.

xdem/volume.py
erikmannerfelt: It's just for calculating the bin sizes if a fraction is given. I think it's okay to use 1-99. Don't you? The bin sizes will only change slightly depending on the numbers.

xdem/volume.py
erikmannerfelt: Implemented @adehecq 's syntax in 5853875.

xdem/volume.py
rhugonnet: Agreed

xdem/volume.py
```diff
@@ -0,0 +1,105 @@
...
+
+    :returns: A Pandas DataFrame with elevation bins and dDEM statistics.
+    """
+    assert ddem.shape == ref_dem.shape
+    # Remove all nans, and flatten the inputs.
+    nan_mask = np.logical_or(
+        np.isnan(ddem) if not isinstance(ddem, np.ma.masked_array) else ddem.mask,
+        np.isnan(ref_dem) if not isinstance(ref_dem, np.ma.masked_array) else ref_dem.mask
+    )
+    ddem = ddem[~nan_mask]
```
rhugonnet: here we still need the values with NaN to quantify the area of each bin;
generally I do:
save the NaNs in the reference DEM -> save the count to a variable
keep the Nans in the dDEM (not NaN in reference DEM) -> save the count at each bin iteration (the interpolation per elevation bin happens here)
at the end, I propagate the mean volume change of the entire glacier to the reference DEM no data area

xdem/volume.py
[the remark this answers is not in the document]
adehecq: I think it's more flexible to use a callable yes. People might want the 25-75 percentile to plot the IQR for example etc. Then of course you can set the ones right now as default.

xdem/volume.py
adehecq: Ok, yes I didn't realize it had just little influence. So it's fine having it fixed.
But if I'm not mistaken, currently, ther is not the option to force bins of equal sizes. Instead of using a constant bin_size, the elevation steps could be calculated as percentiles e.g. 5, 10, 15, ... 90, 95%.

xdem/volume.py
```diff
@@ -0,0 +1,107 @@
...
+        data=np.vstack([
+            medians, means, stds, counts, nmads
+        ]).T,
+        columns=["median", "mean", "std", "count", "nmad"]
+    )
+
+    return output
+
+
+def interpolate_hypsometric_bins(hypsometric_bins: pd.DataFrame, height_column="median", method="polynomial", order=3,
```
adehecq, line 79: nice!

xdem/volume.py
erikmannerfelt: @adehecq it depends on if `normalized_bin_size` is `True` or `False`. If it's false, `bin_size=50` would be parsed as "one bin every 50 metres".

I am open to naming suggestions to make this clearer.

adehecq: Yes, I totally understand the case where normalized_bin_size=False. But if it's True, then the bin_size is still calculated as a fixed elevation size, not a fixed sample size.

adehecq: Well, you can calculate a percentile bin, then use it in your percentile functions:
`pbins = np.arange(min, max, bin_size)`
`bins = np.percentile(mean_dem, pbins)`

xdem/volume.py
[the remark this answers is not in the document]
erikmannerfelt: I remade the function to take any callable in 80d9bae. Now it doesn't give as much per default, but the percentile could e.g. be given by passing `aggregation_function=lambda x: np.percentile(x, 99)`.
```bash
                                             value   count                                                                                                                                                                                 
(225.28848266601562, 275.2884826660156] -19.098885   724.0                                                                                                                                                                                 
(275.2884826660156, 325.2884826660156]  -36.254318  1034.0                                                                                                                                                                                 
(325.2884826660156, 375.2884826660156]  -31.608276   617.0                                                                                                                                                                                 
(375.2884826660156, 425.2884826660156]         NaN     NaN                                                                                                                                                                                 
(425.2884826660156, 475.2884826660156]  -19.853256  1294.0                                                                                                                                                                                 
(475.2884826660156, 525.2884826660156]  -15.140961  1321.0                                                                                                                                                                                 
(525.2884826660156, 575.2884826660156]  -10.407410   997.0                                                                                                                                                                                 
(575.2884826660156, 625.2884826660156]   -3.927429  1021.0                                                                                                                                                                                 
(625.2884826660156, 675.2884826660156]    2.108887   791.0                                                                                                                                                                                 
(675.2884826660156, 725.2884826660156]    1.071594   530.0                                                                                                                                                                                 
(725.2884826660156, 775.2884826660156]    0.579987   296.0                                                                                                                                                                                 
(775.2884826660156, 825.2884826660156]   -1.488831   302.0                                                                                                                                                                                 
(825.2884826660156, 875.2884826660156]    1.046814   282.0                                                                                                                                                                                 
(875.2884826660156, 925.2884826660156]    0.201599   248.0                                                                                                                                                                                 
(925.2884826660156, 975.2884826660156]   -2.662354   161.0
```
@rhugonnet , now the columns are gone! Should the columns still be there, but other aggregate functions could be added if specified? I would consider this most recent approach to be the "cleanest".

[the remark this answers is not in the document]
adehecq: > +1 I'm fine with having all the columns, this will never be a limiting factor in terms of computing

Not so confident as you Romain. For large sample, the median and NMAD could be quite slow to run...
Anyway, I would make at least mean or median + count the defaults.

xdem/volume.py
adehecq: You could save a 'total_count' for example.

rhugonnet:
> For the hypsometry interpolation/extrapolation, I went with the easiest possible approach using the builtin `.interpolate()` method in Pandas.
> 
> One could go a bit further and add polynomial estimation using RANSAC to filter bad bins. `scikit-learn` has an excellent implementation which would be quite easy. It would be one more dependency though..
> 
> What do you guys think?

I'm fine adding scikit-learn with RANSAC, it would be a good addition. Although to be honest I'm not sure it will outperform traditional polynomial by much. I think the issue is mostly with the parametric nature of the polynomial regression, rather than the mitigation of outliers. I had a good performance when implementing 1D lowess, which is nonparametric. It was a while ago and I had to code most of it by hand ([https://github.com/iamdonovan/pyddem/blob/master/pyddem/volint_tools.py#L148](https://github.com/iamdonovan/pyddem/blob/master/pyddem/volint_tools.py#L148)), not sure what is available now.

erikmannerfelt:
I just added a hypsometry area calculation script. The workflow would now be:
```python
ddem_bins = xdem.volume.hypsometric_binning(ddem, reference_dem)
interpolated_bins = xdem.volume.interpolate_hypsometric_bins(ddem_bins)

bin_area = xdem.volume.calculate_hypsometry_area(interpolated_ddem_bins, reference_dem, pixel_size=pixel_size)

# This will be a float
volume_change = (ddem_bins * bin_area).sum()
```

Thoughts?

erikmannerfelt:
> > I just added a hypsometry area calculation script. The workflow would now be:
> > ```python
> > ddem_bins = xdem.volume.hypsometric_binning(ddem, reference_dem)
> > interpolated_bins = xdem.volume.interpolate_hypsometric_bins(ddem_bins)
> > 
> > bin_area = xdem.volume.calculate_hypsometry_area(interpolated_ddem_bins, reference_dem, pixel_size=pixel_size)
> > 
> > # This will be a float
> > volume_change = (ddem_bins * bin_area).sum()
> > ```
> > 
> > 
> > Thoughts?
> 
> It make sense. But what exactly is in ddem_bins and interpolated_ddem_bins (tthere's a typo I think as you called it interpolated_bins elsewhere too)?
> Does it contain the elevation bins and the median/mean dh? Because I don't understand why the median dh is needed in calculate_hypsometry_area, or why the elevation bins are used in the volume calculation...

@adehecq  `ddem_bins` looks like this:
```bash
                                             value   count
(225.28848266601562, 275.2884826660156] -19.098885   724.0
(275.2884826660156, 325.2884826660156]  -36.254318  1034.0
(325.2884826660156, 375.2884826660156]  -31.608276   617.0
(375.2884826660156, 425.2884826660156]  -22.635315   752.0
(425.2884826660156, 475.2884826660156]  -19.853256  1294.0
(475.2884826660156, 525.2884826660156]  -15.140961  1321.0
(525.2884826660156, 575.2884826660156]  -10.407410   997.0
(575.2884826660156, 625.2884826660156]   -3.927429  1021.0
(625.2884826660156, 675.2884826660156]    2.108887   791.0
(675.2884826660156, 725.2884826660156]    1.071594   530.0
(725.2884826660156, 775.2884826660156]    0.579987   296.0
(775.2884826660156, 825.2884826660156]   -1.488831   302.0
(825.2884826660156, 875.2884826660156]    1.046814   282.0
(875.2884826660156, 925.2884826660156]    0.201599   248.0
(925.2884826660156, 975.2884826660156]   -2.662354   161.0
```

Indeed, `interpolated_bins` should have been `interpolated_ddem_bins` or vice versa!

The reason `calculate_hypsometry_area()` needs `ddem_bins` or `interpolated_ddem_bins` is to make a representative DEM:
```python
ddem_func = scipy.interpolate.interp1d(ddem_bins.index.mid, ddem_bins.values, kind="linear", fill_value="extrapolate")

mean_dem = ref_dem - (ddem_func(ref_dem) / 2)
```
Would it make more sense to make `ddem_bins` an optional argument (even though it really should be included at all times)?

adehecq:
> > > I just added a hypsometry area calculation script. The workflow would now be:
> > > ```python
> > > ddem_bins = xdem.volume.hypsometric_binning(ddem, reference_dem)
> > > interpolated_bins = xdem.volume.interpolate_hypsometric_bins(ddem_bins)
> > > 
> > > bin_area = xdem.volume.calculate_hypsometry_area(interpolated_ddem_bins, reference_dem, pixel_size=pixel_size)
> > > 
> > > # This will be a float
> > > volume_change = (ddem_bins * bin_area).sum()
> > > ```
> > > 
> > > 
> > > Thoughts?
> > 
> > 
> > It make sense. But what exactly is in ddem_bins and interpolated_ddem_bins (tthere's a typo I think as you called it interpolated_bins elsewhere too)?
> > Does it contain the elevation bins and the median/mean dh? Because I don't understand why the median dh is needed in calculate_hypsometry_area, or why the elevation bins are used in the volume calculation...
> 
> @adehecq `ddem_bins` looks like this:
> 
> ```shell
>                                              value   count
> (225.28848266601562, 275.2884826660156] -19.098885   724.0
> (275.2884826660156, 325.2884826660156]  -36.254318  1034.0
> (325.2884826660156, 375.2884826660156]  -31.608276   617.0
> (375.2884826660156, 425.2884826660156]  -22.635315   752.0
> (425.2884826660156, 475.2884826660156]  -19.853256  1294.0
> (475.2884826660156, 525.2884826660156]  -15.140961  1321.0
> (525.2884826660156, 575.2884826660156]  -10.407410   997.0
> (575.2884826660156, 625.2884826660156]   -3.927429  1021.0
> (625.2884826660156, 675.2884826660156]    2.108887   791.0
> (675.2884826660156, 725.2884826660156]    1.071594   530.0
> (725.2884826660156, 775.2884826660156]    0.579987   296.0
> (775.2884826660156, 825.2884826660156]   -1.488831   302.0
> (825.2884826660156, 875.2884826660156]    1.046814   282.0
> (875.2884826660156, 925.2884826660156]    0.201599   248.0
> (925.2884826660156, 975.2884826660156]   -2.662354   161.0
> ```
> 
> Indeed, `interpolated_bins` should have been `interpolated_ddem_bins` or vice versa!
> 
> The reason `calculate_hypsometry_area()` needs `ddem_bins` or `interpolated_ddem_bins` is to make a representative DEM:
> 
> ```python
> ddem_func = scipy.interpolate.interp1d(ddem_bins.index.mid, ddem_bins.values, kind="linear", fill_value="extrapolate")
> 
> mean_dem = ref_dem - (ddem_func(ref_dem) / 2)
> ```
> 
> Would it make more sense to make `ddem_bins` an optional argument (even though it really should be included at all times)?

But should calculate_hypsometry_area not only depend on the bin edges and a reference DEM? In that case, I would rather just pass the elevation bins, rather than ddem_bins.
Or do you use the ddem elevation for the calculation of the hypsometry? If so, yes it should be optional, because I believe by default only the reference elevation should be used (an elevation that we trust, as opposed to an elevation that might have outliers).

adehecq:
> If a glacier has lost extremely much at the terminus, its present-day area bins at low elevation will have very small associated areas. The elevation-dependent elevation change relationship will still say that there was a lot of vertical change, but with a tiny area to multiply with, the volume change will be under-represented.

I understand your concern. I don't think there is a best approach so far, so maybe we should have 3 options: use the older DEM, use the newer DEM or use the average of both. But this is up to the user to specify which 'reference' he/she wants. The way it is done now is not very transparent as an elevation is added to what is named 'reference'.
On the mid-term, we should definitely make some comparison between the three options. And maybe think of a better option. I think there might be an even better option than these three, but I need to think about it...

erikmannerfelt:
Oh no!

erikmannerfelt:
> > If a glacier has lost extremely much at the terminus, its present-day area bins at low elevation will have very small associated areas. The elevation-dependent elevation change relationship will still say that there was a lot of vertical change, but with a tiny area to multiply with, the volume change will be under-represented.
> 
> I understand your concern. I don't think there is a best approach so far, so maybe we should have 3 options: use the older DEM, use the newer DEM or use the average of both. But this is up to the user to specify which 'reference' he/she wants. The way it is done now is not very transparent as an elevation is added to what is named 'reference'.
> On the mid-term, we should definitely make some comparison between the three options. And maybe think of a better option. I think there might be an even better option than these three, but I need to think about it...

@adehecq , in c98858e, I added a keyword argument, `timeframe`, to specify "when" the area bins should be from:

```python
# This would yield area bins from the reference DEM only. The default is "mean"
bin_area  = xdem.volume.calculate_hypsometry_area(
                                                  ddem_bins=interpolated_ddem_bins,
                                                  ref_dem=reference_dem,
                                                  pixel_size=pixel_size,
                                                  timeframe="reference"
)

```

rhugonnet:
> @adehecq , in [c98858e](https://github.com/GlacioHack/xdem/commit/c98858ee38d9c56d43c0caa5463d809b1c1f9f45), I added a keyword argument, `timeframe`, to specify "when" the area bins should be from:
> 
> ```python
> # This would yield area bins from the reference DEM only. The default is "mean"
> bin_area  = xdem.volume.calculate_hypsometry_area(
>                                                   ddem_bins=interpolated_ddem_bins,
>                                                   ref_dem=reference_dem,
>                                                   pixel_size=pixel_size,
>                                                   timeframe="reference"
> )
> ```

Looks good, I agree that there is no "best option" despite what Matthias said ;). We can discuss those at some point, but this difference mostly affects the lower bins, and those usually don't have any data gap so the effect will be minor in any case.

erikmannerfelt:
> Looks good, I agree that there is no "best option" despite what Matthias said ;). We can discuss those at some point, but this difference mostly affects the lower bins, and those usually don't have any data gap so the effect will be minor in any case.

I would argue that the mean option is the best! ;) And I will attempt it below. Either way, now everyone has the option to choose. @adehecq is this approach okay with you?

### Why I like the mean
The theoretically optimal approach is arguably to calculate the volume change every year using a hypsometry that fits that exact year. I'll exemplify with the period 1990 to 2009 (like the examples):
```python
volume_change_1995 = hypsometry_1995 * elevation_change_1995
```
With the assumption that the hypsometry and elevation changes linearly between 1990 and 2009, we should calculate the change as:
```python
total_volume_change = pd.Series(..)  # Like the hypsometry

for year in range(1990, 2009 + 1):
     updated_hypsometry = hypsometry_2009 - (elevation_change / (2009 - year))
     yearly_volume_change = updated_hypsometry * (elevation_change / (2009 - 1990))
     total_volume_change += yearly_volume_change
```
Because of the linear assumption, the above script is identical to the simpler:
```python

mean_hypsometry = hypsometry_2009 - (elevation_change / 2)
total_volume_change = mean_hypsometry * elevation_change
```

rhugonnet:
> The theoretically optimal approach is arguably to calculate the volume change every year using a hypsometry that fits that exact year. I'll exemplify with the period 1990 to 2009 (like the examples):

In terms of hypsometry, you are indeed closest to the real glacier hypsometry :+1: 
In terms of gap-filling for the volume change calculation, however, this will have almost no effect. If we bin a little bit lower or higher, the glacier hypsometry will still be consistent for a given period and provide a similar mean elevation change to fill the bin. 
And those details are quite minor against the two issues of:
-  filling gaps in the accumulation areas (not affected by much hypsometry change), and
-  accounting for glacier shrinkage when using elevation bin gap-filling (we'd need to construct time-varying outlines)

adehecq:
> > The theoretically optimal approach is arguably to calculate the volume change every year using a hypsometry that fits that exact year. I'll exemplify with the period 1990 to 2009 (like the examples):
> 
> In terms of hypsometry, you are indeed closest to the real glacier hypsometry +1
> In terms of gap-filling for the volume change calculation, however, this will have almost no effect. If we bin a little bit lower or higher, the glacier hypsometry will still be consistent for a given period and provide a similar mean elevation change to fill the bin.
> And those details are quite minor against the two issues of:
> 
>     * filling gaps in the accumulation areas (not affected by much hypsometry change), and
> 
>     * accounting for glacier shrinkage when using elevation bin gap-filling (we'd need to construct time-varying outlines)

Also I think what's important is that we are consistent, i.e. the hypsometry used for calculating the mean dh is the same as the one used for volume calculation. But It would be nice to have a small notebook to test the different options and discuss teh uncertainties associated with it. Should I add an issue?

rhugonnet:
> > > The theoretically optimal approach is arguably to calculate the volume change every year using a hypsometry that fits that exact year. I'll exemplify with the period 1990 to 2009 (like the examples):
> > 
> > 
> > In terms of hypsometry, you are indeed closest to the real glacier hypsometry +1
> > In terms of gap-filling for the volume change calculation, however, this will have almost no effect. If we bin a little bit lower or higher, the glacier hypsometry will still be consistent for a given period and provide a similar mean elevation change to fill the bin.
> > And those details are quite minor against the two issues of:
> > ```
> > * filling gaps in the accumulation areas (not affected by much hypsometry change), and
> > 
> > * accounting for glacier shrinkage when using elevation bin gap-filling (we'd need to construct time-varying outlines)
> > ```
> 
> Also I think what's important is that we are consistent, i.e. the hypsometry used for calculating the mean dh is the same as the one used for volume calculation. But It would be nice to have a small notebook to test the different options and discuss teh uncertainties associated with it. Should I add an issue?

Sure!

xdem/volume.py
rhugonnet: @adehecq I knew something didn't work with this syntax (tried this in the past)! Using `*3` creates linked objects, so if for instance you do medians[0] = 1 it will change the value of means[0] and stds[0] into 1 as well.
```
medians, means, stds = ( np.zeros(3) * np.nan, ) * 3
means[0] = 1
stds[0]
Out[9]: 1.0
```
@erikmannerfelt Didn't you encounter that issue when updating the syntax? I need to correct this in spstats as well (I was updating similarly on my side)

xdem/volume.py
erikmannerfelt: [I added this as an issue.](https://github.com/GlacioHack/xdem/issues/39)

xdem/volume.py
adehecq: Yes, good to know!!

erikmannerfelt:
Is this approved, @rhugonnet and @adehecq, or should we discuss something more?

erikmannerfelt:
I've added "equal area" binning in 3b33d4b, according to @rhugonnet's approach in #39. Right now, the terminology is a bit all over the place. This is how I've made the options, but feel free to suggest changes:

```python
volume.hypsometric_binning(ddem, ref_dem, bins=50, kind="see below")
```

**"equal_height"**:  The elevation range of each bin is 50 georeferenced units.

**"equal_count"**:  There are 50 bins in total

**"equal_area"**: There are 50 bins in total, and all cover almost exactly the same area (#39)

To exemplify the outputs (the values don't mean much because I've neither coregistered nor masked the DEMs):
```python
In [1]: import xdem

In [2]: ref_dem = xdem.dem.DEM(xdem.examples.FILEPATHS["longyearbyen_ref_dem"])

In [3]: tba_dem = xdem.dem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"])

In [4]: ddem = ref_dem.data - tba_dem.data

In [5]: xdem.volume.hypsometric_binning(ddem, ref_dem.data, bins=10, kind="equal_height")
Out[5]:
                                            value     count
(8.6741304397583, 18.6741304397583]      -1.760605  15780.0
(18.6741304397583, 28.6741304397583]     -2.291516  45596.0
(28.6741304397583, 38.6741304397583]     -1.547386  33645.0
(38.6741304397583, 48.6741304397583]     -1.166397  26651.0
(48.6741304397583, 58.6741304397583]     -1.175062  27800.0
...                                            ...      ...
(978.6741304397583, 988.6741304397583]   -1.051819    265.0
(988.6741304397583, 998.6741304397583]   -0.553040    203.0
(998.6741304397583, 1008.6741304397583]  -0.421906    148.0
(1008.6741304397583, 1018.6741304397583]  0.948212     68.0
(1018.6741304397583, 1028.6741304397583]  0.209137     16.0

[102 rows x 2 columns]

In [6]: xdem.volume.hypsometric_binning(ddem, ref_dem.data, bins=10, kind="equal_count")
Out[6]:
                                            value     count
(8.6741304397583, 109.9881199936731]    -1.807365  242028.0
(109.9881199936731, 211.3021095475879]  -2.855278  142598.0
(211.3021095475879, 312.6160991015027]  -2.900841  183366.0
(312.6160991015027, 413.9300886554175]  -3.097565  177362.0
(413.9300886554175, 515.2440782093322]  -3.637207  160876.0
(515.2440782093322, 616.5580677632471]  -3.866760  144277.0
(616.5580677632471, 717.8720573171619]  -3.371338  121107.0
(717.8720573171619, 819.1860468710767]  -3.312836   90834.0
(819.1860468710767, 920.5000364249914]  -2.082214   41171.0
(920.5000364249914, 1021.8140259789062] -0.364868    8401.0

In [7]: xdem.volume.hypsometric_binning(ddem, ref_dem.data, bins=10, kind="equal_area")
Out[7]:
                                            value     count
(8.6741304397583, 51.949240112304686]    -1.686300  131202.0
(51.949240112304686, 126.75953216552733] -2.185017  131202.0
(126.75953216552733, 216.52808837890626] -2.811653  131202.0
(216.52808837890626, 287.96279296875]    -2.922958  131202.0
(287.96279296875, 362.9493103027344]     -3.012344  131202.0
(362.9493103027344, 439.10780029296876]  -3.240845  131202.0
(439.10780029296876, 523.1372314453124]  -3.678909  131202.0
(523.1372314453124, 615.8244506835937]   -3.877625  131202.0
(615.8244506835937, 726.0829772949219]   -3.362030  131202.0
(726.0829772949219, 1021.8140268789062]  -2.746582  131202.0
```
One potential limitation in this terminology is that "equal_count" and the "count" column are not really related. "equal_count" is an equal bin count, while the "count" column is the pixel/value count for each bin! Is this confusing or should we just leave it as is?

erikmannerfelt:
Crazy how much a 1 minute bathroom break's reflection can matter. 

What about `"fixed_height"`, `"fixed_count"` and `"equal_area"` instead?

Especially the kind `"equal_count"` doesn't make much sense now. Equal to what? (what I mean is that the bin count is fixed!)

erikmannerfelt:
Oh no, I just found a large unintended *feature* in this approach. The "count" is not analogous to "area" when there are nans in the dDEM! That is why I didn't want to mix those terms in the first place, I recall...

What are you guys' opinion on this, @rhugonnet  and @adehecq . To be clear, the "equal_area" binning kind makes it so that bins are created with an equal amount of finite dDEM values, not an equal glacier area. That might be what we want, but it wasn't what I intended!

erikmannerfelt:
Another suggestion for alternate names:
`"equal_height"` like before
`"fixed_bincount"` for fixed bin count with equal height interval
`"equal_count"` for a fixed bin count with equal point counts (thus consistent in terminology with the "count" column)

rhugonnet:
> What are you guys' opinion on this, @rhugonnet and @adehecq . To be clear, the "equal_area" binning kind makes it so that bins are created with an equal amount of finite dDEM values, not an equal glacier area. That might be what we want, but it wasn't what I intended!

Intuitively, I expect the most "performant" one to be a uniform (equal_count) binning accounting for NaNs. That is to say, equal count for the hypsometric binning, independently of what values are passed in the ddem array.
Let me take a bathroom break and see how much it enlightens me on the naming :D

erikmannerfelt:
> 
> Intuitively, I expect the most "performant" one to be a uniform (equal_count) binning accounting for NaNs. That is to say, equal count for the hypsometric binning, independently of what values are passed in the ddem array.
> Let me take a bathroom break and see how much it enlightens me on the naming :D

Please correct me if I'm mistaken, @rhugonnet , but that implies the bins have to be calculated using the reference DEM elevations, not the mean elevations. This may be problematic with large change, like we discussed before:

If a glacier has lost 100 m at 1000-1100 m a.s.l. between 1900 and 2000, the 2000 area will be 0, so the apparent volume change will be 0.

rhugonnet:
FYI, pandas naming:
`qcut` (quantile-cut) = "equal_count"; `cut` (simple cut) = "any bin range customized binning"
Not very useful...

The problem, I think, is that "fixed_bincount" is a sub-category of equal height! And equal_area is actually also based on height binning (although it translate into area). So using "height" or "area" is a bit confusing to me. You have only 1 category (height), and then several ways of dividing it.
It could be as simple as this:
```python
binning_type = ['custom','fixed','quantile','count'] #you can bind by a fixed number of height or areas
binning_width = [[],50 m , 10%, 100]
```
For count, we could let the user one/two-line (elevation range/bin_count) and use "fixed" if desired?

rhugonnet:
> > Intuitively, I expect the most "performant" one to be a uniform (equal_count) binning accounting for NaNs. That is to say, equal count for the hypsometric binning, independently of what values are passed in the ddem array.
> > Let me take a bathroom break and see how much it enlightens me on the naming :D
> 
> Please correct me if I'm mistaken, @rhugonnet , but that implies the bins have to be calculated using the reference DEM elevations, not the mean elevations. This may be problematic with large change, like we discussed before:
> 
> If a glacier has lost 100 m at 1000-1100 m a.s.l. between 1900 and 2000, the 2000 area may be 0, so the apparent volume change will be 0.

I'm not sure this makes any difference: if the elevation change is 0 at lower elevation, you can propagate that 0 to the rest of the bin without any issue. 
One of the advantage of the hypsometric method is to mitigate the effects of no-data in the dDEM. By binning according to an external reference with almost no data gaps, you propagate the change to where you are missing it. So the `count` should absolutely be that of the reference DEM with least data gaps, no?

adehecq:
I would say, Romain's last suggestion is the most instuitive? So 
'fixed' -> fixed bin height
'quantile' -> bin size is based on quantiles
'count' -> a fixed number of bins
If needed to make it clearer, we could have 'fixed' -> 'fixed_height', 'count' -> 'fixed_count'?

erikmannerfelt:
> > Please correct me if I'm mistaken, @rhugonnet , but that implies the bins have to be calculated using the reference DEM elevations, not the mean elevations. This may be problematic with large change, like we discussed before:
> > If a glacier has lost 100 m at 1000-1100 m a.s.l. between 1900 and 2000, the 2000 area may be 0, so the apparent volume change will be 0.
> 
> I'm not sure this makes any difference: if the elevation change is 0 at lower elevation, you can propagate that 0 to the rest of the bin without any issue.
> One of the advantage of the hypsometric method is to mitigate the effects of no-data in the dDEM. By binning according to an external reference with almost no data gaps, you propagate the change to where you are missing it. So the `count` should absolutely be that of the reference DEM with least data gaps, no?

Sorry, @rhugonnet, but I honestly don't understand what you mean. If we have a glacier:
| Elevation | dH | Area_pre | Area_post|
|---------------|:------:|:-------------:|--------------|
| 1000-1100 | -100 | 10000 | 0 |
| 1100-1200 | -70 | 12000 | 5000 |
| 1300-1400 | -50 | 15000 | 10000| 

etc., the integrated volume change in the first elevation bin would be 0, assuming one uses the modern hypsometry. How and where is this propagated?

erikmannerfelt:
> I would say, Romain's last suggestion is the most instuitive? So
> 'fixed' -> fixed bin height
> 'quantile' -> bin size is based on quantiles
> 'count' -> a fixed number of bins
> If needed to make it clearer, we could have 'fixed' -> 'fixed_height', 'count' -> 'fixed_count'?

I agree with "fixed", "quantile" and "count", like @rhugonnet suggested. I'll implement that!

erikmannerfelt:
> > I would say, Romain's last suggestion is the most instuitive? So
> > 'fixed' -> fixed bin height
> > 'quantile' -> bin size is based on quantiles
> > 'count' -> a fixed number of bins
> > If needed to make it clearer, we could have 'fixed' -> 'fixed_height', 'count' -> 'fixed_count'?
> 
> I agree with "fixed", "quantile" and "count", like @rhugonnet suggested. I'll implement that!

Done in 5646dca

rhugonnet:
> Sorry, @rhugonnet, but I honestly don't understand what you mean. If we have a glacier:
> 
> Elevation	dH	Area_pre	Area_post
> 1000-1100	-100	10000	0
> 1100-1200	-70	12000	5000
> 1300-1400	-50	15000	10000
> etc., the integrated volume change in the first elevation bin would be 0, assuming one uses the modern hypsometry. How and where is this propagated?

The date of the reference DEM just shifts all bins, you don't really lose any elevation change information:
If you use a DEM_pre:
> Elevation	dH	Area_pre
> 900-1000	0	0
> 1000-1100	-100	10000
> 1100-1200	-70	10000
> 1300-1400	-50	10000
> 1400-1500	-10	10000
> 1500-1600	0	0

If you use a DEM_post:
> Elevation	dH	Area_pre
> 800-900	0	0
> 900-1000	-100	10000
> 1000-1100	-70	10000
> 1100-1200	-50	10000
> 1300-1400	-10	10000
> 1400-1500	0	0

Hope this makes sense in relation to the comment !

erikmannerfelt:
> > Sorry, @rhugonnet, but I honestly don't understand what you mean. If we have a glacier:
> > Elevation	dH	Area_pre	Area_post
> > 1000-1100	-100	10000	0
> > 1100-1200	-70	12000	5000
> > 1300-1400	-50	15000	10000
> > etc., the integrated volume change in the first elevation bin would be 0, assuming one uses the modern hypsometry. How and where is this propagated?
> 
> The date of the reference DEM just shifts all bins, you don't really lose any elevation change information:
> If you use a DEM_pre:
> 
> > Elevation	dH	Area_pre
> > 900-1000	0	0
> > 1000-1100	-100	10000
> > 1100-1200	-70	10000
> > 1300-1400	-50	10000
> > 1400-1500	-10	10000
> > 1500-1600	0	0
> 
> If you use a DEM_post:
> 
> > Elevation	dH	Area_pre
> > 800-900	0	0
> > 900-1000	-100	10000
> > 1000-1100	-70	10000
> > 1100-1200	-50	10000
> > 1300-1400	-10	10000
> > 1400-1500	0	0
> 
> Hope this makes sense in relation to the comment !

Ooooooh, of course! Wow now I feel stupid. I've made things so hard for myself the whole way

I'll correct this (meaning this will be true: "area" == "count_sum" * "pixel_area") and set the default `calculate_hypsometry_area()` surface to be the reference_dem, just like you've both suggested for a week now!

erikmannerfelt:
Since the latest commit, I daresay this is ready. We solved the issue I had with reference/mean elevations (turns out I was just poorly informed) and the functions are now in a working state. 

Once merged, I can start implementing them in the DEMCollection class.

Is there anything else we should consider first?

adehecq:
> Since the latest commit, I daresay this is ready. We solved the issue I had with reference/mean elevations (turns out I was just poorly informed) and the functions are now in a working state.
> 
> Once merged, I can start implementing them in the DEMCollection class.
> 
> Is there anything else we should consider first?

Feel free to merge and move on with the DEMCollection class. Maybe we should have a meeting next week to make sure we agree on the features to implement in xdem.

erikmannerfelt:
> > Since the latest commit, I daresay this is ready. We solved the issue I had with reference/mean elevations (turns out I was just poorly informed) and the functions are now in a working state.
> > Once merged, I can start implementing them in the DEMCollection class.
> > Is there anything else we should consider first?
> 
> Feel free to merge and move on with the DEMCollection class. Maybe we should have a meeting next week to make sure we agree on the features to implement in xdem.

Okay! Sounds great to have a discussion about moving forward next week.

PR #37, opened 2021-03-10
Spatial statistics: empirical variogram estimation, numerical variogram integration into volume errors

rhugonnet:
Tools for:
- Variogram wrapper: unfortunately the scikit-gstat object is, at this stage, a bit useless (can't fit multi-range, some exception are unstable). So I move empirical variograms to a simple pandas.Dataframe: variance, lag, count.
- Multi-range estimation: need specific binning (for now I will test the "kmeans" option added in skgstat, but we might have to resort to our own binning, although skgstat doesn't allow for customized binning so it will be annoying), with multiple runs (to assess the stability/std of the empirical variogram) that can be run in parallel using multiprocessing (typically 10,000 samples will take several minutes, it can go up pretty fast)
- Integration of variogram/covariance functions over a spatial support: circular integration (spatially continuous ensembles) & double sum of covar based on centroid coordinates (spatially disconnected ensembles).

### DIFF
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -39,22 +39,54 @@ class TestEmpiricalVariogram:
         df = xdem.spstats.get_empirical_variogram(dh=diff.data.flatten()[0:1000],coords=coords[0:1000,:])
 
         # check the wrapper script runs with various inputs
-        df_gsd = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],bin_func='even')
+        # with gsd as input
+        df_gsd = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0])
 
+        # with coords as input, and "uniform" bin_func
         df_coords = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data.flatten(),coords=coords,
                                                                        bin_func='uniform')
 
-        df_1000_bins = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],
-                                                                          bin_func='even',n_lags=1000)
+        # using more bins
+        df_1000_bins = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],n_lags=1000)
 
-        df_sig = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],bin_func='even',
-                                                                    nsamp=1000,nrun=30,nproc=10)
+        # using multiple runs with parallelized function
+        df_sig = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0], nsamp=1000,
+                                                                    nrun=20, nproc=10, maxlag = 10000)
 
-        fun = xdem.spstats.fit_model_sum_vgm(['Sph'],df_sig)
+        # test plotting
+        xdem.spstats.plot_vgm(df_sig)
 
-        fun2 = xdem.spstats.fit_model_sum_vgm(['Sph','Sph','Sph'],emp_vgm_df=df_sig)
+        # single model fit
+        fun, _ = xdem.spstats.fit_model_sum_vgm(['Sph'],df_sig)
+        xdem.spstats.plot_vgm(df_sig,fit_fun=fun)
 
+        # triple model fit
+        fun2, _ = xdem.spstats.fit_model_sum_vgm(['Sph','Sph','Sph'],emp_vgm_df=df_sig)
+        xdem.spstats.plot_vgm(df_sig,fit_fun=fun2)
 
+    def test_neff_estimation(self):
+
+        # test the precision of numerical integration for several spherical models
+
+        # short range
+        crange1 = [10**i for i in range(8)]
+        # long range
+        crange2 = [100*sr for sr in crange1]
+
+        p1 = 0.8
+        p2 = 0.2
+
+        for r1 in crange1:
+            r2 = crange2[crange1.index(r1)]
+
+            # and for any glacier area
+            for area in [10**i for i in range(10)]:
+
+                neff_circ_exact = xdem.spstats.exact_neff_sphsum_circular(area=area,crange1=r1,psill1=p1,crange2=r2
+                                                                          ,psill2=p2)
+                neff_circ_numer = xdem.spstats.neff_circ(area,[(r1,'Sph',p1),(r2,'Sph',p2)])
+
+                assert np.abs(neff_circ_exact-neff_circ_exact)<0.001
 
 
 
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -1,0 +1,49 @@
+"""
+Functions to test the spatial statistics.
+
+"""
+from __future__ import annotations
+import numpy as np
+import xdem
+from xdem import examples
+import geoutils as gu
+
+
+def load_diff() -> tuple[gu.georaster.Raster, np.ndarray] :
+    """Load example files to try coregistration methods with."""
+    examples.download_longyearbyen_examples(overwrite=False)
+
+    reference_raster = gu.georaster.Raster(examples.FILEPATHS["longyearbyen_ref_dem"])
+    to_be_aligned_raster = gu.georaster.Raster(examples.FILEPATHS["longyearbyen_tba_dem"])
+    glacier_mask = gu.geovector.Vector(examples.FILEPATHS["longyearbyen_glacier_outlines"])
+
+    metadata = {}
+    aligned_raster, _ = xdem.coreg.coregister(reference_raster, to_be_aligned_raster, method="amaury", mask=glacier_mask,
+                                     metadata=metadata)
+
+    diff = xdem.spatial_tools.subtract_rasters(reference_raster,aligned_raster)
+    mask = glacier_mask.create_mask(diff)
+
+    return diff, mask
+
+class TestEmpiricalVariogram:
+
+    def test_empirical_variogram(self):
+
+        diff, mask = load_diff()
+
+        x, y = diff.coords(offset='center')
+        coords = np.dstack((x.flatten(), y.flatten())).squeeze()
+
+        # check the base script runs with right input shape
+        df = xdem.spstats.get_empirical_variogram(dh=diff.data.flatten()[0:1000],coords=coords[0:1000,:])
+
+        # check the wrapper script runs with various inputs
+        df_gsd = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],bin_func='even')
+        df_coords = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data.flatten(),coords=coords,bin_func='uniform')
+        df_1000_bins = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],bin_func='even',n_lags=1000)
+        df_sig = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],bin_func='even',
+                                                                    nsamp=1000,nrun=30,nproc=10)
+
+
+
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -40,10 +40,21 @@ class TestEmpiricalVariogram:
 
         # check the wrapper script runs with various inputs
         df_gsd = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],bin_func='even')
-        df_coords = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data.flatten(),coords=coords,bin_func='uniform')
-        df_1000_bins = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],bin_func='even',n_lags=1000)
+
+        df_coords = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data.flatten(),coords=coords,
+                                                                       bin_func='uniform')
+
+        df_1000_bins = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],
+                                                                          bin_func='even',n_lags=1000)
+
         df_sig = xdem.spstats.sample_multirange_empirical_variogram(dh=diff.data,gsd=diff.res[0],bin_func='even',
                                                                     nsamp=1000,nrun=30,nproc=10)
+
+        fun = xdem.spstats.fit_model_sum_vgm(['Sph'],df_sig)
+
+        fun2 = xdem.spstats.fit_model_sum_vgm(['Sph','Sph','Sph'],emp_vgm_df=df_sig)
+
+
 
 
 
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -26,14 +26,16 @@ def load_diff() -> tuple[gu.georaster.Raster, np.ndarray] :
 
     return diff, mask
 
-class TestEmpiricalVariogram:
+class TestVariogram:
 
-    def test_empirical_variogram(self):
+    # check that the scripts are running
+    def test_empirical_fit_variogram_running(self):
 
+        # get some data
         diff, mask = load_diff()
-
         x, y = diff.coords(offset='center')
         coords = np.dstack((x.flatten(), y.flatten())).squeeze()
+
 
         # check the base script runs with right input shape
         df = xdem.spstats.get_empirical_variogram(dh=diff.data.flatten()[0:1000],coords=coords[0:1000,:])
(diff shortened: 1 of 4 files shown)


rhugonnet:
@erikmannerfelt the type "function" in my annotation (for sphinx) isn't recognized by PyCharm, any tip for this?

xdem/spstats.py
```diff
@@ -0,0 +1,507 @@
...
+    :param dh: elevation differences
+    :param coords: coordinates
+    :param method:
+
+    :return: subsets of dh and coords
+    """
+
+    if len(coords) > nsamp:
+        # TODO: maybe we can also introduce something to sample without replacement between all samples?
+        subset = np.random.choice(len(coords), nsamp, replace=False)
```
adehecq, line 78: One issue I had with the random sampling is that my results where not reproducible. For my frontier's paper, I did not save the sampled points, so when I was re-running my scripts, the range/sills varied quite a bit from my initial figures...
This made me think, maybe we could have a way to estimate the uncertainty of the range/sill by running several iterations with random sampling...

adehecq:
> @erikmannerfelt the type "function" in my annotation (for sphinx) isn't recognized by PyCharm, any tip for this?

Is there any Callable type?

rhugonnet:
> > @erikmannerfelt the type "function" in my annotation (for sphinx) isn't recognized by PyCharm, any tip for this?
> 
> Is there any Callable type?

I follow how the types are defined in scipy.integrate and write it like this: 
`def integrate_fun(fun: Union[function,LowLevelCallable],low_b: float ,upp_b: float) -> float:
but the `Union[function,LowLevelCallable]` is not recognized by PyCharm (`function` on its own also), I have red everywhere :'(
`

xdem/spstats.py
```diff
@@ -0,0 +1,507 @@
...
+
+    std_err = np.sqrt(2*np.pi*full_int / area)
+
+    if crange2 is None or psill2 is None or model2 is None:
+        return psill1 / std_err ** 2
+    else:
+        return (psill1 + psill2) / std_err ** 2
+
+
+def integrate_fun(fun: Union[function,LowLevelCallable],low_b: float ,upp_b: float) -> float:
```
erikmannerfelt, line 313: Functions are `Callable`'s in python. Anything that you can put a `()` after is a callable.

```python
from typing import Callable

def apply_function(values: list[float], function: Callable) -> list[float]:
    return [function(value) for value in values]

File: xdem/spstats.py [hunk 3]
```diff
@@ -0,0 +1,507 @@
+"""
+xdem.spstats provides tools to use spatial statistics for elevation change data
+"""
+from __future__ import annotations
+import os
+import numpy as np
+import math as m
+import skgstat as skg
```
erikmannerfelt, line 8: Should this be added to `environment.yml` ?

xdem/spstats.py
```diff
@@ -0,0 +1,507 @@
...
+import ogr
+from functools import partial
+
+def get_empirical_variogram(dh: np.ndarray,coords: np.ndarray, **kwargs) -> pd.DataFrame:
+    """
+    Get empirical variogram from skgstat.Variogram object
+
+    :param dh: elevation differences
+    :param coords: coordinates
+    :return: empirical variogram (variance, lags, counts)
```
erikmannerfelt, line 21: I feel nitpicky now, but isn't the established syntax `:returns:`? I might be wrong, and I don't care that much either!

xdem/spstats.py
```diff
@@ -0,0 +1,507 @@
...
+        # return V.to_DataFrame()
+
+        exp = V.experimental
+        bins = V.bins
+        count = np.zeros(V.n_lags)
+        tmp_count = np.fromiter((g.size for g in V.lag_classes()), dtype=int)
+        count[0:len(tmp_count)] = tmp_count
+
+    # there are still some exceptions not well handled by skgstat
+    except:
```
erikmannerfelt, line 37: A pure `except:` is dangerous, as it could for example be `KeyboardInterrupt`.
I suggest either:
```python
except ValueError, TypeError:
    # Do what you've done below
```
or 
```python
except Exception as exception:
    if "part of the error message" not in str(exception):
        raise exception
    # Do what you've done below.

```

xdem/spstats.py
```diff
@@ -0,0 +1,507 @@
...
+        if 'n_lags' in kwargs.keys():
```
erikmannerfelt, line 38: This could be shortened to:
```python
n_lags = kwargs.get('n_lags') or 10
```
If `kwargs.get('n_lags')` returns None, it defaults to 10!

xdem/spstats.py
```diff
@@ -0,0 +1,507 @@
...
+        coords_sub = coords[subset]
+        dh_sub = dh[subset]
+    else:
+        coords_sub = coords
+        dh_sub = dh
+
+    return dh_sub, coords_sub
+
+def sample_multirange_empirical_variogram(dh: np.ndarray, gsd: float = None, coords: np.ndarray = None,
+                                          nsamp: int = 10000, range_list: list= None, nrun=1, nproc=1, **kwargs) -> pd.DataFrame:
```
erikmannerfelt, line 88: `nrun` and `nproc` are missing type annotations. Do they have to be ints or can they be float?

xdem/spstats.py
```diff
@@ -0,0 +1,507 @@
...
+    if h_equiv > a1_2:
+        std_err = np.sqrt(c1 * a1 ** 2 / (5 * h_equiv ** 2) + c1_2 * a1_2 ** 2 / (5 * h_equiv ** 2))
+    elif (h_equiv < a1_2) and (h_equiv > a1):
+        std_err = np.sqrt(c1 * a1 ** 2 / (5 * h_equiv ** 2) + c1_2 * (1-h_equiv / a1_2+1 / 5 * (h_equiv / a1_2) ** 3))
+    else:
+        std_err = np.sqrt(c1 * (1-h_equiv / a1+1 / 5 * (h_equiv / a1) ** 3) + c1_2 * (1-h_equiv / a1_2+1 / 5 * (h_equiv / a1_2) ** 3))
+
+    return (psill1 + psill2)/std_err**2
+
+def neff_circ(area: float,list_vgm: list) -> float:
```
erikmannerfelt, line 237: Maybe `list` could be more specific? E.g. `list[Union[float, int]]`. Or wait, what exactly is the content of this list??

xdem/spstats.py
```diff
@@ -0,0 +1,507 @@
...
+    :param fun: function
+    :param low_b: lower bound
+    :param upp_b: upper bound
+
+    :return: integral
+    """
+
+    return integrate.quad(fun,low_b,upp_b)[0]
+
+def cov(h: float,crange: float,model: str='Sph',psill: float=1.,kappa: float=1/2,nugget: float=0) -> function:
```
erikmannerfelt, line 325: Here, the type annotation should be `-> Callable`

xdem/spstats.py
```diff
@@ -0,0 +1,507 @@
...
+
+    :param std: standard deviation
+    :param Neff: maximum number of effective samples
+    :param neff: number of effective samples
+
+    :return: standard error
+    """
+    return std * np.sqrt(1 / Neff * (Neff - neff) / Neff)
+
+def std_err(std: float, Neff: float) -> float:
```
erikmannerfelt, line 387: Why is `Neff` capitalized? Looks like it is a class.

erikmannerfelt (commented):
Nice!! I have some comments, but most are pretty small.

xdem/spstats.py
rhugonnet: this is the syntax that PyCharm pops out automatically when I start a function description!

xdem/spstats.py
rhugonnet: The issue with scikit-gstat was an unexpected `ZeroDivisionError` in recent versions (hopefully resolved by now).
Added this.

xdem/spstats.py
rhugonnet: Nice tip! Love learning those =D

xdem/spstats.py
rhugonnet: Added

xdem/spstats.py
rhugonnet: Thanks! I couldn't remember the syntax to show the details of what should be inside the list

xdem/spstats.py
rhugonnet: Done.

xdem/spstats.py
erikmannerfelt: `list[subtype]` is a default syntax since python=3.9, but can be used in previous versions by including `from __future__ import annotations`. 

I've found that mypy (a static type checker) doesn't understand this in the latest release! Therefore, I install mypy from source (`pip install git+https://github.com/python/mypy.git`), which has this addition. I don't know from where PyCharm gets its type checks, but maybe it's a setting somewhere? 

Either way, I would consider that behaviour a bug, since your syntax is perfectly valid since maybe python=3.5 or something.

tests/test_spstats.py
```diff
@@ -0,0 +1,49 @@
...
+
+    reference_raster = gu.georaster.Raster(examples.FILEPATHS["longyearbyen_ref_dem"])
+    to_be_aligned_raster = gu.georaster.Raster(examples.FILEPATHS["longyearbyen_tba_dem"])
+    glacier_mask = gu.geovector.Vector(examples.FILEPATHS["longyearbyen_glacier_outlines"])
+
+    metadata = {}
+    aligned_raster, _ = xdem.coreg.coregister(reference_raster, to_be_aligned_raster, method="amaury", mask=glacier_mask,
+                                     metadata=metadata)
+
+    diff = xdem.spatial_tools.subtract_rasters(reference_raster,aligned_raster)
```
erikmannerfelt, line 24: This is the most correct way to do it, but for speed, I suggest adding `resampling_method="nearest"` as a keyword argument. These specific DEMs are already in the same grid, so it will speed up the test slightly!

xdem/spstats.py
```diff
@@ -0,0 +1,702 @@
...
+    """
+    Distance between two lat/lon coordinates projected on a spheroid
+    :param tup1: lon/lat coordinates of first point
+    :param tup2: lon/lat coordinates of second point
+
+    :return: distance
+    """
+
+    # approximate radius of earth in km
+    R = 6373000
```
erikmannerfelt, line 497: Does this hardcoded number change the results significantly?

Either way, maybe it could be nice with a reference? A quick wikipedia-ing showed that it for example deviates from the WGS84 ellipsoid.

xdem/spstats.py
```diff
@@ -0,0 +1,702 @@
...
+    dlat = lat2 - lat1
+
+    a = m.sin(dlat / 2)**2 + m.cos(lat1) * m.cos(lat2) * m.sin(dlon / 2)**2
+    c = 2 * m.atan2(m.sqrt(a), m.sqrt(1 - a))
+
+    distance = R * c
+
+    return distance
+
+def kernel_sph(xi: float, x0: float, a1: float) -> float:
```
erikmannerfelt, line 514: This is just an idea, but could `@np.vectorize` be used to speed up some calculations significantly?

The decorator allows for arrays to be passed to the function directly and does element-wise operations in near-C speed.

Maybe we can try it in a future PR!

erikmannerfelt (commented):
Nice!

Another suggestion: It would be really nice to have an example in the README for variogram calculations with default settings. Could this be a worthy addition?

tests/test_spstats.py
adehecq: If the two DEMs are on the same grid, there should be no interpolation then and the type of interpolation should not matter. Is there a check for this in substract_rasters?

xdem/spstats.py
adehecq: This should at least be an argument with a default value.

adehecq:
> Nice!
> 
> Another suggestion: It would be really nice to have an example in the README for variogram calculations with default settings. Could this be a worthy addition?

Agreed! There is a lot of things in this PR so it's difficult to check it all, but it would be nice to be able to test the different features. Anyway, at some point we should definitely write some documentation/how-to.

xdem/spstats.py
```diff
@@ -0,0 +1,702 @@
...
+    :param nproc: number of processing cores
+
+    :return: empirical variogram (variance, lags, counts)
+    """
+
+    # checks
+    dh = dh.squeeze()
+    if coords is None and gsd is None:
+        raise TypeError('Must provide either coordinates or ground sampling distance.')
+    elif gsd is not None and len(dh.shape) == 1:
```
adehecq, line 109: use dh.ndim rather than len(dh.shape)

xdem/spstats.py
```diff
@@ -0,0 +1,702 @@
...
+                                          nsamp: int = 10000, range_list: list= None, nrun: int=1, nproc: int=1,
+                                          **kwargs) -> pd.DataFrame:
+
+    """
+    Wrapper to sample multi-range empirical variograms from the data.
+    If no option is passed, a varying binning is used with adapted ranges and data subsampling
+
+    :param dh: elevation differences
+    :param gsd: ground sampling distance (if array is 2D on structured grid)
+    :param coords: coordinates (if array is 1D)
```
adehecq, line 96: You should clarify in the docstring what is teh expected shape of coords,

xdem/spstats.py
```diff
@@ -0,0 +1,702 @@
...
+        kwargs.update({'bin_func':'even'})
+    if 'n_lags' not in kwargs.keys():
+        kwargs.update({'n_lags':100})
+
+    # estimate variogram
+    if nrun == 1:
+        # subsetting
+        dh_sub, coords_sub = random_subset(dh, coords, nsamp)
+        # getting empirical variogram
+        print(dh_sub.shape)
```
adehecq, line 162: The print statement should have some text to explain what is printed.

xdem/spstats.py
rhugonnet: Added the reference + argument as default

xdem/spstats.py
rhugonnet: Debug print, removed.

xdem/spstats.py
rhugonnet: Done

tests/test_spstats.py
erikmannerfelt: @adehecq that is not checked but maybe it should be. I just call `Raster.reproject(...., resampling=resampling_method)`.

I experimented a bit with geoutils:
```python

import time

import geoutils as gu
import rasterio as rio

# I use the KH-9 DEM you sent me which is quite big
dem = gu.georaster.Raster("/home/erik/Downloads/DZB1208-500032L004006/DZB1208-500032L004006-DEM.tif")

# I reproject the DEM to its own exact bounds using cubic spline
# This is a proxy for if a DEM with the exact same bounds as a reference DEM should (not) be reprojected.
time_before = time.time()
dem2 = dem.reproject(dem, resampling=rio.warp.Resampling.cubic_spline)
time_after = time.time()
print(f"{time_after - time_before:.2f} s")

```
Which returned: 
```
4.98 s
```
I would argue that this is a limitation in either `georaster` or `rasterio`. Why would resampling be done if the source and destination bounds/crs/resolution is the same??

I see two approaches to circumvent this: either rewrite the `subtract_rasters()` function a little, post an associated issue on `geoutils`, or both. What do you think?

rhugonnet: > > Nice!
> > Another suggestion: It would be really nice to have an example in the README for variogram calculations with default settings. Could this be a worthy addition?
> 
> Agreed! There is a lot of things in this PR so it's difficult to check it all, but it would be nice to be able to test the different features. Anyway, at some point we should definitely write some documentation/how-to.

Yes, this will come when all is done! For now, trying to make everything consistent, working and finish the related tests :)

xdem/spstats.py
erikmannerfelt: @rhugonnet I can't see that addition. Is it in a commit that is not yet pushed?

tests/test_spstats.py
rhugonnet: > @adehecq that is not checked but maybe it should be. I just call `Raster.reproject(...., resampling=resampling_method)`.
> 
> I experimented a bit with geoutils:
> 
> ```python
> import time
> 
> import geoutils as gu
> import rasterio as rio
> 
> # I use the KH-9 DEM you sent me which is quite big
> dem = gu.georaster.Raster("/home/erik/Downloads/DZB1208-500032L004006/DZB1208-500032L004006-DEM.tif")
> 
> # I reproject the DEM to its own exact bounds using cubic spline
> # This is a proxy for if a DEM with the exact same bounds as a reference DEM should (not) be reprojected.
> time_before = time.time()
> dem2 = dem.reproject(dem, resampling=rio.warp.Resampling.cubic_spline)
> time_after = time.time()
> print(f"{time_after - time_before:.2f} s")
> ```
> 
> Which returned:
> 
> ```
> 4.98 s
> ```
> 
> I would argue that this is a limitation in either `georaster` or `rasterio`. Why would resampling be done if the source and destination bounds/crs/resolution is the same??
> 
> I see two approaches to circumvent this: either rewrite the `subtract_rasters()` function a little, post an associated issue on `geoutils`, or both. What do you think?

I think this should be raised as an issue in geoutils

erikmannerfelt: @rhugonnet [I added an issue in geoutils](https://github.com/GlacioHack/GeoUtils/issues/102)!

adehecq: Great, that's what I was going to suggest!

xdem/spstats.py
rhugonnet: Indeed everything runs perfectly with `from __future__ import annotations`.
I'll wait for PyCharm to correct the syntax correction bug. Thanks!

xdem/spstats.py
rhugonnet: Added issue #40

PR #42, opened 2021-03-11
DEMCollection and dDEM classes

erikmannerfelt:
Two convenience classes; `dDEM` and `tDEM` are added.

### dDEM
A subclass of `geoutils.georaster.Raster`.
Contains `start_time`, `end_time` and `error` attributes, as well as a couple of new functions.

### tDEM
A **t**emporal **DEM** collection.
This class is primarily used to generate and handle `dDEM`s

### Example
```python
import xdem
import geoutils as gu
import datetime

# Load some example data
dem_2009 = gu.georaster.Raster(xdem.examples.FILEPATHS["longyearbyen_ref_dem"])
dem_1990 = gu.georaster.Raster(xdem.examples.FILEPATHS["longyearbyen_tba_dem"])
glaciers_1990 = gu.geovector.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines"])

# Extract the Scott Turnerbreen glacier as an example and make a boolean mask from it.
scott_1990 = gu.geovector.Vector(glaciers_1990.ds.loc[glaciers_1990.ds["NAME"] == "Scott Turnerbreen"])
mask = scott_1990.create_mask(dem_2009) == 255).reshape(dem_2009).data.shape

# Fake a 2060 DEM by making the glacier 30 m smaller.
dem_2060 = dem_2009.copy()
dem_2060[mask] -= 30

# Create the temporal DEM collection with appropriate timestamps.
tdem = xdem.volume.tDEM(
    dems=[dem_1990, dem_2009, dem_2060],
    timestamps=[datetime.datetime(year, 8, 1) for year in [1990, 2009, 2060]],
    reference_dem=1  # This can be either an index of the 'dems' list or the instance of the DEM.
)

# Create the dDEMs. These are both returned, and saved in the class (`self.ddems`)
tdem.subtract_dems(resampling_method="nearest")

# Calculate the cumulative dH within the Scott Turnerbreen mask
cumulative_dh = tdem.get_cumulative_dh(mask=mask)

print(cumulative_dh)

```
prints:
```
1990-08-01     0.000000
2009-08-01   -13.721097
2060-08-01   -57.442195
dtype: float64
```
### Steps ahead
1. Add functionality for temporally varying glacier outlines
2. Add error calculations
3. Add DEM/dDEM interpolation shortcuts.

```

What do you think about this so far?

### CODE REVIEW

File: xdem/volume.py [hunk 1]
```diff
@@ -0,0 +1,194 @@
+from __future__ import annotations
```
rhugonnet, line 1: I would move Classes to different python files than functions, for readability (otherwise we'll get over 2000 lines for sure). Maybe tdem.py and ddem.py

erikmannerfelt: I can move them! @adehecq what do you think of `tdem.py` and `ddem.py`? I'm okay with that. Or maybe a submodule for this? 

```
xdem/submodule/tdem.py
xdem/submodule/ddem.py
```

xdem/volume.py
```diff
@@ -0,0 +1,194 @@
...
+            start_time=start_time,
+            end_time=end_time,
+            error=error,
+        )
+
+
+class tDEM:
+    """A temporal collection of DEMs."""
+
+    def __init__(self, dems: list[gu.georaster.Raster], timestamps: list[datetime.datetime],
```
rhugonnet, line 73: Preferably a list of xdem.dem.DEM objects. Basically the same object, but possible timestamp/metadata/vref attached

xdem/__init__.py
```diff
@@ -1,5 +1 @@
-from . import coreg
-from . import spatial_tools
-from . import dem
-from . import examples
-from . import spstats
\ No newline at end of file
+from . import coreg, dem, examples, spatial_tools, volume, spstats
```
rhugonnet, line 1: good point here haha

xdem/volume.py
erikmannerfelt: Totally agree and I just completely forgot that. I'll fix it. I still think it would be nice to accept Raster objects, but then the `timestamps` argument is needed.

xdem/__init__.py
erikmannerfelt: I just let my formatters go nuts! Apparently this is "pep8"-y

xdem/volume.py
erikmannerfelt: Wait I misunderstood. The DEM class doesn't have a timestamp attached right now. Scratch the optionality of the `timestamps` argument.

rhugonnet:
Great work, I think the idea and structure is there. In particular the Class methods are great!

For the tDEM and dDEM classes themselves - everything related to __init__() - I suggest we move all to a composition primarily based on `xarray` for the following reasons:
- xarray has a great handling of datetime (np.datetime64) through similar abilites as pandas but in N-dimensions.
- xarray will hold all variables with the proper referenced dimensions (e.g., a stack of 10 DEMs of size (x=100,y=100,t=10) accessible through ds.h and a reference DEM of size (x=100,y=100) accessible through ds.ref in the same Dataset object.
- the time dimension (and other attributes, such as vertical referencing) will be properly linked to each slice of the objects.
- we can pass down the crs to the xarray dataset, and this will save to netCDF to visualize/share the georeferenced time stack/dDEM stack directly.
- we can easily implement chunking/parallel computing on the xarray.Dataset through Dask
- the list goes on... ;D

xdem/volume.py
rhugonnet: The DEM class inherits from the SatImg class which has a .datetime attribute! See my general comment on passing the datetimes directly to xarray.

erikmannerfelt: @rhugonnet I changed the default tDEM collection to be of `DEM` classes in 4538bee. It still takes `Raster`s, but will convert them in the `__init__()`

I just saw your datetime comment. I'll experiment with that now.

erikmannerfelt:
@rhugonnet I have heard lots of people talk about the advantages of `xarray` (or is it a few people talking lots about it?? Often hard to tell the difference!). I've never touched it myself.

I like the idea of keeping dependencies low, and this would be yet another framework for people to learn. I am not completely against it though. @adehecq what do you think?

xdem/volume.py
erikmannerfelt: @rhugonnet I added `.datetime` attribute parsing in df80f81.

The tDEM instantiation is now extremely simple, assuming the datetime attribute is set:
```python

tdem = tDEM([dem_2009, dem_1990])  # The first DEM in the list is parsed as the reference

```

rhugonnet: Those classes are gonna be so neat =D !

erikmannerfelt:
> I'm curious: when do you use `Optional[]` in annotations? Doesn't the `= None` already fulfills that message?

`Optional[Type]` is a shortcut for `Union[None, Type]`. In the same way that `def func(value: int = 1.1)` is technically incorrect, so is `def func(value: float = None)`. I understand the argument that python should be able to understand the latter context and convert it implicitly to `def func(value: Optional[float] = None)`, but I'm sure there's a good reason why (one that I don't know!).

The current approach is similar to other languages, for example rust:
```rust
fn func(value: f32, optional_value: Option<f32>) -> f32 {
    // do stuff
}

func(1.0, None)
func(1.0, 1.1)
```
(in rust, arguments can't have default values, so you have to specify `None` explicitly. This is to reduce unexpected behaviour I think)

erikmannerfelt:
I'm looking forward to hear from you what you think of this! I've started on interpolation functionality for dDEMs. Right now, I only have one linear spatial interpolation using `scipy.interpolate.griddata()` which is really slow, but it serves as a proof of concept. @adehecq , as soon as you have the interpolation stuff ready, we can implement it here!

```python
# Load everything like the example in the top

# Create the dDEMs. These are both returned, and saved in the class (`self.ddems`)
tdem.subtract_dems(resampling_method="nearest")

# Since the dDEM has no gaps, the filled_data is equal to the data attribute
assert tdem.ddems[0].filled_data is tdem.ddems[0].data

# Generate 10000 NaN values randomly in the dDEM                                                      
tdem.ddems[0].data[np.random.randint(0, tdem.ddems[0].data.shape[0], 100),                                      
                   np.random.randint(0, tdem.ddems[0].data.shape[1], 100)] = np.nan

# Now the dDEM has gaps, but the _filled_data attribute is empty, so this is now None
assert tdem.ddems[0].filled_data is None

# This will in the future be done concurrently!
tdem.interpolate_ddems(method="linear")

# Now the dDEM data attribute has NaNs, but the filled_data is available
assert tdem.ddems[0].filled_data is not None and np.count_nonzero(np.isnan(tdem.ddems[0])) > 0

# Calculate the cumulative dH within the Scott Turnerbreen mask
# This will warn if the analysis is done on the unfilled data (nans_ok=False; if filled_data is None)
cumulative_dh = tdem.get_cumulative_dh(mask=mask, nans_ok=False)
```

So the most important part I guess is that the `dDEM` class (which is now a subclass of `DEM`, not `Raster`) has a normal `.data` attribute, which may or may not have NaNs, and a `.filled_data` which will never have NaNs (assuming it exists).

Thoughts about this?

adehecq:
> I'm looking forward to hear from you what you think of this! I've started on interpolation functionality for dDEMs. Right now, I only have one linear spatial interpolation using `scipy.interpolate.griddata()` which is really slow, but it serves as a proof of concept. @adehecq , as soon as you have the interpolation stuff ready, we can implement it here!
> 
> ```python
> # Load everything like the example in the top
> 
> # Create the dDEMs. These are both returned, and saved in the class (`self.ddems`)
> tdem.subtract_dems(resampling_method="nearest")
> 
> # Since the dDEM has no gaps, the filled_data is equal to the data attribute
> assert tdem.ddems[0].filled_data is tdem.ddems[0].data
> 
> # Generate 10000 NaN values randomly in the dDEM                                                      
> tdem.ddems[0].data[np.random.randint(0, tdem.ddems[0].data.shape[0], 100),                                      
>                    np.random.randint(0, tdem.ddems[0].data.shape[1], 100)] = np.nan
> 
> # Now the dDEM has gaps, but the _filled_data attribute is empty, so this is now None
> assert tdem.ddems[0].filled_data is None
> 
> # This will in the future be done concurrently!
> tdem.interpolate_ddems(method="linear")
> 
> # Now the dDEM data attribute has NaNs, but the filled_data is available
> assert tdem.ddems[0].filled_data is not None and np.count_nonzero(np.isnan(tdem.ddems[0])) > 0
> 
> # Calculate the cumulative dH within the Scott Turnerbreen mask
> # This will warn if the analysis is done on the unfilled data (nans_ok=False; if filled_data is None)
> cumulative_dh = tdem.get_cumulative_dh(mask=mask, nans_ok=False)
> ```
> 
> So the most important part I guess is that the `dDEM` class (which is now a subclass of `DEM`, not `Raster`) has a normal `.data` attribute, which may or may not have NaNs, and a `.filled_data` which will never have NaNs (assuming it exists).
> 
> Thoughts about this?

I will add some interpolation functionalities yes, but I won't have time to do this today I believe... But Erik, if you have gap-filling/resampling scripts somewhere, please put them in the google drive so I can have a look.
But in the meantime, you could also just use rasterio's fill: https://rasterio.readthedocs.io/en/latest/api/rasterio.fill.html

Regarding the data vs filled_data, it seems like a good idea. But how will you treat the case where there is no NaN in data? Will you duplicate the data, or just have filled_data point to data?

erikmannerfelt:
> I will add some interpolation functionalities yes, but I won't have time to do this today I believe... But Erik, if you have gap-filling/resampling scripts somewhere, please put them in the google drive so I can have a look.
> But in the meantime, you could also just use rasterio's fill: https://rasterio.readthedocs.io/en/latest/api/rasterio.fill.html
> 
> Regarding the data vs filled_data, it seems like a good idea. But how will you treat the case where there is no NaN in data? Will you duplicate the data, or just have filled_data point to data?

Ah, I had forgotten about rasterio fill. I'll have a look when I next have time (probably not today).

`filled_data` is `data` if `data` contains no NaNs.

erikmannerfelt:
Wow, that was an easy fix... @adehecq, I replaced the `scipy.interpolate.griddata()` with `rasterio.fill.fillnodata()` in 1c4ada6 and it works like 10 times faster.

erikmannerfelt:
I've now added volume change and temporally variable (glacier) outlines support!

```python
import datetime

import geoutils as gu
import xdem

# Read 1990 and 2009 (ish) DEMs and glacier outlines.
dem_2009 = xdem.dem.DEM(xdem.examples.FILEPATHS["longyearbyen_ref_dem"])                                            
dem_1990 = xdem.dem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"])                                            
outlines_1990 = gu.geovector.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines"])                       
outlines_2010 = gu.geovector.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines_2010"])


# Fake a 2060 DEM
mask_2010 = (scott_2010.create_mask(self.dem_2009) == 255).reshape(self.dem_2009.data.shape)                                                                                                                                         
dem_2060 = self.dem_2009.copy()                                                                                 
dem_2060.data[mask_2010] -= 30

# This can also be parsed automatically if DEM.datetime exists on all given DEMs
timestamps = [datetime.datetime(1990, 8, 1), datetime.datetime(2009, 8, 1), datetime.datetime(2060, 8, 1)]

# Extract the outlines for Scott Turnerbreen                                                                                                             
scott_1990 = gu.geovector.Vector(outlines_1990.ds.loc[outlines_1990.ds["NAME"] == "Scott Turnerbreen"])                                                                                                               
scott_2010 = gu.geovector.Vector(outlines_2010.ds.loc[outlines_2010.ds["NAME"] == "Scott Turnerbreen"])


# Create a tDEM object with the DEMs and outlines for 1990 and 2010
tdem = xdem.volume.tDEM(                                                                                        
              [self.dem_1990, self.dem_2009, dem_2060],                                                                   
              timestamps=timestamps,                                                                                      
              outlines=dict(zip(timestamps[:2], [scott_1990, scott_2010])),                                               
              reference_dem=1                                                                                             
)

# Create and interpolate dDEMs
tdem.subtract_dems(resampling_method="nearest")
tdem.interpolate_ddems(method="linear")

# Calculate cumulative dH and dV series
# The values are masked with the outlines as described by the self.get_ddem_mask() method.
# In short; if outlines exist from the start and end time of a dDEM, the union of the two masks are used.
# If only outlines from the start year exists, it will use that.
# If only one outline was provided, it will always use that as a mask
# Otherwise, it will just use the whole dDEM
# A custom mask can always be provided with the mask= keyword argument. Then, outlines are ignored.
cumulative_dh = tdem.get_cumulative_series(kind="dh")                                                           
cumulative_dv = tdem.get_cumulative_series(kind="dv")

print(cumulative_dh)
print(cumulative_dv)

```
prints the following:

```python
1990-08-01     0.000000
2009-08-01   -13.379259
2060-08-01   -43.379259       
dtype: float64

1990-08-01    0.000000e+00
2009-08-01   -5.676017e+07
2060-08-01   -1.543202e+08
dtype: float64
```

What do you think?

xdem/volume.py
adehecq: Why not define ddem and tdem in the same file as DEM? So I guess xdem?

adehecq:
Looks promising, but I have several questions:
- for now I don't see a lot of interest in the dDEM class, if one simply provides it with a single raster. Why not provide two rasters (DEMs) and have methods such as difference, coregister, calculate volume etc
- how does tDEM scale with size? Loading into memory each DEM can be impossible in some cases, and here we basically double the amount of data with the differences
- it seems like tDEM would also not apply well for Romain's work, for example if you have a series of DEMs and want to fit an elevation trend or so
My idea of tDEM was that, upon initialization, it would reproject all DEMs to a common grid and save the data cube into an xarray with chunks. This would allow to perform operations such as differencing or volume change calculation on a large volume with chunks.

erikmannerfelt:
> Looks promising, but I have several questions:
> 
>     * for now I don't see a lot of interest in the dDEM class, if one simply provides it with a single raster. Why not provide two rasters (DEMs) and have methods such as difference, coregister, calculate volume etc
> 
>     * how does tDEM scale with size? Loading into memory each DEM can be impossible in some cases, and here we basically double the amount of data with the differences
> 
>     * it seems like tDEM would also not apply well for Romain's work, for example if you have a series of DEMs and want to fit an elevation trend or so
>       My idea of tDEM was that, upon initialization, it would reproject all DEMs to a common grid and save the data cube into an xarray with chunks. This would allow to perform operations such as differencing or volume change calculation on a large volume with chunks.

* The dDEM class handles interpolation and will handle error assessment. I don't think that counts as uninteresting. I would argue that coregistration has nothing to do with DEM subtraction (although dDEMs are of course sometimes an intermediate step in some methods), so this should be independent on the dDEM class.
* I tried loading 10 instances of the same KH-9 DEM you sent me, and the memory usage from start to finish was about 12 GiB. @rhugonnet and I discussed that it would be nice to have the core functionality in place before worrying too much about the performance part. Of course, these go slightly hand in hand, as we shouldn't create something that cannot be optimised later (if it needs to be optimised), so I am open to approaches that would reduce this. One very easy optimisation could be to use float32 values in the dDEMs  as default instead of the current float64's.
* Maybe @rhugonnet can answer this? How much memory is too much? Reprojecting DEMs to fit each other perfectly and then merging them into a stack sounds great. This has to be after coregistration, however, as the DEMs would otherwise be projected twice (which I would argue is risky because it will modify the data unnecessarily). My aim with this tDEM object was to make it as simple as possible to coregister, subtract, and assess DEMs. Your approach sounds great for the two second points, but I struggle to see where the first comes in.

Perhaps we should think of renaming these things? The "tDEM" object I create can be renamed to a `DEMCollection`, which handles coregistration and reprojection. The class can have a method called `DEMCollection.create_tdem()`, which converts it to @adehecq's "tDEM" version.

I guess this is the risk one faces by developing a tool with no specific use-case in mind: different people's needs complicates the others' needs. My use case is comparing, say, up to five DEMs with extents of less than or equal to 10000x10000 pixels, for which my approach would fit perfectly. I think the most relevant question we have to ask ourselves is: what will others require? Will most others require blazing fast 100000x100000x2000 temporal DEM comparisons, or will most just need to subtract two differently projected DEMs?

I think we need a thorough discussion on this moving forward. I won't do anything else on this until we have a plan. I've already spent a week on this so I don't want to spend more time only to have it completely revised.

adehecq:
>
>     * The dDEM class handles interpolation and will handle error assessment. I don't think that counts as uninteresting. I would argue that coregistration has nothing to do with DEM subtraction (although dDEMs are of course sometimes an intermediate step in some methods), so this should be independent on the dDEM class.
> 
I was not saying that it does not have an interest per se, but I do not see a lot of benefit of having it as a separate class of Raster or DEM. We could have an interpolation method in Raster that would return a filled version of the DEM. My point was that I would find it more useful if one would provide two rasters, rather than a wrapper around a single raster, which already exists with Raster/DEM. If we provide both DEM rasters, then we can include methods such as coregistration (which would just be a wrapper around coreg functions) and subtract. That makes it a lot more interesting as a separate class.

>     * I tried loading 10 instances of the same KH-9 DEM you sent me, and the memory usage from start to finish was about 12 GiB. @rhugonnet and I discussed that it would be nice to have the core functionality in place before worrying too much about the performance part. Of course, these go slightly hand in hand, as we shouldn't create something that cannot be optimised later (if it needs to be optimised), so I am open to approaches that would reduce this. One very easy optimisation could be to use float32 values in the dDEMs  as default instead of the current float64's.
> 
Good to know! And yes, I think the default should be float32. I've never encountered situations where float64 was needed for this kind of calculation...

>     * Maybe @rhugonnet can answer this? How much memory is too much? Reprojecting DEMs to fit each other perfectly and then merging them into a stack sounds great. This has to be after coregistration, however, as the DEMs would otherwise be projected twice (which I would argue is risky because it will modify the data unnecessarily). My aim with this tDEM object was to make it as simple as possible to coregister, subtract, and assess DEMs. Your approach sounds great for the two second points, but I struggle to see where the first comes in.
> 
Yes, I thought that the idea of tDEM (or DEMCollection) was to accept any DEM, make sure that all DEMs would be co-registered to each other and resampled to a common grid, to ease further operations. Of course, saving to an xarray could be an option for large datasets only. In this sense, I do not think the current framework would need a lot of re-writing, it would mostly be about having an additional method to create the DEM stack and save it as xarray.
> 
> Perhaps we should think of renaming these things? The "tDEM" object I create can be renamed to a `DEMCollection`, which handles coregistration and reprojection. The class can have a method called `DEMCollection.create_tdem()`, which converts it to @adehecq's "tDEM" version.
> 
> I guess this is the risk one faces by developing a tool with no specific use-case in mind: different people's needs complicates the others' needs. My use case is comparing, say, up to five DEMs with extents of less than or equal to 10000x10000 pixels, for which my approach would fit perfectly. I think the most relevant question we have to ask ourselves is: what will others require? Will most others require blazing fast 100000x100000x2000 temporal DEM comparisons, or will most just need to subtract two differently projected DEMs?
> 
> I think we need a thorough discussion on this moving forward. I won't do anything else on this until we have a plan. I've already spent a week on this so I don't want to spend more time only to have it completely revised.

Yes, we can talk about it tomorrow, since we're supposed to meet already. Sorry if that felt like huge criticism, but it was not. I think you could more or less merge that PR and we can move from there.

rhugonnet:
> I think we need a thorough discussion on this moving forward. I won't do anything else on this until we have a plan. I've already spent a week on this so I don't want to spend more time only to have it completely revised.

Yes, sorry I'm not more active this week! I have other things to tend to...

I agree we should press pause on these aspects. As I said last Friday, I will get back to you two on a structure + its relation to existing tools, as I am most familiar with those. After working nearly 3 years on these things (either DEM time series, sequential or simple DEM differences), I have a good idea of how to put this together. This will make things easier. I should have more time early next week.

I think it's great progress and we'll definitely be able to use the work you've being doing moving forward :+1:

erikmannerfelt:
Thanks for the replies, @adehecq and @rhugonnet.

I am relieved to hear we seem to be more or less on the same page, and I appreciate the clarifications above! Let's indeed discuss these things in person. I'm having three talks within 4 work-days so I'm quite stressed, which is maybe why I took this a bit too personal, even though I know I shouldn't!

rhugonnet:
Good luck with the talks!!

erikmannerfelt:
I now changed the `tDEM` name to `DEMCollection` and moved it and `dDEM` to `xdem/dem.py`.

I imported all the `dem.py` classes in the `__init__.py` to simplify the syntax a bit:
```python
import xdem

# Previously: (note: this still works)
dem = xdem.dem.DEM("path/to/file")

# Now:
dem = xdem.DEM("path/to/file")
dems = xdem.DEMCollection(list_of_dems)
```
Do you agree on this latest change, @rhugonnet  and @adehecq ?

Also, @rhugonnet, I couldn't get any of your tests to pass using the latest main. Do they pass for you?

adehecq:
> I now changed the `tDEM` name to `DEMCollection` and moved it and `dDEM` to `xdem/dem.py`.
> 
> I imported all the `dem.py` classes in the `__init__.py` to simplify the syntax a bit:
> 
> ```python
> import xdem
> 
> # Previously: (note: this still works)
> dem = xdem.dem.DEM("path/to/file")
> 
> # Now:
> dem = xdem.DEM("path/to/file")
> dems = xdem.DEMCollection(list_of_dems)
> ```
> 
> Do you agree on this latest change, @rhugonnet and @adehecq ?
> 

I'm happy with the new names and setup!

erikmannerfelt:
> 
> I'm happy with the new names and setup!

Nice! Maybe I can merge once I have an okay from @rhugonnet and clarification with the tests?

erikmannerfelt:
In 0154255, I added something quite exciting. Pandas/GeoPandas supports queries made within them, so Vectors support that as well by this slightly hacky approach:
```python
all_outlines = gu.Vector("...")

scott_outline = all_outlines.copy()
scott_outline.ds = scott_outline.ds.query("NAME == 'Scott Turnerbreen'")
```
The syntax here refers to the column called `NAME` which must be equal to the string `'Scott Turnerbreen'`. Maybe I'll add `Vector.query()` as a PR for geoutils. What do you think, @adehecq and @rhugonnet?

These queries are now implemented in the `DEMCollection object`:
```python
In [1]: import datetime

In [2]: import geoutils as gu

In [3]: import xdem

# Read 1990 and 2009 (ish) DEMs and glacier outlines.
In [4]: dem_2009 = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_ref_dem"], datetime=datetime.datetime(2009, 8, 1))

In [5]: dem_1990 = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"], datetime=datetime.datetime(1990, 8, 1))

In [6]: outlines = {
    datetime.datetime(1990, 8, 1): gu.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines"]),
    datetime.datetime(2009, 8, 1): gu.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines_2010"]),
}


In [7]: dems = xdem.DEMCollection(dems=[dem_2009, dem_1990], outlines=outlines)

In [8]: dems.subtract_dems()

In [9]: dems.interpolate_ddems()


# Get dV for Scott Turnerbreen
In [10]: dems.get_cumulative_series(kind="dv", outlines_filter="NAME == 'Scott Turnerbreen'")
Out [10]:
1990-08-01    0.000000e+00
2009-08-01    -5.676017e+07
dtype: float64

# Get dV for the neighbouring Ayerbreen
In [11]: dems.get_cumulative_series(kind="dv", outlines_filter="NAME == 'Ayerbreen'")
Out [11]:
1990-08-01    0.000000e+00
2009-08-01    -3.909687e+07
dtype: float64
```
This addition means that dDEMs are only calculated once for an entire area, and then one can cherry-pick the glaciers to look at. The glacier-specific mask can for any purpose be extracted by:
```python
dems.get_ddem_mask(dems.ddems[0], outlines_filter="whatever filter")  # This returns a numpy ndarray
```

rhugonnet (approved):
Read through everything again, it looks good! :)

rhugonnet:
I'm still a bit afraid the dem.py will end up being ~5000 lines long if we keep all in the same place, I think a `ddem.py` and "demcollection.py" would make more sense, but we can worry about this later ;)

rhugonnet:
@erikmannerfelt On the tests, all is passing on my side:
`============================ 18 passed, 32 warnings in 507.15s (0:08:27) ============================`
Some are taking too long (10 min for testing is too much), we'll have to work on quicker tests.

erikmannerfelt:
I agree with what you say @rhugonnet, and I've now moved the dDEM and DEMCollection objects to two separate modules! I'll merge now.

Issue #43, opened 2021-03-11
DEM should be possible to create from a Raster object


erikmannerfelt:
I could submit a PR, but my aggressive formatting would mess up your code, @rhugonnet  ;)

If the user has a `Raster` and wants to convert this to a `DEM`, there is currently no easy way to do it. One hacky way is:
```python
dem = xdem.dem.DEM.from_array(raster.data, raster.transform, raster.crs, raster.nodata)
```
A quite simple conversion could be done in a `.from_raster()` static method. I recently learned this trick, and it's great because no copying is done in memory whatsoever!:
```python
class DEM(...):
    # all the old stuff
    def from_raster(raster: gu.georaster.Raster, vref_name=None, vref_grid=None, silent=False):
        dem = DEM.__new__(DEM)
        dem.__dict__ = raster.__dict__

        # user input
        dem.vref = vref_name
        dem.vref_grid = vref_grid
        dem.ccrs = None

        # trying to get vref from product name (priority to user input)
        # dem.__parse_vref_from_fn(silent=silent)
        return dem
```
This kind of works for me, but I just now realised that it won't work perfectly because it's supposed to be initialised through a `SatelliteImage`, so it doesn't have all the right attributes.

Anyway, this would be a really nice addition!

rhugonnet:
Right now, I think this works:
```python
import geoutils as gu
import xdem
fn_img = xdem.examples.FILEPATHS["longyearbyen_ref_dem"]
r = gu.georaster.Raster(fn_img)
dem = xdem.dem.DEM(r.filename)
```
Which yields:
```
No metadata could be read from filename.
type(dem)
Out[7]: xdem.dem.DEM
```
It goes through the "init" again, but this is not an issue as those run instantly (manipulating strings). If we had to migrate some more problematic, larger objects, we'd have to use a specific function.
However we would have a problem if `self.filename` does not exist (a Raster copied in memory for instance). Then everything would fail. I guess it would be useful to store the original filename in `georaster.Raster`, maybe under `._filename`?

erikmannerfelt:
> Right now, I think this works:
> 
> ```python
> import geoutils as gu
> import xdem
> fn_img = xdem.examples.FILEPATHS["longyearbyen_ref_dem"]
> r = gu.georaster.Raster(fn_img)
> dem = xdem.dem.DEM(r.filename)
> ```
> 
> Which yields:
> 
> ```
> No metadata could be read from filename.
> type(dem)
[...]

PR #46, opened 2021-03-11
Add copy method for DEM class (overrides SatImg/Raster copy)

### LINKED ISSUE #44: DEM.copy() returns a Raster instance
The `.copy()` method, which is inherited by the `Raster` class, does not work as expected on the DEM class:

```python
In [1]: import xdem

In [2]: dem = xdem.dem.DEM("examples/Longyearbyen/data/DEM_2009_ref.tif")

In [3]: dem_copy = dem.copy()

In [4]: assert isinstance(dem_copy, xdem.dem.DEM)
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-4-54169263df41> in <module>
----> 1 assert isinstance(dem_copy, xdem.dem.DEM)

AssertionError: 

In [5]: type(dem_copy)
Out[5]: geoutils.georaster.Raster

In [6]: assert isinstance(dem, xdem.dem.DEM)
```
This could be argued to be a flaw in geoutils, but could also be fixed here.

rhugonnet:
Add subclass copy method (requires satImg_copy PR to work) + associated tests
Resolves #44


rhugonnet:
Ups, also added a test on the fit for spstat (still not 100% there on git fluency)

xdem/dem.py
```diff
@@ -71,6 +72,20 @@ def __init__(self, filename, vref_name=None, vref_grid=None, silent=False, **kwa
         # trying to get vref from product name (priority to user input)
         self.__parse_vref_from_fn(silent=silent)
 
+    def __copy__(self):
```
erikmannerfelt, line 75: This will yield unexpected outcomes if the `.data` attribute has been changed after loading. It will also not work if the DEM is created in memory (e.g. by `DEM.from_array()`). See my comment in a related issue:

> Right now, I think this works:
> 
> ```python
> import geoutils as gu
> import xdem
> fn_img = xdem.examples.FILEPATHS["longyearbyen_ref_dem"]
> r = gu.georaster.Raster(fn_img)
> dem = xdem.dem.DEM(r.filename)
> ```
> 
> Which yields:
> 
> ```
> No metadata could be read from filename.
> type(dem)
> Out[7]: xdem.dem.DEM
> ```
> 
> It goes through the "init" again, but this is not an issue as those run instantly (manipulating strings). If we had to migrate some more problematic, larger objects, we'd have to use a specific function.
> However we would have a problem if `self.filename` did not exist (a Raster copied in memory for instance). Then everything would fail. I guess it would be useful to store the original filename in `georaster.Raster`, maybe under `._filename`? Thoughts on this? @adehecq is getting better at understanding those underdash attributes :D

That only works assuming the Raster object has not been changed from disk:
```python
In [1]: import xdem

In [2]: import geoutils as gu

In [3]: raster = gu.georaster.Raster("examples/Longyearbyen/data/DEM_2009_ref.tif")

In [5]: raster_downsampled = raster.reproject(raster, dst_res=10)

In [6]: dem_downsampled = xdem.dem.DEM(raster_downsampled.filename)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-b3cc9f821625> in <module>
----> 1 dem_downsampled = xdem.dem.DEM(raster_downsampled.filename)

~/Projects/ETH/GlacioHack/xdem/xdem/dem.py in __init__(self, filename, vref_name, vref_grid, silent, **kwargs)
     59         """
     60 
---> 61         super().__init__(filename, **kwargs)
     62 
     63         if self.nbands > 1:

~/.local/share/conda/xdem/lib/python3.8/site-packages/geoutils/satimg.py in __init__(self, filename, attrs, load_data, bands, as_memfile, read_from_fn, datetime, tile_name, satellite, sensor, product, version, read_from_meta, fn_meta, silent)
    256         """
    257 
--> 258         super().__init__(filename, attrs=attrs, load_data=load_data, bands=bands, as_memfile=as_memfile)
    259 
    260         #TODO: maybe the Raster class should have an "original filename" attribute that doesn't get erased during

~/.local/share/conda/xdem/lib/python3.8/site-packages/geoutils/georaster.py in __init__(self, filename, bands, load_data, downsampl, masked, attrs, as_memfile)
     93         # Don't recognise the input, so stop here.
     94         else:
---> 95             raise ValueError('filename argument not recognised.')
     96 
     97         self._read_attrs(attrs)

ValueError: filename argument not recognised.

In [7]: raster_downsampled.filename
```
To make matters worse, it may sometimes work, but will give unexpected outcomes:
```python
In [16]: raster = gu.georaster.Raster("examples/Longyearbyen/data/DEM_2009_ref.tif")

In [17]: raster.data += 5

In [18]: dem = xdem.dem.DEM(raster.filename)

In [19]: assert np.mean(np.abs(raster.data - dem.data)) < 0.001
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-19-cd99e5be1816> in <module>
----> 1 assert np.mean(np.abs(raster.data - dem.data)) < 0.001

AssertionError:
```
The `.data` attribute of Raster can be modified while retaining the reference to the filename on disk. The `Raster.matches_disk` attribute is not even updated to be `False` (I'll submit an issue for this)!

I think the only reliable approach right now is my first hacky suggestion.

This problem will propagate to the `copy()` method in #46

_Originally posted by @erikmannerfelt in https://github.com/GlacioHack/xdem/issues/43#issuecomment-797300145_

tests/test_dem.py
```diff
@@ -22,6 +23,35 @@ def test_load(self):
         fn_img = xdem.examples.FILEPATHS["longyearbyen_ref_dem"]
         img = DEM(fn_img)
 
+    def test_copy(self):
```
erikmannerfelt, line 26: A test should be added to assert the consistency of the `.data` attribute.

A quick version would be:
```python
dem1 = DEM("path/to/file")
dem1.data += 5

dem2 = dem1.copy()

assert np.mean(np.abs(dem1.data - dem2.data)) < 0.001
```
And I suspect the above test will fail right now.

tests/test_spstats.py
```diff
@@ -66,6 +67,30 @@ def test_empirical_fit_variogram_running(self):
...
+        y = models.spherical(x,r=r1,c0=ps1) + models.spherical(x,r=r2,c0=ps2) \
+            + models.spherical(x,r=r3,c0=ps3)
+
+        sig = 0.025
+        y_noise = np.random.normal(0,sig,size=len(x))
+
+        y_simu = y + y_noise
+        sigma = np.ones(len(x))*sig
+
+        df = pd.DataFrame()
```
erikmannerfelt, line 85: Will instantiating DataFrames without a dtype not trigger a warning since a specific Pandas version? Or is that maybe just for Series maybe...

xdem/dem.py
```diff
@@ -71,6 +72,20 @@ def __init__(self, filename, vref_name=None, vref_grid=None, silent=False, **kwa
...
+        new_satimg = super().copy()
+        new_satimg.filename = self.filename
+        # those attributes are immutable, including pyproj.CRS
+        # name_attrs = ['vref','vref_grid','ccrs']
+        # for attrs in name_attrs:
+        #     setattr(new_dem, attrs, getattr(self, attrs))
+
+        return DEM(new_satimg.filename,silent=True)
+
+    def copy(self):
```
erikmannerfelt, line 86: Docstring!!

erikmannerfelt (changes requested):
I think there are some large limitations in this suggested approach...

erikmannerfelt:
A temporary fix could be to just do:
```python

def copy(self):
     new_dem = copy.copy(self)
     new_dem.data = self.data.copy()
     return new_dem

```

tests/test_dem.py
rhugonnet: This test is done further down, the two data are consistent:
```python
# check data array
        assert np.count_nonzero(~img.data == img2.data) == 0
        # check dataset_mask array
        assert np.count_nonzero(~img2.dataset_mask() == img2.dataset_mask()) == 0
```

tests/test_spstats.py
rhugonnet: No warnings ;)

xdem/dem.py
rhugonnet: I'm not sure I understand the issue for `copy()` (it's a valid one for passing a Raster to a DEM though).
As far as I know, the issue does not propagate to copy because:
- `Raster.copy()` is based on `Raster.from_array()` which create a new memory file based on the array.
- `SatImg/DEM.__copy__()` is based on `super().copy()` from Raster or SatImg, which thus creates that new memory file.
- `SatImg/DEM.copy()` uses the copy package to call the shallow `__copy__()` method (it's just a wrapper).
Can you elaborate on the underlying issue?

tests/test_dem.py
erikmannerfelt: But you haven't checked what happens if `img.data` is modified, right? If `img2` is loaded from `img.filename`, then it probably won't contain the in-memory modifications to `img.data`

xdem/dem.py
erikmannerfelt: This is the exact same issue as I commented on in the test file:

But you haven't checked what happens if `img.data` is modified, right? If `img2` is loaded from `img.filename`, then it probably won't contain the in-memory modifications to `img.data`

_Originally posted by @erikmannerfelt in https://github.com/GlacioHack/xdem/pull/46#discussion_r593013514_

tests/test_spstats.py
erikmannerfelt: I just tried and it's fine in my version.

xdem/dem.py
rhugonnet: I will double-check, but this shouldn't be an issue because it's the copy that is passed to the class `__init__()`, and filename is just copied (in `__copy__()`):
```python
new_satimg = super().copy()
new_satimg.filename = self.filename
```

rhugonnet: I see what you mean! Basically we need to add an assertion in `SatelliteImage.__init__()` and `DEM.__init__()` in case data is already loaded so that it doesn't call from `super()` again.

erikmannerfelt: Yeah, that should probably work!

erikmannerfelt: But I would suggest adding the test that I suggested (`dem1.data += 5` etc.) to make sure that things work expectedly.

rhugonnet: But wait... this line in `georaster.Raster.__init__()` should prevent any reloading from filename, no?
```python
 # Or, image is already a Memory File.
        elif isinstance(filename, rio.io.MemoryFile):
            self.filename = None
            self.ds = filename.open()
```
So I'm not sure there's any issue.

erikmannerfelt: I'm honestly a little lost now haha. As long as my proposed test passes, I would accept any approach to fix it!

erikmannerfelt:
(I accidentally added this comment in the wrong repo first!)

With @adehecq 's PR, I think the copy method works without your PR, @rhugonnet.

This is with the latest geoutils and xdem/main
```python
In [1]: import xdem

In [2]: import numpy as np

In [3]: dem = xdem.dem.DEM("examples/Longyearbyen/data/DEM_2009_ref.tif")

In [4]: dem2 = dem.copy()

In [5]: dem.data += 1

In [6]: assert not np.array_equal(dem.data, dem2.data, equal_nan=True)
```
This passes, meaning the copy method should work correctly.

rhugonnet:
> EDIT: I am too fast to judge sometimes. Sorry... This PR is still required:
> 
> ```python
> In [7]: dem.set_vref(vref_name='WGS84')
> 
> In [8]: dem3 = dem.copy()
> 
> In [9]: dem3.vref
> ```
> 
> The vref is not copied.

Haha no worries, I wrote this in Slack but it probably wasn't very clear.
OK, now I have:
- squash-merged Amaury's PR into geoutils
- merged with mine,
- modified the `__copy__`  Subclass methods to simply return the object out of `super().copy()` (corrected by Amaury), 
- added a loop for Subclass attributes to be copied (all immutable so it's a simple = for now). Because of this immutable/mutable characteristic that cannot be planned, we might have to add some specific copy() methods to newer objects in the future. Thus I didn't not use the "dict" attribute solution of @erikmannerfelt and kept the hardcoded list of immutable attributes. This list of attributes is written only once just before the start of the class, and is called to be used by copy(), or by test methods (as suggested by Amaury).
- kept the `copy()` wrapper based on the package copy.
- implemented your suggestions in test_georaster, test_satimg and test_dem (not in this PR) : `np.array_equal` + `r1 is not r2`. The list of attributes to check on is imported directly from the Classes. For georaster, 3 attributes cannot be checked by a simple __eq__ statement and have to be remove from list for obvious reasons (e.g., driver).

erikmannerfelt (approved):

tests/test_dem.py
```diff
@@ -22,6 +25,48 @@ def test_load(self):
...
+        """
+              Test that the copy method works as expected for satimg. In particular
```
adehecq, line 30: satimg should be replaced by dem

tests/test_dem.py
```diff
@@ -22,6 +25,48 @@ def test_load(self):
...
+
+        # Check dataset_mask array
+        assert np.all(r.data.mask == r2.data.mask)
+
+        # Check that if r.data is modified, it does not affect r2.data
+        r.data += 5
+        assert not np.array_equal(r.data, r2.data, equal_nan=True)
+
+        # Check that both have same output type
+        assert type(r) == type(r2)
```
adehecq, line 68: This is probably redundant with l 44.

adehecq:
Made some minor comments, but otherwise looks good!

adehecq (approved):

tests/test_dem.py
rhugonnet: Done

rhugonnet:
Resolves #46

rhugonnet:
And https://github.com/GlacioHack/GeoUtils/pull/116 resolves #44

Issue #48, opened 2021-03-14
Dependency handling needs revision


erikmannerfelt:
In `setup.py`, `richdem` is set as an extra_requires, but it is needed to `import xdem`. In addition, `scikit-gstat` is needed. `geoutils` obviously a requirement, but since it is not yet on PyPi, it won't be downloaded by `pip install xdem`. I don't know exactly how to fix that (apart from uploading geoutils on PyPi, of course)...

Right now, the "correct" way to install xdem without conda is quite complex:
```bash
pip install scikit-gstat
pip install richdem

pip install git+github.com/GlacioHack/GeoUtils.git
pip install git+github.com/GlacioHack/xdem.git  # or pip install xdem :)
```
Skipping any of the above steps leads in an `ImportError` when trying to `import xdem`.

If we would add `scikit-gstat` and `richdem` as requirements, this would be simpler. Question is, however, if we want them as mandatory requirements? As far as I'm aware, `richdem` only provides "nice to have" functions right now, unless @rhugonnet uses them more intricately somewhere? `scikit-gstat` is mandatory for variograms, right? Could `geoutils` be replaced somehow with `git+git+github.com/GlacioHack/GeoUtils.git` so that `pip` automatically fetches the latest master when trying to install `xdem`?

I know that worrying about "ease of install" in such an early stage of development is kind of dumb. I feel, however, that it will be easier to keep up with these issues instead of delaying them to a "super-fix" later!

PR #49, opened 2021-03-14
Fix 'pip install' to install all dependencies

erikmannerfelt:
`pip install git+https://github.com/GlacioHack/xdem.git` is now enough to install xdem! See #48 for what was previously needed on a clean install.

```bash
❯ conda create -n clean_env python ipython
❯ conda activate clean_env
❯ pip install git+https://github.com/GlacioHack/xdem.git

❯ ipython

Python 3.9.0 | packaged by conda-forge | (default, Nov 26 2020, 07:57:39)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.19.0 -- An enhanced Interactive Python. Type '?' for help.
                                                                                                                                                                                                                    
In [1]: import geoutils
ERROR 4: Unable to open EPSG support file gcs.csv.  Try setting the GDAL_DATA environment variable to point to the directory containing EPSG csv files.

In [2]: import xdem
```

I guess the geoutils error is because of GDAL... It's always GDAL!

I will mark this as a draft, because I've had to make both `richdem` and `scikit-gstat` into required dependencies. We should discuss this before merging!

`pip install xdem` will work accordingly as soon as we've marked a GitHub release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(name='xdem',
       license='BSD-3',
       packages=['xdem'],
       install_requires=['numpy', 'scipy', 'rasterio', 'geopandas',
-                        'pyproj', 'tqdm', 'geoutils @ https://github.com/GlacioHack/GeoUtils/tarball/master', 'richdem', 'scikit-gstat'],
+                        'pyproj', 'tqdm', 'geoutils @ https://github.com/GlacioHack/GeoUtils/tarball/master', 'scikit-gstat'],
       extras_require={'rioxarray': ['rioxarray'], 'richdem': ['richdem'], 'pdal': ['pdal'], 'opencv': ['opencv']},
       scripts=[],
       zip_safe=False)
--- a/setup.py
+++ b/setup.py
@@ -12,8 +12,10 @@ setup(name='xdem',
       author='The GlacioHack Team',
       license='BSD-3',
       packages=['xdem'],
-      install_requires=['numpy', 'scipy', 'rasterio', 'geopandas', 'pyproj', 'tqdm', 'geoutils'],
+      install_requires=['numpy', 'scipy', 'rasterio', 'geopandas',
+                        'pyproj', 'tqdm', 'geoutils', 'richdem', 'scikit-gstat'],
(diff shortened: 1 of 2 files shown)


rhugonnet:
> I guess the geoutils error is because of GDAL... It's always GDAL!
> 
> I will mark this as a draft, because I've had to make both `richdem` and `scikit-gstat` into required dependencies. We should discuss this before merging!

Good thinking, we should keep setup.py updated.
Yes, making GDAL bindings work simply with pip requirements can be a bit of a nightmare... And it's really not worth spending too much time because, even if it works temporarily, it can break again a few months later.
The only robust solution I know of is to go through the conda-forge channel with an environment file, like the one we provide :)

rhugonnet (approved):

erikmannerfelt:
Thanks for the input @rhugonnet! A dream come true would be to get rid of GDAL at some point in the future... But that's a beast of a job I presume!

@adehecq , what do you think of the new required dependencies?

rhugonnet:
> Thanks for the input @rhugonnet! A dream come true would be to get rid of GDAL at some point in the future... But that's a beast of a job I presume!

Well `rio` is "just" a convenience GDAL wrapper, and GDAL is the basis of all georeferenced analysis. So I'm not sure we can ever get rid of it, it's at the core of our calculations, but maybe they'll get better at managing dependencies with other packages/setup of environment variables :)

adehecq:
> Thanks for the input @rhugonnet! A dream come true would be to get rid of GDAL at some point in the future... But that's a beast of a job I presume!
> 
> @adehecq , what do you think of the new required dependencies?

Looks good!
And yes, as Romain said, GDAL is underneath most geospatial libraries, so you'll never get rid of it, even if you're not directly importing it!! ;-)

erikmannerfelt:
> > Thanks for the input @rhugonnet! A dream come true would be to get rid of GDAL at some point in the future... But that's a beast of a job I presume!
> > @adehecq , what do you think of the new required dependencies?
> 
> Looks good!
> And yes, as Romain said, GDAL is underneath most geospatial libraries, so you'll never get rid of it, even if you're not directly importing it!! ;-)

What I meant, @rhugonnet and @adehecq, was that I dream of a future where it's not required anymore!! It could be called free-DAL!!

Issue #53, opened 2021-03-14
Homogenize annotation and typing syntax


rhugonnet:
Right now it's still a mix. I'm up for homogenizing for in-function annotations everywhere (Erik's good at it, but I'm slowly learning ;D)

erikmannerfelt:
I support this!

The linters I use are: `mypy` (git version), `prospector` (`pylint` + `pep8` + `McCabe complexity`) and `pydocstyle`. The formatters I use are: `isort` and `autopep8`.

The latest release of `mypy` (December 2020) strangely doesn't like the `list[subtype]` syntax, even after including `from __future__ import annotations`. I therefore use the latest source version instead, which has included a fix (`pip install git+https://github.com/python/mypy.git`).

### Linters

**mypy** is a static type helper, for example giving an error if you write `def func(value: int = 1.1) -> float` (1.1 is not an int), or try to subsequently use the function in an incorrect manner: `"hello" + func(3)` (str + float doesn't work).

**pydocstyle** makes sure docstrings are more or less correctly formatted. It doesn't enforce e.g. sphinx formatting, so maybe there's another tool to add here?

**prospector** is a convenience linter that implements three linters in one:

**pylint** is a general-purpose linter for common issues. For example, it enforces the correct use of "snake_case" and "CamelCase" naming conventions.

**pep8** warns you about line lengths, weird whitespace, wrong indentation, correct line spacing between functions, etc.

**McCabe complexity** makes sure that functions and classes are not overly complex, e.g. if a function has more than X defined variables, maybe it should be split in two. I am particularly bad at this..

### Formatters

**isort** sorts imports alphabetically and divides them in three categories: "builtin imports", "third party imports", "local imports" (if the import is a directory in your cwd for example).

**autopep8** helps with automatically conforming your code to the pep8 standard.

I've set the `autopep8` and `pylint` maximum column width to 120 lines instead of 80 ([read Linus Torvalds' take on this!](https://lkml.org/lkml/2020/5/29/1038))



Should we agree on a combination of linters and formatters? I know some repos have commit/push/merge hooks that format the code. Since we are only ~three active developers right now, maybe this is overly complex at the time being, but we could consider something similar in the future.

adehecq:
Maybe the issue should be renamed "Write a contribution guideline" with all this info ;-)
https://docs.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors

adehecq:
[...]

PR #56, opened 2021-03-15
Create CONTRIBUTING.md

erikmannerfelt:
As suggested by @adehecq  in #53.

Opinions on the text?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,21 @@
+## How to contribute
+Contributors of `xdem` should attempt to conform to pep8 coding standards.
+An exception to the standard is having a 120 max character line length (instead of 80).
+
+Suggested linters are:
+1. prospector
+2. mypy (git version)
+3. pydocstyle
+
+Suggested formatters are:
+1. autopep8
+2. isort
+
+These can all be installed with this command:
+```bash
+pip install prospector git+https://github.com/mypy/mypy.git pydocstyle autopep8 isort
+```
+Note that your text editor of choice will also need to be configured with these tools (and max character length changed).
+
+Contributions are welcome in the form of pull requests (PRs) to the main branch.
+At least one test per feature (in the associated `tests/test_*.py` file) should be included in the PR, but more than one is suggested.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,8 +49,8 @@ Clone the git repo and create a conda environment
 ```bash
 git clone https://github.com/GlacioHack/xdem.git
 cd xdem
-conda create -f environment.yml  # add '-n customname' if you want.
-conda activate glacio  # or any other name specified above
+conda create -f environment.yml  # add '-n custom_name' if you want.
+conda activate xdem  # or any other name specified above
 pip install -e .  # Install xdem
 ```
 The linters and formatters mentioned above are recommended to install now.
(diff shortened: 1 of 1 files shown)


adehecq:
Maybe something in-between the relatively simple rasterio's guide: https://rasterio.readthedocs.io/en/latest/contributing.html
and very detailed pandas' guide: https://pandas.pydata.org/docs/development/contributing.html

rhugonnet:
Looks good!

CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
+# How to contribute
+
+## Rights
+The MIT license (see LICENSE) applies to all contributions.
+
+## Issue Conventions
+When submitting bugs, please include the following information:
+* Operating system type and version (Windows 10 / Ubuntu 18.04 etc.).
```
adehecq, line 8: Windows 10? Yeks! ;-)

CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+Some features may require later versions of Python (3.6+) to function correctly.
+
+### Setup
+
+Clone the git repo and create a conda environment
+```bash
+git clone https://github.com/GlacioHack/xdem.git
+cd xdem
+conda create -f environment.yml  # add '-n customname' if you want.
+conda activate glacio  # or any other name specified above
```
adehecq, line 53: Isn't the default name for the env 'xdem'? (from the environment file)

CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+pip install -e .  # Install xdem
+```
+The linters and formatters mentioned above are recommended to install now.
+
+### Running the tests
+To run the entire test suite, run pytest in the current directory:
```
adehecq, line 59: Actually, it's recommended not to run from the base directory as it could hide issues with the import.

erikmannerfelt: Do you have a reference for this? I always run it in the current directory. This is also implicitly recommended by [rasterio](https://rasterio.readthedocs.io/en/latest/contributing.html#running-the-tests).

CONTRIBUTING.md
erikmannerfelt: I forgot that we changed that! Corrected in a390749.

CONTRIBUTING.md
adehecq: No I don't have a reference, but if the package is not properly installed, in particular submodules etc, imports may succeed if you run from the base directory, but fail from another directory.

erikmannerfelt: So what do you propose instead?
```bash
cd ../  # To the parent directory
pytest xdem
```
Like that?

adehecq: Yes I think that's safer. Could also run `pytest xdem/tests` to avoid ambiguity, in case other files with *test* are added somewhere else later.

erikmannerfelt: May I suggest a compromise. I'll add:

> It is also recommended to try the tests from the parent directory, to validate that import statements work as they should:
> ```bash
> cd ../  # Change to the parent directory
> pytest xdem
> ```

I would argue against your suggestion to recommend `pytest xdem/tests`, for the exact same reason as you say. Python scripts should never start with `test_` if they aren't intended as tests, because that can indeed be ambiguous.

adehecq: As you wish! It's all mostly details anyway.
I call a lot of scripts test_* because I'm testing an idea! But not in geoutils or xdem.

PR #57, opened 2021-03-15
Update Class input name + fix tests and copy signature

rhugonnet:
Same as geoutils for input name: https://github.com/GlacioHack/GeoUtils/pull/128
Noticed `copy()` signature didn't match upstream functions (needed the new_array parameter), which created an error. This is fixed. Removed `__copy__` method that wasn't used all the way up to `Raster` (and is a bit useless in the end, probably no one is going to use the copy module to copy our data when there is a copy() class method).


xdem/dem.py
```diff
@@ -83,21 +82,16 @@ def __init__(self, filename, vref_name=None, vref_grid=None, silent=False, **kwa
         # trying to get vref from product name (priority to user input)
         self.__parse_vref_from_fn(silent=silent)
 
-    def __copy__(self):
```
erikmannerfelt, line 86: Same question about side-effects as in GeoUtils.

erikmannerfelt: ... which I got answered!

erikmannerfelt (approved):

PR #60, opened 2021-03-19
Some suggestions to make nuth_kaab and deramping more consistent with ICP results

adehecq:
- made changes to deramping so it works with masked Raster arrays
- added a function to calculate a rotation, translation between two PCs. This was initially thought as a way to convert a "Ramp" into a transformation matrix (which seems to work!), but I'm curious to see how it performs on real point-clouds.
@erikmannerfelt if you have a chance to take a look, I'm curious to hear what you think?


rhugonnet:
@erikmannerfelt @adehecq where are we on this, and in relation to #87? I'll dive into the details only if it is still needed

erikmannerfelt:
@rhugonnet the small corrections have been implemented in different PRs and I am working on the superinpose implementation with the `Coreg` subclass (see #94). Therefore, I will close this PR.

PR #68, opened 2021-03-23
Local and regional hypsometric dDEM interpolation

erikmannerfelt:
I've added local and hypsometric interpolation for the `dDEM` and `DEMCollection` classes. They are functional right now, but will be given some improvements. Hence, I will create a draft PR.

The regional approach works as follows:

1. Mask out the glacier areas using a provided mask or glacier outlines
2. Calculate a regional elevation change gradient from the masked dDEM.
3. Apply the gradient model to the reference DEM to get an "idealized dDEM"
4. Calculate the difference between the idealized dDEM and the actual dDEM, and linearly interpolate it.
5. Correct the idealized dDEM with the difference in 4) to get the final interpolated dDEM

The local approach does approximately the same thing, but it calculates it for each `Vector.ds` row separately. So point nr 2-5 are done iteratively for each glacier in the `Vector`.

If you consider that 4-5 should be optional, I can certainly make it so.

An example with a DEMCollection:
```python
from datetime import datetime

import xdem                                                                                                     
import geoutils as gu                                                                                           
                                                                         
                                                                                                                          
xdem.examples.download_longyearbyen_examples(overwrite=false)                                                   
                                                                                                                          
dem_2009 = xdem.dem(xdem.examples.filepaths["longyearbyen_ref_dem"], datetime=datetime(2009, 8, 1))                                            
dem_1990 = xdem.dem(xdem.examples.filepaths["longyearbyen_tba_dem"], datetime=datetime(1990, 8, 1))                                            
outlines_1990 = gu.Vector(xdem.examples.filepaths["longyearbyen_glacier_outlines"])

dems = xdem.DEMCollection([dem_2009, dem_1990], outlines=outlines_1990)

dems.subtract_dems()
dems.interpolate_ddems(method="local_hypsometric")

```

I've also tried to describe these methods as easy as possible, and I've experimented with matplotlib figures to visualise this! I would love some input on the text or just the general structure. [Here's a link to the documentation of the associated branch on my fork](https://xdem-erikm.readthedocs.io/en/spatial_hypsometric/comparison.html).

### DIFF
--- a/docs/source/comparison.rst
+++ b/docs/source/comparison.rst
@@ -103,6 +103,48 @@ Linear spatial interpolation (also often called bilinear interpolation) of dDEMs
 
         ddem.interpolate(method="linear")
 
+
+.. plot::
+        
+        import xdem
+        import geoutils as gu
+        dem_2009 = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_ref_dem"])
+        dem_1990 = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"])
+        outlines_1990 = gu.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines"])
+
+        ddem = xdem.dDEM(
+                xdem.spatial_tools.subtract_rasters(dem_2009, dem_1990, resampling_method="nearest"),
+                start_time=np.datetime64("1990-08-01"),
+                end_time=np.datetime64("2009-08-01")
+        )
+        # The example DEMs are void-free, so let's make some random voids.
+        ddem.data.mask = np.zeros_like(ddem.data, dtype=bool)  # Reset the mask
+        # Introduce 50000 nans randomly throughout the dDEM.
+        ddem.data.mask.ravel()[np.random.choice(ddem.data.size, 50000, replace=False)] = True
+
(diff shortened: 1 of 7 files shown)


erikmannerfelt:
Things on the TODO list are right now:

* Add functionality to customize the binning call, e.g. binsize, binning approach etc. (note: the customization already exists, it's just not available from the `dDEM.interpolate()` method).
* Set local hypsometry quality threshold for when to revert to regional hypsometry (for glaciers with terrible coverage)
* Add normalized regional hypsometric interpolation to use the normalized glacier elevation range instead of the absolute.
* Possibly make the local hypsometric approach faster. It's not terribly slow, but not blazing fast either.
* Add @rhugonnet's uncertainty based binning approaches?

adehecq:
@erikmannerfelt is this something we could merge into the main branch? I would like to test/use some of the functionalities for my own analysis.

erikmannerfelt:
Yes please! I thought the conversation plateaued because we were considering alternate approaches to this, but I would be more than happy to consider this as a (possibly temporary) working addition.

adehecq (approved):
Ok as it is now. Maybe save all unresolved comments as issues?

erikmannerfelt:
I will!

PR #71, opened 2021-04-05
Modular `coreg.py` revision

erikmannerfelt:
The entire coregistration structure has been revisited in this PR, with modularity and simplicity in focus.
All approaches are now sublasses of the `Coreg` class, which is heavily inspired by the structure employed in `scikit-learn`. Chaining coregistration functions is as simple as adding them using the "+" operator, or by explicitly constructing a `CoregPipeline` from a `list[Coreg]`.

Right now, I've added a lot of code, and have marked the old stuff with `DeprecationWarning`s. 
Later (or in this PR), I'll remove all of the old code.

### Data
In the following examples, I will use the same data, defined below:
```python
# Download the necessary data. This may take a few minutes.
xdem.examples.download_longyearbyen_examples(overwrite=False)

### Load the data using xdem and geoutils (could be with rasterio and geopandas instead)
# Load a reference DEM from 2009
reference_dem = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_ref_dem"])
# Load a moderately well aligned DEM from 1990
dem_to_be_aligned = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"]).reproject(reference_dem)
# Load glacier outlines from 1990. This will act as the unstable ground.
glacier_outlines = gu.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines"])


# Prepare the inputs for coregistration.
ref_data = reference_dem.data  # This is a numpy 2D array/masked_array
tba_data = dem_to_be_aligned.data  + 50 # This is a numpy 2D array/masked_array with an artificial bias
mask = glacier_outlines.create_mask(reference_dem) != 255  # This is a boolean numpy 2D array where inliers=True
transform = reference_dem.transform  # This is a rio.transform.Affine object.
```
# Syntax
All coregistration approaches have the same parent class; `Coreg`, which provides the following methods:
* `.fit()` for estimating the transform.                                                                              
* `.apply()` for applying the transform to a DEM.                                                                     
* `.apply_pts()` for applying the transform to a set of 3D points.                                                    
* `.to_matrix()` to convert the transform to a 4x4 transformation matrix, if possible.

The coregistration classes can be stacked sequentially into a `CoregPipeline` using two types of syntax:
```python
step1 = coreg.BiasCorr()
step2 = coreg.ICP()

pipeline = coreg.CoregPipeline([step1, step2])
same_pipeline = step1 + step2
```
The coregistration methods are no longer dependent on `GeoUtils`, to hopefully appeal to a larger audience.

DEMs are `np.ndarray`s or `np.masked_arrays`, masks are boolean `np.ndarray`s, and transforms are `rio.transform.Affine` objects.

I envision the `DEMCollection` object to be a `GeoUtils`-centred interface to coregistration, for example with a `DEMCollection.coregister()` function or similar.

# Examples

## Standard Nuth and Kääb (2011)
This now has a bias correction built in, in case people would use only this one.

```python
from xdem import coreg

nuth_kaab = coreg.NuthKaab(max_iterations=50, error_threshold=0.05)

# Fit the data to a suitable x/y/z offset.
nuth_kaab.fit(ref_data, tba_data, transform=transform, mask=mask)

# Apply the transformation to the data (or any other data)
aligned_dem = nuth_kaab.apply(tba_data, transform=transform)
```

## ICP with bias correction
I like ICP, but it sometimes works poorly if there is a large bias between the products.
As a solution, I will chain a bias correction and ICP to run sequentially:
```python


pipeline = coreg.BiasCorr() + coreg.ICP()

# This will first estimate the bias, apply it, and subsequently estimate the ICP transform 
pipeline.fit(ref_data, tba_data, mask=mask, transform=transform)

# This sequentially applies the Coreg.apply() functions
aligned_dem_array = pipeline.apply(tba_data, transform=transform)

# This prints the combined matrix of all steps
print(pipeline.to_matrix())
```
outputs:
```
[[ 9.99999990e-01  6.10542619e-05 -1.29476318e-04  3.39646143e+00]
 [-6.10697122e-05  9.99999991e-01 -1.19328652e-04  1.73416704e+00]
 [ 1.29469031e-04  1.19336558e-04  9.99999984e-01 -5.48167431e+01]
 [ 0.00000000e+00  0.00000000e+00  0.00000000e+00  1.00000000e+00]]
```
Note the bias (index 2,3) (-4.8 m when removing the artificial bias) which is close to what I get without the bias addition (-5.5 m). I presume that the difference of 7 dm is because ICP is generally bad with biases, so the `BiasCorr + ICP` approach is better.

## Ultimate "bias + ICP + Nuth and Kääb (2011)" combo
@adehecq  and I had some discussions around his KH-9 DEMs.
Nuth and Kääb (2011) is not enough because there are considerable rotations involved, but ICP stops at local minima and provides only sub-acceptable results.
Plus, ICP needs a bias correction since the KH-9 DEMs were something like 1800 m off in the start!
Here's a pipeline that does all of it:
```python
pipeline = coreg.BiasCorr() + coreg.ICP() + coreg.NuthKaab()

pipeline.fit(ref_data, tba_data, mask=mask, transform=transform)

print(pipeline.to_matrix())
```
prints this for the Longyearbyen examples:
```
[[ 9.99999990e-01  6.10542619e-05 -1.29476318e-04 -2.52099627e+00]
 [-6.10697122e-05  9.99999991e-01 -1.19328652e-04 -9.10047542e+00]
 [ 1.29469031e-04  1.19336558e-04  9.99999984e-01 -5.43112118e+01]
 [ 0.00000000e+00  0.00000000e+00  0.00000000e+00  1.00000000e+00]]
```
There's quite a large difference in the x-component (index 1,3; -9.1 m here vs. 1.73 m above) which fits well with the fact that we saw a considerable east-west offset when only using ICP.
While the above step takes a while, I feel like this would be the go-to combination with poorly aligned datasets.

# Making more Coreg sublasses
I wrote all of this with the intention of it being possible to extend by others.
The user interface is defined by the `Coreg` class, which formats the data, and provides it to hidden methods which are defined by the subclass. Making a new subclass therefore only requires the `__init__()`  method and four others:
```python
class NewCoregApproach(Coreg):
    def __init__(self, param1, param2):
        self.param1 = param1
        self.param2 = param2

        # This is where all the metadata and fitting data are stored. 
        # I couldn't find a way to automatically make the attribute, so it has to be repeated in every subclass..
        self._meta: dict[str, Any] = {}

    def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optional[rio.transform.Affine], weights: Optional[np.ndarray]):
        ## do the fitting magic here, and save it in self._meta

    def _apply_func(dem: np.ndarray, transform: rio.transform.Affine) -> np.ndarray:
          # do the apply magic here and return the result
          # 'dem' will always be a np.ndarray with NaNs as nodata values

    def _apply_pts_func(self, coords: np.ndarray) -> np.ndarray:
        # do the apply-to-points magic here and return the result
        # 'coords' will always have the shape (N, 3)

    def _to_matrix_func(self) -> np.ndarray:
        # do the to_matrix magic here and return the result
```
For a functional example, please see `BiasCorr` as this is the simplest formulation.

And before anyone asks, `_apply_func` always takes a `np.ndarray`, but the `.apply()` userspace function takes either `np.ndarray`s or `np.masked_arrays` and returns the same type as its input.

# Documentation
I have spent quite some time and effort to make the documentation clear for people. You can see examples and explanations that I have written in my fork:
https://xdem-erikm.readthedocs.io/en/coreg_class/index.html

As of right now (05/04/2021), the page index in readthedocs has gone haywire. I don't know what that's about, but the rest looks as it should. (EDIT: [This seems to be an issue that affects others as well](https://github.com/readthedocs/readthedocs.org/issues/8070))

# On the todo list
* Add a `.error()` method or property to all functions. NMAD? But NMAD doesn't show potential biases.
* Allow for @adehecq 's filters to be joined in a pipeline, e.g `coreg.ICP() + filters.NMAD(3) + coreg.NuthKaab()`
* Improve performance on `CoregPipeline.apply()` by combining matrices instead of calling `.apply()` on every `Coreg` class inside. This would mean only one reprojection is done once. It would only work when `.to_matrix()` is supported of course (e.g. `Deramp(degree=2)` cannot be described as a matrix).
* Maybe move `Deramp` and `BiasCorr` into `xdem/biascorr.py`. I propose that they should still be a `Coreg` subclass, however.
* Add subsampling argument to the `.fit()` method to improve performance at the cost of quality. (added in 
593ef00)
* Test the classes on more difficult data than the "best-case" Longyearbyen examples.
* Add `verbose` and/or `callback` arguments for progress updates.

Hope you like it! It's taken quite some time to get right, but I am pretty happy about the result. Especially with the `CoregPipeline` class!

### DIFF
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -166,7 +166,7 @@ class TestCoregClass:
         matrix = biascorr.to_matrix()
         assert matrix[2, 3] == bias, matrix
 
-        # Cehck that the first z coordinate is now the bias
+        # Check that the first z coordinate is now the bias
         assert biascorr.apply_pts(self.points)[0, 2] == biascorr._meta["bias"]
 
         # Apply the model to correct the DEM
@@ -197,61 +197,75 @@ class TestCoregClass:
         shifted_dem[:, :pixel_shift] = np.nan
         shifted_dem += bias
 
+        # Fit the synthesized shifted DEM to the original
         nuth_kaab.fit(self.ref.data.squeeze(), shifted_dem, transform=self.ref.transform)
 
+        # Make sure that the estimated offsets are similar to what was synthesized.
         assert abs(nuth_kaab._meta["offset_east_px"] - pixel_shift) < 0.03
         assert abs(nuth_kaab._meta["offset_north_px"]) < 0.03
         assert abs(nuth_kaab._meta["bias"] + bias) < 0.03
 
+        # Apply the estimated shift to "revert the DEM" to its original state.
         unshifted_dem = nuth_kaab.apply(shifted_dem, transform=self.ref.transform)
+        # Measure the difference (should be more or less zero)
         diff = np.asarray(self.ref.data.squeeze() - unshifted_dem)
 
+        # Check that the median is very close to zero
         assert np.abs(np.nanmedian(diff)) < 0.01
+        # Check that the RMSE is low
         assert np.sqrt(np.nanmean(np.square(diff))) < 1
 
+        # Transform some arbitrary points.
         transformed_points = nuth_kaab.apply_pts(self.points)
 
+        # Check that the x shift is close to the pixel_shift * image resolution
         assert abs((transformed_points[0, 0] - self.points[0, 0]) + pixel_shift * self.ref.res[0]) < 0.1
+        # Check that the z shift is close to the original bias.
         assert abs((transformed_points[0, 2] - self.points[0, 2]) + bias) < 0.1
 
     def test_deramping(self):
         warnings.simplefilter("error")
 
+        # Try a 1st degree deramping.
         deramp = coreg.Deramp(degree=1)
 
+        # Fit the data
         deramp.fit(**self.fit_params)
 
+        # Apply the deramping to a DEm
         deramped_dem = deramp.apply(self.tba.data, self.ref.transform)
 
+        # Get the periglacial offset after deramping
         periglacial_offset = (self.ref.data.squeeze() - deramped_dem)[~self.mask.squeeze()]
+        # Get the periglacial offset before deramping
         pre_offset = (self.ref.data - self.tba.data).squeeze()[~self.mask]
 
         # Check that the error improved
         assert np.abs(np.mean(periglacial_offset)) < np.abs(np.mean(pre_offset))
 
         # Check that the mean periglacial offset is low
-        assert -1 < np.mean(periglacial_offset) < 1
+        assert np.abs(np.mean(periglacial_offset)) < 1
 
+        # Try a 0 degree deramp (basically bias correction)
         deramp0 = coreg.Deramp(degree=0)
-
         deramp0.fit(self.ref.data, self.tba.data, ~self.mask, transform=self.ref.transform)
 
+        # Check that only one coefficient exists (y = x + a => coefficients=["a"])
         assert len(deramp0._meta["coefficients"]) == 1
+        # Extract said bias
         bias = deramp0._meta["coefficients"][0]
 
-        # Make sure to_matrix does not throw an error.
+        # Make sure to_matrix does not throw an error. It will for higher degree deramps
         deramp0.to_matrix()
 
-        # Create some 3D coordinates with Z coordinates being 0
-        points = np.array([[1, 2, 3, 4], [1, 2, 3, 4], [0, 0, 0, 0]], dtype="float64").T
-
-        assert deramp0.apply_pts(points)[0, 2] == bias
+        # Check that the apply_pts would apply a z shift equal to the bias
+        assert deramp0.apply_pts(self.points)[0, 2] == bias
 
     def test_icp_opencv(self):
         warnings.simplefilter("error")
 
+        # Do a fast an dirty 3 iteration ICP just to make sure it doesn't error out.
         icp = coreg.ICP(max_iterations=3)
-
         icp.fit(self.ref.data, self.tba.data, ~self.mask, transform=self.ref.transform)
 
         aligned_dem = icp.apply(self.tba.data, self.ref.transform)
@@ -261,18 +275,21 @@ class TestCoregClass:
     def test_pipeline(self):
         warnings.simplefilter("error")
 
+        # Create a pipeline from two coreg methods.
         pipeline = coreg.CoregPipeline([coreg.BiasCorr(), coreg.ICP(max_iterations=3)])
-
         pipeline.fit(**self.fit_params)
 
         aligned_dem = pipeline.apply(self.tba.data, self.ref.transform)
 
         assert aligned_dem.shape == self.ref.data.squeeze().shape
 
+        # Make a new pipeline with two bias correction approaches.
         pipeline2 = coreg.CoregPipeline([coreg.BiasCorr(), coreg.BiasCorr()])
+        # Set both "estimated" biases to be 1
         pipeline2.pipeline[0]._meta["bias"] = 1
         pipeline2.pipeline[1]._meta["bias"] = 1
 
+        # Assert that the combined bias is 2
         pipeline2.to_matrix()[2, 3] == 2.0
 
     def test_coreg_add(self):
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import copy
 import os
 import tempfile
+import time
 import warnings
 from typing import Any
 
@@ -304,3 +305,45 @@ class TestCoregClass:
         # Try to add two CoregPipelines
         bias5 = bias3 + bias3
         assert bias5.to_matrix()[2, 3] == bias * 4
+
+    def test_subsample(self):
+        warnings.simplefilter("error")
+
+        # Test subsampled bias correction
+        bias_sub = coreg.BiasCorr()
+
+        # Fit the bias using 50% of the unmasked data using a fraction
+        bias_sub.fit(**self.fit_params, subsample=0.5)
+        # Do the same but specify the pixel count instead.
+        # They are not perfectly equal (np.count_nonzero(self.mask) // 2 would be exact)
+        # But this would just repeat the subsample code, so that makes little sense to test.
+        bias_sub.fit(**self.fit_params, subsample=self.tba.data.size // 2)
+
+        # Do full bias corr to compare
+        bias_full = coreg.BiasCorr()
+        bias_full.fit(**self.fit_params)
+
+        # Check that the estimated biases are similar
+        assert abs(bias_sub._meta["bias"] - bias_full._meta["bias"]) < 0.1
+
+        # Test ICP with subsampling
+        icp_full = coreg.ICP(max_iterations=20)
+        icp_sub = coreg.ICP(max_iterations=20)
+
+        # Measure the start and stop time to get the duration
+        start_time = time.time()
+        icp_full.fit(**self.fit_params)
+        icp_full_duration = time.time() - start_time
+
+        # Do the same with 50% subsampling
+        start_time = time.time()
+        icp_sub.fit(**self.fit_params, subsample=0.5)
+        icp_sub_duration = time.time() - start_time
+
+        # Make sure that the subsampling increased performance
+        assert icp_full_duration > icp_sub_duration
+
+        # Calculate the difference in the full vs. subsampled ICP matrices
+        matrix_diff = np.abs(icp_full.to_matrix() - icp_sub.to_matrix())
+        # Check that the x/y/z differences do not exceed 30cm
+        assert np.count_nonzero(matrix_diff > 0.3) == 0
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -6,13 +6,18 @@ Author(s):
 """
 from __future__ import annotations
 
+import copy
 import os
 import tempfile
+import warnings
 from typing import Any
 
 import geoutils as gu
+import numpy as np
 
-from xdem import coreg, examples
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    from xdem import coreg, examples
 
 
 def load_examples() -> tuple[gu.georaster.Raster, gu.georaster.Raster, gu.geovector.Vector]:
@@ -126,3 +131,47 @@ def test_only_paths():
     to_be_aligned_raster = examples.FILEPATHS["longyearbyen_tba_dem"]
 
     coreg.coregister(reference_raster, to_be_aligned_raster)
+
+
+class TestCoregClass:
+    ref, tba, outlines = load_examples()  # Load example reference, to-be-aligned and mask.
+    mask = outlines.create_mask(ref) == 255
+
+    def test_bias(self):
+        warnings.simplefilter("error")
+
+        # Create a bias correction instance
+        biascorr = coreg.BiasCorr()
+        # Fit the bias model to the data
+        biascorr.fit(reference_dem=self.ref.data, dem_to_be_aligned=self.tba.data, mask=self.mask)
+
+        # Check that a bias was found.
+        assert biascorr._meta.get("bias") is not None
+        assert biascorr._meta["bias"] != 0.0
+
+        # Copy the bias to see if it changes in the test (it shouldn't)
+        bias = copy.copy(biascorr._meta["bias"])
+
+        # Check that the to_matrix function works as it should
+        matrix = biascorr.to_matrix()
+        assert matrix[2, 3] == bias, matrix
+
+        # Create some 3D coordinates with Z coordinates being 0
+        points = np.array([[1, 2, 3, 4], [1, 2, 3, 4], [0, 0, 0, 0]], dtype="float64").T
+        # Cehck that the first z coordinate is now the bias
+        assert biascorr.apply_pts(points)[0, 2] == biascorr._meta["bias"]
+
+        # Apply the model to correct the DEM
+        tba_unbiased = biascorr.apply(self.tba.data, None)
+
+        # Create a new bias correction model
+        biascorr2 = coreg.BiasCorr()
+        # Check that this is indeed a new object
+        assert biascorr is not biascorr2
+        # Fit the corrected DEM to see if the bias will be close to or at zero
+        biascorr2.fit(reference_dem=self.ref.data, dem_to_be_aligned=tba_unbiased, mask=self.mask)
+        # Test the bias
+        assert abs(biascorr2._meta.get("bias")) < 0.01
+
+        # Check that the original model's bias has not changed (the _meta dicts are two different objects)
+        assert biascorr._meta["bias"] == bias
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -18,7 +18,7 @@ import numpy as np
 
 with warnings.catch_warnings():
     warnings.simplefilter("ignore")
-    from xdem import coreg, examples
+    from xdem import coreg, examples, spatial_tools
 
 
 def load_examples() -> tuple[gu.georaster.Raster, gu.georaster.Raster, gu.geovector.Vector]:
@@ -248,7 +248,7 @@ class TestCoregClass:
 
         # Try a 0 degree deramp (basically bias correction)
         deramp0 = coreg.Deramp(degree=0)
-        deramp0.fit(self.ref.data, self.tba.data, ~self.inlier_mask, transform=self.ref.transform)
+        deramp0.fit(**self.fit_params)
 
         # Check that only one coefficient exists (y = x + a => coefficients=["a"])
         assert len(deramp0._meta["coefficients"]) == 1
@@ -266,7 +266,7 @@ class TestCoregClass:
 
         # Do a fast an dirty 3 iteration ICP just to make sure it doesn't error out.
         icp = coreg.ICP(max_iterations=3)
-        icp.fit(self.ref.data, self.tba.data, ~self.inlier_mask, transform=self.ref.transform)
+        icp.fit(**self.fit_params)
 
         aligned_dem = icp.apply(self.tba.data, self.ref.transform)
 
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -175,3 +175,25 @@ class TestCoregClass:
 
         # Check that the original model's bias has not changed (the _meta dicts are two different objects)
         assert biascorr._meta["bias"] == bias
+
+    def test_icp_opencv(self):
+        warnings.simplefilter("error")
+
+        icp = coreg.ICP(max_iterations=3)
+
+        icp.fit(self.ref.data, self.tba.data, ~self.mask, transform=self.ref.transform)
+
+        aligned_dem = icp.apply(self.tba.data, self.ref.transform)
+
+        assert aligned_dem.shape == self.ref.data.squeeze().shape
+
+    def test_pipeline(self):
+        warnings.simplefilter("error")
+
+        pipeline = coreg.CoregPipeline([coreg.BiasCorr(), coreg.ICP(max_iterations=3)])
+
+        pipeline.fit(self.ref.data, self.tba.data, ~self.mask, transform=self.ref.transform)
+
+        aligned_dem = pipeline.apply(self.tba.data, self.ref.transform)
+
+        assert aligned_dem.shape == self.ref.data.squeeze().shape
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -137,13 +137,22 @@ class TestCoregClass:
     ref, tba, outlines = load_examples()  # Load example reference, to-be-aligned and mask.
     mask = outlines.create_mask(ref) == 255
 
+    fit_params = dict(
+        reference_dem=ref.data,
+        dem_to_be_aligned=tba.data,
+        mask=~mask,
+        transform=ref.transform
+    )
+    # Create some 3D coordinates with Z coordinates being 0 to try the apply_pts functions.
+    points = np.array([[1, 2, 3, 4], [1, 2, 3, 4], [0, 0, 0, 0]], dtype="float64").T
+
     def test_bias(self):
         warnings.simplefilter("error")
 
         # Create a bias correction instance
         biascorr = coreg.BiasCorr()
         # Fit the bias model to the data
-        biascorr.fit(reference_dem=self.ref.data, dem_to_be_aligned=self.tba.data, mask=self.mask)
+        biascorr.fit(**self.fit_params)
 
         # Check that a bias was found.
         assert biascorr._meta.get("bias") is not None
@@ -156,10 +165,8 @@ class TestCoregClass:
         matrix = biascorr.to_matrix()
         assert matrix[2, 3] == bias, matrix
 
-        # Create some 3D coordinates with Z coordinates being 0
-        points = np.array([[1, 2, 3, 4], [1, 2, 3, 4], [0, 0, 0, 0]], dtype="float64").T
         # Cehck that the first z coordinate is now the bias
-        assert biascorr.apply_pts(points)[0, 2] == biascorr._meta["bias"]
+        assert biascorr.apply_pts(self.points)[0, 2] == biascorr._meta["bias"]
 
         # Apply the model to correct the DEM
         tba_unbiased = biascorr.apply(self.tba.data, None)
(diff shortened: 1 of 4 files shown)


rhugonnet:
Wow this is truly amazing !! And glad to see I'm not the only one working during Easter ;)
I already had a first read, and I will come back to look at all the details later this afternoon :smiley:

tests/test_coreg.py
```diff
@@ -126,3 +131,176 @@ def test_only_paths():
...
+        # Check that a bias was found.
+        assert biascorr._meta.get("bias") is not None
+        assert biascorr._meta["bias"] != 0.0
+
+        # Copy the bias to see if it changes in the test (it shouldn't)
+        bias = copy.copy(biascorr._meta["bias"])
+
+        # Check that the to_matrix function works as it should
+        matrix = biascorr.to_matrix()
+        assert matrix[2, 3] == bias, matrix
```
rhugonnet, line 166: How does the `,` works for the `==` assertion?

tests/test_coreg.py
```diff
@@ -126,3 +131,176 @@ def test_only_paths():
...
+
+        # Cehck that the first z coordinate is now the bias
```
rhugonnet, line 168: Typo

tests/test_coreg.py
```diff
@@ -126,3 +131,176 @@ def test_only_paths():
...
+        warnings.simplefilter("error")
+
+        nuth_kaab = coreg.NuthKaab(max_iterations=10)
+
+        # Synthesize a shifted and vertically offset DEM
+        pixel_shift = 2
+        bias = 5
+        shifted_dem = self.ref.data.squeeze().copy()
+        shifted_dem[:, pixel_shift:] = shifted_dem[:, :-pixel_shift]
+        shifted_dem[:, :pixel_shift] = np.nan
```
rhugonnet, line 196: I guess that's a fast way of doing it!

tests/test_coreg.py
```diff
@@ -126,3 +131,176 @@ def test_only_paths():
...
+
+        nuth_kaab.fit(self.ref.data.squeeze(), shifted_dem, transform=self.ref.transform)
+
+        assert abs(nuth_kaab._meta["offset_east_px"] - pixel_shift) < 0.03
+        assert abs(nuth_kaab._meta["offset_north_px"]) < 0.03
+        assert abs(nuth_kaab._meta["bias"] + bias) < 0.03
+
+        unshifted_dem = nuth_kaab.apply(shifted_dem, transform=self.ref.transform)
+        diff = np.asarray(self.ref.data.squeeze() - unshifted_dem)
+
```
rhugonnet, line 207: Comments are getting scarce down there, looks like my code ;)

rhugonnet (commented):
OK, I've read through `coreg.py` twice now!! (and once for all the rest).

C'est vraiment génial :partying_face: ! Well thought and organized. Really impressed (and learned quite a lot) with this synergy of `Coreg` and `CoregPipeline`, embedded in the recursivity of certain Class methods like `__add__` and in the overriden Subclass methods like`_fit_func` etc. 
All this is going to be extremely useful for practicality, reproducibility, modularity... It's a dream come true, really!

I had almost no specific comment because all seems really great as it is. We can debug/solve issues if we start having some when using it!

Moving forward, I have two general comments:
- I'm thinking that maybe it would be better to have the full co-registration methods (and all their sub-methods) lie outside of the Classes, for both readability (they would live next to their associated subfunctions) and reuse outside class methods (otherwise you absolutely need a Coreg object to run those). This would also make it easier to grasp for user less familiar with classes. The class methods would then always be wrapper quite easy to write:
```python
class NuthKaab(Coreg)
    def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optional[rio.transform.Affine], weights: Optional[np.ndarray], **kwargs):
        """Estimate the x/y/z offset between two DEMs."""

        offset_east, offset_north, bias = nuth_kaab(ref_dem, tba_dem, transform, weights, **kwargs)

        self._meta["offset_east_px"] = offset_east
        self._meta["offset_north_px"] = offset_north
        self._meta["bias"] = bias
```
- My second thought is that we could extend the pipeline object to encompass even more methods than just the ones in `coreg.py` (you probably already had this in mind, Erik!). For instance it would be great to be able to do:
```python
step1 = filters.NMAD()
step2 = coreg.BiasCorr()
step3. = filters.MedianKernel()
step4 = coreg.ICP()
step5 = biascorr.AlongTrack()

pipeline = step1 + step2 + step3 + step4 + step5
```
I don't know what would be best there... probably to have classes for `BiasCorr`, `Filters` and `Coreg` object, and one `CoregPipeline` (or other naming) to englobe them all? (one ring to rule them all :elf: )

In any case, again superb work! :+1: 
Can't wait to start contributing to and using all this (I will be active again very soon, started drowning a bit again with some data sharing for the global study last week, but it's almost all done now!).

rhugonnet (approved):

adehecq:
I will also try to have a close look at all this today, and hopefully test it locally. I will come back with more feedback!

xdem/coreg.py
```diff
@@ -1063,3 +1081,604 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+
+        with warnings.catch_warnings():
+            # Deprecation warning from pytransform3d. Let's hope that is fixed in the near future.
+            warnings.filterwarnings("ignore", message="`np.float` is a deprecated alias for the builtin `float`")
+            for i, coreg in enumerate(self.pipeline):
+                new_matrix = coreg.to_matrix()
+
+                transform_mgr.add_transform(i, i + 1, new_matrix)
+
+            return transform_mgr.get_transform(0, len(self.pipeline))
```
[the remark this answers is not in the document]
erikmannerfelt: I am a bit afraid of matrices haha, so that's why I decided on using pytransform3d. Indeed, the syntax becomes a bit convoluted here, but I think (keywork think, I am not certain) it is necessary. Basically what we want is to merge transforms "A -> B -> C" into just "A-C" (in my case, I've used integer indices instead: "0 -> 1 -> 2" into "0-2"). 

Maybe there's an easier way to do this, but pytransform3d makes sure it is done right.

tests/test_coreg.py
erikmannerfelt: Everything after the comma gets printed in the `AssertionError`:
```python
assert 1 == 2, "One turns out to be different than two"
```
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-1-d33f3b548b72> in <module>
----> 1 assert 1 == 2, "One turns out to be different than two"

AssertionError: One turns out to be different than two
```
Just printing the `matrix` variable is not very helpful any more, I confess. It was for me to check it visually when I developed it!

tests/test_coreg.py
erikmannerfelt: Haha! Added better docs in b569902

tests/test_coreg.py
erikmannerfelt: Fixed in b569902

erikmannerfelt:
Thank you @rhugonnet for the great and motivating feedback!!!

Answers to your general comments:

1. It can definitely be made as you suggest. Then maybe all coreg functions should return a matrix to be compatible with each other (no x_px, y_px, z_m). Then we could have one function to apply a matrix to a DEM. Right now, `NuthKaab` and `ICP` applies shifts differently, and I don't know which is best. 
2. I totally agree on the filter + coreg merging! It's point nr. 2 on my todo list in the top. How about `CoregPipeline` is renamed to just `Pipeline` (just like scikit-learn, by the way)? Then the `Filter.__add__()` method could return a `Pipeline` instance as well.

adehecq:
First question, in your example above, you wrote:
`dem_to_be_aligned = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"]).resample(reference_dem)`
did you mean reproject, rather than resample?

erikmannerfelt:
> First question, in your example above, you wrote:
> `dem_to_be_aligned = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"]).resample(reference_dem)`
> did you mean reproject, rather than resample?

@adehecq, I found another typo in my example. The mask should be `!= 255`, not `== 255` because the inliers should be `True`, not the outliers.

That is actually something we should discuss at some point. Is a mask `True` for the inliers (read: "these are the areas to include") or for the outliers (read: "these are the areas to exclude")??

adehecq:
It could be nice to have a verbose/plot option to print some results or plot figures. Should these options be accepted by all fit functions?

erikmannerfelt:
> It could be nice to have a verbose/plot option to print some results or plot figures. Should these options be accepted by all fit functions?

Ah, that should have been on my todo list. Either a `verbose` flag, a `callback`, or both.

adehecq:
> > First question, in your example above, you wrote:
> > `dem_to_be_aligned = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"]).resample(reference_dem)`
> > did you mean reproject, rather than resample?
> 
> @adehecq, I found another typo in my example. The mask should be `!= 255`, not `== 255` because the inliers should be `True`, not the outliers.
> 
> That is actually something we should discuss at some point. Is a mask `True` for the inliers (read: "these are the areas to include") or for the outliers (read: "these are the areas to exclude")??
> 
> EDIT: I've fixed the example above.

Ok cool, it explains why I had a 1000 m bias!! Now Nuth & Kaab seem to work as expected.
As for the masks, the logic of Vectors.create_mask is that the inside of the polygons is True, which is the most logical I believe. But for the coreg functions, it could be ambiguous yes... Intuitively, I would understand 'mask' as 'ignore these points. Maybe we could rename mask as 'inlier_mask'?

erikmannerfelt:
> > > First question, in your example above, you wrote:
> > > `dem_to_be_aligned = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"]).resample(reference_dem)`
> > > did you mean reproject, rather than resample?
> > 
> > 
> > @adehecq, I found another typo in my example. The mask should be `!= 255`, not `== 255` because the inliers should be `True`, not the outliers.
> > That is actually something we should discuss at some point. Is a mask `True` for the inliers (read: "these are the areas to include") or for the outliers (read: "these are the areas to exclude")??
> > EDIT: I've fixed the example above.
> 
> Ok cool, it explains why I had a 1000 m bias!! Now Nuth & Kaab seem to work as expected.
> As for the masks, the logic of Vectors.create_mask is that the inside of the polygons is True, which is the most logical I believe. But for the coreg functions, it could be ambiguous yes... Intuitively, I would understand 'mask' as 'ignore these points. Maybe we could rename mask as 'inlier_mask'?

Great idea. I changed it in 843b7b0

adehecq:
Interestingly, in your examples Bias + ICP + NuthKaab performs less well than NuthKaab alone...
That's because ICP introduces some warping of the DEM that NuthKaab cannot correct :-/

erikmannerfelt:
> Interestingly, in your examples Bias + ICP + NuthKaab performs less well than NuthKaab alone...
> That's because ICP introduces some warping of the DEM that NuthKaab cannot correct :-/

Huh.. I have a hunch why this might be. I'll get on it after lunch.

erikmannerfelt:
I think the problem is because I use `scipy.interpolate.griddata` which honestly just doesn't work that well. I tried `NuthKaab` and `BiasCorr + ICP + NuthKaab` and tried changing the interpolation type.


| Approach | griddata interpolation | NMAD (m) | time (s) |
|----------------|--------------------------------|----------|--------|
| NuthKaab |  - | 3.10 | 7.95 |
| BiasCorr + ICP + NuthKaab | nearest (original) | 3.71 | 27.91 |
| BiasCorr + ICP + NuthKaab | linear | 3.49 | 49.82 | 
| BiasCorr + ICP + NuthKaab | cubic | 3.50 | 54.0 | 

Note the improvement in error if linear or cubic interpolation is done instead of nearest neighbor, but the time it takes increases by almost 2x.
The reason interpolation is needed is because I right now apply a matrix by converting the DEM to a point cloud, then applying the transform, and re-gridding it (using `scipy.interpolate.griddata`).
@adehecq  or @rhugonnet , if you know a better approach for applying a transformation matrix to a DEM, I would be delighted to implement that instead!

It could be in a general:
```python
def apply_matrix(dem: np.ndarray, transform: rio.transform.Affine, matrix: np.ndarray) -> np.ndarray:
    # do matrix magic
    return transformed_dem
``` 
which could be used for any purpose.

adehecq:
> I think the problem is because I use `scipy.interpolate.griddata` which honestly just doesn't work that well. I tried `NuthKaab` and `BiasCorr + ICP + NuthKaab` and tried changing the interpolation type.
> Approach 	griddata interpolation 	NMAD (m) 	time (s)
> NuthKaab 	- 	3.10 	7.95
> BiasCorr + ICP + NuthKaab 	nearest (original) 	3.71 	27.91
> BiasCorr + ICP + NuthKaab 	linear 	3.49 	49.82
> BiasCorr + ICP + NuthKaab 	cubic 	3.50 	54.0
> 
> Note the improvement in error if linear or cubic interpolation is done instead of nearest neighbor, but the time it takes increases by almost 2x.

Ok, good to know. But I don't think the resampling is the issue here. I looked at the DEM diff in both cases and because ICP allow also for rotation, the quality of the coregistration was not as good over the entire area.

> The reason interpolation is needed is because I right now apply a matrix by converting the DEM to a point cloud, then applying the transform, and re-gridding it (using `scipy.interpolate.griddata`).
> @adehecq or @rhugonnet , if you know a better approach for applying a transformation matrix to a DEM, I would be delighted to implement that instead!

I think this approach makes sense as it is more generic. For warping an image, I use scikit warp (https://scikit-image.org/docs/dev/api/skimage.transform.html#skimage.transform.warp) or I saw that OpenCV has also an AffineWarp that could be worth looking at.
For a simple shift, another option is simply to update the georeference information, like done by Raster.shift.

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
     )
 
     return aligned_raster, error
+
+
+def _transform_to_bounds_and_res(dem: np.ndarray,
```
adehecq, line 1086: Maybe just pass dem.shape rather than the whole array (in case one does not want to load a large array in the future)

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+
+class Coreg:
+    _meta: Optional[dict[str, Any]] = None  # All __init__ functions should instantiate an empty dict.
+    _fit_called = False  # Flag to check if the .fit() method has been called.
+
+    def __init__(self):
+        """This function should have been overwritten by subclassing."""
+        raise ValueError("Coreg class should not be instantiated directly.")
+
+    def fit(self, reference_dem: Union[np.ndarray, np.ma.masked_array],
```
adehecq, line 1103: Something I wanted to mention, but we should make sure that the functions do not only accept DEMs, but also point clouds. Many people use ICESat/ICESat-2 data as well for coregistration. For most applications, a raster is not needed.
I've added a note in the Project's tab about this.

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+    def _apply_pts_func(self, coords: np.ndarray):
+        """Apply the bias to a given coordinate array."""
+        new_coords = coords.copy()
+        new_coords[:, 2] += self._apply_func(coords[:, 2], None)  # type: ignore
+
+        return new_coords
+
+    def _to_matrix_func(self) -> np.ndarray:
+        """Convert the bias to a transform matrix."""
+        empty_matrix = np.array([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 1]], dtype=float)
```
adehecq, line 1264: You could use
`np.diag(np.ones(4))`

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+    Iterative Closest Point DEM coregistration.
+
+    Estimates a rigid transform (rotation + translation) between two DEMs.
+
+    Requires 'opencv'
+    """
+
+    def __init__(self, max_iterations=100, tolerance=0.05, rejection_scale=2.5, num_levels=6):  # pylint: disable=super-init-not-called
+        """
+        Instantiate an ICP coregistration object.
```
adehecq, line 1282: Maybe put a link to the PDAL ICP documentation for details about the meaning of the parameters?

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+        full_mask = (~ref_mask & ~tba_mask & (np.asarray(inlier_mask) if inlier_mask is not None else True)).squeeze()
+
+        # If subsample is not equal to one, subsampling should be performed.
+        if subsample != 1.0:
+            # If subsample is less than one, it is parsed as a fraction (e.g. 0.8 => retain 80% of the values)
+            if subsample < 1.0:
+                subsample = int(np.count_nonzero(full_mask) * (1 - subsample))
+
+            # Randomly pick N inliers in the full_mask where N=subsample
+            random_falses = np.random.choice(np.argwhere(full_mask.flatten()).squeeze(), int(subsample), replace=False)
```
adehecq, line 1144: It's great to think about a subsampling functionality. However, I see that the amount of data that is loaded is still the same. I would favor an option where only a fraction of the data was loaded instead, to reduce memory usage, but maybe that requires too much re-working?

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+    def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optional[rio.transform.Affine],
+                  weights: Optional[np.ndarray]):
+        """Estimate the rigid transform from tba_dem to ref_dem."""
+        if weights is not None:
+            warnings.warn("ICP was given weights, but does not support it.")
+
+        bounds, resolution = _transform_to_bounds_and_res(ref_dem, transform)
+        points: dict[str, np.ndarray] = {}
+        # Generate the x and y coordinates for the reference_dem
+        x_coords, y_coords = np.meshgrid(
```
adehecq, line 1306: See my comment about subsampling. It could be done at this stage for example, to reduce memory usage.

xdem/coreg.py
adehecq: Well, opencv's documentation, I forgot you use OpenCV's ICP now!

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+        icp = cv2.ppf_match_3d_ICP(self.max_iterations, self.tolerance, self.rejection_scale, self.num_levels)
+        _, residual, matrix = icp.registerModelToScene(points["tba"], points["ref"])
+
+        assert residual < 1000, f"ICP coregistration failed: residual={residual}, threshold: 1000"
+
+        self._meta["matrix"] = matrix
+
+    def _apply_func(self, dem: np.ndarray, transform: rio.transform.Affine) -> np.ndarray:
+        """Apply the coregistration matrix to a DEM."""
+        bounds, resolution = _transform_to_bounds_and_res(dem, transform)
```
adehecq, line 1343: This call seem to almost always be followed by meshgrid to generate a coordinate grid. Maybe have a specific function for this? I think in case for example we implement a subsampling in the future, this would have to be replaced only once.
The only cases where _transform_to_bounds is used is, I believe, to get the resolution, which is actually contained in the transform itself.

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+            coreg._fit_called = True
+
+            tba_dem_mod = coreg._apply_func(tba_dem_mod, transform)
+
+    def _apply_func(self, dem: np.ndarray, transform: rio.transform.Affine) -> np.ndarray:
+        """Apply the coregistration steps sequentially to a DEM."""
+        dem_mod = dem.copy()
+
+        for coreg in self.pipeline:
+            dem_mod = coreg.apply(dem_mod, transform)
```
adehecq, line 1518: That's really fine for now, but in the future, I think this could be done in one step using some matrix magic! ;-)

xdem/coreg.py
```diff
@@ -1063,3 +1081,604 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+        """Apply the coregistration steps sequentially to a set of points."""
+        coords_mod = coords.copy()
+
+        for coreg in self.pipeline:
+            coords_mod = coreg._apply_pts_func(coords_mod)
+
+        return coords_mod
+
+    def _to_matrix_func(self) -> np.ndarray:
+        """Try to join the coregistration steps to a single transformation matrix."""
```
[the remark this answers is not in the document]
adehecq: I agree!

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+        self.max_iterations = max_iterations
+        self.error_threshold = error_threshold
+
+        self._meta: dict[str, Any] = {}
+
+    def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optional[rio.transform.Affine],
+                  weights: Optional[np.ndarray]):
+        """Estimate the x/y/z offset between two DEMs."""
+        bounds, resolution = _transform_to_bounds_and_res(ref_dem, transform)
+        verbose = False  # TODO: Make this an argument somewhere.
```
adehecq, line 1589: Should there be an attribute in Coreg called verbose, that could be passed to each function?

xdem/coreg.py
```diff
@@ -1063,3 +1081,604 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+            # Calculate new elevations from the offset x- and y-coordinates
+            new_elevation = elevation_function(y=east_grid + offset_east, x=north_grid - offset_north)
+
+            # Set NaNs where NaNs were in the original data
+            new_nans = nodata_function(y=east_grid + offset_east, x=north_grid - offset_north)
+            new_elevation[new_nans >= 1] = np.nan
+
+            # Assign the newly calculated elevations to the aligned_dem
+            aligned_dem = new_elevation
+
```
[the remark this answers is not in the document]
adehecq: Mmmh, interesting point...
Regarding point 1, I believe that once Coreg is well setup, most users won't care at all about subfunctions etc. With the suggested documentation, I find it much easier to use than having to look at individual functions, which might possibly have different structures etc. So that's rather a no on this point.
Regarding point 3, I think it's good to have flexibility and allowing to re-use the functions if possible. On the other hand, most users really won't dive that far down in the functionalities. The question is, will we need to access these functions ourselves? Right now I can't think of a test case. If so, I think we can still work around it by calling the class and running the _fit_func and _apply_func functions and reading the output in _meta?
Right now, the only issue I see is that these functions only work with raster data, no point cloud. If we decide to extract the base functions out of the class, I think they should at least accept 3-D and 2-D arrays.

xdem/coreg.py
```diff
@@ -1063,3 +1081,620 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
...
+        """Apply the estimated x/y/z offsets to a DEM."""
+        bounds, resolution = _transform_to_bounds_and_res(dem, transform)
+        scaling_factor = self._meta["resolution"] / resolution
+
+        # Make index grids for the east and north dimensions
+        east_grid = np.arange(dem.shape[1]) * scaling_factor
+        north_grid = np.arange(dem.shape[0]) * scaling_factor
+
+        # Make a function to estimate the DEM (used to construct an offset DEM)
+        elevation_function = scipy.interpolate.RectBivariateSpline(x=north_grid, y=east_grid,
```
adehecq, line 1662: This is redundant with the _fit_func. Should it not be a separate method?

tests/test_coreg.py
```diff
@@ -126,3 +132,235 @@ def test_only_paths():
...
+        # Do full bias corr to compare
+        bias_full = coreg.BiasCorr()
+        bias_full.fit(**self.fit_params)
+
+        # Check that the estimated biases are similar
+        assert abs(bias_sub._meta["bias"] - bias_full._meta["bias"]) < 0.1
+
+        # Test ICP with subsampling
+        icp_full = coreg.ICP(max_iterations=20)
+        icp_sub = coreg.ICP(max_iterations=20)
```
adehecq, line 348: missing a subsample here?

adehecq:
Finally managed to go through all of it!!
Fantastic work Erik! I really like the way it's going. This should make DEM coregistration easy for everyone!
I made some suggestions along the way, but I think this PR should be merged sooner than later. Anything you cannot take into account rapidly can be added as issue or in the projects' tab.

adehecq (approved):

xdem/coreg.py
erikmannerfelt: I'm pretty sure the resolution is not contained in the transform. [See the rio docs](https://rasterio.readthedocs.io/en/latest/api/rasterio.transform.html#rasterio.transform.array_bounds) where the height and width have to be supplied each time a raster-specific operation is made. I believe the Affine transform only contains information of the upper left corner, the scale, and the rotation, but I might be wrong.

Anyway, I cleaned up the code a bit and made a function for x_coords and y_coords  creation, so we can more easily change this in the future if we want to (56a4d1b)

xdem/coreg.py
erikmannerfelt: Mind if we add this as an issue for future improvement?

xdem/coreg.py
erikmannerfelt: Added in 56a4d1b

xdem/coreg.py
erikmannerfelt: Cool! Added in 56a4d1b

xdem/coreg.py
erikmannerfelt: Let's have this as a future improvement! I see this working great with ICP and NuthKaab, but not as easily with dDEM-dependent analyses: Deramp and BiasCorr. I'm sure there's a way, though, and we can explore that in the future!

xdem/coreg.py
erikmannerfelt: Implemented in 2ea683b

xdem/coreg.py
erikmannerfelt: Indeed, I can add an issue for it.

Problems arise with intermediate non-matrixable steps such as Deramp. It could be implemented in a way where all "matrixable" transforms that occur in a row are merged:

pipeline = [A, B, C (matrix not supported), D, E, F]

apply:
A-B, C, D-F

xdem/coreg.py
erikmannerfelt: It is not really redundant. The elevation function is made to apply the sub-pixel shift.

It could theoretically be replaced with a future `apply_matrix()` function that we were discussing, but right now, there's no better way to apply the sub-pixel shift as far as I'm aware.

tests/test_coreg.py
erikmannerfelt: Subsampling is done in the `icp_sub.fit()` step. It's only the `.fit()` method that uses subsampling.

xdem/coreg.py
erikmannerfelt: Added `verbose`  argument and associated callbacks in 1f3e735

erikmannerfelt:
Thank you both for your great feedback! As suggested by @adehecq, maybe we can merge this and add the todo's as issues?

xdem/coreg.py
adehecq: The resolution is in transform[0] and transform[4]. But I agree, to get the bounds, you need the shape. 
My reply contained a typo, what I meant is that in at least one occasion, _transform_to_bounds_and_res was not followed by meshgrid, and in that case, only the resolution was needed, not the bounds, making the call to this function useless.

xdem/coreg.py
adehecq: Yes, but as discussed somewhere else with @rhugonnet, I think all transformations that cannot be converted into a matrix format should go into biascorr.py (this includes deramp) and should be handled a bit differently. So I wouldn't worry to much about deramp here.

xdem/coreg.py
adehecq: Sure!

xdem/coreg.py
adehecq: Oh that would work with deramp and biascorr for sure. My own dem_coregistration.py script was able to handle point data in hdf5 format.
All basic functions actually only need a 1D array of (x, y, z) values. Except of course for NuthKaab were we need to calculate the slope/aspect before "destructuring" the raster. But get_hozitonal_shift does not require 2-D data.

xdem/coreg.py
adehecq: What I meant is that the same 'elevation_function' is created in _fit_func and _apply_func. Maybe it could be written as a separate method of this class... Not a huge deal, but this makes things easier to maintain.
Note that I'm the one to blame here, as these are mostly inspired from my code!

xdem/coreg.py
```diff
@@ -1083,11 +1083,11 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
     return aligned_raster, error
 
 
-def _transform_to_bounds_and_res(dem: np.ndarray,
+def _transform_to_bounds_and_res(shape: tuple[int, int],
```
adehecq, line 1086: Maybe a quick docstring? Mostly to explain what shape is

adehecq: Never mind, it was in a later commit!

adehecq:
> Thank you both for your great feedback! As suggested by @adehecq, maybe we can merge this and add the todo's as issues?

Good to go for me!

PR #83, opened 2021-04-08
Make hyspometric interpolation methods more gap-friendly

adehecq:
These are some fix I had to make to run the local hypsometric interpolation on a ddem with very large gaps, for example with outlines falling completely outside of valid range.
@erikmannerfelt please check if what I've made makes sense and we can discuss how to better account for such cases.

### DIFF
--- a/xdem/ddem.py
+++ b/xdem/ddem.py
@@ -157,7 +157,7 @@ class dDEM(xdem.dem.DEM):   # pylint: disable=invalid-name
                         interpolated_ddem,
                         reference_elevation.data,
                         mask=feature_mask
-                    ).data
+                    )#.data   # This causes an issue at l 172: NotImplementedError: sub-views are not implemented
                 except ValueError as exception:
                     # Skip the feature if too few glacier values exist.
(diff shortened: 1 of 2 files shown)


xdem/ddem.py
```diff
@@ -157,7 +157,7 @@ def interpolate(self, method: str = "linear",
                         interpolated_ddem,
                         reference_elevation.data,
                         mask=feature_mask
-                    ).data
+                    )#.data   # This causes an issue at l 172: NotImplementedError: sub-views are not implemented
```
erikmannerfelt, line 160: This entire statement should probably be encased in a `np.asarray(...)` instead. `xdem.volume.hypsometric_interpolation()` should return a masked_array (but we want ndarrays in this case), which is why I took the data property. 

Unfortunately, `np.ndarray.data` is something completely different from `np.ma.masked_array.data`!!

erikmannerfelt (commented):
Great! I have one suggestion to be a bit more future-proof, but otherwise it looks good.

adehecq:
Solves #84

erikmannerfelt (approved):
Great!!

PR #85, opened 2021-04-12
Improvements to interpolation methods

adehecq:


erikmannerfelt (approved):
Looks great!

xdem/spatial_tools.py
```diff
@@ -11,6 +11,55 @@
...
 
+def get_mask(array: Union[np.ndarray, np.ma.masked_array]) -> np.ndarray:
+    """
+    Return the mask of invalid values, whether array is a ndarray with NaNs or a np.ma.masked_array.
+
+    :param array: Input array.
+
+    :returns invalid_mask: boolean array, True where array is masked or Nan.
+    """
+    return (array.mask | np.isnan(array.data)) if isinstance(array, np.ma.masked_array) else np.isnan(array)
```
erikmannerfelt, line 22: Funny, I had just made a similar function in coreg.py! I suggest that this is named `_get_mask()` instead to hide it from the public API.

xdem/spatial_tools.py
```diff
@@ -11,6 +11,55 @@
...
+
+
+def get_array_and_mask(array: Union[np.ndarray, np.ma.masked_array]) -> (np.ndarray, np.ndarray):
```
erikmannerfelt, line 25: Same here with the underscore to hide it from the public API

xdem/spatial_tools.py
```diff
@@ -11,6 +11,55 @@
...
+        array = array.astype('float32')
+
+    # Convert into a regular ndarray and convert invalid values to NaN
+    array_data = np.asarray(array)
+    array_data[invalid_mask] = np.nan
+
+    return array_data, invalid_mask
+
+
+def get_valid_extent(array: Union[np.ndarray, np.ma.masked_array]) -> tuple:
```
erikmannerfelt, line 50: Underscore suggestion to hide from public API

erikmannerfelt (commented):
Great! I just suggest setting an underscore before the new methods to hide them from the public API.

erikmannerfelt:
Oh, and I was thinking to myself before whether `~np.isfinite` should be used instead of `np.isnan` to remove `inf`s as well?

xdem/volume.py
```diff
@@ -283,3 +284,135 @@ def hypsometric_interpolation(voided_ddem: Union[np.ndarray, np.ma.masked_array]
...
+    The algorithm loops through each features in the vector file.
+
+    The dDEM is assumed to have been created as "voided_ddem = reference_dem - other_dem".
+
+    :param voided_ddem: A dDEM with voids (either an array with nans or a masked array).
+    :param ref_dem: The reference DEM in the dDEM comparison.
+    :param mask: A raster of same shape as voided_ddem and ref_dem, containing a diferent non-0 pixel value for \
+each geometry on which to loop.
+    :param min_coverage: The minimum coverage fraction to be considered for interpolation. Default is 0.2.
+    :param plot: Set to True to display intermediate plots.
```
erikmannerfelt, line 304: Could you add a `:returns:` line? Would be nice to explain what it returns. Something along:

```
:returns: A dDEM which is interpolated within the mask bounds for all areas filling the min_coverage criterion. 
```
or something more concise.

xdem/volume.py
```diff
@@ -283,3 +284,135 @@ def hypsometric_interpolation(voided_ddem: Union[np.ndarray, np.ma.masked_array]
...
+        valid_pixels = np.count_nonzero(local_inlier_mask)
+        coverage[k] = valid_pixels/float(total_pixels)
+
+    # Filter geometries with too little coverage
+    valid_geometry_index = geometry_index[coverage >= min_coverage]
+    print("Found {:d} geometries with sufficient coverage".format(len(valid_geometry_index)))
+
+    idealized_ddem = -9999 * np.ones_like(dem)
+
+    for k, index in enumerate(valid_geometry_index):
```
erikmannerfelt, line 350: Could be nice in the future to experiment with multiprocessing/multithreading here.

erikmannerfelt (commented):
Cool! It's slight repetition of my `xdem.volume.hypsometric_interpolation()` (which I envisioned to be used in both regional and local). I can suggest a simplification in a PR later.

adehecq:
To do:
- add some tests to the new functions
- add other scripts to extrapolate hypsometric bins, especially in the upper part where observations are often missing: one idea is to get a regional hypsometric curve and fit scaled versions of local hypsometric curves to follow the regional curve. Any suggestion on how to best do this @erikmannerfelt @rhugonnet ?

xdem/spatial_tools.py
rhugonnet: we should replace `np.isnan` intp `np.isfinite` here or we might run into errors in some cases (it's rare but happens, I had to do this in pybob a couple years ago)

xdem/volume.py
```diff
@@ -175,23 +233,45 @@ def calculate_hypsometry_area(ddem_bins: Union[pd.Series, pd.DataFrame], ref_dem
     return output
 
 
-def linear_interpolation(array: Union[np.ndarray, np.ma.masked_array]) -> np.ndarray:
+def linear_interpolation(array: Union[np.ndarray, np.ma.masked_array], max_search_distance: int = 10,
+                         extrapolate: bool = False, force_fill: bool = False) -> np.ndarray:
     """
-    Interpolate a 2D array using bilinear interpolation.
+    Interpolate a 2D array using rasterio's fillnodata.
```
rhugonnet, line 239: so is this `rasterio.fill.fillnodata` still a black box, or we now know a bit more what it does?!

xdem/volume.py
```diff
@@ -175,23 +233,45 @@ def calculate_hypsometry_area(ddem_bins: Union[pd.Series, pd.DataFrame], ref_dem
...
-    nan_mask = (array.mask | np.isnan(array.data)) if isinstance(array, np.ma.masked_array) else np.isnan(array)
+    nan_mask = xdem.spatial_tools.get_mask(array)
+
+    interpolated_array = rasterio.fill.fillnodata(array.copy(), mask=(~nan_mask).astype("uint8"),
+                                                  max_search_distance=max_search_distance)
 
-    interpolated_array = rasterio.fill.fillnodata(array.copy(), mask=(~nan_mask).astype("uint8"))
+    # Remove extrapolated values
+    if not extrapolate:
+        interp_mask = cv2.morphologyEx((~nan_mask).squeeze().astype('uint8'), cv2.MORPH_CLOSE,
```
rhugonnet, line 257: I'd need a comment here to understand exactly what this does.
I'm not sure we need a closing, getting the dilation of `nan_mask` should be sufficient, no?

xdem/volume.py
```diff
@@ -266,3 +342,146 @@ def hypsometric_interpolation(voided_ddem: Union[np.ndarray, np.ma.masked_array]
     assert output is not None
 
     return output
+
+
+def local_hypsometric_interpolation(voided_ddem: Union[np.ndarray, np.ma.masked_array],
```
rhugonnet, line 347: In terms of performance/structure, I suggest moving away from the method of using a `mask` with multiple values for each feature of a geometry. It's almost instantaneous to generate a mask for a single feature, so the loop over shapefile features should be outside this function (especially given that this is the `local_hypsometric_method`).
In particular, it will be more easy/efficient to subdivide tasks and parallelize using multiprocessing this way (echoes Erik's comment below),

rhugonnet (commented):
> add other scripts to extrapolate hypsometric bins, especially in the upper part where observations are often missing: one idea is to get a regional hypsometric curve and fit scaled versions of local hypsometric curves to follow the regional curve. Any suggestion on how to best do this @erikmannerfelt @rhugonnet ?

I see several approaches for the regional part.
0/ **Existing:** fill with the mean regional signal at a given elevation (Fanny's paper, and in Bob's paper, I don't like this much and I quote Bob's results here: _the global hypsometric methods are furthest from the true values, which is perhaps not surprising in a region with a high variability of glacier elevation changes_)
1/ **Easy:** loop through all local hypsometric `pd.DataFrame`, normalize those by their hypsometry and apply the area-weighted mean dhdt at the same normalized hypsometry to gap fill: at least a bit more representative of the regional signal independently of the elevation location at the scale of each glacier, but can still be off due to high inter-regional variability
2/ **Best**: loop through all local hypsometric `pd.DataFrame`, normalize by both hypsometry and elevation change and derive a mean normalized signal for the region; then, for glacier with local hypsometric (maybe with a criteria covering at least X% of the area), de-normalize by an estimated range of elevation change rate. If the rate at high elevation is not available (accumulation area full of voids), you can use the signal at half of the normalized elevation range, or maybe the best would be to use the mean elevation change rate at highest elevation captured in the region (essentially, you scale the regional hypsometric signal only by the dhdt in the ablation area of each glacier, keeping the accumulation dhdt similar than where measured for others in the region).

rhugonnet:
For 2/, the best would be a Bayesian approach, but you don't have time for this right now.
@adehecq I think that the detailed steps should be:
a) Normalize all local hypsometric estimation with more than X% coverage in accumulation area by both hypsometry and elevation change rate.
b) Derive area-weighted, mean hypsometric regional normalized signal from everything in a).
c) Derive area-weighted, mean regional elevation change rate at highest elevation from everything in a).
d) For glaciers with more than X% coverage in ablation area, de-normalize the regional signal of b) with the maximum dhdt of the glacier and the dhdt at highest elevation of c).
e) For glaciers with less than X% coverage in ablation area (glacier with pretty much no coverage), use with the mean elevation change rate in the region? (or for a more beautiful visual rendering, fill with the mean hypsometric signal without normalization)

What do you think?

xdem/volume.py
erikmannerfelt: It's using [gdal_fillnodata](https://gdal.org/api/gdal_alg.html#_CPPv414GDALFillNodata15GDALRasterBandH15GDALRasterBandHdiiPPc16GDALProgressFuncPv). Reading the documentation, it actually looks to be an IDW approach. Or am I wrong??

xdem/volume.py
erikmannerfelt: Interesting. @adehecq  and I actually argued (without empiric basis) for the opposite, that we wanted to avoid multiple rasterize calls to improve performance! Should we benchmark the two approaches, or do you already know that your suggested approach (and the one I am already using for the `dDEM.interpolate("local_hypsometric")`) is more efficient, @rhugonnet, or should we try to figure that out?

I guess the most important part is that it is understandable for the user. 

What if, instead of giving a `mask` with integers representing the glaciers, the function just takes an iterable of `shapely.geometry.base.BaseGeometry` instances and will fill every outline that meets the `min_coverage` criterion?

erikmannerfelt:
> > add other scripts to extrapolate hypsometric bins, especially in the upper part where observations are often missing: one idea is to get a regional hypsometric curve and fit scaled versions of local hypsometric curves to follow the regional curve. Any suggestion on how to best do this @erikmannerfelt @rhugonnet ?
> 
> I see several approaches for the regional part.
> 0/ **Existing:** fill with the mean regional signal at a given elevation (Fanny's paper, and in Bob's paper, I don't like this much and I quote Bob's results here: _the global hypsometric methods are furthest from the true values, which is perhaps not surprising in a region with a high variability of glacier elevation changes_)
> 1/ **Easy:** loop through all local hypsometric `pd.DataFrame`, normalize those by their hypsometry and apply the area-weighted mean dhdt at the same normalized hypsometry to gap fill: at least a bit more representative of the regional signal independently of the elevation location at the scale of each glacier, but can still be off due to high inter-regional variability
> 2/ **Best**: loop through all local hypsometric `pd.DataFrame`, normalize by both hypsometry and elevation change and derive a mean normalized signal for the region; then, for glacier with local hypsometric (maybe with a criteria covering at least X% of the area), de-normalize by an estimated range of elevation change rate. If the rate at high elevation is not available (accumulation area full of voids), you can use the signal at half of the normalized elevation range, or maybe the best would be to use the mean elevation change rate at highest elevation captured in the region (essentially, you scale the regional hypsometric signal only by the dhdt in the ablation area of each glacier, keeping the accumulation dhdt similar than where measured for others in the region).

@rhugonnet as you may have inferred from your comment, we have the number "0" (`xdem.volume.hypsometric_interpolation`), which just tries to find the trend in all glacier values. It's not very good, as you mentioned.

I could get on the other two in a PR. It would be nice if this was merged first, however.

xdem/spatial_tools.py
adehecq: Decided to leave it as it is for now.

xdem/volume.py
adehecq: Yes, the documentation says inverse distance weighting, so that's what it should be.

xdem/volume.py
adehecq: It will remove all pixels that have information provided from one side only, i.e. extrapolated.
The closing makes sure that gaps the size of the search distance are kept, but all areas that have grown artificially are not.

xdem/volume.py
adehecq: It would be worth testing both approaches indeed.
But I think the approach I implemented works better for a large raster with small geometries. In that case, I disagree with you @rhugonnet that generating a mask is almost instantaneous, as one need to create e.g. a 20k x 20k array each time for only a few valid pixels...
Ideally, for each geometry, one would crop the raster, create the associated mask and pass it to the generic hypsometric approach.

adehecq: I've added an issue to remember to test both approaches.

xdem/volume.py
adehecq: Added to issue #96

xdem/spatial_tools.py
adehecq: Added in c4c7e20

adehecq:
> I could get on the other two in a PR. It would be nice if this was merged first, however.

@erikmannerfelt I copied most of the important stuff into issues, so feel free to merge this PR and work on the issues if you'd like. If you have time on Monday that would be great! Otherwise, I'll leave this open until I have time to look at it.

Issue #82, opened 2021-04-08
Add a general-purpose `apply_matrix()` function for DEMs


erikmannerfelt:
In #71, we discussed errors associated to DEM gridding, since ICP (and possibly more future approaches) need to convert the DEM into a point cloud. ICP currently does:

1. Convert DEM to point cloud (preparation)
2. Match the ref and tba point clouds (estimates the rigid transform)
3. Re-grid the tba DEM after transforming it (applies the rigid transform) 

Step 3 currently uses `scipy.interpolate.griddata` which is extremely slow! A faster and more reliable approach would be fantastic to have. Something with the approximate signature:

```python
def apply_matrix(dem: np.ndarray, transform: rio.transform.Affine, matrix: np.ndarray) -> np.ndarray:
    # do matrix magic
    return transformed_dem

``` 

One idea I had was to describe the 3D rigid transform as a 2D similarity transform and a linear vertical correction:

A 3D rigid transform from above (from 2D) would include horizontal translation, horizontal rotation, horizontal scaling (due to rotation-induced perspective differences), and elevation (bias + rotation) differences. As far as I understand, the first 3 parts could be described as a 2D similarity transform, and the final as a deramp.

The advantage of such an approach is that the DEM never has to be converted to a point cloud and regridded; it just needs to be moved and twisted a bit. I guess the most difficult part of this, however, would be to mathematically do the conversion...

This general-purpose function is integral for #79 to implement.

I am also all-ears to other alternatives.

erikmannerfelt:
In my sleep, I came up with a hacky solution that should work for now:

1. Transform a set of arbitrary points using the 3D rigid transform.
2. Estimate a 2D (horizontal) SimilarityTransform using `skimage.transform.estimate_transform(orig_points2D, trans_points2D, "similarity")`
3. Estimate a (vertical) deramp using the trusty xdem tools
4. Apply the SimilarityTransform using `skimage` and then apply the deramp.

The great advantage of this is that the input DEM is never converted into a point cloud, and thus doesn't need regridding, but is just warped (`skimage`) and value corrected (`deramp`). 

Of course, the best would be to estimate the SimilarityTransform and Deramp mathematically instead of empirically deriving them, but that's beyond my math knowledge!

erikmannerfelt:
Added in #87

Issue #79, opened 2021-04-08
Improve `CoregPipeline.apply()` and `.fit()` performance by merging matrices


erikmannerfelt:
If parts or all steps of a pipeline can be described as matrices, the current apply loop is inefficient and might induce resampling-related errors:

https://github.com/GlacioHack/xdem/blob/82b87ebb3dd3393d8bc854ca1b0d8d0c6791fb0d/xdem/coreg.py#L1531

A better approach would be to merge transformation matrices (when possible) and therefore do fewer resampling iterations, potentially lowering the error and the processing time.

A conceptual example is:

1. A) BiasCorr (matrixable)
2. B) ICP (matrixable)
3. C) NuthKaab (matrixable)
4. D) Deramp(degree=3) (not matrixable)

Currently, the pipeline would modify the DEM four times. By combining the matrices of A, B and C, only two modifications are needed:

1. A-C
2. D

This could be done by first implementing a property of `Coreg`, for example called `is_rigid` which is True if it can be represented as a rigid transform (4x4 matrix). Then, the following pseudocode could work:

```python
dem_mod = dem.copy()

matrices: list[np.ndarray] = []
for coreg in self.pipeline:
    if coreg.is_rigid:
        # Append the matrix to the list of matrices to be applied at the same time.
        matrices.append(coreg.to_matrix())
    else:   # This marks when matrices can no longer be added, and need to be applied before the non-rigid apply is done
        # Use matrix-magic to combine the matrices
        combined_matrix = combine_matrix(matrices)
        # Use more matrix-magic to transform the DEM
        apply_matrix(dem_mod, combined_matrix)
        # Reset the matrices list
        matrices.clear()
        # Apply the non-rigid transform
        coreg.apply(dem_mod, transform=transform)

if len(matrices) > 0:
     # do the same as above to apply all the matrices
```
where the `combine_matrix()` and `apply_matrix()` functions need to be created.

PR #87, opened 2021-04-13
General `apply_matrix()` function

erikmannerfelt:
Brought up in #82 and other issues, we need a general-purpose function for applying transformation matrices to DEMs. 

Currently, this is not done consistently. The `NuthKaab` uses `scipy.interpolate.RectBivariateSpline()` (works well but only horizontally) and `ICP` converts the DEM into a point cloud, transforms it, and re-grids it with `scipy.interpolate.griddata()` (works but often introduces weird artefacts and is terribly slow).

This function aims to standardize the way DEMs are transformed. It works in a few steps:

1. Convert the DEM into a PC (not for gridding, only analysis)
2. Transform the PC using the matrix
3. Estimate and apply a deramp from the elevation difference of each point in the PC vs. transformed PC
4. Convert the horizontal transformed coordinates to index coordinates (indices specify where the pixels should be displaced)
5. Transform the DEM using the new indices.
6. Transform the nodata mask using the same warping.

It sounds complex, but it works really well! It is hundreds of times faster than the `scipy.interpolate.griddata` approach and it works with any rigid transform.

### It opens up for many new improvements:

1. The pipeline optimization suggested in #79 can be implemented (matrices can be merged and then a transformation is only made once)
2. As found by @adehecq, my "ultimate" `BiasCorr + ICP + NuthKaab` currently does not actually  improve the results compared to only `NuthKaab` because of the `ICP.apply()` call. This new function should fix that.
3. The `Coreg` class can be simplified slightly, since every transformation that can be explained by a rigid transform could automatically use the `apply_matrix()` function.
4. Generic `Coreg` objects could be created from a `Coreg.from_matrix()` constructor method or similar. This may be sought if a transformation matrix is already known and the user wants to apply it to a DEM, e.g.:
```python
known_x_offset = 4.9  # m
matrix = np.diag(np.ones(4))
matrix[0, 3] = known_x_offset

coreg = xdem.coreg.Coreg.from_matrix(matrix)

coreg.apply(dem, transform=transform)
```
Another possible constructor method could be:
```python
coreg = xdem.coreg.Coreg.from_translation(x=4.9)  # This creates a 4x4 matrix like the example above.
coreg.apply(dem, transform=transform)
```

### Before this is merged:
1. The `Coreg.apply()` methods should use this function
2. More tests are needed. Right now, I've tried with translations, biases and rotations, but not all at the same time for example.


xdem/coreg.py
```diff
@@ -1712,3 +1714,144 @@ def _to_matrix_func(self) -> np.ndarray:
...
+    """Will be replaced by Amaury's version once that is merged."""
+    nans = ~np.isfinite(np.asarray(array))
+
+    if isinstance(array, np.ma.masked_array):
+        nans = nans | array.mask
+
+    return nans
+
+
+def apply_matrix(dem: np.ndarray, transform: rio.transform.Affine, matrix: np.ndarray, invert: bool = False,
```
rhugonnet, line 1739: This is great work, yet I keep thinking there must be a way to apply all of this simultaneously (and that would probably be more efficient computationally). There's nothing like this in pytransform3d?

rhugonnet (dismissed):
Great work, another step forward towards consistency and performance. :+1:

tests/test_coreg.py
```diff
@@ -366,3 +368,101 @@ def test_subsample(self):
...
+        matrix = np.diag(np.ones(4, dtype=float))
+        matrix[0, 3] = pixel_shift * self.tba.res[0]
+        matrix[2, 3] = -bias
+
+        transformed_dem = coreg.apply_matrix(shifted_dem.data.squeeze(), self.ref.transform, matrix)
+
+        diff = np.asarray(self.ref.data.squeeze() - transformed_dem)
+
+        # Check that the median is very close to zero
+        assert np.abs(np.nanmedian(diff)) < 0.05
```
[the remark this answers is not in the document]
erikmannerfelt: In a best case scenario, yes. But since cubic interpolation is done, there is actually a small difference. There could be a check in the start to see if the shift is by an integer amount (whereby it will only shift appropriately) or if the more complex warp should be made. Does that sound okay?

tests/test_coreg.py
```diff
@@ -366,3 +368,101 @@ def test_subsample(self):
...
+        # Check that the NMAD is low
+        assert spatial_tools.nmad(diff) < 3
```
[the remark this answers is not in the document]
erikmannerfelt: Again, it will not be exactly equal due to the warping.

tests/test_coreg.py
```diff
@@ -366,3 +368,101 @@ def test_subsample(self):
...
+            return matrix
+
+        rotation = 4
+        rotated_dem = coreg.apply_matrix(
+            self.ref.data.squeeze(),
+            self.ref.transform,
+            rotation_matrix(rotation),
+        )
+        # Make sure that the rotated DEM is way off.
+        assert np.abs(np.nanmedian(rotated_dem - self.ref.data.data)) > 400
```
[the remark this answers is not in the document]
erikmannerfelt: The DEM is about 20 km tall, so the theoretical difference at the other end after rotating is:
```python
In [1]: 20000 * np.sin(np.deg2rad(4))
Out [1]: 1395.129474882506
```
meaning the midpoint should be off by about half; ~700 m.

tests/test_coreg.py
```diff
@@ -366,3 +368,101 @@ def test_subsample(self):
...
+
+        # Apply a rotation in the opposite direction
+        unrotated_dem = coreg.apply_matrix(
+            rotated_dem,
+            self.ref.transform,
+            rotation_matrix(-rotation * 0.989)  # This is not exactly -rotation, probably due to displaced pixels.
```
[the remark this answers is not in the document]
erikmannerfelt: Yeah... I don't know where that is from either...  `apply_matrix()` estimates a deramp. I'd love any suggestions to figure out the cause behind this magic number!

tests/test_coreg.py
```diff
@@ -366,3 +368,101 @@ def test_subsample(self):
...
+            plt.xlim(*extent[:2])
+            plt.ylim(*extent[2:])
+            plt.subplot(155)
+            plt.title("Original vs. de-rotated")
+            plt.hist(diff[np.isfinite(diff)], bins=np.linspace(-10, 10, 100))
+            plt.tight_layout(w_pad=0.05)
+            plt.show()
+
+        # Check that the median is very close to zero
+        assert np.abs(np.nanmedian(diff)) < 0.5
```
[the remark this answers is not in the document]
erikmannerfelt: It's often around 0.05, so I could definitely reduce the threshold a bit. It's just to make sure that the tests pass!

xdem/coreg.py
```diff
@@ -1712,3 +1714,144 @@ def _to_matrix_func(self) -> np.ndarray:
...
+        nan_mask.astype("uint8"),
+        inds,
+        order=resampling_order,
+        mode="constant",
+        cval=1,
+        preserve_range=True
+    ) > 0.5  # Due to different interpolation approaches, everything above 0.5 is assumed to be 1 (True)
+
+    # Apply the transformed nan_mask
+    transformed_dem[tr_nan_mask] = np.nan
```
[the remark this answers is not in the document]
erikmannerfelt: Sorry, but could you elaborate here? The warp function has a fill value of 1 (`cval=1`), meaning all values outside the frame (near the border) should be 1 (True).

xdem/coreg.py
erikmannerfelt: Thanks! I am certain there may be a simpler way mathematically:

"3D rigid transform"  -> "2D projective transform" + "pixel value correction"

I'm just not that good at math, unfortunately!

erikmannerfelt:
I made multiple additions. First, the Coreg class can be created from a matrix or translation:
```python
known_x_offset = 4.9  # m
matrix = np.diag(np.ones(4))
matrix[0, 3] = known_x_offset

coreg = xdem.coreg.Coreg.from_matrix(matrix)

coreg.apply(dem, transform=transform)

```
or:
```python

coreg = xdem.coreg.Coreg.from_translation(x_off=4.9)  # This creates a 4x4 matrix like the example above.
coreg.apply(dem, transform=transform)
```
This may come in handy in future `.save()` and `.load()` functions (in case coregistration takes time and one wants to store the transform for later).

## Coreg subclassing simplification
Second, the new `apply_matrix()` function is now used by default in the `.apply()` method.  If `._apply_func()` has been implemented, it will use this instead.

The same default behaviour is made for `apply_pts()` method which will try to apply the coregistration as a rigid transform if `._apply_pts_func()` has not been defined (which overrides the default behaviour).

Finally, if a matrix exists in `self._meta["matrix"]`, the method `._to_matrix_func()` is not needed, and `.to_matrix()` will just find the dictionary version instead.

What this all means is that Coreg subclassing is now really easy.

For example, the `ICP` subclass only overrides the `.__init__()` and `._fit_func()` methods, and everything else is inherited! Similarly, `NuthKaab` only implements three methods: `.__init__()`, `._fit_func()` and  `_to_matrix_func()`. This improvement will hopefully make future extensions much easier, and will probably make maintaining everything trivial. Finally, it's less code so that's great.

rhugonnet:
> What this all means is that Coreg subclassing is now really easy.

Really neat. When an implementation of code structure means less code, less duplication, and same results, I guess it means that it's the right one :wink: 
Good job!

tests/test_coreg.py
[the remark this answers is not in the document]
rhugonnet: Mmmh.. yes I guess it would be good to avoid resampling the DEM by too much when it's not necessary! Bilinear resampling has a good behaviour for this (if the shift is an integer, the resampling is right on the point, so same value). 
Ultimately this should be the choice of the user, so we can probably leave as is.

tests/test_coreg.py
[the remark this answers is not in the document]
rhugonnet: Indeed (we could add it in the comment of the test)

tests/test_coreg.py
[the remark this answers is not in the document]
rhugonnet: :+1:

xdem/coreg.py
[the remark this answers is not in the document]
rhugonnet: What I'm trying to say is that when you create a `filled_dem` with `-9999` at the NaN values (or anything else) in order to run `skimage.transform.warp`, those will be used in the interpolations with resampling_order> 0. Because of this, it will create a bias for non-NaN values that are close to NaN values (within 3 pixels for a resampling order of 3, for instance).
As a consequence, it could be a good idea to extend the NaN mask using a binary dilation with the same order than the resampling method used (like it is done when deriving slope, for instance).
Or maybe I am missing something? (I didn't use skimage much)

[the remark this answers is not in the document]
erikmannerfelt: But if the nan mask is dilated, won't it be smaller and smaller for every iteration? So: 
```python
while True:
    dem = apply_matrix(dem, transform, empty_matrix)
```
Would eventually mean that `dem` is empty, right?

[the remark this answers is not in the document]
rhugonnet: Yes, that's the behaviour! GDAL behaves similarly. If you resample a raster with resampling_order>0 interpolation 10 times, the NaNs will propagate into bigger and bigger data gaps and eventually you will be left with nothing.
It should be only an option for sure, but someone might want to avoid using data at the edges after resampling?

I'm still not sure I get 100% how skimage deals with the nodata here: are the filled values of "-9999" actually used in the resampling and then masked later? This would create strong edge effects by including the -9999 in the resampling.

rhugonnet (dismissed):
All good! Seeing you moving forward so much, I can't wait to get back to xdem! Everything else has been so time-consuming recently...

We can just open a "suggestion" issues for my unresolved comments: 1/ NaN dilation issue + 2/ Apply rigid transform in one go.
On 2/, there is no such method provided in point cloud libraries for rigid transform? This seems surprising to me... or maybe I missed part of the problem! I'm not too familiar with opencv and such.
I was looking around a bit and stumbled upon this: https://github.com/nghiaho12/rigid_transform_3D
Would this be useful?

xdem/coreg.py
```diff
@@ -1084,7 +1086,7 @@ def coregister(reference_raster: Union[str, gu.georaster.Raster], to_be_aligned_
     return aligned_raster, error
 
 
-def _transform_to_bounds_and_res(shape: tuple[int, int],
+def _transform_to_bounds_and_res(shape: tuple[int, ...],
```
adehecq, line 1089: What do the '...' mean?

xdem/coreg.py
```diff
@@ -1205,16 +1227,79 @@ def apply_pts(self, coords: np.ndarray) -> np.ndarray:
...
+        return transformed_points
+
+    @property
+    def is_rigid(self) -> bool:
+        """Check if the transform be explained by a rigid transform."""
+        # _is_rigid is found by seeing if to_matrix() raises an error.
+        # If this hasn't been done yet, it will be None
+        if self._is_rigid is None:
+            try:  # See if to_matrix() raises an error.
+                self.to_matrix()
```
adehecq, line 1253: I understand from this that your definition of 'rigid' is that it an be expressed as a 4x4 matrix?
Scaling can be expressed using a 4x4 matrix, but is not a rigid transformation, so I'm not sure the term is appropriate here. Or did I miss something?

erikmannerfelt: Good point, `is_rigid` doesn't work... What about `is_helmert` or something (after the [Helmert transform](https://en.wikipedia.org/wiki/Helmert_transformation)). Open to other suggestions!

What it should mean is exactly what you ask; "can it be described as a 4x4 matrix?"

xdem/coreg.py
erikmannerfelt: It means the tuple can have an arbitrary length. The shape of an ndarray can't be known by the linter before running, so it needs to be arbitrary.

The linters are also stupid the way that you can't write:
```python
dem_shape: tuple[int, int] = dem.shape[:2]
```
Slicing doesn't work, even though it really should...

Although you could write:
```python
dem_shape: tuple[int, int] = (dem.shape[0], dem.shape[1])
```
but at that point, it just felt so stupid that I reverted to `tuple[int, ...]` as the type...

xdem/coreg.py
rhugonnet: Or `is_affine`?

rhugonnet: Source: https://www.euclideanspace.com/maths/geometry/affine/matrix4x4/index.htm

erikmannerfelt:
> All good! Seeing you moving forward so much, I can't wait to get back to xdem! Everything else has been so time-consuming recently...
> 
> We can just open a "suggestion" issues for my unresolved comments: 1/ NaN dilation issue + 2/ Apply rigid transform in one go.
> On 2/, there is no such method provided in point cloud libraries for rigid transform? This seems surprising to me... or maybe I missed part of the problem! I'm not too familiar with opencv and such.
> I was looking around a bit and stumbled upon this: https://github.com/nghiaho12/rigid_transform_3D
> Would this be useful?

Thanks for the suggestions, @rhugonnet! I can definitely add 1/ as an issue. Regarding 2/ (and your suggested package), the problem is that 3D gridding takes a lot of time, so this `apply_matrix()` only does 2D gridding and then a value correction. The script you link needs a re-gridding afterward.

xdem/coreg.py
erikmannerfelt: Hmm, maybe that could work... It just shouldn't be mixed up with 2D affine transforms, like `rio.transform.Affine`, which is very different from a 3D "affine" transform.

xdem/coreg.py
```diff
@@ -1664,55 +1715,154 @@ def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optiona
...
-        shifted_dem[new_nans >= 1] = np.nan
+def invert_matrix(matrix: np.ndarray) -> np.ndarray:
+    """Invert a transformation matrix."""
+    with warnings.catch_warnings():
+        # Deprecation warning from pytransform3d. Let's hope that is fixed in the near future.
+        warnings.filterwarnings("ignore", message="`np.float` is a deprecated alias for the builtin `float`")
 
-        shifted_dem += self._meta["bias"]
+        # Invert the transform if wanted.
+        return pytransform3d.transformations.invert_transform(matrix)
```
adehecq, line 1738: Does it differ from numpy.linalg.inv?

xdem/coreg.py
```diff
@@ -1664,55 +1715,154 @@ def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optiona
...
+            f"`{resampling}` is not a valid resampling mode."
+            " Choices: [`nearest`, `bilinear`, `cubic`] or an integer."
+        )
+    # Copy the DEM to make sure the original is not modified, and convert it into an ndarray
+    demc = np.array(dem)
+
+    # Check if the matrix only contains a Z correction. In that case, only shift the DEM values by the bias.
+    empty_matrix = np.diag(np.ones(4, float))
+    matrix_diff = matrix - empty_matrix
+    if abs(matrix_diff[matrix_diff != matrix_diff[2, 3]].mean()) < 1e-4:
```
adehecq, line 1776: Would that not cause a problem if the matrix has other elements that are equal to element [2, 3]?

adehecq: Also, should it not be strictly equal to 0? The 1e-4 value is somewhat arbitrary here...
In the worse case, we're just doing extra work for nothing.

xdem/coreg.py
erikmannerfelt: Seems like it's the same! I'll change to `np.linalg.inv`
```python
In [1]: import numpy as np

In [2]: import pytransform3d.transformations

In [3]: m = pytransform3d.transformations.random_transform()
In [4]: m
Out[4]: 
array([[-0.69885409, -0.18464974, -0.69101913,  1.45427351],
       [-0.57588809,  0.71823881,  0.39049446,  0.76103773],
       [ 0.42421206,  0.67084834, -0.60828171,  0.12167502],
       [ 0.        ,  0.        ,  0.        ,  1.        ]])

In [5]: assert np.array_equal(pytransform3d.transformations.invert_transform(m), np.linalg.inv(m))
```

xdem/coreg.py
```diff
@@ -1664,55 +1715,154 @@ def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optiona
...
+        return demc + matrix[2, 3]
+
+    # Temporary. Should probably be removed.
+    demc[demc == -9999] = np.nan
```
adehecq, line 1780: why is that?

xdem/coreg.py
erikmannerfelt: I tried to find a syntax for "slice all values except for this one" but I couldn't find out how to!

xdem/coreg.py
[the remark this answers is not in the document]
adehecq: Yes, definitely something worth checking and testing. Have you tried running the function on a DEM with gaps to see how it behaves?

xdem/coreg.py
```diff
@@ -1664,55 +1715,154 @@ def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optiona
...
 
-    def _apply_pts_func(self, coords: np.ndarray) -> np.ndarray:
-        """Apply the estimated x/y/z offsets to a set of points."""
-        offset_east = self._meta["offset_east_px"] * self._meta["resolution"]
-        offset_north = self._meta["offset_north_px"] * self._meta["resolution"]
+def apply_matrix(dem: np.ndarray, transform: rio.transform.Affine, matrix: np.ndarray, invert: bool = False,
+                 centroid: Optional[tuple[float, float, float]] = None,
+                 resampling: Union[int, str] = "cubic") -> np.ndarray:
+    """
+    Apply a transformation matrix to a DEM.
```
adehecq, line 1745: Can you maybe explain here the different steps to get the result? It is definitely non trivial, and probably not very standard.

adehecq:
Looks very promising! Some parts are still a bit of black magic, but hopefully we can improve it later!
Some minor suggestions in the code, but the overall structure seem good!

tests/test_coreg.py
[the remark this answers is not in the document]
adehecq: Yes, if you think the issue comes from cubic interpolation, why not test it with bilinear interpolation here? I agree, somehow it should be exactly equal...

xdem/coreg.py
erikmannerfelt: Removed in 5c5777e

xdem/coreg.py
erikmannerfelt: Changed the invert_matrix to be a bit more useful in 5c5777e and also sanity-check the matrix. Therefore, pytransform3d is still necessary, and can be kept as is.

tests/test_coreg.py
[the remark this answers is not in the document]
adehecq: Isn't this because of your center of rotation? If I understand your code correctly, the centroid is at the bottom left corner, so this will affect your result.
Should the centroid not be at the center of the DEM?

xdem/coreg.py
erikmannerfelt: I tried to explain in 0c9d568. Please let me know if you have any suggestions for improvement!

tests/test_coreg.py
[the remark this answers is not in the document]
erikmannerfelt: Bilinear worked much better. I've set that as the default from now on in b1ffc81.

xdem/coreg.py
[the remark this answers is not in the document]
erikmannerfelt: I've added a `dilate_mask` argument in 7095024, @rhugonnet.

I also changed the temporary nodata value to be the nanmedian (instead of -9999), which I guess should reduce the risk of extreme outliers.

The tests include nans, @adehecq, so that does not seem to be a problem.

xdem/coreg.py
erikmannerfelt: Changed `is_rigid` to `is_affine` in be9e9fa.

tests/test_coreg.py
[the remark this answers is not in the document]
erikmannerfelt: Changed so the centroid is the centre coordinate of the raster bounds in 07bf199.

erikmannerfelt:
Thanks for your feedback, @adehecq and @rhugonnet! The function is definitely improved thanks to your suggestions. Let me know what you think of its structure whenever you have time!

adehecq (approved):
I've forgotten the details of it now that time passed, but I checked the commits after my last comments and it seems reasonable. I think we can accept as it is for now, and if issues arise, we'll handle it later.

Issue #78, opened 2021-04-08
Move `coreg.BiasCorr` and `coreg.Deramp` to `biascorr.py`?


erikmannerfelt:
To better fit the names of the script files, these may fit better in `biascorr.py`. 

I strongly suggest that they still inherit from the `Coreg` class, however, to be fully compatible with each other. I suggest the same for e.g. `biascorr.AlongTrack()`.

rhugonnet:
Fixed by #158

PR #91, opened 2021-04-13
Add elevation scale correction `Coreg` approach.

erikmannerfelt:
Sometimes when working with nadir photogrammetric reconstructions, dDEMs can show elevation-dependent differences due to a miscalculated focal length. Obviously, that is a sign that one should probably redo the photogrammetry! If that is not possible, however, this correction may be used instead.

This function estimates a linear or polynomial model with altitude and corrects for it. Example usage:

```python
dem = xdem.DEM(examples.FILEPATHS["longyearbyen_ref_dem"])

factor = 1.2   # Simulated scale offset
scaled_dem = dem.data * factor

zcorr = xdem.coreg.ZScaleCorr(degree=1)
zcorr.fit(dem.data, scaled_dem)

unscaled_dem = xcorr.apply(dem.data) 

diff = (dem.data - unscaled_dem).filled(np.nan)
assert np.abs(np.nanmedian(diff)) < 0.01
```
This passes, as shown by the tests.

As per #78, maybe this should be moved to `biascorr.py` in the future.

### DIFF
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -13,6 +13,7 @@ import time
 import warnings
 from typing import Any
 
+import cv2
 import geoutils as gu
 import numpy as np
 
@@ -366,3 +367,62 @@ class TestCoregClass:
         matrix_diff = np.abs(icp_full.to_matrix() - icp_sub.to_matrix())
         # Check that the x/y/z differences do not exceed 30cm
         assert np.count_nonzero(matrix_diff > 0.3) == 0
+
+    def test_z_scale_corr(self):
+        warnings.simplefilter("error")
+
+        # Instantiate a Z scale correction object
+        zcorr = coreg.ZScaleCorr()
+
+        # This is the z-scale to multiply the DEM with.
+        factor = 1.2
+        scaled_dem = self.ref.data * factor
+
+        # Fit the correction
+        zcorr.fit(self.ref.data, scaled_dem)
+
+        # Apply the correction
+        unscaled_dem = zcorr.apply(scaled_dem, None)
+
+        # Make sure the difference is now minimal
+        diff = (self.ref.data - unscaled_dem).filled(np.nan)
+        assert np.abs(np.nanmedian(diff)) < 0.01
+
+        # Create a spatially correlated error field to mess with the algorithm a bit.
+        corr_size = int(self.ref.data.shape[2] / 100)
+        error_field = cv2.resize(
+            cv2.GaussianBlur(
+                np.repeat(np.repeat(
+                    np.random.randint(0, 255, (self.ref.data.shape[1]//corr_size,
+                                               self.ref.data.shape[2]//corr_size), dtype='uint8'),
+                    corr_size, axis=0), corr_size, axis=1),
+                ksize=(2*corr_size + 1, 2*corr_size + 1),
+                sigmaX=corr_size) / 255,
+            dsize=(self.ref.data.shape[2], self.ref.data.shape[1])
+        )
+
+        # Create 50000 random nans
+        dem_with_nans = self.ref.data.copy()
+        dem_with_nans.mask = np.zeros_like(dem_with_nans, dtype=bool)
+        dem_with_nans.mask.ravel()[np.random.choice(dem_with_nans.data.size, 50000, replace=False)] = True
+
+        # Add spatially correlated errors in the order of +- 5 m
+        dem_with_nans += error_field * 3
+
+        # Try the fit now with the messed up DEM as reference.
+        zcorr.fit(dem_with_nans, scaled_dem)
+        unscaled_dem = zcorr.apply(scaled_dem, None)
+        diff = (dem_with_nans - unscaled_dem).filled(np.nan)
+        assert np.abs(np.nanmedian(diff)) < 0.05
+
+        # Try a second-degree scaling
+        scaled_dem = 1e-4 * self.ref.data ** 2 + 300 + self.ref.data * factor
+
+        # Try to correct using a nonlinear correction.
+        zcorr_nonlinear = coreg.ZScaleCorr(degree=2)
+        zcorr_nonlinear.fit(dem_with_nans, scaled_dem)
+
+        # Make sure the difference is minimal
+        unscaled_dem = zcorr_nonlinear.apply(scaled_dem, None)
+        diff = (dem_with_nans - unscaled_dem).filled(np.nan)
+        assert np.abs(np.nanmedian(diff)) < 0.05
--- a/xdem/coreg.py
+++ b/xdem/coreg.py
@@ -1721,7 +1721,7 @@ class ZScaleCorr(Coreg):
     Often useful for nadir image DEM correction, where the focal length is slightly miscalculated.
 
     DISCLAIMER: This function may introduce error when correcting non-photogrammetric biases.
-    See Gardelle et al. (2012) (Figure 2), http://dx.doi.org/10.3189/2012jog11j175.
+    See Gardelle et al. (2012) (Figure 2), http://dx.doi.org/10.3189/2012jog11j175, for curvature-related biases.
     """
 
     def __init__(self, degree=1, bin_count=100):
--- a/xdem/coreg.py
+++ b/xdem/coreg.py
@@ -1719,6 +1719,9 @@ class ZScaleCorr(Coreg):
     Correct linear or nonlinear elevation scale errors.
 
     Often useful for nadir image DEM correction, where the focal length is slightly miscalculated.
+
+    DISCLAIMER: This function may introduce error when correcting non-photogrammetric biases.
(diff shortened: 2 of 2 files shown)


rhugonnet:
I'll get up to speed on all that's been happening on GitHub later tonight! :)

A scientific comment in the meantime:
Elevation-dependent biases exist in almost all dDEMs combining different instruments, and are generally mostly due to resolution. At different resolution, DEMs have biases that originate from how they capture elevations in places of high-curvature (e.g., a lower resolution DEM cuts the peaks and fills the cavities, while a higher one resolves them). Because high elevations have more peaks than cavities, the biases are very asymmetrical and you end up with a big elevation-dependent bias when you bin by elevation (especially on stable terrain, removing glaciers), but this bias is actually only specific to those areas near high-curvatures (that do not include glaciers). 
Consequently, applying an elevation-dependent bias-correction in dDEMs is, most of the time, **super dangerous** !

It will be good to have the tool available nonetheless :wink:. We'll also want to implement a curvature bias-correction that bins by curvature, even though those do not perform well either (it's the areas directly nearby high-curvatures that are affected, not only the high-curvatures points themselves).

tests/test_coreg.py
```diff
@@ -366,3 +367,62 @@ def test_subsample(self):
...
+        dem_with_nans.mask.ravel()[np.random.choice(dem_with_nans.data.size, 50000, replace=False)] = True
+
+        # Add spatially correlated errors in the order of +- 5 m
+        dem_with_nans += error_field * 3
+
+        # Try the fit now with the messed up DEM as reference.
+        zcorr.fit(dem_with_nans, scaled_dem)
+        unscaled_dem = zcorr.apply(scaled_dem, None)
+        diff = (dem_with_nans - unscaled_dem).filled(np.nan)
+        assert np.abs(np.nanmedian(diff)) < 0.05
```
adehecq, line 416: Is this value large enough to work in all cases? Due to the random noise generated, this could fail in some cases, no?

tests/test_coreg.py
```diff
@@ -366,3 +367,62 @@ def test_subsample(self):
...
+        scaled_dem = 1e-4 * self.ref.data ** 2 + 300 + self.ref.data * factor
+
+        # Try to correct using a nonlinear correction.
+        zcorr_nonlinear = coreg.ZScaleCorr(degree=2)
+        zcorr_nonlinear.fit(dem_with_nans, scaled_dem)
+
+        # Make sure the difference is minimal
+        unscaled_dem = zcorr_nonlinear.apply(scaled_dem, None)
+        diff = (dem_with_nans - unscaled_dem).filled(np.nan)
+        assert np.abs(np.nanmedian(diff)) < 0.05
```
adehecq, line 428: Same here

adehecq:
It seems all good to me, just some minor comments.
Regarding Romain's comment: yes, as Romain says, it could be dangerous to apply this correction without much care. Maybe we could add a note in the docstring to note use it to correct for curvature biases.
@rhugonnet, I guess this issue will be discussed in your next paper(s)? Would be interesting to add a reference to a paper if possible.

tests/test_coreg.py
erikmannerfelt: I tried running the tests in a loop ~30 times and it worked every time.

tests/test_coreg.py
erikmannerfelt: See above.

erikmannerfelt:
> It seems all good to me, just some minor comments.
> Regarding Romain's comment: yes, as Romain says, it could be dangerous to apply this correction without much care. Maybe we could add a note in the docstring to note use it to correct for curvature biases.
> @rhugonnet, I guess this issue will be discussed in your next paper(s)? Would be interesting to add a reference to a paper if possible.

Yeah, if I get a reference, I can add it in the docstring. It will make an automatic link in the reference API.

rhugonnet:
> Yeah, if I get a reference, I can add it in the docstring. It will make an automatic link in the reference API.

For the elevation-dependent bias, you can quote Nuth and Kääb (2011), Figure 3. It's one of the first time it was introduced in a methodological paper.
For the fact that it is dangerous to use this "correction" because it originates generally from curvature biases, this was later described in Gardelle et al. (2012) (Figure 2), http://dx.doi.org/10.3189/2012jog11j175

rhugonnet:
> @rhugonnet, I guess this issue will be discussed in your next paper(s)? Would be interesting to add a reference to a paper if possible.

It will be briefly discussed in the DEM uncertainty paper. However improving corrections of curvature-related biases is complex... it would be a completely different piece of work, really.
So for now, unfortunately, we can only say those biases exist and that we don't know how to fully account for them :/

erikmannerfelt:
> > Yeah, if I get a reference, I can add it in the docstring. It will make an automatic link in the reference API.
> 
> For the elevation-dependent bias, you can quote Nuth and Kääb (2011), Figure 3. It's one of the first time it was introduced in a methodological paper.
> For the fact that it is dangerous to use this "correction" because it originates generally from curvature biases, this was later described in Gardelle et al. (2012) (Figure 2), http://dx.doi.org/10.3189/2012jog11j175

Thanks for the link, @rhugonnet! 

As far as I understand from the paper, elevation dependent biases on DEMs should not be accounted for without first understanding their origin, right? In terms of nadir photogrammetry, there will be a near-linear scale difference, which I think is quite distinctly different from the SAR-based DEMs in that paper. So maybe that should be clarified, instead of just saying "this approach is risky"?

rhugonnet:
> Thanks for the link, @rhugonnet!
> 
> As far as I understand from the paper, elevation dependent biases on DEMs should not be accounted for without first understanding their origin, right? In terms of nadir photogrammetry, there will be a near-linear scale difference, which I think is quite distinctly different from the SAR-based DEMs in that paper. So maybe that should be clarified, instead of just saying "this approach is risky"?

DEMS being SAR- or photogrammetry-based doesn't make a difference when it comes resolution-dependent biases. 
I'm not familiar for the issue with nadir photogrammetry, but if the reference DEM used to bias-correct the nadir photogrammetry DEM has different native resolution (not the reported resolution, but the inherent one that we don't know), then the same problem will arise. It will also contain important curvature-related biases in an elevation-based correction, that will wrongly shift the entire glacier surface because of the high-curvature stable terrain at high elevations.

So I think we should definitely keep "this approach is risky" in any case, and also clarifying why!

erikmannerfelt:
@rhugonnet  is this disclaimer okay?

https://github.com/GlacioHack/xdem/blob/9435428a5dc97d5da1197b68d19c1bf3a6a376c0/xdem/coreg.py#L1723
https://github.com/GlacioHack/xdem/blob/9435428a5dc97d5da1197b68d19c1bf3a6a376c0/xdem/coreg.py#L1724

tests/test_coreg.py
erikmannerfelt: I ran it in another 200 iteration loop and it passed every time.

erikmannerfelt:
Here is an example of a semi-linearly scaled DEM, where I got the inspiration for creating this function. 
image
Data courtesy of Jacob Bendle.

rhugonnet:
> @rhugonnet is this disclaimer okay?
> 
> https://github.com/GlacioHack/xdem/blob/9435428a5dc97d5da1197b68d19c1bf3a6a376c0/xdem/coreg.py#L1723
> 
> 
> https://github.com/GlacioHack/xdem/blob/9435428a5dc97d5da1197b68d19c1bf3a6a376c0/xdem/coreg.py#L1724

Yes this is perfect! Maybe you can add "due to curvature-related biases" so that the reader knows what to look at in Gardelle et al. (2012).

**For the image:**
It's hard to judge with just this, but my guess from the elevation change map is that it looks a bit dangerous... :worried:  
The very bluish stable terrain near glaciers (high curvatures) stays bluish after correction. This means you have shifted for an elevation-averaged bias, that is actually not resolving fully the bias in those high-curvature "stable" areas, meaning there is areas of lower-curvatures with lower bias on stable terrain. As the surface of the glacier is almost all low-curvature, and so it means that it has likely been shifted similarly (and wrongly) because of this different, elevation-dependent bias in high-curvatures stable areas.
What we could introduce as a robust option would be an elevation-dependent correction based on low-curvatures only. We can derive curvatures rapidly with richDEM now! (it was hard to find anything in Python before, really)

erikmannerfelt:
Fixed the disclaimer, @rhugonnet . Good to know that it might not be that simple! I may find another example for when this is actually needed...

rhugonnet (approved):

adehecq:
> Here is an example of a semi-linearly scaled DEM, where I got the inspiration for creating this function.
> image
> Data courtesy of Jacob Bendle.

Yes, I agree with @rhugonnet here, it could simply be a curvature bias here... So this could come from the fact that both DEMs have a different **effective** resolution. For example, if this is a photogrammetry based DEM, what is the size of the correlation windows used to generate the disparity map. If they differ between both DEMs, it will create this kind of artifacts. The best way to get rid of them is to downsample both DEMs at the lowest effective resolution.

PR #92, opened 2021-04-14
Fix typo which made docs fail

erikmannerfelt:
Docs failed to build in 470ad6a, hehe...

I have successfully built this PR in my fork, so it works now: https://readthedocs.org/projects/xdem-erikm/builds/13507007/

Maybe we should add an action soon to make sure our PRs don't mess up the docs?

### DIFF
--- a/docs/source/coregistration.rst
+++ b/docs/source/coregistration.rst
@@ -102,8 +102,8 @@ The loop is stopped either when the maximum iteration limit is reached, or when 
         ddem_pre = (dem_2009.data - dem_1990.data).filled(np.nan).squeeze()
         ddem_post = (dem_2009.data - dem_coreg).filled(np.nan).squeeze()
 
-        nmad_pre = xdem.spatial_tools.nmad(ddem_pre[inlier_mask])
-        nmad_post = xdem.spatial_tools.nmad(ddem_post[inlier_mask])
+        nmad_pre = xdem.spatial_tools.nmad(ddem_pre[inlier_mask.squeeze()])
+        nmad_post = xdem.spatial_tools.nmad(ddem_post[inlier_mask.squeeze()])
 
         vlim = 20
         plt.figure(figsize=(8, 5))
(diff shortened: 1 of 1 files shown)


rhugonnet:
> Maybe we should add an action soon to make sure our PRs don't mess up the docs?

How can we make sure of this?

erikmannerfelt:
Good question hehe. 

Readthedocs and GitHub actions don't really work well together. However, a test could be added to check the error code of the sphinx build. I'll search around for some different approaches.

So that would probably be a step after CI I guess.

rhugonnet:
> Good question hehe.
> 
> Readthedocs and GitHub actions don't really work well together. However, a test could be added to check the error code of the sphinx build. I'll search around for some different approaches.
> 
> So that would probably be a step after CI I guess.

For CI, I'm still a bit worried about the conda installation of a big environment, especially with the download of proj-data. I guess it will take 10 min to run the CI at every commit to check tests are passing... or am I wrong?
 
For Readthedocs, do you have short guidelines for Amaury and I on what can or can't break it? What we should do/avoid doing. (I'm not super familiar with it)

erikmannerfelt:
> 
> For CI, I'm still a bit worried about the conda installation of a big environment, especially with the download of proj-data. I guess it will take 10 min to run the CI at every commit to check tests are passing... or am I wrong?
> 
> For Readthedocs, do you have short guidelines for Amaury and I on what can or can't break it? What we should do/avoid doing. (I'm not super familiar with it)

For CI, indeed testing might take like 10 minutes. We don't need to apply the rule that 'they need to pass in order to merge' (at least yet). It would still be a great addition to just know whether the latest main passes or not!

For Readthedocs, maybe we could have a summary somewhere? In the docs themselves? I have two suggestions to make sure you're not making breaking changes:
1. Try it locally. I usually do `cd docs && make html && cd ../` for building (here, errors might show up) and then `xdg-open docs/build/html/index.html` and look around to see that things look okay.
2. Enable a readthedocs website for your own fork. I have https://xdem-erikm.readthedocs.io where I can try to build different branches of my fork before merging to upstream. This takes longer than building locally, but I've found some issues that only turn up here, unfortunately.

rhugonnet (approved):

adehecq:
Maybe we should meet one day to discuss how to build and write the documentation. I'm still newbie at this!

rhugonnet:
Me too :)

PR #93, opened 2021-04-14
Use `xdem-data` repo for example data

erikmannerfelt:
As you may have noticed, testing and doc-building takes a very long time because data are quite inefficiently downloaded from the source.

On my computer, this download takes about 3 minutes. When using GitHub Actions, it takes around 13 minutes on their servers! 

I have created a new repo: https://github.com/erikmannerfelt/xdem-data, which could replace the way we fetch data. There, scripts exist for fetching the data from source, but the modified data are in the repo itself. Therefore, `xdem` can get the data by just fetching the latest `xdem-data` tarball and extracting whatever is needed.

Downloading the Longyearbyen data takes 2.23 s on my computer. That's a 81x performance increase!

Above just making docs, CI and general testing easier, this would ensure that testing data are not modified without us knowing about it. Imagine in 3 years if tests suddenly start failing because the Norwegian Polar Institute has decided to change their link structure...

If you think this is not a terrible idea, I can change ownership of the `xdem-data` repo to GlacioHack, further solidifying its future.

### DIFF
--- a/xdem/examples.py
+++ b/xdem/examples.py
@@ -34,7 +34,7 @@ def download_longyearbyen_examples(overwrite: bool = False):
     # Static commit hash to be bumped every time it needs to be.
     commit = "321f84d5a67666f45a196a31a2697e22bfaf3c59"
     # The URL from which to download the repository
-    url = f"https://github.com/erikmannerfelt/xdem-data/tarball/main#commit={commit}"
+    url = f"https://github.com/GlacioHack/xdem-data/tarball/main#commit={commit}"
 
     # Create a temporary directory to extract the tarball in.
     temp_dir = tempfile.TemporaryDirectory()
--- a/xdem/examples.py
+++ b/xdem/examples.py
@@ -31,8 +31,10 @@ def download_longyearbyen_examples(overwrite: bool = False):
         print("Datasets exist")
         return
 
+    # Static commit hash to be bumped every time it needs to be.
+    commit = "321f84d5a67666f45a196a31a2697e22bfaf3c59"
     # The URL from which to download the repository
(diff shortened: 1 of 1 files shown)


fmaussion:
A short note about this (just stumbled upon this by chance): it is the aproach used by salem and OGGM as well. A very important thing to do in the main repo (i.e. here) is that a specific version of the code should always fetch the same commit from the data repository.

See [OGGM for an example](https://github.com/OGGM/oggm/blob/9d83c8e84f7660524841ffb016c5c5bd71cce662/oggm/utils/_downloads.py#L72) of how we do this.

The situation this is preventing is that old versions of the code which used to work and test properly will continue to do so, *even if the data repository changes*. This happens often when you delete files which aren't used in the tests anymore, or replace them, etc.

erikmannerfelt:
That's a great point, @fmaussion! Thanks for your input.

Either in the forms of releases, tags or commit hashes I guess it could work. Releases would be my preference, I think, but it's not really based on much.

fmaussion:
My experience is that commit hashes are much easier and as searchable as tags. tags are useful for humans, not so for machines.

erikmannerfelt:
Now the data are always downloaded from the specified commit hash of `xdem-data/main`. Thanks again for the input, @fmaussion!

rhugonnet:
This looks great! Thanks a lot Erik and @fmaussion !
It was also taking ages on my side, I guess my network is closer to that if GitHub Actions than of yours, Erik.

Running **xdem** with CI is becoming a lot more sexy, suddenly :grinning:

rhugonnet (approved):

erikmannerfelt:
Lovely! Since I have an approval, I'll move the ownership to GlacioHack and then merge when it works.

PR #95, opened 2021-04-16
Improve merge_rasters

adehecq:
Add new optional features to `xdem.spatial_tools.merge_rasters`:
- **include_ref**: set to False will not include reference in the merging
- **use_ref_bounds**: set to True will use ref's bounds rather than max bounds

Fix bug:
- merge_bounding_boxes was checking that bounds were a multiple of resolution, whereas only the extent should be a multiple. @erikmannerfelt you might want to check that

Update the tests accordingly.

### DIFF
--- a/xdem/spatial_tools.py
+++ b/xdem/spatial_tools.py
@@ -98,12 +98,10 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
         for key in "bottom", "left":
             max_bounds[key] = np.nanmin([max_bounds[key], bound.__getattribute__(key)])
 
-    for key in max_bounds:
-        modulo = max_bounds[key] % resolution
-        max_bounds[key] -= modulo
-
-        if key in ["right", "top"] and modulo > 0:
-            max_bounds[key] += resolution
+    # Make sure that extent is a multiple of resolution
+    for key1, key2 in zip(("left", "bottom"), ("right", "top")):
+        modulo = (max_bounds[key2] - max_bounds[key1]) % resolution
+        max_bounds[key2] += modulo
 
     return rio.coords.BoundingBox(**max_bounds)
 
--- a/xdem/spatial_tools.py
+++ b/xdem/spatial_tools.py
@@ -109,7 +109,8 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
 
 
 def merge_rasters(rasters: list[gu.georaster.Raster], reference: int = 0, merge_algorithm: Callable = np.nanmean,
-                  resampling_method: Union[str, rio.warp.Resampling] = "nearest") -> gu.georaster.Raster:
+                  resampling_method: Union[str, rio.warp.Resampling] = "bilinear", include_ref = True,
+                  use_ref_bounds = False) -> gu.georaster.Raster:
     """
     Merge a list of rasters into one larger raster.
 
@@ -119,6 +120,8 @@ def merge_rasters(rasters: list[gu.georaster.Raster], reference: int = 0, merge_
     :param reference: The reference index (defaults to the first raster in the list).
     :param merge_algorithm: The algorithm to merge the rasters with. Defaults to the mean.
     :param resampling_method: The resampling method for the raster reprojections.
+    :param include_ref: If False, will not merge the reference with the others
+    :param use_ref_bounds: If True, will use reference bounds, otherwise will use maximum bounds of all rasters.
 
     :returns: The merged raster with the same parameters (excl. bounds) as the reference.
     """
@@ -131,17 +134,23 @@ def merge_rasters(rasters: list[gu.georaster.Raster], reference: int = 0, merge_
         resampling_method = resampling_method_from_str(resampling_method)
 
     reference_raster = rasters[reference]
-    # Find the maximum covering bounding box
-    max_bounds = merge_bounding_boxes([raster.bounds for raster in rasters], resolution=reference_raster.res[0])
+
+    if use_ref_bounds:
+        max_bounds = reference_raster.bounds
+    else:
+        # Find the maximum covering bounding box
+        max_bounds = merge_bounding_boxes([raster.bounds for raster in rasters], resolution=reference_raster.res[0])
 
     # Make a data list and add all of the reprojected rasters into it.
     data: list[np.ndarray] = []
-    for raster in rasters:
+    rasters_to_merge = rasters if include_ref else rasters[1:]
+    for raster in rasters_to_merge:
         reprojected_raster = raster.reproject(
             dst_bounds=max_bounds,
+            dst_res=reference_raster.res,
             dst_crs=reference_raster.crs,
             dtype=reference_raster.data.dtype,
-            nodata=reference_raster.nodata
+            nodata=np.nan  # quick fix, should use get_mask when other PR is merged (this could fail if dtype is int)
         )
         data.append(reprojected_raster.data.squeeze())
 
--- a/xdem/spatial_tools.py
+++ b/xdem/spatial_tools.py
@@ -106,9 +106,9 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
     return rio.coords.BoundingBox(**max_bounds)
 
 
-def merge_rasters(rasters: list[gu.georaster.Raster], reference: int = 0,
-                  merge_algorithm: Union[Callable, list(Callable)] = np.nanmean,
-                  resampling_method: Union[str, rio.warp.Resampling] = "bilinear", include_ref = True,
+def merge_rasters(rasters: list[gu.georaster.Raster], reference: Union[int, gu.Raster] = 0,
+                  merge_algorithm: Union[Callable, list[Callable]] = np.nanmean,
+                  resampling_method: Union[str, rio.warp.Resampling] = "bilinear",
                   use_ref_bounds = False) -> gu.georaster.Raster:
     """
     Merge a list of rasters into one larger raster.
@@ -118,12 +118,12 @@ def merge_rasters(rasters: list[gu.georaster.Raster], reference: int = 0,
     merge_rasters it will be loaded for reprojection and deleted, therefore avoiding duplication and \
     optimizing memory usage.
 
-    :param rasters: A list of geoutils Raster objects.
-    :param reference: The reference index (defaults to the first raster in the list).
+    :param rasters: A list of geoutils Raster objects to be merged.
+    :param reference: The reference index, in case the reference is to be merged, or a separate Raster object \
+ in case the reference should not be merged. Defaults to the first raster in the list.
     :param merge_algorithm: The algorithm, or list of algorithms, to merge the rasters with. Defaults to the mean.\
 If several algorithms are provided, each result is returned as a separate band.
     :param resampling_method: The resampling method for the raster reprojections.
-    :param include_ref: If False, will not merge the reference with the others.
     :param use_ref_bounds: If True, will use reference bounds, otherwise will use maximum bounds of all rasters.
 
     :returns: The merged raster with the same parameters (excl. bounds) as the reference.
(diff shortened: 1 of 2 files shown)


xdem/spatial_tools.py
```diff
@@ -98,65 +98,101 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
...
+    for key1, key2 in zip(("left", "bottom"), ("right", "top")):
+        modulo = (max_bounds[key2] - max_bounds[key1]) % resolution
+        max_bounds[key2] += modulo
 
     return rio.coords.BoundingBox(**max_bounds)
 
 
-def merge_rasters(rasters: list[gu.georaster.Raster], reference: int = 0, merge_algorithm: Callable = np.nanmean,
-                  resampling_method: Union[str, rio.warp.Resampling] = "nearest") -> gu.georaster.Raster:
+def merge_rasters(rasters: list[gu.georaster.Raster], reference: int = 0,
```
erikmannerfelt, line 109: I think the functionality with using a non-included reference is great. I do believe that the syntax can be a bit more clear, however. 

Right now, the list one gives would be `[ref (which will not be included), nonref0, nonref1, nonref2 ... ]` which I think will take a while for people to get. What about having the `reference` argument either take a list index (like it does currently) or take a separate raster? This way, all given rasters in the `rasters` list will always be merged. This would make `include_ref` redundant but `use_ref_bounds` is still great to have. The function signature would then be:

```python
def merge_rasters(rasters: list[gu.Raster], reference: Union[int, gu.Raster] = 0, merge_algorithm, Union[Callable, list[Callable]] = np.nanmean, resampling_method: Union[str, rio.warp.Resampling] = "bilinear", use_ref_bounds: bool = False) -> gu.Raster:
```
What do you think about that?

xdem/spatial_tools.py
```diff
@@ -98,65 +98,101 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
...
+                  merge_algorithm: Union[Callable, list(Callable)] = np.nanmean,
```
erikmannerfelt, line 110: It's `list[Callable]`, not `list(Callable)` (the latter would instantiate a list. The former creates a subtype).

erikmannerfelt (changes requested):
I think the implementation looks perfect!

I have some comments on the function call which would be nice if we could discuss. Also a typing typo.

xdem/spatial_tools.py
```diff
@@ -98,65 +98,101 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
         for key in "bottom", "left":
             max_bounds[key] = np.nanmin([max_bounds[key], bound.__getattribute__(key)])
 
-    for key in max_bounds:
-        modulo = max_bounds[key] % resolution
-        max_bounds[key] -= modulo
-
-        if key in ["right", "top"] and modulo > 0:
-            max_bounds[key] += resolution
+    # Make sure that extent is a multiple of resolution
```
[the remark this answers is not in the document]
adehecq: Not sure I understand. This is what is done right now, i.e. the extent is a multiple of resolution. What you had is that each bounds were a multiple of resolution, which is not really required.

xdem/spatial_tools.py
adehecq: Sounds good, I'll implement that!

xdem/spatial_tools.py
adehecq: Good catch!

xdem/spatial_tools.py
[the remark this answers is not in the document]
erikmannerfelt: Sorry I meant that I prefer that each bounding coordinate is a multiple, e.g. that a resolution of 5 would always have the left coordinate to end with 0 or 5 and so on. Makes gridding and merging so much easier! Again, however, it's not great to have in xdem, so I support your decision of removing it!

xdem/spatial_tools.py
adehecq: Fixed in 2a40981

xdem/spatial_tools.py
adehecq: Added in 2a40981

adehecq:
@erikmannerfelt, I realized that in my current case I actually would like to merge the dh rates ((ref - dem)/ddate) rather than the DEMs themselves.
Right now, doing that would require doing the reprojection and loop outside of merge_rasters, making the latter a bit useless.
Would this feature be interesting in xdem? If so, could that be an option in merge_rasters, or should it rather be a separate function?
It would be easy to add an option to calculate the difference with the ref. However, calculating the time difference would require passing the dates. This could possibly be a ddem or tdem function.
In all cases, I wonder if there should not be a subfunction of merge_rasters that would only do the reprojecting and looping, and pass the 3D array for additional processing. What are your thoughts on that?

erikmannerfelt:
> @erikmannerfelt, I realized that in my current case I actually would like to merge the dh rates ((ref - dem)/ddate) rather than the DEMs themselves.
> Right now, doing that would require doing the reprojection and loop outside of merge_rasters, making the latter a bit useless.
> Would this feature be interesting in xdem? If so, could that be an option in merge_rasters, or should it rather be a separate function?
> It would be easy to add an option to calculate the difference with the ref. However, calculating the time difference would require passing the dates. This could possibly be a ddem or tdem function.
> In all cases, I wonder if there should not be a subfunction of merge_rasters that would only do the reprojecting and looping, and pass the 3D array for additional processing. What are your thoughts on that?


I would love that function in the terrA project! I feel like that is a separate (and super useful) function, however.

erikmannerfelt (approved):
Breathtaking!

adehecq:
> > In all cases, I wonder if there should not be a subfunction of merge_rasters that would only do the reprojecting and looping, and pass the 3D array for additional processing. What are your thoughts on that?
> 
> I would love that function in the terrA project! I feel like that is a separate (and super useful) function, however.

So I'm working on a stack_rasters function that will just do the looping and reprojecting and returns the stacked array and output bounds.
I can probably add a simple option to return the difference rather than DEM as well.
On the longer term, this could be modified, or duplicated, to save the output in a file that could be easily opened by xarray.

xdem/spatial_tools.py
```diff
@@ -106,39 +106,24 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
     return rio.coords.BoundingBox(**max_bounds)
 
 
-def merge_rasters(rasters: list[gu.georaster.Raster], reference: Union[int, gu.Raster] = 0,
-                  merge_algorithm: Union[Callable, list[Callable]] = np.nanmean,
+def stack_rasters(rasters: list[gu.georaster.Raster], reference: Union[int, gu.Raster] = 0,
                   resampling_method: Union[str, rio.warp.Resampling] = "bilinear",
-                  use_ref_bounds = False) -> gu.georaster.Raster:
+                  use_ref_bounds = False) -> tuple[gu.georaster.Raster, rio.coords.BoundingBox]:
```
erikmannerfelt, line 111: I know that you've merged now, but why is it returning the bounds? What about `Raster.bounds` ??

xdem/spatial_tools.py
```diff
@@ -98,65 +98,142 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
...
-            raise exception
-        merged_data = np.apply_along_axis(merge_algorithm, axis=0, arr=data)
+        # Remove unloaded rasters
+        if not raster.is_loaded:
+            raster._data = None
+
+    # Convert to numpy array
+    data = np.asarray(data)
+
+    return data, dst_bounds
```
erikmannerfelt, line 172: Wait, so it's returning a `np.ndarray`, not a `gu.Raster`?

xdem/spatial_tools.py
```diff
@@ -98,65 +98,142 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
...
+
+        # Reproject to reference grid
         reprojected_raster = raster.reproject(
-            dst_bounds=max_bounds,
+            dst_bounds=dst_bounds,
+            dst_res=reference_raster.res,
             dst_crs=reference_raster.crs,
             dtype=reference_raster.data.dtype,
-            nodata=reference_raster.nodata
+            nodata=np.nan  # quick fix, should use get_mask when other PR is merged (this could fail if dtype is int)
```
erikmannerfelt, line 161: Was this the PR you just merged?

xdem/spatial_tools.py
adehecq: My mistake, it's actually returning a np.ndarray, as stated in the docstring, but wrongly indicated in the type.
Now wondering which one makes most sense, returning an array with references, or a Raster... What do you think?

xdem/spatial_tools.py
```diff
@@ -98,65 +98,142 @@ def merge_bounding_boxes(bounds: list[rio.coords.BoundingBox], resolution: float
...
 
 
-def merge_rasters(rasters: list[gu.georaster.Raster], reference: int = 0, merge_algorithm: Callable = np.nanmean,
-                  resampling_method: Union[str, rio.warp.Resampling] = "nearest") -> gu.georaster.Raster:
+def stack_rasters(rasters: list[gu.georaster.Raster], reference: Union[int, gu.Raster] = 0,
+                  resampling_method: Union[str, rio.warp.Resampling] = "bilinear",
+                  use_ref_bounds = False) -> tuple[gu.georaster.Raster, rio.coords.BoundingBox]:
     """
-    Merge a list of rasters into one larger raster.
+    Stack a list of rasters into a common grid as a 3D np array with nodata set to Nan.
```
erikmannerfelt, line 114: Could you write the output shape here somewhere to make it easier to understand what it does? Something alike:
```
If `reference` is an index, the output will have a shape of (N, height, width) where N is len(rasters) - 1 and height and width is equal to rasters[0].shape.
If `reference` is another raster, N will be equal to len(rasters), and height and width will be equal to reference.shape .
```
Do I understand the shapes correctly?

xdem/spatial_tools.py
erikmannerfelt: Right! Maybe it's worth returning a Raster? Since we have the shiny Raster.split_bands() functionality, I think it makes sense!

xdem/spatial_tools.py
erikmannerfelt: Mentioned above.

xdem/spatial_tools.py
adehecq: Yes. I needed the get_mask function...

xdem/spatial_tools.py
adehecq: Implemented in 001f86a

xdem/spatial_tools.py
adehecq: Actually, N will always be the same, as all items in rasters are now stacked (since reference is external). So there is no distinction between both.
Additionally, it's difficult to state what the output shape will be. If use_ref_bounds is True, then yes it will have the same dimensions, otherwise not as it will depend on the max bounds. Would something like this help?

> 
> If use_ref_bounds is True, output will have the shape (N, height, width) where N is len(rasters) and height and width is equal to reference's shape.
> If use_ref_bounds is False, output will have the shape (N, height2, width2) where N is len(rasters) and height2 and width2 are set based on reference's resolution and the maximum extent of all rasters.

erikmannerfelt: Wait, so if this is run:
```python
stacked = stack_rasters([ref, nonref0, nonref1], reference=0)
```
`stacked.shape[0]` will be 2??

erikmannerfelt (commented):
Nice! I have one question on the shapes (might just misunderstand), but everything else is fantastic!

xdem/spatial_tools.py
adehecq: No, as stated in the new docstring, stacked.shape[0] is **always** equal to len(rasters). Now the way 'reference' is excluded is by setting it to a separate dataset rather than an index. Is that clear?

erikmannerfelt: I see!

erikmannerfelt (approved):

PR #99, opened 2021-04-19
Removed all deprecated coreg functions.

erikmannerfelt:
All of the old coreg functions (marked with `DeprecationWarning`s) have been implemented as `Coreg` subclasses. They are therefore completely redundant, and we can remove them. The only function with a `DeprecationWarning` that I have not removed is the `deramping()` function that found a use in #87.


rhugonnet (approved):
Perfect!
Closely related to this: what do you think about having the core structure of the coreg/biascorr functions, and their subfunctions, living in a structured fashion outside the Subclasses? I think this would be somewhat preferable as discussed in #71

erikmannerfelt:
> Perfect!
> Closely related to this: what do you think about having the core structure of the coreg/biascorr functions, and their subfunctions, living in a structured fashion outside the Subclasses? I think this would be somewhat preferable as discussed in #71

Thanks! Perhaps we can discuss this in video at some point? I get your point, and see both pros and cons with it.

PR #75, opened 2021-04-08
Added missing 'import numpy as np' statements to code examples

erikmannerfelt:
Found by @adehecq

PR #101, opened 2021-04-21
Improve documentation, add documentation testing, and make all examples into scripts.

erikmannerfelt:
This was a hard one, including environment variables and all kinds of weird errors!

### Fixed warnings and incorrect docstrings
There were quite many warnings before when building the documentation. I've made sure that all of them have been fixed.

### Documentation build tests
I've added tests for all example scripts (I'll get more to that in a second) and for building the documentation. Now, theoretically, ReadTheDocs should always build if the build check on GitHub passes. Makes for easier maintainability in the long term, as a PR that passes the build check should be compatible with ReadTheDocs (right now, builds might pass but then the documentation fails to build!).

### More future-frendly examples structure.
Finally, since we already have had trouble with outdated examples, all example code is now in `docs/source/code/*.py`. This has many advantages:
1. Linters and formatters can locally help with the scripts to make sure that they're correct.
2. It's easier to debug, as the code can be run without rebuilding the entire documentation.
3. Inline code in rst is awkward to write
4. It allows testing the code externally (if an example is outdated, it won't pass the build check!)

### See the result
If you're interested, you can see how the documentation will look here:
https://xdem-erikm.readthedocs.io/en/improve_docs/index.html

### DIFF
--- a/docs/source/comparison.rst
+++ b/docs/source/comparison.rst
@@ -121,6 +121,7 @@ Now, we can easily calculate the elevation or volume change between the DEMs, fo
 
 which will return a Pandas Series:
 
-.. program-output:: python code/comparison_print_cumulative_dh.py 
+.. program-output:: $PYTHON code/comparison_print_cumulative_dh.py 
+        :shell:
 
 `See here for the outline filtering syntax <https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html>`_.
--- a/docs/source/comparison.rst
+++ b/docs/source/comparison.rst
@@ -9,35 +9,9 @@ DEM subtraction and volume change
 
 Example data in this chapter are loaded as follows:
 
-.. code-block:: python
+.. literalinclude:: code/comparison.py
+        :lines: 5-33
 
-        from datetime import datetime        
-
-        import geoutils as gu
-        import xdem
-
-        # Download the necessary data. This may take a few minutes.
-        xdem.examples.download_longyearbyen_examples(overwrite=False)
-
-        # Load a reference DEM from 2009
-        dem_2009 = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_ref_dem"], datetime=datetime(2009, 8, 1))
-        # Load a DEM from 1990
-        dem_1990 = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_tba_dem"], datetime=datetime(1990, 8, 1))
-        # Load glacier outlines from 1990.
-        glaciers_1990 = gu.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines"])
-        glaciers_2010 = gu.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines_2010"])
-        
-        # Make a dictionary of glacier outlines where the key represents the associated date.
-        outlines = {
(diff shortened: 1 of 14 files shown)


erikmannerfelt:
This partly tackles #53 as invalid docstrings will show up in the build check (as warnings if they are minor, and failed builds if they are major).

#75 and #92 were only needed because we didn't have a good validation technique for the documentation examples. This should solve it.

docs/source/comparison.rst
```diff
@@ -9,35 +9,9 @@ DEM subtraction and volume change
 
 Example data in this chapter are loaded as follows:
 
-.. code-block:: python
+.. literalinclude:: code/comparison.py
```
rhugonnet, line 12: so `literalinclude` is kind of the key here :D

docs/source/conf.py
```diff
@@ -59,10 +66,15 @@
...
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+# html_static_path = ['_static']  # Commented out as we have no custom static data
 
 
 def run_apidoc(_):
+    """
+    Make sure readthedocs finds the module.
+
+    Maybe this is not needed?
```
rhugonnet, line 76: time will tell !

tests/test_docs.py
```diff
@@ -0,0 +1,63 @@
+import os
+import shutil
+import subprocess
+import sys
+import warnings
+
+
+class TestDocs:
```
rhugonnet, line 8: very nice

rhugonnet (approved):
Nothing to say except that this is going to be super useful!

I'm wondering if this kind of "good practice" won't eventually (in 5-year time?) be included in a more user-friendly way to GitHub. I hope so.
The advantage is that we really get to know how this all works (well.. mostly you Erik!)

erikmannerfelt:
Thank you for the feedback, @rhugonnet!

I wanted to see if other packages used similar testing, so I checked rasterio.

1. They recommend installing rio=0.24 and gdal=1.11 in their [installation page](https://rasterio.readthedocs.io/en/latest/installation.html#windows)
2. I haven't found any incorrect syntax, but many statements are only really pseudocode. [See here](https://rasterio.readthedocs.io/en/latest/topics/switch.html#format-drivers); what are the kwargs!? (turns out they're width, height, cound, transform, dtype and crs...)
3. There are some typos [here and there](https://rasterio.readthedocs.io/en/latest/topics/switch.html#valid-data-masks); `dtype-uint8` in the output instead of `dtype=uint8`.

adehecq:
> Thank you for the feedback, @rhugonnet!
> 
> I wanted to see if other packages used similar testing, so I checked rasterio.
> 
>     1. They recommend installing rio=0.24 and gdal=1.11 in their [installation page](https://rasterio.readthedocs.io/en/latest/installation.html#windows)
> 
>     2. I haven't found any incorrect syntax, but many statements are only really pseudocode. [See here](https://rasterio.readthedocs.io/en/latest/topics/switch.html#format-drivers); what are the kwargs!? (turns out they're width, height, cound, transform, dtype and crs...)
> 
>     3. There are some typos [here and there](https://rasterio.readthedocs.io/en/latest/topics/switch.html#valid-data-masks); `dtype-uint8` in the output instead of `dtype=uint8`.

Does this mean we're now doing better than rasterio?! :-)

Issue #102, opened 2021-04-22
Add a hillshade function


adehecq:
After discussion with @erikmannerfelt, we agreed to include the hillshade function of [earthpy](https://earthpy.readthedocs.io/en/latest/gallery_vignettes/plot_dem_hillshade.html) to xdem.
This seem to show different results than gdaldem hillshade by default, but can be solved by a zFactor of ~10-50.

Without scaling factor
https://files.slack.com/files-pri/T01D6EMSU0J-F01V0U7FWAH/figure_1.png

With scaling factor 50
https://files.slack.com/files-pri/T01D6EMSU0J-F01V47N24JX/figure_1.png

erikmannerfelt:
To be clear, we discussed that the earthpy function could use some improvements, and we therefore decided to make our own. Other would be intellectual theft, right!?

erikmannerfelt:
Added in #104

PR #104, opened 2021-04-22
Add hillshade function.

erikmannerfelt:
This was harder than I thought!

Initially inspired by [earthpy](https://earthpy.readthedocs.io/en/latest/_modules/earthpy/spatial.html) (since we wanted to avoid more dependencies), I set on to remake their function in the style of xdem. Turns out, however, that their hillshade function is not accurate! 

As noted by @adehecq  in #102, the input DEM of `earthpy.spatial.hillshade()` needed to be divided by an (at the time) arbitrary value to get the right z-scale. Turns out, hillshades need the resolution to generate properly! Makes sense in hindsight, considering that the `np.gradient` just  gets the pixel differences, so the horizontal scale is needed for a proper z-scale.

So here we have it: our very own hillshade function. I've made sure that it works well with nans/ masked_arrays as well. The syntax is pretty straightforward:
```python
import xdem

dem = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_ref_dem"])

hillshade = xdem.spatial_tools.hillshade(
    dem=dem.data,  # This can be an array or masked array. Required.
    resolution=dem.res,  # This is either a float or a sequence of floats. Required.
    azimuth=315,  # This is in degrees going clockwise from north between 0 and 360. Defaults to 315
    altitude=45,  # This is in degrees between 0 and 90. 90 is straight up. Defaults to 45
    z_factor=1  # Vertical exaggeration. Can be any positive finite float. Defaults to 1
)
```
The function returns a `float32` array of shape `(height, width)` with values ranging from 0 to 255.
Thus, it can safely be saved as an `uint8`, but that simplification is up to the user.

In the test, I quantitatively compare with a hillshade generated by GDAL to make sure that they are approximately equal, which they are! I also tested on other DEMs I had lying around, and they were good as well.

### DIFF
--- a/tests/test_spatial_tools.py
+++ b/tests/test_spatial_tools.py
@@ -13,6 +13,7 @@ import warnings
 import geoutils as gu
 import numpy as np
 import rasterio as rio
+from osgeo import gdal
 
 import xdem
 from xdem import examples
@@ -67,10 +68,16 @@ def test_hillshade():
     def make_gdal_hillshade(filepath) -> np.ndarray:
         temp_dir = tempfile.TemporaryDirectory()
         temp_hillshade_path = os.path.join(temp_dir.name, "hillshade.tif")
-        gdal_commands = ["gdaldem", "hillshade",
-                         filepath, temp_hillshade_path,
-                         "-az", "315", "-alt", "45"]
-        subprocess.run(gdal_commands, check=True, stdout=subprocess.PIPE)
+        # gdal_commands = ["gdaldem", "hillshade",
+        #                 filepath, temp_hillshade_path,
+        #                 "-az", "315", "-alt", "45"]
+        #subprocess.run(gdal_commands, check=True, stdout=subprocess.PIPE)
+        gdal.DEMProcessing(
+            destName=temp_hillshade_path,
+            srcDS=filepath,
+            processing="hillshade",
+            options=gdal.DEMProcessingOptions(azimuth=315, altitude=45)
+        )
 
         data = gu.Raster(temp_hillshade_path).data
         temp_dir.cleanup()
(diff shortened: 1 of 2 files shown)


tests/test_spatial_tools.py
```diff
@@ -53,3 +58,57 @@ def test_merge_rasters():
...
+
+
+def test_hillshade():
+    """Test the hillshade algorithm, partly by comparing it to the GDAL hillshade function."""
+    warnings.simplefilter("error")
+
+    def make_gdal_hillshade(filepath) -> np.ndarray:
+        temp_dir = tempfile.TemporaryDirectory()
+        temp_hillshade_path = os.path.join(temp_dir.name, "hillshade.tif")
+        gdal_commands = ["gdaldem", "hillshade",
```
adehecq, line 70: Note that big of a deal, but you can call this directly in Python with gdal.DEMProcessing !

adehecq (dismissed):
Fantastic, well done !!

adehecq:
So the issue with earthpy was from a misuse on my side, or from a bug in their implementation? In the latter case, it's probably worth raising an issue on their repo!

tests/test_spatial_tools.py
erikmannerfelt: That may actually be a big deal, because the GitHub Actions CI don't have GDAL in their PATH and I don't know how to fix this, hence the ugly "only test if gdal is found" assertion in the test.

erikmannerfelt:
> So the issue with earthpy was from a misuse on my side, or from a bug in their implementation? In the latter case, it's probably worth raising an issue on their repo!

It may be best to raise an issue. You didn't do anything wrong! The required input for the DEM to the hillshade in earthpy is that the pixel value range has the same scale as the pixels. So if you have a 1x1 m DEM with elevations expressed in m, it would work. If you have a 0.5x1 m DEM, you're gonna have a bad time.

erikmannerfelt:
Thanks for your recommendation of the GDAL python bindings, @adehecq! Now the full test suite can be run without a problem on GitHub's servers!

tests/test_spatial_tools.py
```diff
@@ -13,6 +13,7 @@
 import geoutils as gu
 import numpy as np
 import rasterio as rio
+from osgeo import gdal
```
adehecq, line 16: rasterio strongly [discourage](https://rasterio.readthedocs.io/en/latest/topics/switch.html) importing rio and gdal in the same script... This could lead to massive bugs at some point. But I guess it's ok if gdal is only used in the hillshade test...

adehecq: Actually, maybe gdal should only be imported within the make_gdal_hillshade function?

erikmannerfelt: Moved it to the function in 4d33081

adehecq (approved):

Issue #105, opened 2021-04-23
volume.interpolate_hypsometric_bins does not return proper pandas dataframe


adehecq:
When running `volume.interpolate_hypsometric_bins`, the output is different from the input. In particular, it does not contain the attribute "value" which can cause issues when the output is passed to other xdem.volume functions like `calculate_hypsometry_area`.
The output should be consistent.

adehecq:
This change will impact the scipy.interpolate.interp1d calls in hypsometric_interpolation and local_hyspometric_interpolation -> these lines must be updated.

erikmannerfelt:
I'm on it!

PR #106, opened 2021-04-26
Fix interpolate_hypsometric_bins

erikmannerfelt:
As seen by @adehecq  in #105, the function in volume.py did not return a dataframe similar to the input dataframe. This is fixed, and I've added some extra documentation.


adehecq (approved):
Thanks a lot for fixing this! I had a fix to this locally, but with the rush lately, I didn't have time to push it.
Looks all good to me!

PR #118, opened 2021-05-12
Fix hypsometric interpolation for large gaps

adehecq:
Several error arise when running the several hypsometric interpolation methods, when there are too few valid pixels.
This PR tries to solve those issues.

### DIFF
--- a/xdem/volume.py
+++ b/xdem/volume.py
@@ -121,8 +121,7 @@ def interpolate_hypsometric_bins(hypsometric_bins: pd.DataFrame, value_column="v
         bins.loc[bins_under_threshold, value_column] = np.nan
 
     # Count number of valid (finite) values
-    values = bins[value_column]
-    nvalids = len(values[np.isfinite(values)])
+    nvalids = np.count_nonzero(np.isfinite(bins[value_column]))
 
     if nvalids <= order + 1:
         # Cannot interpolate -> leave as it is
@@ -363,7 +362,7 @@ def hypsometric_interpolation(voided_ddem: Union[np.ndarray, np.ma.masked_array]
 def local_hypsometric_interpolation(voided_ddem: Union[np.ndarray, np.ma.masked_array],
                                     ref_dem: Union[np.ndarray, np.ma.masked_array],
                                     mask: np.ndarray, min_coverage: float = 0.2,
-                                    count_threshold: Optional[int] = 1, nodata: float = -9999,
+                                    count_threshold: Optional[int] = 1, nodata: Union[float, int] = -9999,
                                     plot: bool = False) -> np.ma.masked_array:
     """
     Interpolate a dDEM using local hypsometric interpolation.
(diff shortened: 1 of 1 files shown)


xdem/volume.py
```diff
@@ -120,8 +120,17 @@ def interpolate_hypsometric_bins(hypsometric_bins: pd.DataFrame, value_column="v
         bins_under_threshold = bins["count"] < count_threshold
         bins.loc[bins_under_threshold, value_column] = np.nan
 
-    # Interpolate all bins that are NaN.
-    bins[value_column] = bins[value_column].interpolate(method=method, order=order, limit_direction="both")
+    # Count number of valid (finite) values
+    values = bins[value_column]
+    nvalids = len(values[np.isfinite(values)])
```
erikmannerfelt, line 125: Could this not be simplified to `nvalids = np.count_nonzero(np.isfinite(bins[value_column]))`?

xdem/volume.py
```diff
@@ -354,7 +363,8 @@ def hypsometric_interpolation(voided_ddem: Union[np.ndarray, np.ma.masked_array]
 def local_hypsometric_interpolation(voided_ddem: Union[np.ndarray, np.ma.masked_array],
                                     ref_dem: Union[np.ndarray, np.ma.masked_array],
                                     mask: np.ndarray, min_coverage: float = 0.2,
-                                    count_threshold: Optional[int] = 1, plot: bool = False) -> np.ma.masked_array:
+                                    count_threshold: Optional[int] = 1, nodata: float = -9999,
```
erikmannerfelt, line 366: Pet peeve for linters: -9999 is an int ;) -9999. or -9999.0 is better!

xdem/volume.py
```diff
@@ -367,6 +377,7 @@ def local_hypsometric_interpolation(voided_ddem: Union[np.ndarray, np.ma.masked_
 each geometry on which to loop.
     :param min_coverage: Optional. The minimum coverage fraction to be considered for interpolation.
     :param count_threshold: Optional. A pixel count threshold to exclude during the hypsometric curve fit.
+    :param nodara: Optional. No data value to be used for the output masked_array.
```
erikmannerfelt, line 380: Typo: nodara!

erikmannerfelt (commented):
Looking great! I just have a few tiny comments.

erikmannerfelt:
Also, the tests fail because of a change in geoutils. We can rerun them when #117  is merged.

xdem/volume.py
erikmannerfelt: Or `bins[value_column].count()` which counts the occurrence of finite values.

adehecq: The last option looks great! The only issue is that by default pandas apparently [does not count inf as a NaN](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.count.html), unless specifically set. So I will stick to the first case, which is clearer.

xdem/volume.py
adehecq: Yes, -9999 is an int, and I want it to be so, but nodata does not have to be an it. So I guess what I should do is add int as a possible type...

adehecq:
@erikmannerfelt, let me know if you have an issue with commit f7d8d0a.
The spatial interpolation should not be needed, as the regional hypsometric curve should enable filling all gaps (at least once we include extrapolation in case some upper/lower bins are missing) and it's not really the expected behavior.
Plus it was leaving large gaps with the recently changed default linear interpolation radius which is much smaller.

rhugonnet (approved):
Looks good to me!

erikmannerfelt (approved):
Looking great! The tests are failing unrelated to your fixes, so you can bypass the check if you want. Otherwise, wait for #121  to be merged and then sync to upstream.

adehecq:
> Looking great! The tests are failing unrelated to your fixes, so you can bypass the check if you want. Otherwise, wait for #121 to be merged and then sync to upstream.

Ok, I bypassed the tests and merged. Hopefully there was no hidden test failure!

Issue #119, opened 2021-05-13
Conda instructions README and `environment.yml` does not install geoutils


erikmannerfelt:
The README is outdated, for example saying `conda activate glacio` while it should be `conda activate xdem`.

Also, when installing xdem via conda, it will not install geoutils. This needs to be added in the `environment.yml` (with a --force-reinstall flag until geoutils has a good release system).

PR #120, opened 2021-05-13
Update readme and environment.yml to make installation easier with conda+pip

erikmannerfelt:
Solves #119

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -23,7 +23,3 @@ dependencies:
     - scikit-gstat
     - pytransform3d
     - https://github.com/GlacioHack/GeoUtils/tarball/master
-#    - "--editable=git+https://github.com/GlacioHack/GeoUtils.git#egg=GeoUtils"
-#    - "--editable=git+https://github.com/GlacioHack/xdem.git#egg=xdem"
-variables:
-    - PIP_FORCE_REINSTALL=1  # Since geoutils has no versioning system, this is needed.
--- a/environment.yml
+++ b/environment.yml
@@ -22,5 +22,8 @@ dependencies:
     - richdem
     - scikit-gstat
     - pytransform3d
+    - https://github.com/GlacioHack/GeoUtils/tarball/master
 #    - "--editable=git+https://github.com/GlacioHack/GeoUtils.git#egg=GeoUtils"
 #    - "--editable=git+https://github.com/GlacioHack/xdem.git#egg=xdem"
+variables:
+    - PIP_FORCE_REINSTALL=1  # Since geoutils has no versioning system, this is needed.
--- a/setup.py
+++ b/setup.py
@@ -1,22 +1,45 @@
-from os import path
+import os
+import subprocess
+import sys
 
 from setuptools import setup
+from setuptools.command.install import install
 
-FULLVERSION = '0.0.1'
+FULLVERSION = '0.0.2-2'
 VERSION = FULLVERSION
+
+class PostInstall(install):
+    """Post install script to install github dependencies."""
+
+    pkgs = "git+https://github.com/GlacioHack/GeoUtils.git"
+
+
+    def run(self):
+        install.run(self)
+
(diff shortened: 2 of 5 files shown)


erikmannerfelt:
I also included some improvements to work with pypi while I was at it.

setup.py
```diff
@@ -1,22 +1,45 @@
...
 setup(name='xdem',
       version=FULLVERSION,
-      description='',
-      url='',
+      description='Set of tools to manipulate Digital Elevation Models (DEMs) ',
+      long_description=LONG_DESCRIPTION,
+      long_description_content_type="text/markdown",
+      url='https://github.com/GlacioHack/xdem',
       author='The GlacioHack Team',
+      author_email="john@doe.com",
```
rhugonnet, line 34: is john@doe.com really necessary here? :O

setup.py
```diff
@@ -1,22 +1,45 @@
...
+import sys
 
 from setuptools import setup
+from setuptools.command.install import install
 
-FULLVERSION = '0.0.1'
+FULLVERSION = '0.0.2-2'
 VERSION = FULLVERSION
 
+class PostInstall(install):
```
rhugonnet, line 11: can you add a short comment on what this does (for posterity)? :D

rhugonnet (approved):
Great :)
Nice spotting for the "glacio" conda environment name instead of "xdem"

setup.py
erikmannerfelt: Unfortunately yes... I don't remember how to reproduce the error, but I had to add an email address for everything to work!

erikmannerfelt: Added a small comment for the future in a7734de

setup.py
erikmannerfelt: Added in a7734de

environment.yml
```diff
@@ -22,5 +22,8 @@ dependencies:
     - richdem
     - scikit-gstat
     - pytransform3d
+    - https://github.com/GlacioHack/GeoUtils/tarball/master
```
adehecq, line 25: We should check that if we have a version of GeoUtils installed already in editable mode, it does not get overwritten with this environment.

setup.py
```diff
@@ -1,22 +1,52 @@
...
+      long_description_content_type="text/markdown",
+      url='https://github.com/GlacioHack/xdem',
       author='The GlacioHack Team',
+      author_email="john@doe.com",  # This is needed for PyPi unfortunately.
       license='BSD-3',
       packages=['xdem'],
       install_requires=['numpy', 'scipy', 'rasterio', 'geopandas',
-                        'pyproj', 'tqdm', 'geoutils @ https://github.com/GlacioHack/GeoUtils/tarball/master', 'scikit-gstat', 'scikit-image'],
+                        'pyproj', 'tqdm', 'scikit-gstat', 'scikit-image'],
       extras_require={'rioxarray': ['rioxarray'], 'richdem': ['richdem'], 'pdal': [
```
adehecq, line 46: This makes me think. In the setup.py it's possible to mention optional dependencies. Is it possible to do that in the conda environment file too? Or should we have a "base_environement.yml" and "opt_environment.yml"?

adehecq (dismissed):
Apart from some thoughts, this seems good!

erikmannerfelt:
I unfortunately had to undo the hacky solution with the post-install script, because it had all kinds of side-effects! One included the proj-data not being installed correctly (probably because of a mix between pip and conda) so the CI testing failed constantly. This should also solve the question by @adehecq  if it overrides an already installed instance of geoutils (it shouldn't anymore)

setup.py
erikmannerfelt: It's unfortunately not possible.. One would indeed have to have an external yml or txt file for that.

environment.yml
erikmannerfelt: This works now with the latest commits.

adehecq (approved):

Issue #114, opened 2021-05-07
Allow access to internal `Coreg` centroid if it exists


erikmannerfelt:
When rotational components of a transformation matrix are involved, the centroid (the pivot point) has to be known. `coreg.ICP` uses the centre of a DEM as the centroid, and this is saved in `ICP._meta["centroid"]`. The `_meta` attribute is prepended with a `_` to mark it as protected (people shouldn't mess around with it). The centroid is however important to extract sometimes, and a user should not need to access the protected member.

 Just like the `Coreg.to_matrix()` function tries to fetch `Coreg._meta["matrix"]`, a `Coreg.centroid` property or function could try to fetch `Coreg._meta["centroid"]`, and may fail or (preferably) return None if it doesn't exist.

Issue #113, opened 2021-05-07
`coreg.ICP` may raise confusing cv2.error instead of a builtin error


erikmannerfelt:
Opencv's errors are often a little obscure, so it's best to replace them with normal `ValueError`s or similar.

In case `coreg.ICP` is met with a DEM with only nans, it will fail with an error message that could have been worse, but it can still be improved.

Minimal example:
```python
In [1]: import xdem                                                                                                                                                                                                                        
                                                                                                                                                                                                                                           
In [2]: import rasterio as rio              

In [3]: import numpy as np                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                                                                                                                                                 
In [4]: affine = rio.transform.from_origin(0, 0, 1, 1)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                         
In [5]: dem1 = np.ones((20, 20), dtype=float)                                                                                                                                                                                              
                                                                                                                                                                                                                                           
In [6]: dem2 = np.ones((20, 20), dtype=float) + np.nan  # create an only-nans DEM                                                                                                
                                                                                                                                                                                                                                           
In [7]: icp = xdem.coreg.ICP()                                                                                                                                                                                                             
                                                                                                                                                                                                                                           
In [8]: icp.fit(dem1, dem2, transform=affine)                                                                                                                                                                                             
---------------------------------------------------------------------------                                                                                                                                                                
error                                     Traceback (most recent call last)                                                                                                                                                                
<ipython-input-10-c9212f5bc980> in <module>                                                                                                                                                                                                
----> 1 icp.fit(dem1, dem2, transform=affine)                                                                                                                                                                                              
                                                                                                                                                                                                                                           
~/.local/share/conda/terradem/lib/python3.9/site-packages/xdem/coreg.py in fit(self, reference_dem, dem_to_be_aligned, inlier_mask, transform, weights, subsample, verbose)                                                                
    507                                                                                                                                                                                                                                    
    508         # Run the associated fitting function                                                                                                                                                                                      
--> 509         self._fit_func(ref_dem=ref_dem, tba_dem=tba_dem, transform=transform, weights=weights, verbose=verbose)                                                                                                                    
    510                                                                                                                                                                                                                                    
    511         # Flag that the fitting function has been called.

[...]

Issue #112, opened 2021-05-06
`coreg.BiasCorr` warns for all-nans but doesn't fail


erikmannerfelt:
`BiasCorr` should arguably fail if it encounters a DEM with only nans. Right now, a numpy warning occurs which is not very useful in identifying the problem.

Minimal example:

```python
In [1]: import xdem

In [2]: import numpy as np

In [3]: biascorr = xdem.coreg.BiasCorr()

In [4]: dem1 = np.ones((50, 50))

In [5]: dem2 = np.ones((50, 50)) + np.nan

In [6]: biascorr.fit(dem1, dem2, transform=None)
/home/erik/.local/share/conda/xdem/lib/python3.8/site-packages/numpy/lib/function_base.py:380: RuntimeWarning: Mean of empty slice.
  avg = a.mean(axis)
/home/erik/.local/share/conda/xdem/lib/python3.8/site-packages/numpy/core/_methods.py:188: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
```
A sanity check should be added to make sure that the function fails instead of giving quite useless warnings!

Issue #110, opened 2021-04-30
Validate new `apply_matrix()` function further.


erikmannerfelt:
As mentioned in #87, this function is quite convoluted and could potentially produce odd results in cases that tests don't yet properly cover.

@adehecq mentioned potential issues that may arise here:
https://github.com/GlacioHack/xdem/blob/8faaef1185dcd910a65183def5c4d2a10d8c280e/xdem/coreg.py#L1127
> Would that not cause a problem if the matrix has other elements that are equal to element [2, 3]?
> Also, should it not be strictly equal to 0? The 1e-4 value is somewhat arbitrary here...
> In the worse case, we're just doing extra work for nothing.

and suggested:
>Not the simplest way, but you could (make a copy,) set the unwanted value to 0 and calculate the mean.

It should be validated if this is a problem, and if so, we should try to fix it.

Other stress tests could be:
* Introduce test cases with many nans
* Validate the accuracy of sub-pixel shifts.
* Run a transformation back and forth many times and see if any bias drift happens.
* Try more extreme transformations (I would argue that 10 degrees of rotation is already extreme, but we could make it worse!)

erikmannerfelt:
It seems like it struggles with many nans. This has to be checked further!

PR #121, opened 2021-05-13
Consistency fixes, and added error/residual methods to Coreg class.

erikmannerfelt:
We are getting quite many issues, and many of them are relatively small. I therefore decided to just tackle a few in one PR!

* Solves #114:  A `Coreg.centroid()` method now exists which fetches the centroid if it exists.
* Solves #113: A more descriptive `ValueError` replaces the cv2 error.
* Solves #112: All `Coreg` approaches now raise an error if only nans are encountered.
* Tackles #110: The better "check if matrix is only vertical" suggestion by @adehecq is implemented.
* Tackles #89: `spatial_tools.get_array_and_mask()` now converts to (height, width) and raises an error if it can't be converted.
* Solves #76: A `Coreg.residuals()` method is added. Then a `Coreg.error()`method is added for preset error measures made form the residuals.
* Solves #66: A check now exists for the `spatial_tools.nmad()` to work properly with masked_arrays. (previously, the mask was ignored)
* Solves #122: There is a clear performance increase on my laptop, but not always on GitHub's CI. Therefore, I changed the test to be "make sure that it's faster or not much worse with subsampling."

### DIFF
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -330,7 +330,8 @@ class TestCoregClass:
         icp_sub_duration = time.time() - start_time
 
         # Make sure that the subsampling increased performance
-        assert icp_full_duration > icp_sub_duration
+        # Temporarily add a fallback assertion that if it's slower, it shouldn't be much slower (2021-05-17).
+        assert icp_full_duration > icp_sub_duration or (abs(icp_full_duration - icp_sub_duration) < 1)
 
         # Calculate the difference in the full vs. subsampled ICP matrices
         matrix_diff = np.abs(icp_full.to_matrix() - icp_sub.to_matrix())
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -17,6 +17,7 @@ import cv2
 import geoutils as gu
 import numpy as np
 import pytest
+import rasterio as rio
 import pytransform3d.transformations
 
 with warnings.catch_warnings():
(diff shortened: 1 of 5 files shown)


xdem/coreg.py
```diff
@@ -596,6 +593,90 @@ def to_matrix(self) -> np.ndarray:
...
+        # Create the full inlier mask (manual inliers plus non-nans)
+        full_mask = (~ref_mask) & np.isfinite(aligned_dem) & inlier_mask
+
+        # Calculate the DEM difference
+        diff = ref_arr - aligned_dem
+
+        # Return the difference values within the full inlier mask
+        return diff[full_mask]
+
+    def error(self, reference_dem: Union[np.ndarray, np.ma.masked_array],
```
rhugonnet, line 638: it's often quite useful to know how many pixels were used to derive the statistics in the residuals, should we have a "count" option as well somewhere in here?

xdem/coreg.py
```diff
@@ -596,6 +593,90 @@ def to_matrix(self) -> np.ndarray:
...
+
+        if error_type == "nmad":
+            error = xdem.spatial_tools.nmad(residuals)
+        elif error_type == "median":
+            error = np.median(residuals)
+        elif error_type == "mean":
+            error = np.mean(residuals)
+        elif error_type == "std":
+            error = np.std(residuals)
+        elif error_type == "rms":
```
rhugonnet, line 673: we could add the Mean Absolute Error, less sensitive to outliers (https://medium.com/human-in-a-machine-world/mae-and-rmse-which-metric-is-better-e60ac3bde13d)

rhugonnet (dismissed):
Very nice list of fixes :+1:

xdem/coreg.py
erikmannerfelt: Added in 43b1d3e

xdem/coreg.py
erikmannerfelt: Added in 43b1d3e

tests/test_coreg.py
```diff
@@ -109,6 +110,51 @@ def test_bias(self):
...
+    def test_all_nans(self):
+        """Check that the coregistration approaches fail gracefully when given only nans."""
+        dem1 = np.ones((50, 50), dtype=float)
+        dem2 = dem1.copy() + np.nan
+        affine = rio.transform.from_origin(0, 0, 1, 1)
+
+        biascorr = coreg.BiasCorr()
+        icp = coreg.ICP()
+        
+        pytest.raises(ValueError, biascorr.fit, dem1, dem2, transform=affine)
```
adehecq, line 122: Does this function just check that the method biascorr.fit with arguments dem1, dem2 and tranform raises a ValueError??

xdem/coreg.py
```diff
@@ -596,6 +593,99 @@ def to_matrix(self) -> np.ndarray:
...
+
+        # Calculate the DEM difference
+        diff = ref_arr - aligned_dem
+
+        # Return the difference values within the full inlier mask
+        return diff[full_mask]
+
+    def error(self, reference_dem: Union[np.ndarray, np.ma.masked_array],
+              dem_to_be_aligned: Union[np.ndarray, np.ma.masked_array],
+              error_type: str = "nmad",
```
adehecq, line 640: Looks really great! Just one comment is that sometimes one might want more than just one statistics, and it is counter-productive to calculate the residuals each time. So it would be nice to be able to request several statistics at once rather than only one...

xdem/ddem.py
```diff
@@ -137,16 +137,13 @@ def interpolate(self, method: str = "linear",
...
 
-            nans = np.isnan(np.asarray(self.data)) | (self.data.mask
-                                                      if isinstance(self.data, np.ma.masked_array)
-                                                      else False)
-            interpolated_ddem = np.where(nans, np.nan, np.asarray(self.data))
+            interpolated_ddem, nans = xdem.spatial_tools.get_array_and_mask(self.data.copy())
             entries = mask.ds[mask.ds.intersects(shapely.geometry.box(*self.bounds))]
 
-            ddem_mask = nans.copy()
+            ddem_mask = nans.copy().squeeze()
```
adehecq, line 143: I thought in this new implementation, the squeeze was not needed anymore?

adehecq:
Very nice fixes!

tests/test_coreg.py
erikmannerfelt: Yes!

PR #123, opened 2021-05-18
Add dev-environment.yml

erikmannerfelt:
In order to simplify installation, this may be of use. The full installation should now be:
```bash
git clone git+https://github.com/GlacioHack/xdem.git && cd xdem
conda env create -f dev-environment.yml
conda activate xdem-dev
pip install -e .
pytest -n4
```
I tried to do `--editable=.` to install xdem, but it didn't work for some reason... We should merge #120  before this as it may change the syntax slightly.

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -1,0 +1,37 @@
+name: xdem-dev
+
+channels:
+  - conda-forge
+
+dependencies:
+        - python>=3
+        - proj>=7.2
+        - geopandas
+        - ipython
+        - numba
+        - matplotlib
+        - pdal
+        - opencv
+        - pyproj
+        - rasterio
+        - scipy
+        - tqdm
+        - scikit-image
+        - pip
+        - proj-data
+
+        # Development-specific
+        - pytest
+        - pytest-xdist
+        - sphinx
+        - numpydoc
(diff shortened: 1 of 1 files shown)


adehecq:
I think instead of duplicating the dependencies in dev, we should have a "base-environment.yml" with the basic requirements needed to run xdem. Then a "dev-environment.yml" which would should be used to update the environment.with `conda env update`

erikmannerfelt:
> I think instead of duplicating the dependencies in dev, we should have a "base-environment.yml" with the basic requirements needed to run xdem. Then a "dev-environment.yml" which would should be used to update the environment.with `conda env update`

The main advantage of conda environments is the dependency solving, so it's always better to use an `environment.yml` over `conda install firstpackage && conda install secondpackage` etc. Therefore, running `conda update -n xdem -f dev-environment.yml` is more or less against conda's premise, as the dependencies of `base-environment.yml` cant't be modified even if they should be (you'll get that weird "found conflicts!" thing that takes 85080923904 years to complete).

So it's probably best to have a full `dev-environment.yml` with all dependencies, or just keep it like we have it (the dev stuff is installed with pip).

erikmannerfelt:
This is now obsolete in favour of #133.

PR #125, opened 2021-05-19
Add a filter module and several filters.

adehecq:
I've added several filters to the newly created filters module. For now it features:
- several implementations of the Gaussian filters (openCV, scipy)
- all filters are insensitive to NaNs, i.e. the filtered value ignore NaNs
- an outlier filtering based on a threshold distance to the Gaussian filtered raster

There are some differences between opencv's and scipy's Gaussian filters that I don't understand, in particular regarding the kernel size. So for now, the kernel size is left to default, which seem to yield similar results for both implementations. This is not a big deal as most users won't care about the kernel size, only on the sigma value.
For example:
```
V1 = scipy.ndimage.gaussian_filter(V, sigma=10, truncate=4)
V2 = cv.GaussianBlur(V, (40, 40), sigmaX=10, borderType=cv.BORDER_REFLECT)
```
do not yield the same results, while:
```
V2 = cv.GaussianBlur(V, (65, 65), sigmaX=10, borderType=cv.BORDER_REFLECT)
```
does approximately... I don't understand where the value of 65 comes from.

Note that in most cases, opencv's implementation is faster (up to 4x), but it seems to depend on the value of sigma, so I leave both options.

TO DO:
- median filters
- min/max filters

### DIFF
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -1,3 +1,8 @@
+"""
+Functions to test the filters.py submodule.
+"""
+from __future__ import annotations
+
 import numpy as np
 import pytest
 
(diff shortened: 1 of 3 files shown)


adehecq:
@erikmannerfelt is there another place than in the __init__ file where I need to add the new filters module?

erikmannerfelt:
> @erikmannerfelt is there another place than in the **init** file where I need to add the new filters module?

It would be nice with a documentation page, but other than that, no!

Also, the tests should pass now if you just merge from main.

erikmannerfelt (approved):
Again, documentation would have been nice, but we can fix that later if you want!

PR #131, opened 2021-05-26
Change URLs in accordance with geoutils master->main rename.

erikmannerfelt:
The default branch name in GeoUtils has been changed from `master` to `main`, so the download URLs have to be changed accordingly.


adehecq (approved):
Simple!

PR #133, opened 2021-05-27
Update dependencies and version

erikmannerfelt:
* All dependencies are now on conda-forge!
* Removed `richdem` from the `environment.yml` as it is an optional dependency
* Removed `pdal` as it is no longer in use.
* Set minimum python version to 3.7 (because of type annotations)
* Ran `black` on `setup.py` since formatting was all over the place.
* The `build` workflow no longer uses pip to install dependencies.
* Prepared `build` workflow to build on Windows and MacOS (didn't enable it yet).
* Updated `README.md` accordingly.

In 0c3c06f, I also added:
* Documentation versioning now dynamically follows the setup.py version (it was hardcoded before..)
* Added manual dispatch to PyPi publishing workflow (see https://github.com/GlacioHack/GeoUtils/pull/192)
* Bumped version to `v0.0.3` (bugfixes on all fronts)

In aa3485f, I added a `dev-environment.yml`, making #123  obsolete.

In ccb4885 and 82b702b:
* Added missing dev requirement
* Fixed test in `test_coreg.py` that sometimes fails randomly!
* Simplified doc building test, making it easier to read, and easier to work with on Windows.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,30 +11,47 @@ on:
 
 jobs:
   build:
+    name: ${{ matrix.os }}, python ${{ matrix.python-version}}
+    runs-on: ${{ matrix.os }}
 
-    runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        os: ["ubuntu-latest"]
+        python-version: ["3.7", "3.8", "3.9"]
+
+
+    # Run all shells using bash (including Windows)
+    defaults:
+      run:
+        shell: bash -l {0}
 
     steps:
     - uses: actions/checkout@v2
+
+    # Set up the conda-forge environment with mamba
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+      uses: conda-incubator/setup-miniconda@v2
       with:
+        auto-update-conda: true
         python-version: ${{ matrix.python-version }}
+        mamba-version: "*"
+        channels: conda-forge
+        channel-priority: strict
+        environment-file: environment.yml
+        activate-environment: xdem
+
+    # Install the dependencies that are not in the environment file
     - name: Install dependencies
       run: |
-        $CONDA/bin/conda install -c conda-forge mamba -y
-        $CONDA/bin/mamba env update --file environment.yml --name base
-        $CONDA/bin/pip install -r dev-requirements.txt
-        $CONDA/bin/pip install .
+        mamba install sphinx numpydoc sphinx_rtd_theme sphinx-autodoc-typehints sphinxcontrib-programoutput flake8 pytest pylint
+        pip install . --no-dependencies
+
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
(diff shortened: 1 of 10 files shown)


.github/workflows/python-package.yml
```diff
@@ -11,30 +11,47 @@ on:
...
+        activate-environment: xdem
+
+    # Install the dependencies that are not in the environment file
     - name: Install dependencies
       run: |
-        $CONDA/bin/conda install -c conda-forge mamba -y
-        $CONDA/bin/mamba env update --file environment.yml --name base
-        $CONDA/bin/pip install -r dev-requirements.txt
-        $CONDA/bin/pip install .
+        mamba install sphinx numpydoc sphinx_rtd_theme sphinx-autodoc-typehints sphinxcontrib-programoutput flake8 pytest pylint
```
adehecq, line 46: Is it better to have it hard-coded here rather than in an environment.yml?

docs/source/conf.py
```diff
@@ -25,7 +25,7 @@
 author = 'xdem contributors'
 
 # The full version, including alpha/beta/rc tags
-release = "0.0.1"
+release = xdem.version.version
```
adehecq, line 28: Nice!

setup.py
```diff
@@ -5,31 +5,44 @@
 from setuptools import setup
 from setuptools.command.install import install
 
-FULLVERSION = '0.0.2-2'
+FULLVERSION = "0.0.3"
```
adehecq, line 8: I see this is hard-coded here. So does it have to be updated each time? Or where should we modify the version number?

setup.py
```diff
@@ -5,31 +5,44 @@
...
+    license="BSD-3",
+    packages=["xdem"],
+    install_requires=[
+        "numpy",
+        "scipy",
+        "rasterio",
+        "geopandas",
+        "pyproj",
+        "tqdm",
+        "scikit-gstat",
```
adehecq, line 32: Ideally this would also be an optional dependency as most users will be not necessarily be interested in the spstats part... I think it adds up a lot of additional dependencies no?

setup.py
erikmannerfelt: Yes, and there's no apparent way to fix this. It's here and in the GitHub release part. Everything else is propagated from them.

adehecq:
Great!
The github parts are still a mystery to me, so I trust you on this!
I made some comments, but feel free to report them as issue or keep them in mind for later if you need to merge this PR.

adehecq (dismissed):

setup.py
erikmannerfelt: In its current state, the spstats dependencies are required. We could discuss in the future if that should be the case (@rhugonnet), but for now, this is how it should be!

.github/workflows/python-package.yml
erikmannerfelt: Added a `dev-environment.yml` in aa3485f!

erikmannerfelt:
I just saw that the documentation doesn't build now on readthedocs. This needs to be fixed before merging

erikmannerfelt:
readthedocs is happy again! @adehecq

erikmannerfelt:
I'll merge this as it was approved by @adehecq and I had to dismiss his approval as readthedocs started complaining. But now that's fixed!

PR #135, opened 2021-05-28
Solves several issues with Nuth&Kaab coregistration

adehecq:


adehecq:
@erikmannerfelt Should I just get rid of the progress bar? It looks ugly when the iterations are being stopped, which is the case most of the time I would assume now. And it would make handling the verbose easier. Right now I allow two levels of verbose...
image

xdem/coreg.py
```diff
@@ -1074,26 +1074,31 @@ class NuthKaab(Coreg):
...
-        self.error_threshold = error_threshold
+        self.offset_threshold = offset_threshold
 
         super().__init__()
 
     def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optional[rio.transform.Affine],
-                  weights: Optional[np.ndarray], verbose: bool = False):
+                  weights: Optional[np.ndarray], verbose: Union[bool, int] = False):
         """Estimate the x/y/z offset between two DEMs."""
+        if verbose > 0:
```
rhugonnet, line 1092: you can simply write `if verbose`

rhugonnet: I just saw you added a second level of verbose, my bad!

adehecq:
Left to do:
- add option to plot figures of intermediate results
- check why coreg does not work on some own DEMs, while it works with old geoutils...

rhugonnet:
Should we think of a reproducible format of plotting function we can use for any coreg/bias correction (start/end at least, and during iteration for some)?
For saving plots, I'd recommend putting everything in a beautiful PDF (like this https://github.com/iamdonovan/pybob/blob/master/pybob/coreg_tools.py#L444)

erikmannerfelt:
@adehecq sure, the progress bar can go! I still recommend tqdm without specifying the total, as it nicely updates on the iteration count, number of iterations per second, the total elapsed time, and any other optional text (such as the current NMAD).

adehecq:
> @adehecq sure, the progress bar can go! I still recommend tqdm without specifying the total, as it nicely updates on the iteration count, number of iterations per second, the total elapsed time, and any other optional text (such as the current NMAD).

What do you mean exactly?
Currently the loop is as follow:
`for i in trange(self.max_iterations, disable=not progressbar):`
If the progress bar is enabled and I print any statement within the loop, then the progress bar and prints intertwine and it looks ugly. But if there is a better way to do it, let me know!

erikmannerfelt:
> > @adehecq sure, the progress bar can go! I still recommend tqdm without specifying the total, as it nicely updates on the iteration count, number of iterations per second, the total elapsed time, and any other optional text (such as the current NMAD).
> 
> What do you mean exactly?
> Currently the loop is as follow:
> `for i in trange(self.max_iterations, disable=not progressbar):`
> If the progress bar is enabled and I print any statement within the loop, then the progress bar and prints intertwine and it looks ugly. But if there is a better way to do it, let me know!

Use the `tqdm` class:
```python
>>> from tqdm import tqdm
>>> import time
>>> progress_bar = tqdm()
>>> for i in range(1000):
...     time.sleep(1)
...     progress_bar.update()
...     progress_bar.desc = f"Running iteration {i}"
... progress_bar.close()
Running iteration 0: 2it [00:35, 14.99s/it]
```
It could also be used in a `with` context:
```python
with tqdm() as progress_bar:
    for i in range(1000):
        time.sleep(1)
        progress_bar.update()
        progress_bar.desc = f"Running iteration {i}"
```

erikmannerfelt:
It's great that you're finding these issues! Can you find a way to improve the tests, so that we know for certain that it works?

xdem/coreg.py
```diff
@@ -892,13 +892,15 @@ class Deramp(Coreg):
     Estimates an n-D polynomial between the difference of two DEMs.
     """
 
-    def __init__(self, degree: int = 1):
+    def __init__(self, degree: int = 1, max_points: int = 5e5):
```
erikmannerfelt, line 895: Nice! It kind of reinvents the `subsample` argument for `self.fit()`. On the other hand, the `subsample` argument cannot currently subsample differently for different parts in the pipeline. 

I'll add an issue for that, and we can have it like you implemented right now (but it would be nice to have automatically for all `Coreg` subclasses!)

rhugonnet: +1

adehecq: My idea was that this argument would be used by the subsampling functionality directly. I could just rename "max_point" into "subsample" if that makes it more consistent.

erikmannerfelt: Sounds good!

erikmannerfelt: I added the `Coreg` implementation as an issue in #137

rhugonnet: It would be good to have an optional argument that'd make a call to a subsampling function that is the same everywhere (we have one in `spstats_tools` too).  For instance =None would mean no subsampling.
Maybe it would best fit in `spatial_tools`?

erikmannerfelt: That's a great idea!

adehecq: Yes, that would be even better!

adehecq:
> Use the `tqdm` class:
> 
> ```python
> >>> from tqdm import tqdm
> >>> import time
> >>> progress_bar = tqdm()
> >>> for i in range(1000):
> ...     time.sleep(1)
> ...     progress_bar.update()
> ...     progress_bar.desc = f"Running iteration {i}"
> ... progress_bar.close()
> Running iteration 0: 2it [00:35, 14.99s/it]
> ```
> 
> It could also be used in a `with` context:
> 
> ```python
> with tqdm() as progress_bar:
>     for i in range(1000):
>         time.sleep(1)
>         progress_bar.update()
>         progress_bar.desc = f"Running iteration {i}"
> ```

It was even simpler than this. I found out that tqdm objects have a write method to print statements on screen. See dc93dfc.
But I'm not really convinced by the usefulness of the progressbar now that each iteration is quick and the algorithm usually converges in 2-3 iterations...

xdem/coreg.py
adehecq: Damn, that function is harder than I expected to write! 
I want to have an option to return the indices of the array that are randomly extracted (for example if willing to extract the same random sample from different DEMs), but I have issues getting those indexes in a general case for a 1D to 3D array, when there are masked values.
This is what I have so far:
```
def subsample_raster(array: Union[np.ndarray, np.ma.masked_array], subsample: Union[float, int],
                     return_indices: bool = False) -> np.ndarray:
    """
    Randomly subsample a 1D or 2D array by a subsampling factor, taking only non NaN/masked values.

    :param subsample: If < 1, will be considered a fraction of valid pixels to extract.
If > 1 will be considered the number of pixels to extract.
    :param return_indices: If set to True, will also return the extracted indices.
 
    :returns: The subsampled array (1D)
    """
    # Get number of points to extract
    if (subsample <= 1) & (subsample >= 0):
        npoints = subsample * np.size(array)
    elif subsample > 1:
        npoints = int(subsample)

    # Remove invalid values and flatten array
    orig_shape = array.shape
    mask = get_mask(array)
    array = array[~mask]

    # Checks that array and npoints are correct
    assert np.ndim(array) == 1
    if npoints >= len(array):
        npoints = len(array)

    # Randomly extract npoints without replacement
    indices = np.random.choice(len(array), npoints, replace=False)
```
But `indices` is for now a flat array, whose length is not necessarily equal to the number of pixels in `array` or `mask`... I don't know how to easily get back to the original pixels coordinates. The easiest I see is to create an array containing the pixel position in each dimension, but for a 3D array, I would basically create three arrays of the same size as my input, which could possibly use a lot of memory!
Any ideas welcome.

adehecq:
Left to do:
- improve tests, in particular to include more NaNs, and test several subfunctions like get_horizontal_shift
- add a raster subsampling function
This PR took a lot more time than I though. Maybe for the sake of fixing the major bugs, should we merge the PR as it is, @erikmannerfelt and @rhugonnet ? Then I can add the above point to the issue list.

xdem/coreg.py
```diff
@@ -892,15 +892,17 @@ class Deramp(Coreg):
...
 
-    def __init__(self, degree: int = 1, max_points: int = 5e5):
+    def __init__(self, degree: int = 1, subsample: Union[int, float] = 5e5):
         """
         Instantiate a deramping correction object.
 
         :param degree: The polynomial degree to estimate. degree=0 is a simple bias correction.
-        :param max_points: The maximum number of random points to extract for fitting (for speed-up).
+        :param subsample: Factor for subsampling the input raster for speed-up.
+        If < 1, will be considered a fraction of valid pixels to extract.
```
erikmannerfelt, line 901: What if it's 1.0? Maybe better to say "If <= 1, will be considered a fraction ..."

xdem/coreg.py
```diff
@@ -947,12 +949,22 @@ def residuals(coefs: np.ndarray, x_coords: np.ndarray, y_coords: np.ndarray, tar
             print("Estimating deramp function...")
 
         # reduce number of elements for speed
-        if len(x_coords) > self.max_points:
-            indices = np.random.randint(0, len(x_coords) - 1, self.max_points)
+        # Get number of points to extract
+        max_points = len(x_coords)
```
erikmannerfelt, line 953: `len(arr)` is always equal to `arr.shape[0]`. Should it not be `.size`?
```python
In [1]: import numpy as np

In [2]: np.ones((4, 4))
Out[2]:
array([[1., 1., 1., 1.],
       [1., 1., 1., 1.],
       [1., 1., 1., 1.],
       [1., 1., 1., 1.]])

In [3]: len(np.ones((4, 4)))
Out[3]: 4

In [4]: np.ones((4, 4)).size
Out[4]: 16
```

xdem/coreg.py
erikmannerfelt: The already existing subsample script solves this I think:
https://github.com/GlacioHack/xdem/blob/74cf3528daa3bc35b853281e66dad05eb1b7a2f8/xdem/coreg.py#L498

It takes a random choice of row/col indices that are finite.

erikmannerfelt:
> Left to do:
> 
>     * improve tests, in particular to include more NaNs, and test several subfunctions like get_horizontal_shift
> 
>     * add option for plotting some outputs
> 
>     * add a raster subsampling function
>       This PR took a lot more time than I though. Maybe for the sake of fixing the major bugs, should we merge the PR as it is, @erikmannerfelt and @rhugonnet ? Then I can add the above point to the issue list.

I agree that your points can be issues for later instead! I just have some questions and then I am happy

xdem/coreg.py
adehecq: I didn't know this existed! I need to check that out. Anyway, this should be a separate function as we discussed.

The issue with how it is implemented now is that it masks pixels. This raises two issues:
- it does not reduce the number of points, so we're possibly duplicating the data. Ideally, we should pass a flattened 1D array to the underlying _fit_func, but then they all need to accept 1d arrays. This will not work for Nuth&Kaab yet as we need a raster to calculate the slope/aspect.
- masking random pixels will have a terrible effect on the Nuth & Kaab as the slope/aspect need to be calculated, which involve using neighbor pixels that might possibly be masked...

adehecq: This also explains my question to you regarding the inliers_mask @erikmannerfelt. Currently, by default subsample = 1, which means inlier_mask is never used!

xdem/coreg.py
adehecq: x_coords is a 1D array so it does not really matter here. But for clarity I will replace with size!

xdem/coreg.py
adehecq: Solved in 2ba3b5d

xdem/coreg.py
adehecq: Solved in 2ba3b5d

xdem/coreg.py
adehecq: There's an issue with your subsampling method... rows and cols are reversed!
This is why we should have a separate function for this and test it.
Also, for now it is only implemented for a 2D case. I would like it to be more generic.

erikmannerfelt (approved):
Nice!

xdem/coreg.py
adehecq: I found an easy way to code this subsampling function. I will add it as an issue and work on it in a different PR.

Issue #132, opened 2021-05-27
Add `verbose` argument to `xdem.spstats.patches_method` and/or change prints to `tqdm`


erikmannerfelt:
Whenever tests fail, we get this beautiful output:
```
Datasets exist
Number of valid pixels: 1138155
Found valid cadrant 1 (maximum: 1000)
Found valid cadrant 2 (maximum: 1000)
Found valid cadrant 3 (maximum: 1000)
Found valid cadrant 4 (maximum: 1000)
Found valid cadrant 5 (maximum: 1000)
Found valid cadrant 6 (maximum: 1000)
Found valid cadrant 7 (maximum: 1000)
Found valid cadrant 8 (maximum: 1000)
Found valid cadrant 9 (maximum: 1000)
Found valid cadrant 10 (maximum: 1000)
Found valid cadrant 11 (maximum: 1000)
Found valid cadrant 12 (maximum: 1000)
Found valid cadrant 13 (maximum: 1000)
Found valid cadrant 14 (maximum: 1000)
Found valid cadrant 15 (maximum: 1000)
Found valid cadrant 16 (maximum: 1000)
Found valid cadrant 17 (maximum: 1000)
Found valid cadrant 18 (maximum: 1000)
Found valid cadrant 19 (maximum: 1000)
Found valid cadrant 20 (maximum: 1000)
Found valid cadrant 21 (maximum: 1000)
Found valid cadrant 22 (maximum: 1000)
Found valid cadrant 23 (maximum: 1000)
Found valid cadrant 24 (maximum: 1000)
Found valid cadrant 25 (maximum: 1000)
Found valid cadrant 26 (maximum: 1000)
Found valid cadrant 27 (maximum: 1000)
Found valid cadrant 28 (maximum: 1000)
Found valid cadrant 29 (maximum: 1000)
Found valid cadrant 30 (maximum: 1000)
Found valid cadrant 31 (maximum: 1000)
Found valid cadrant 32 (maximum: 1000)
Found valid cadrant 33 (maximum: 1000)
Found valid cadrant 34 (maximum: 1000)
Found valid cadrant 35 (maximum: 1000)
Found valid cadrant 36 (maximum: 1000)
Found valid cadrant 37 (maximum: 1000)
Found valid cadrant 38 (maximum: 1000)
Found valid cadrant 39 (maximum: 1000)
Found valid cadrant 40 (maximum: 1000)
Found valid cadrant 41 (maximum: 1000)
[...]

Issue #45, opened 2021-03-11
Implement a method to de-bias the random.choice sampling for pairwise distances in variogram estimation


rhugonnet:
Add option to sample_multirange_empirical_variogram():
Pick random point in raster + sample in rings of varying sizes to allow for a good distribution of samples for all distances and increase speed

rhugonnet:
Closed by #136

PR #136, opened 2021-06-01
Improve sampling for variogram estimation: circular and ring sampling + tests

rhugonnet:
Added ring/circular shape masking to subsample rasters.
This will extremely useful to better estimate empirical variograms from the raster samples. Because we generally have millions of samples in a raster, we can't calculate the millions^{squared} pairwise differences. Sampling 10,000 points is already quite a long calculation (~1min). If we subsample within rings of different radius, we limit the samples to similar pairwise distances, thus improving the empirical estimation with less unused samples.
TODO: write the parent function to subsample with the ring method, iteratively, for different radius sizes.

### DIFF
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -162,39 +162,53 @@ class TestSubSampling:
         """Test that the circular masking works as intended"""
 
         # using default (center should be [2,2], radius 2)
-        circ = xdem.spstats.create_circular_mask(5,5)
-        circ2 = xdem.spstats.create_circular_mask(5,5,center=[2,2],radius=2)
+        circ = xdem.spstats.create_circular_mask((5,5))
+        circ2 = xdem.spstats.create_circular_mask((5,5),center=[2,2],radius=2)
 
+        # check default center and radius are derived properly
         assert np.array_equal(circ,circ2)
 
+        # check mask
+        # masking is not inclusive, i.e. exactly radius=2 won't include the 2nd pixel from the center, but radius>2 will
+        eq_circ = np.zeros((5,5), dtype=bool)
+        eq_circ[1:4,1:4]=True
+        assert np.array_equal(circ,eq_circ)
+
         # check distance is not a multiple of pixels (more accurate subsampling)
-
         # will create a 1-pixel mask around the center
-        circ3 = xdem.spstats.create_circular_mask(5,5,center=[1,1],radius=1)
-        # will create a square mask (<1.5 pixel) around the center
-        circ4 = xdem.spstats.create_circular_mask(5,5,center=[1,1],radius=1.5)
+        circ3 = xdem.spstats.create_circular_mask((5,5),center=[1,1],radius=1)
 
+        eq_circ3 = np.zeros((5,5), dtype=bool)
+        eq_circ3[1,1] = True
+        assert np.array_equal(circ3, eq_circ3)
+
+        # will create a square mask (<1.5 pixel) around the center
+        circ4 = xdem.spstats.create_circular_mask((5,5),center=[1,1],radius=1.5)
+        # should not be the same as radius = 1
         assert not np.array_equal(circ3,circ4)
 
 
     def test_ring_masking(self):
         """Test that the ring masking works as intended"""
 
-        # by default, the mask is only False (ring of size 0)
-        ring1 = xdem.spstats.create_ring_mask(5,5)
+        # by default, the mask is only an outside circle (ring of size 0)
+        ring1 = xdem.spstats.create_ring_mask((5,5))
+        circ1 = xdem.spstats.create_circular_mask((5,5))
 
-        assert np.array_equal(ring1,np.zeros((5,5)))
+        assert np.array_equal(ring1,circ1)
 
         # test rings with different inner radius
-        ring2 = xdem.spstats.create_ring_mask(5,5,in_radius=1,out_radius=2)
-        ring3 = xdem.spstats.create_ring_mask(5,5,in_radius=0,out_radius=2)
-        ring4 = xdem.spstats.create_ring_mask(5,5,in_radius=1.5,out_radius=2)
+        ring2 = xdem.spstats.create_ring_mask((5,5),in_radius=1,out_radius=2)
+        ring3 = xdem.spstats.create_ring_mask((5,5),in_radius=0,out_radius=2)
+        ring4 = xdem.spstats.create_ring_mask((5,5),in_radius=1.5,out_radius=2)
 
         assert np.logical_and(~np.array_equal(ring2,ring3),~np.array_equal(ring3,ring4))
 
-        eq_ring3 = xdem.spstats.create_circular_mask(5,5)
-        eq_ring3[2,2] = 0
-        assert np.array_equal(ring3,eq_ring3)
+        # check default
+        eq_ring2 = np.zeros((5,5), dtype=bool)
+        eq_ring2[1:4,1:4] = True
+        eq_ring2[2,2] = False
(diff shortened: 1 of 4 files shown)


erikmannerfelt:
Nice! Especially the `verbose` setting ;) 

Impressively old branch haha! 129 commits behind main!

rhugonnet:
@erikmannerfelt Any idea why mamba installation fails? I don't get it, nothing should have changed on this

erikmannerfelt:
@rhugonnet, if you check the log, it says "Connection broken" on that step. It's a bug with GitHub (not related to xdem) where it sometimes just randomly fails... When that happens, you have to trigger a re-run of all jobs!

I just triggered a re-run.

tests/test_spstats.py
```diff
@@ -18,11 +19,10 @@
     warnings.filterwarnings("ignore", category=DeprecationWarning)
     from skgstat import models
 
-
 PLOT = False
 
-
-def load_diff() -> tuple[gu.georaster.Raster, np.ndarray]:
+def load_diff() -> Tuple[gu.georaster.Raster, np.ndarray] :
```
erikmannerfelt, line 24: Why did you change this from `tuple` to `Tuple`? If a linter suggests it, the linter is outdated.

rhugonnet: I does this to me for `list` and `tuple`, and wants me to import `List` and `Tuple` from annotations instead. Are those two similarly outdated?

rhugonnet:
> @rhugonnet, if you check the log, it says "Connection broken" on that step. It's a bug with GitHub (not related to xdem) where it sometimes just randomly fails... When that happens, you have to trigger a re-run of all jobs!
> 
> I just triggered a re-run.

Thanks, good to know! Didn't see it on the log (I did not think it could happen actually).

xdem/spstats.py
```diff
@@ -91,6 +90,73 @@ def random_subset(dh: np.ndarray, coords: np.ndarray, nsamp: int):
 
     return dh_sub, coords_sub
 
+def create_circular_mask(h: float, w: float, center: Optional[list[float]] = None, radius: Optional[float] = None) -> np.ndarray:
+    """
+    Create circular mask on a raster
+
+    :param h: height of array
```
adehecq, line 97: Maybe use a shape directly? It's simpler as one can use `array.shape` as input instead.

xdem/spstats.py
```diff
@@ -91,6 +90,73 @@ def random_subset(dh: np.ndarray, coords: np.ndarray, nsamp: int):
 
     return dh_sub, coords_sub
 
+def create_circular_mask(h: float, w: float, center: Optional[list[float]] = None, radius: Optional[float] = None) -> np.ndarray:
```
adehecq, line 93: Just found out there is a skimage function that does just that! https://www.kite.com/python/docs/skimage.draw.circle

xdem/spstats.py
```diff
@@ -91,6 +90,73 @@ def random_subset(dh: np.ndarray, coords: np.ndarray, nsamp: int):
...
+    """
+    Create circular mask on a raster
+
+    :param h: height of array
+    :param w: width of array
+    :param center: center
```
adehecq, line 99: Good to mention what the default will be, as it's unclear from the function definition.

xdem/spstats.py
```diff
@@ -91,6 +90,73 @@ def random_subset(dh: np.ndarray, coords: np.ndarray, nsamp: int):
...
+
+    mask = dist_from_center <= radius
+
+    return mask
+
+def create_ring_mask(h: float, w: float, center: Optional[list[float]] = None, in_radius: float = 0., out_radius: float = 0.) -> np.ndarray:
+    """
+    Create ring mask on a raster
+
+    :param h: height of array
```
adehecq, line 120: Same points as for the previous functions about the arguments.

xdem/spstats.py
```diff
@@ -91,6 +90,73 @@ def random_subset(dh: np.ndarray, coords: np.ndarray, nsamp: int):
...
+    :param w: width of array
+    :param center: center
+    :param in_radius: inside radius
+    :param out_radius: outside radius
+    :return:
+    """
+
+    mask_inside = create_circular_mask(h,w,center=center,radius=in_radius)
+    mask_outside = create_circular_mask(h,w,center=center,radius=out_radius)
```
adehecq, line 129: By default, the in_radius and out_radius are equal, so will it not return an empty mask?? I would avoid this as it can be confusing.

tests/test_spstats.py
```diff
@@ -156,6 +156,46 @@ def test_neff_estimation(self):
...
+class TestSubSampling:
+
+    def test_circular_masking(self):
+        """Test that the circular masking works as intended"""
+
+        # using default (center should be [2,2], radius 2)
+        circ = xdem.spstats.create_circular_mask(5,5)
+        circ2 = xdem.spstats.create_circular_mask(5,5,center=[2,2],radius=2)
+
+        assert np.array_equal(circ,circ2)
```
adehecq, line 168: You could also test that it is exactly equal to the matrix you expect i.e. 0 and 1 with the correct shape. This would be a more rigorous test!

tests/test_spstats.py
```diff
@@ -156,6 +156,46 @@ def test_neff_estimation(self):
...
+
+    def test_ring_masking(self):
+        """Test that the ring masking works as intended"""
+
+        # by default, the mask is only False (ring of size 0)
+        ring1 = xdem.spstats.create_ring_mask(5,5)
+
+        assert np.array_equal(ring1,np.zeros((5,5)))
+
+        # test rings with different inner radius
```
adehecq, line 188: Same comment, you could create the matrix as it should look like and compare that they are the same!

xdem/spstats.py
erikmannerfelt: If this is done, the type would be `tuple[int, ...]` as an array's number of dimensions is never known by the type hinting. I just mention it here because I've been stuck on it myself!

rhugonnet: Done

xdem/spstats.py
rhugonnet: Actually `skimage.draw.circle` is being deprecated for `skimage.draw.disk`, which also takes ~3 lines to write because it only outputs coordinates. 
I added the skimage solution too as it's only 3 lines, it behaves slightly differently (interval not inclusive, so radius=2 won't create a circle of 2 pixels, you need to input above, 2.000001 works for instance)

xdem/spstats.py
rhugonnet: Done

xdem/spstats.py
rhugonnet: Now the defaults return a full circle with `inside_radius` of 0

tests/test_spstats.py
rhugonnet: Added

tests/test_spstats.py
rhugonnet: Added

tests/test_spstats.py
rhugonnet: Fixed

rhugonnet:
@erikmannerfelt @adehecq We can probably merge this too, and add the usage of those sampling functions in another PR for more clarity

rhugonnet:
I need an approval here to move on to the next steps :pleading_face:

erikmannerfelt:
You need to solve the merge conflicts first, @rhugonnet!

erikmannerfelt (approved):

rhugonnet:
Resolves #132
Resolves #45

PR #138, opened 2021-06-02
Add more badges!

adehecq:
Badges are fun, badges look nice, here come more!
To see what it looks like: https://github.com/adehecq/xdem/tree/badges


erikmannerfelt (approved):
Nice!

adehecq:
We could possibly add the PyPi badge, but then two lines, one for conda-forge, one for PyPi would make more sense.
Here's the link to the PyPI badge: https://img.shields.io/pypi/v/xdem.svg

PR #141, opened 2021-06-03
Add a function for subsampling functionality

adehecq:
As mentioned in PR #135, a shared subsampling function would be useful.
 
**Done:**
 - created the function
 - added tests
 
**To do:**
- add tests with option 'return_indices"
- replace with this function where needed

### DIFF
--- a/xdem/spatial_tools.py
+++ b/xdem/spatial_tools.py
@@ -3,14 +3,14 @@ A set of basic operations to be run on 2D arrays and DEMs.
 """
 
 from __future__ import annotations
+
 from typing import Callable, Union
 
+import geoutils as gu
 import numpy as np
 import rasterio as rio
 import rasterio.warp
 from tqdm import tqdm
-
-import geoutils as gu
 
 
 def get_mask(array: Union[np.ndarray, np.ma.masked_array]) -> np.ndarray:
@@ -399,8 +399,9 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
     return np.clip(255 * (shaded + 0.6) / 1.84, 0, 255).astype("float32")
 
 
-def subsample_raster(array: Union[np.ndarray, np.ma.masked_array], subsample: Union[float, int],
-                     return_indices: bool = False) -> np.ndarray:
+def subsample_raster(
+    array: Union[np.ndarray, np.ma.masked_array], subsample: Union[float, int], return_indices: bool = False
+) -> np.ndarray:
     """
     Randomly subsample a 1D or 2D array by a subsampling factor, taking only non NaN/masked values.
 
--- a/xdem/spatial_tools.py
+++ b/xdem/spatial_tools.py
@@ -1,6 +1,7 @@
 """
-
+A set of basic operations to be run on 2D arrays and DEMs.
 """
+
 from __future__ import annotations
 from typing import Callable, Union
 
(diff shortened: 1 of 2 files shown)


rhugonnet:
So should we merge this? @adehecq

xdem/spatial_tools.py
```diff
@@ -396,3 +396,42 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
     # Return the hillshade, scaled to uint8 ranges.
     # The output is scaled by "(x + 0.6) / 1.84" to make it more similar to GDAL.
     return np.clip(255 * (shaded + 0.6) / 1.84, 0, 255).astype("float32")
+
+
+def subsample_raster(array: Union[np.ndarray, np.ma.masked_array], subsample: Union[float, int],
```
erikmannerfelt, line 401: The typing is still a bit off in this.. It says that it returns an ndarray, but it may also return a tuple of ndarrays. I suggest that if `return_indices=True`, it only returns the indices, not the array (which is easily created by the user). That way, the return type is consistent.

xdem/spatial_tools.py
```diff
@@ -396,3 +396,42 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+        npoints = int(subsample)
+    else:
+        raise ValueError("`subsample` must be > 0")
+
+    # Remove invalid values and flatten array
+    mask = get_mask(array) #-> need to remove .squeeze in get_mask
+    valids = np.argwhere(~mask.flatten()).squeeze()
+
+    # Checks that array and npoints are correct
+    assert np.ndim(valids) == 1
```
erikmannerfelt, line 425: This will raise an AssertionError without explanation. Could one be added? Python's AssertionErrors are inherently a little annoying I think because they don't tell the user anything!!

xdem/spatial_tools.py
```diff
@@ -396,3 +396,42 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
     # The output is scaled by "(x + 0.6) / 1.84" to make it more similar to GDAL.
     return np.clip(255 * (shaded + 0.6) / 1.84, 0, 255).astype("float32")
+
+
+def subsample_raster(array: Union[np.ndarray, np.ma.masked_array], subsample: Union[float, int],
+                     return_indices: bool = False) -> np.ndarray:
+    """
+    Randomly subsample a 1D or 2D array by a subsampling factor, taking only non NaN/masked values.
+
+    :param subsample: If < 1, will be considered a fraction of valid pixels to extract.
```
erikmannerfelt, line 406: If <= 1 (to be consistent with the functionality)

erikmannerfelt (commented):
Sorry for not having reviewed this earlier. I just have a few small comments that I feel are to a lesser or higher degree important.

xdem/spatial_tools.py
adehecq: Good point, but actually, I don't think this error should ever happen. It's just to make sure that the array has been flattened correctly and has the expected shape.
I can write "Something is wrong with array dimension, check input data and shape"

xdem/spatial_tools.py
adehecq: Good catch!

adehecq: Done in 05f5e40

xdem/spatial_tools.py
adehecq: Done in 05f5e40

xdem/spatial_tools.py
adehecq: Done in 90bd065

erikmannerfelt:
Is this ready for approval? It looks great now in my book!

rhugonnet:
> Is this ready for approval? It looks great now in my book!

+1

adehecq:
> Is this ready for approval? It looks great now in my book!

Yes, I just need to check that the tests pass and run the linters.

rhugonnet (approved):
Not much changed by the linters!

adehecq:
> Not much changed by the linters!

Of course, I naturally write good code! ;-)

Issue #142, opened 2021-06-04
Add piece-wise DEM coregistration


erikmannerfelt:
In photogrammetric workflows, there might be nonlinear errors that need to be corrected for in the bundle adjustment using a good spread of GCPs. In that case, a single transformation matrix cannot explain it, so simple coregistration is not enough.

In https://github.com/erikmannerfelt/VAW-IllgrabenDroneProcessing, I made a script to extract 3D parts of a point cloud and coregister only that part. It made for the coregistration results to be explainable by markers in Metashape. We should have such a function here!

I know that @adehecq  could do with such a function as well with one of his students.

Thoughs?

PR #143, opened 2021-06-04
Add support for nonlinear coregistration

erikmannerfelt:
Solves #142.

# The new PiecewiseCoreg object
In many cases, biases are not systematic throughout the DEM, so a nonlinear correction is required. This new class takes an arbitrary Coreg class and makes it nonlinear!

The class takes two required arguments; the `Coreg` object to apply non-lienarly and the amount of `subdivisions` to calculate it within. In short, what it does is:
###  To get the coregistration fit:
1. Subdivide the DEM in N (`=subdivison`) equally sized chunks.
2. For each chunk, run the coregistration separately, and save the results.
### To apply the coregistration:
1. Take all the coregistration results and describe them as  X/Y/Z shifts in the middle of the grid. These shifts are then spatially interpolated to generate three offset fields.
2. Apply the Z shifts as a simple `dem - shift` correction
3. Apply the X/Y shifts using `skimage.transform.warp`

Below is an example of the Nuth & Kääb approach, running it "globally" and in 64 (8x8) chunks. What we can see quite visibly in the "global" correction (middle figure) is that there is some kind of mosaic seam near 525000 m east. This cannot be accounted for in a "global" approach. The 64 chunk version, however, brings the error down considerably! There are still some weridnesses, but the NMAD has gone down from 2.5 m to 1.7 m. In the upper left corner, you see the Z correction that was estimated from the 64 chunks.

image

Another big feature is that the approach is multithreaded, so running coregistration 64 times took just 4 seconds on my computer. I think I've done it in a memory-saving way (some copying is needed, but it's only done for each grid separately, so it's relatively small for each thread; then it gets garbage-collected).

I've attached the code for the figure below.

## Subdivide array
Subdividing a 2D array into chunks was harder than I thought! I looked everywhere and didn't find a good function that did "divide my 2D array into N unique parts that are more or less identical in size" (the "more or less" occurs because the shape can be odd). Therefore, I made my own!

I thought it  might have some other uses, so I'll present it here:
```python
>>> subdivide_array((4, 4), 4)
array([[0, 0, 1, 1],
       [0, 0, 1, 1],
       [2, 2, 3, 3],
       [2, 2, 3, 3]])
```
Here, we tell it "divide the contents of a 4x4 matrix into four unique groups". It will try to divide the groups so that the count is equal, but this only applies when the `count` argument is divisible by the product of the shape, e.g:
```python
>>> subdivide_array((5, 4), 3)
array([[0, 0, 0, 0],
       [0, 0, 0, 0],
       [1, 1, 2, 2],
       [1, 1, 2, 2],
       [1, 1, 2, 2]])
```

## Unreated fixes
Some tests that shouldn't fail started failing for me, so I had to make some fixes.
1.  The `test_apply_matrix` is now a separate test for a few reasons: it ensures the inputs weren't mutated by other functions, it didn't fit the `TestCoregClass` name, and keeping it separate will make it slightly easier to extend going forward.
2. Documentation is now generated in multiple threads. On my laptop, it reduces the time from ~90s to ~30s!
3. The GitHub actions are now testing in multiple threads. Two cores are available, so it goes about twice as fast! Dependency solving is still the slowest part, but it's now down to about 8 minutes from 10 minutes before. Testing only takes about two minutes now.
4. The `Coreg` and `CoregPipeline` classes now have a `copy()` method with associated tests.
5. `Coreg.error()` can now accept a list of error statistics to calculate, as per @adehecq's suggestion in #121

## TODO:
* Test the function on more datasets
* Create some way of using these results as synthetic GCPs in photogrammetric software such as Metashape (this may not fit in xdem, but in a separate toolbox perhaps)
* The new `warp_dem()` function slightly repeats what's in `apply_matrix()`. I think `apply_matrix()` can be improved by using `warp_dem()` under the hood, but that's for a future PR!

### DIFF
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -544,3 +544,56 @@ def test_apply_matrix():
     # Check that the NMAD is low
     assert spatial_tools.nmad(diff) < 5
     print(np.nanmedian(diff), spatial_tools.nmad(diff))
+
+
+def test_warp_dem():
+
+    source_coords = np.array(
+            [
+                [1000, 1000, 200],
+                [1200, 1200, 200],
+                [1300, 1300, 200],
+                [1250, 1450, 200]
+            ]
+    )
+
+    dest_coords = source_coords.copy()
+    dest_coords[0, 0] += 1
+    dest_coords[1, 0] += 2
+    dest_coords[2, 0] += 3
+    dest_coords[3, 0] += 2.5
+
+
+    transform = rio.transform.from_origin(1000, 1500, 1, 1)
+
+    shape = (500, 500)
+    corr_size = int(shape[1] / 30)
+    dem = cv2.resize(
+        cv2.GaussianBlur(
+            np.repeat(np.repeat(
+                np.random.randint(0, 255, (shape[0]//corr_size,
+                                           shape[1]//corr_size), dtype='uint8'),
+                corr_size, axis=0), corr_size, axis=1),
+            ksize=(2*corr_size + 1, 2*corr_size + 1),
+            sigmaX=corr_size) / 255,
+        dsize=(shape[1], shape[0])
+    ) * 200
+
+    
+    transformed_dem = coreg.warp_dem(
+            dem=dem,
+            transform=transform,
+            source_coords=source_coords,
+            destination_coords=dest_coords,
+            resampling_order=1
+    )
+
+    return
+
+    import matplotlib.pyplot as plt
+
+    plt.imshow(dem)
+    plt.show()
+
+    assert source_coords.shape == (2, 3)
+
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -73,6 +73,37 @@ class TestCoregClass:
             if "non-finite values" not in str(exception):
                 raise exception
 
+    @pytest.mark.parametrize("coreg_class", [coreg.BiasCorr, coreg.ICP, coreg.NuthKaab])
+    def test_copy(self, coreg_class: coreg.Coreg):
+        """Test that copying work expectedly (that no attributes still share references)."""
+        warnings.simplefilter("error")
+
+        # Create a coreg instance and copy it.
+        corr = coreg_class()
+        corr_copy = corr.copy()
+
+        # Assign some attributes and metadata after copying
+        corr.foo = "bar"
+        corr._meta["hello"] = "there"
+        # Make sure these don't appear in the copy
+        assert corr_copy._meta != corr._meta
+        assert not hasattr(corr_copy, "foo")
+
+        # Create a pipeline, add some metadata, and copy it
+        pipeline = coreg_class() + coreg_class()
+        pipeline.pipeline[0]._meta["shouldexist"] = True
+
+        pipeline_copy = pipeline.copy()
+
+        # Add some more metadata after copying (this should not be transferred)
+        pipeline._meta["hello"] = "there"
+        pipeline_copy.pipeline[0]._meta["foo"] = "bar"
+
+        assert pipeline._meta != pipeline_copy._meta
+        assert pipeline.pipeline[0]._meta != pipeline_copy.pipeline[0]._meta
+        assert pipeline_copy.pipeline[0]._meta["shouldexist"]
+
+
     def test_bias(self):
         warnings.simplefilter("error")
 
@@ -151,8 +182,6 @@ class TestCoregClass:
         # Create random noise and see if the standard deviation is equal (it should)
         dem3 = dem1 + np.random.random(size=dem1.size).reshape(dem1.shape)
         assert abs(biascorr.error(dem1, dem3, transform=affine, error_type="std") - np.std(dem3)) < 1e-6
-
-
 
 
     def test_nuth_kaab(self):
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -418,10 +418,12 @@ class TestCoregClass:
         # Validate that all values are different
         assert np.unique(z_diff).size == z_diff.size
 
-
         # Validate that the PiecewiseCoreg doesn't accept uninstantiated Coreg classes
         with pytest.raises(ValueError, match="instantiated Coreg subclass"):
             coreg.PiecewiseCoreg(coreg=coreg.BiasCorr, subdivision=1)  # type: ignore
+
+
+        transformed_dem = piecewise.apply(self.tba.data, self.tba.transform)
 
 def test_apply_matrix():
     warnings.simplefilter("error")
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -248,7 +248,7 @@ class TestCoregClass:
         warnings.simplefilter("error")
 
         # Create a pipeline from two coreg methods.
-        pipeline = coreg.CoregPipeline([coreg.BiasCorr(), coreg.ICP(max_iterations=3)])
+        pipeline = coreg.CoregPipeline([coreg.BiasCorr(), coreg.NuthKaab()])
         pipeline.fit(**self.fit_params)
 
         aligned_dem = pipeline.apply(self.tba.data, self.ref.transform)
@@ -315,151 +315,30 @@ class TestCoregClass:
         # Check that the estimated biases are similar
         assert abs(bias_sub._meta["bias"] - bias_full._meta["bias"]) < 0.1
 
-        # Test ICP with subsampling
-        icp_full = coreg.ICP(max_iterations=20)
-        icp_sub = coreg.ICP(max_iterations=20)
+        # Test NuthKaab with subsampling
+        nuthkaab_full = coreg.NuthKaab()
+        nuthkaab_sub = coreg.NuthKaab()
 
         # Measure the start and stop time to get the duration
         start_time = time.time()
-        icp_full.fit(**self.fit_params)
+        nuthkaab_full.fit(**self.fit_params)
         icp_full_duration = time.time() - start_time
 
         # Do the same with 50% subsampling
         start_time = time.time()
-        icp_sub.fit(**self.fit_params, subsample=0.5)
+        nuthkaab_sub.fit(**self.fit_params, subsample=0.5)
         icp_sub_duration = time.time() - start_time
 
         # Make sure that the subsampling increased performance
         # Temporarily add a fallback assertion that if it's slower, it shouldn't be much slower (2021-05-17).
         # This doesn't work with GitHub's CI, but it works locally. I'm disabling this for now (2021-05-20).
-        #assert icp_full_duration > icp_sub_duration or (abs(icp_full_duration - icp_sub_duration) < 1)
+        assert icp_full_duration > icp_sub_duration or (abs(icp_full_duration - icp_sub_duration) < 1)
 
-        # Calculate the difference in the full vs. subsampled ICP matrices
-        matrix_diff = np.abs(icp_full.to_matrix() - icp_sub.to_matrix())
+        # Calculate the difference in the full vs. subsampled matrices
+        matrix_diff = np.abs(nuthkaab_full.to_matrix() - nuthkaab_sub.to_matrix())
         # Check that the x/y/z differences do not exceed 30cm
         assert np.count_nonzero(matrix_diff > 0.3) == 0
 
-    def test_apply_matrix(self):
-        warnings.simplefilter("error")
-        # This should maybe be its own function, but would just repeat the data loading procedure..
-
-        # Test only bias (it should just apply the bias and not make anything else)
-        bias = 5
-        matrix = np.diag(np.ones(4, float))
-        matrix[2, 3] = bias
-        transformed_dem = coreg.apply_matrix(self.ref.data.squeeze(), self.ref.transform, matrix)
-        reverted_dem = transformed_dem - bias
-
-        # Check that the revered DEM has the exact same values as the initial one
-        # (resampling is not an exact science, so this will only apply for bias corrections)
-        assert np.nanmedian(reverted_dem) == np.nanmedian(np.asarray(self.ref.data))
-
-        # Synthesize a shifted and vertically offset DEM
-        pixel_shift = 11
-        bias = 5
-        shifted_dem = self.ref.data.squeeze().copy()
-        shifted_dem[:, pixel_shift:] = shifted_dem[:, :-pixel_shift]
-        shifted_dem[:, :pixel_shift] = np.nan
-        shifted_dem += bias
-
-        matrix = np.diag(np.ones(4, dtype=float))
-        matrix[0, 3] = pixel_shift * self.tba.res[0]
-        matrix[2, 3] = -bias
-
-        transformed_dem = coreg.apply_matrix(shifted_dem.data.squeeze(),
-                                             self.ref.transform, matrix, resampling="bilinear")
-
-        # Dilate the mask a bit to ensure that edge pixels are removed.
-        transformed_dem_dilated = coreg.apply_matrix(
-            shifted_dem.data.squeeze(),
-            self.ref.transform, matrix, resampling="bilinear", dilate_mask=True)
-        # Validate that some pixels were removed.
-        assert np.count_nonzero(np.isfinite(transformed_dem)) > np.count_nonzero(np.isfinite(transformed_dem_dilated))
-
-        diff = np.asarray(self.ref.data.squeeze() - transformed_dem)
-
-        # Check that the median is very close to zero
-        assert np.abs(np.nanmedian(diff)) < 0.01
-        # Check that the NMAD is low
-        assert spatial_tools.nmad(diff) < 0.01
-
-        def rotation_matrix(rotation=30):
-            rotation = np.deg2rad(rotation)
-            matrix = np.array([
-                [1, 0, 0, 0],
-                [0, np.cos(rotation), -np.sin(rotation), 0],
-                [0, np.sin(rotation), np.cos(rotation), 0],
-                [0, 0, 0, 1]
-            ])
-            return matrix
-
-        rotation = 4
-        centroid = [np.mean([self.ref.bounds.left, self.ref.bounds.right]), np.mean(
-            [self.ref.bounds.top, self.ref.bounds.bottom]), self.ref.data.mean()]
-        rotated_dem = coreg.apply_matrix(
-            self.ref.data.squeeze(),
-            self.ref.transform,
-            rotation_matrix(rotation),
-            centroid=centroid
-        )
-        # Make sure that the rotated DEM is way off, but is centered around the same approximate point.
-        assert np.abs(np.nanmedian(rotated_dem - self.ref.data.data)) < 1
-        assert spatial_tools.nmad(rotated_dem - self.ref.data.data) > 500
-
-        # Apply a rotation in the opposite direction
-        unrotated_dem = coreg.apply_matrix(
-            rotated_dem,
-            self.ref.transform,
-            rotation_matrix(-rotation * 0.99),
-            centroid=centroid
-        ) + 4.0  # TODO: Check why the 0.99 rotation and +4 biases were introduced.
-
-        diff = np.asarray(self.ref.data.squeeze() - unrotated_dem)
-
-        if False:
-            import matplotlib.pyplot as plt
-
-            vmin = 0
-            vmax = 1500
-            extent = (self.ref.bounds.left, self.ref.bounds.right, self.ref.bounds.bottom, self.ref.bounds.top)
-            plot_params = dict(
-                extent=extent,
-                vmin=vmin,
-                vmax=vmax
-            )
-            plt.figure(figsize=(22, 4), dpi=100)
-            plt.subplot(151)
-            plt.title("Original")
-            plt.imshow(self.ref.data.squeeze(), **plot_params)
-            plt.xlim(*extent[:2])
-            plt.ylim(*extent[2:])
-            plt.subplot(152)
-            plt.title(f"Rotated {rotation} degrees")
-            plt.imshow(rotated_dem, **plot_params)
-            plt.xlim(*extent[:2])
-            plt.ylim(*extent[2:])
-            plt.subplot(153)
-            plt.title(f"De-rotated {-rotation} degrees")
-            plt.imshow(unrotated_dem, **plot_params)
-            plt.xlim(*extent[:2])
-            plt.ylim(*extent[2:])
-            plt.subplot(154)
-            plt.title("Original vs. de-rotated")
-            plt.imshow(diff, extent=extent, vmin=-10, vmax=10, cmap="coolwarm_r")
-            plt.colorbar()
-            plt.xlim(*extent[:2])
-            plt.ylim(*extent[2:])
-            plt.subplot(155)
-            plt.title("Original vs. de-rotated")
-            plt.hist(diff[np.isfinite(diff)], bins=np.linspace(-10, 10, 100))
-            plt.tight_layout(w_pad=0.05)
-            plt.show()
-
-        # Check that the median is very close to zero
-        assert np.abs(np.nanmedian(diff)) < 0.5
-        # Check that the NMAD is low
-        assert spatial_tools.nmad(diff) < 5
-        print(np.nanmedian(diff), spatial_tools.nmad(diff))
 
     def test_z_scale_corr(self):
         warnings.simplefilter("error")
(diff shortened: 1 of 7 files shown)


erikmannerfelt:
For some reason, I'm not allowed to upload a .py file! Here's a .txt version of the python script for the figure.

[nuth_kaab_64.txt](https://github.com/GlacioHack/xdem/files/6639068/nuth_kaab_64.txt)

xdem/spatial_tools.py
```diff
@@ -49,7 +51,7 @@ def get_array_and_mask(array: Union[np.ndarray, np.ma.masked_array], check_shape
         array = array.astype('float32')
 
     # Convert into a regular ndarray and convert invalid values to NaN
-    array_data = np.asarray(array).squeeze()
+    array_data = np.array(array).squeeze()
```
adehecq, line 54: Would that not make a copy of the array?
We should make sure that we do not copy the data when it's not needed.

xdem/spatial_tools.py
```diff
@@ -396,3 +398,102 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+    """
+    close_cube = int(np.sqrt(size))
+
+    if close_cube ** 2 == size:
+        return (close_cube, close_cube)
+
+    height = close_cube
+    width = close_cube
+    # Iteratively increase height, then width, until the size is larger or equal to the input size
+    for i in range(int(1e8)):
```
adehecq, line 434: I don't think this number of iterations is ever used...
First, you find N such that N^2 < size < (N+1)^2 (here N is your close_cube).
Then your final result is always either N^2, N*(N+1) or (N+1)^2, depending on which one is larger than size first.

erikmannerfelt: Good point! I avoid `while True` because I read somewhere that NASA avoids it haha (to avoid endless loops). That loop should never finish; it's just a precaution for potential flaws in my programming.

xdem/spatial_tools.py
```diff
@@ -396,3 +398,102 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+
+    if len(shape) != 2:
+        raise ValueError(f"Expected a 2D shape, got {len(shape)}D shape: {shape}")
+
+    # Generate a small grid of indices, with the same unique count as 'count'
+    rect = _get_closest_rectangle(count)
+    small_indices = np.pad(np.arange(count), np.prod(rect) - count, mode="edge")[:np.prod(rect)].reshape(rect)
+
+    # Upscale the grid to fit the output shape using nearest neighbour scaling.
+    indices = skimage.transform.resize(small_indices, shape, order=0, preserve_range=True).astype(int)
```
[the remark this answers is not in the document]
erikmannerfelt: That works for easily divisible sizes, yes, but not for e.g. odd sizes. I wanted to make something that works in all cases

xdem/spatial_tools.py
erikmannerfelt: I can undo that, but I did it because values were modified inplace:
```python
ref_dem = ...  # a numpy array
tba_dem = ... # a numpy array
mask = ... # a numpy array

coreg.fit(ref_dem, tba_dem, mask=mask, transform=None) 

# Now ref_dem and tba_dem have been modified by the mask, since they are modified inplace.


```

xdem/coreg.py
```diff
@@ -1442,3 +1482,457 @@ def _to_matrix_func(self) -> np.ndarray:
...
+
+        self._meta["coreg_meta"] = []
+
+
+
+    def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: rio.transform.Affine | None,
+                  weights: np.ndarray | None, verbose: bool = False):
+        """Fit the coreg approach for each subdivision."""
+
+        groups = self.subdivide_array(tba_dem.shape)
```
adehecq, line 1527: If I'm correct, this will duplicate the size of the DEM right?
Have you considered using views and strides, instead of copying the data? Here's an implementation that could be useful maybe? https://github.com/ilastik/lazyflow/blob/master/lazyflow/utility/blockwise_view.py

xdem/spatial_tools.py
erikmannerfelt: I don't think it's good practice to modify the inputs when they shouldn't be modified. Copying is the easiest way out. In the long term, we could maybe find out a way to ensure immutability without copying.

xdem/coreg.py
erikmannerfelt: You don't know how long I searched for exactly that in a simple way!!

Either way, I like this approach as the block indices can be exported and used for further analysis. This suggested approach is definitely more memory-efficient, but it won't be easy to use further, such as in visualization, debugging, and validating the outcome.

xdem/spatial_tools.py
adehecq: Ok yes, we should certainly not modify the input DEMs in that case. But get_array_and_mask is now used many times in the code, and in some cases, it may not be wished to copy the data...
So maybe a copy after the call to get_array_and_mask should be favored? Or an option 'copy' in the function call?

xdem/spatial_tools.py
adehecq: Ok, but my point was also that the loop is probably not needed. You only need to test N, N*(N+1) and (N+1)^2.

adehecq: So you loop only twice instead of 1e8!

xdem/coreg.py
adehecq: Haha, I feel you. I've done this kind of block dividing before and I've searched for at least 5 min for the right keywords to finally find the word I was looking for ('stride'), which I found in this nice example.

xdem/coreg.py
```diff
@@ -1442,3 +1482,457 @@ def _to_matrix_func(self) -> np.ndarray:
...
+
+        indices = np.unique(groups)
+
+        progress_bar = tqdm(total=indices.size, desc="Coregistering chunks", disable=(not verbose))
+
+        def coregister(i: int) -> dict[str, Any] | BaseException:
+            """Coregister a chunk in a thread-safe way."""
+            inlier_mask = groups == i
+
+            # Find the corresponding slice of the inlier_mask to subset the data
```
adehecq, line 1537: Looks somewhat similar (although more succinct?) than the function I added in spatial_tools: https://github.com/GlacioHack/xdem/blob/c60330abf5847be3654411374ae5bdf2a3d151d2/xdem/spatial_tools.py#L58

xdem/coreg.py
```diff
@@ -1442,3 +1482,457 @@ def _to_matrix_func(self) -> np.ndarray:
...
+            * {x,y,z}_off: The calculated offset in georeferenced units.
+            * inlier_count: The number of pixels that were inliers in the chunk.
+            * nmad: The NMAD after coregistration.
+            * median: The bias after coregistration.
+
+        :raises ValueError: If no coregistration results exist yet.
+
+        :returns: A dataframe of statistics for each chunk.
+        """
+        points = self.to_points()
```
adehecq, line 1690: How long does this function take? You're calling it to fit the function, to get the statistics etc. I guess some calculations are done several times.
Is there a way to avoid repeat some of the calculation?

xdem/coreg.py
```diff
@@ -1442,3 +1482,457 @@ def _to_matrix_func(self) -> np.ndarray:
...
+        >>> transform = rio.transform.from_origin(0, 5, 1, 1)
+        >>> source_coords = np.array([
+        ...     [0, 0, 0],
+        ...     [5, 0, 0],
+        ...     [5, 5, 0],
+        ...     [0, 5, 0]
+        ... ]).astype("float32")
+        >>> destination_coords = source_coords.copy()
+        >>> destination_coords[3, 2] = 4
+        >>> warped_dem = warp_dem(
```
adehecq, line 1819: I do not really understand this example...

tests/test_coreg.py
```diff
@@ -519,3 +427,324 @@ def test_z_scale_corr(self):
...
+        with pytest.raises(AssertionError, match="No coreg results exist.*"):
+            piecewise.to_points()
+    
+        piecewise.fit(**self.fit_params)
+        points = piecewise.to_points()
+
+        # Validate that the number of points is equal to the amount of subdivisions.
+        assert points.shape[0] == subdivision
+
+        # Validate that the points to not represent only the same location.
```
adehecq, line 460: typo?

tests/test_coreg.py
```diff
@@ -519,3 +427,324 @@ def test_z_scale_corr(self):
...
+        plt.show()
+
+    # Check that the median is very close to zero
+    assert np.abs(np.nanmedian(diff)) < 0.5
+    # Check that the NMAD is low
+    assert spatial_tools.nmad(diff) < 5
+    print(np.nanmedian(diff), spatial_tools.nmad(diff))
+
+
+def generate_random_field(shape: tuple[int, int], corr_size: int) -> np.ndarray:
```
adehecq, line 615: Should it be stored somewhere else, outside of the tests? Such a function could be useful for tutorials for example maybe?

tests/test_coreg.py
```diff
@@ -519,3 +427,324 @@ def test_z_scale_corr(self):
...
+    dest_coords[1, 2] = elev_shift
+    warped_dem = coreg.warp_dem(
+        dem=small_dem,
+        transform=small_transform,
+        source_coords=source_coords,
+        destination_coords=dest_coords,
+        resampling="linear",
+    )
+
+    # The warped DEM should have
```
adehecq, line 676: Missing text?

adehecq:
This looks very nice and interesting! But that's a lot to ingest and it's difficult to assess the efficiency of the implementation, so it will require some testing on larger data set.
Maybe I should merge PR #141 before this is merge, as this may cause conflicts? I was hoping to use the function in the coreg.py, but it can be added as an issue instead.

xdem/coreg.py
erikmannerfelt: Ah yes, there it is! I looked in geoutils and didn't find it so I gave up. But that's exactly what I needed.

xdem/coreg.py
erikmannerfelt: It takes four seconds on my laptop. Without multiple threads, it takes about a minute.

xdem/spatial_tools.py
erikmannerfelt: Could you please elaborate on what you mean, e.g. though a minimal code example? I'm unfortunately not sure of what you mean!

xdem/spatial_tools.py
erikmannerfelt: Copying is made if a masked_array is supplied, so it would be more consistent to copy by default... I see your point though that this is a potentially un-performant approach. We should have a benchmark system to validate the effects of this! In the tests, there's no apparent effect, but we do these operations on relatively small rasters..

erikmannerfelt: For now, I can add a `copy` argument, but say in parentheses that masked_arrays are always copied.

erikmannerfelt: But wait a minute, the `get_array_and_mask()` is currently very inconsistent in whether it will modify the input or not:

1. View if given a float array
2. Copy if given an integer array
3. View if given a float masked_array with `False` as its mask
4. Copy if given an integer array
4. Copy if given a float masked_array with a mask

Is this really what we want?? I get that it's good to return views as there is less memory duplication involved, but in this case, it really seems difficult to know if it's a view or a copy, no? What do you think, @rhugonnet?

xdem/coreg.py
erikmannerfelt: Removed it in ffbf761 as exemplifying it intuitively on that scale doesn't really make sense.

tests/test_coreg.py
erikmannerfelt: Fixed in 37dddbc

tests/test_coreg.py
erikmannerfelt: Fixed in 37dddbc

tests/test_coreg.py
erikmannerfelt: Added `xdem/misc.py` for all kinds of small functions like that, and moved the semi-random function there in 366b722

erikmannerfelt:
"Piece-wise" should probably be changed to "Block-wise", as per @adehecq's suggestion!

rhugonnet:
> "Piece-wise" should probably be changed to "Block-wise", as per @adehecq's suggestion!

+1 with @adehecq's suggestion
Piece-wise is usually employed when using for instance 10 different DEMs and finding the best common vertical/horizontal shift to align them all simultaneously. No need for a single "reference", in that case.
Our Icelandic colleagues, like Tomas Johanneson and Joaquin Belart, have a great workflow for this that could maybe be integrated to xdem at some point! :)

rhugonnet:
I will review the PR later today!

erikmannerfelt:
Thanks, @rhugonnet! In open3d, that is called "multiway registration" (I'm not preferring the name; just pointing toward another implementation!): http://www.open3d.org/docs/release/tutorial/pipelines/multiway_registration.html

xdem/coreg.py
rhugonnet: What about using dask blocks that were made for numpy and directly support chunks?
https://docs.dask.org/en/latest/array-stack.html#block
Dask will also support overlap of blocks when it is needed for some computations:
https://docs.dask.org/en/latest/array-overlap.html#map-a-function-across-blocks

xdem/spatial_tools.py
rhugonnet: Nice spotting this, not good indeed. 
The copy argument seems like a good idea! Otherwise we could simply return views systematically, as those are more memory-efficient, and we copy the objects ourselves when it is needed in a certain processing chain?

xdem/spatial_tools.py
```diff
@@ -396,3 +398,102 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+        if (height * width) >= size:
+            return (height, width)
+
+    else:
+        raise ValueError("Close rectangle could not be found after 1e8 iterations.")
+
+
+
+
+def subdivide_array(shape: tuple[int, ...], count: int) -> np.ndarray:
```
rhugonnet, line 449: FYI: 
Commonly used block splitter/stitcher set of functions for numpy arrays to use with multiprocessing: https://github.com/iamdonovan/pyddem/blob/master/pyddem/fit_tools.py#L1366
Same for overlapping blocks (patchify/unpatchify):
https://github.com/iamdonovan/pyddem/blob/master/pyddem/fit_tools.py#L1389
But we should not reinvent the wheel on this, and probably switch to functions developed efficiently in Dask at some later point! :)

xdem/coreg.py
```diff
@@ -1442,3 +1482,426 @@ def _to_matrix_func(self) -> np.ndarray:
...
+                    points[:, dim, 1] - points[:, dim, 0],
+                    function="linear",
+                )
+
+            new_coords[:, dim] += model(coords[:, 0], coords[:, 1])
+
+        return new_coords
+
+
+def warp_dem(
```
rhugonnet, line 1781: I'm not sure I grasp the addition of `warp_dem` in relation to `geoutils.reproject()` ?

rhugonnet (commented):
Great implementation! This is something I thought about many times, but implementing was a pain as it was always too specific. Within the frame of the Coreg pipelines and classes, it simply gives birth to so many more interesting possibilities :partying_face: 
Amazing :+1: 
I'm only worried about two things:
- The blocks might need overlapping to avoid introducing edge effects for many type of coregistration methods, such as horizontal shifts.
- There is existing, efficient tools made for parallel processing/lazy computing which have been designed to improve both the speed and the memory usage of such methods, specifically for numpy arrays in **Dask**. We'll need to think about switching to those soon, otherwise our methods will become obsolete quite rapidly!

Regarding the figure example you present: what is the exact time of acquisition for those DEMs? did you consider that the residuals of the Nuth and Kääb could be due to different amounts of snow-height in the surrounding of the glaciers?
Fig. S5 here: https://static-content.springer.com/esm/art%3A10.1038%2Fs41586-021-03436-z/MediaObjects/41586_2021_3436_MOESM1_ESM.pdf

xdem/coreg.py
erikmannerfelt: Thanks a lot for the feedback, @rhugonnet!

>     * The blocks might need overlapping to avoid introducing edge effects for many type of coregistration methods, such as horizontal shifts.
I'm not sure this is needed, as the blocks are not transformed separately, but a transformation is made using a thin-plate spline approximation. Here's a figure of it that I found on the internets:



Basically, the transformation matrix is described as an X/Y/Z shift near-center of every block, so for e.g. a subdivision of 64, we have 64 points telling an algorithm "these points here should rather be X px to the right, Y px north, etc.). Then I use `xdem.coreg.warp_dem` to do a "puppet warp" (as it's called in Photoshop) on the DEM, along with a vertical correction as well (which you can see in the upper left plot in the figure). This means that there will be no seams between the blocks, as the warping is performed on the whole raster at the same time. 

Does that make sense?

>     * There is existing, efficient tools made for parallel processing/lazy computing which have been designed to improve both the speed and the memory usage of such methods, specifically for numpy arrays in **Dask**. We'll need to think about switching to those soon, otherwise our methods will become obsolete quite rapidly!
Interesting! Do we already have Dask as a dependency? `concurrent.futures` works pretty well together with some copying of only the chunks' interiors (I never copy the whole raster in a thread), so I would say that this is pretty solid. But if there's a much better way, we should of course embrace it. I just wonder, is Dask a heavy dependency and should it be required? What do you think, @adehecq?

> Regarding the figure example you present: what is the exact time of acquisition for those DEMs? did you consider that the residuals of the Nuth and Kääb could be due to different amounts of snow-height in the surrounding of the glaciers?
> Fig. S5 here: https://static-content.springer.com/esm/art%3A10.1038%2Fs41586-021-03436-z/MediaObjects/41586_2021_3436_MOESM1_ESM.pdf

I have used the images myself for my MSc thesis, and the snow cover is almost identical. There might be a tiny bit more in 2009, but the potential difference would be on decimeter level. Here, we see differences of +2 m to -4 m, which is way too much!

erikmannerfelt:
Thanks a lot for the feedback, @rhugonnet!

>     * The blocks might need overlapping to avoid introducing edge effects for many type of coregistration methods, such as horizontal shifts.
I'm not sure this is needed, as the blocks are not transformed separately, but a transformation is made using a thin-plate spline approximation. Here's a figure of it that I found on the internets:



Basically, the transformation matrix is described as an X/Y/Z shift near-center of every block, so for e.g. a subdivision of 64, we have 64 points telling an algorithm "these points here should rather be X px to the right, Y px north, etc.). Then I use `xdem.coreg.warp_dem` to do a "puppet warp" (as it's called in Photoshop) on the DEM, along with a vertical correction as well (which you can see in the upper left plot in the figure). This means that there will be no seams between the blocks, as the warping is performed on the whole raster at the same time. 

Does that make sense?

>     * There is existing, efficient tools made for parallel processing/lazy computing which have been designed to improve both the speed and the memory usage of such methods, specifically for numpy arrays in **Dask**. We'll need to think about switching to those soon, otherwise our methods will become obsolete quite rapidly!
Interesting! Do we already have Dask as a dependency? `concurrent.futures` works pretty well together with some copying of only the chunks' interiors (I never copy the whole raster in a thread), so I would say that this is pretty solid. But if there's a much better way, we should of course embrace it. I just wonder, is Dask a heavy dependency and should it be required? What do you think, @adehecq?

> Regarding the figure example you present: what is the exact time of acquisition for those DEMs? did you consider that the residuals of the Nuth and Kääb could be due to different amounts of snow-height in the surrounding of the glaciers?
> Fig. S5 here: https://static-content.springer.com/esm/art%3A10.1038%2Fs41586-021-03436-z/MediaObjects/41586_2021_3436_MOESM1_ESM.pdf

I have used the images myself for my MSc thesis, and the snow cover is almost identical. There might be a tiny bit more in 2009, but the potential difference would be on decimeter level. Here, we see differences of +2 m to -4 m, which is way too much!

xdem/spatial_tools.py
erikmannerfelt: They look very simple, which would be nice! I must however admit that I have no idea what's going on in them since there are no comments. Might you explain what the inputs/outputs look like?

tests/test_coreg.py
```diff
@@ -519,3 +427,302 @@ def test_z_scale_corr(self):
...
+    ref, tba, outlines = load_examples()  # Load example reference, to-be-aligned and mask.
+
+    # Test only bias (it should just apply the bias and not make anything else)
+    bias = 5
+    matrix = np.diag(np.ones(4, float))
+    matrix[2, 3] = bias
+    transformed_dem = coreg.apply_matrix(ref.data.squeeze(), ref.transform, matrix)
+    reverted_dem = transformed_dem - bias
+
+    # Check that the revered DEM has the exact same values as the initial one
```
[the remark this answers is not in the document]
erikmannerfelt: No, I truly revere the DEMs. They are what keep me employed!

(just kidding! Thanks for seeing it!)

erikmannerfelt:
>Nice spotting this, not good indeed.
The copy argument seems like a good idea! Otherwise we could simply return views systematically, as those are more memory-efficient, and we copy the objects ourselves when it is needed in a certain processing chain?

What about if the `get_array_and_mask()` has a `copy` argument which is `True` by default? If set to `False`, it will try to return a view, but triggers a warning if that was not possible. How does that sound, @adehecq  and @rhugonnet?

adehecq:
> Thanks a lot for the feedback, @rhugonnet!
> 
> > ```
> > * The blocks might need overlapping to avoid introducing edge effects for many type of coregistration methods, such as horizontal shifts.
> > ```
> 
> I'm not sure this is needed, as the blocks are not transformed separately, but a transformation is made using a thin-plate spline approximation. Here's a figure of it that I found on the internets:
> 
> 
> 
> Basically, the transformation matrix is described as an X/Y/Z shift near-center of every block, so for e.g. a subdivision of 64, we have 64 points telling an algorithm "these points here should rather be X px to the right, Y px north, etc.). Then I use `xdem.coreg.warp_dem` to do a "puppet warp" (as it's called in Photoshop) on the DEM, along with a vertical correction as well (which you can see in the upper left plot in the figure). This means that there will be no seams between the blocks, as the warping is performed on the whole raster at the same time.
> 
> Does that make sense?
> 

I agree with @erikmannerfelt, the current implementation should not have edge effects, it will smoothly interpolate the shift.

> > ```
> > * There is existing, efficient tools made for parallel processing/lazy computing which have been designed to improve both the speed and the memory usage of such methods, specifically for numpy arrays in **Dask**. We'll need to think about switching to those soon, otherwise our methods will become obsolete quite rapidly!
> > ```
> 
> Interesting! Do we already have Dask as a dependency? `concurrent.futures` works pretty well together with some copying of only the chunks' interiors (I never copy the whole raster in a thread), so I would say that this is pretty solid. But if there's a much better way, we should of course embrace it. I just wonder, is Dask a heavy dependency and should it be required? What do you think, @adehecq?

I think Dask is great and we should use it more! I'm just concerned that if we have two inputs in different grids/projections, we might have to do the reprojection beforehand, save to disk and then load the data as a Dask array. Unless xarray has implemented a warp function??
> 
> > Regarding the figure example you present: what is the exact time of acquisition for those DEMs? did you consider that the residuals of the Nuth and Kääb could be due to different amounts of snow-height in the surrounding of the glaciers?
> > Fig. S5 here: https://static-content.springer.com/esm/art%3A10.1038%2Fs41586-021-03436-z/MediaObjects/41586_2021_3436_MOESM1_ESM.pdf
> 
> I have used the images myself for my MSc thesis, and the snow cover is almost identical. There might be a tiny bit more in 2009, but the potential difference would be on decimeter level. Here, we see differences of +2 m to -4 m, which is way too much!

adehecq:
> > Nice spotting this, not good indeed.
> > The copy argument seems like a good idea! Otherwise we could simply return views systematically, as those are more memory-efficient, and we copy the objects ourselves when it is needed in a certain processing chain?
> 
> What about if the `get_array_and_mask()` has a `copy` argument which is `True` by default? If set to `False`, it will try to return a view, but triggers a warning if that was not possible. How does that sound, @adehecq and @rhugonnet?

Yes, the issue with returning views by default is that it may not be possible sometimes, for example for Byte types that cannot use NaNs...
Ideally all functions would work with masked arrays, so we don't need to use NaNs..
But in the mean time, I guess it's better to be consistent and return a copy as suggested by @erikmannerfelt. But if doing so, we need to go back to all calls to this function to make sure that copy=False whenever possible.

xdem/spatial_tools.py
adehecq: See my comment below. It's not possible to return views for arrays that do not accept NaNs...

erikmannerfelt:
> 
> Yes, the issue with returning views by default is that it may not be possible sometimes, for example for Byte types that cannot use NaNs...
> Ideally all functions would work with masked arrays, so we don't need to use NaNs..
> But in the mean time, I guess it's better to be consistent and return a copy as suggested by @erikmannerfelt. But if doing so, we need to go back to all calls to this function to make sure that copy=False whenever possible.

I'll add that then! I may be wrong but I don't think there is any place where one can consistently get views. The problem is that, as I mentioned above, views can only be returned for float arrays, masked_arrays without a mask,  int arrays without nodatas, and int masked_arrays without a mask. In most cases, we have to presume that a mask may exist, so returning views will seldom be possible.

xdem/spatial_tools.py
adehecq: Let me state it again differently. If size is the total number of pixels in your "close to cube", then we can always find an integer, let's call it N for simplicity (it's called close_cube in your code) that verifies:
_N^2 <= size < (N+1)^2_.
You find N as you do with `int(np.sqrt(size))`. So the possibilities for your "close to cube" are a rectangle of dimensions _N x N_, _N x (N+1)_ or _(N+1) x (N+1)_. Just pick the one among these three that is larger than size (by construction, the 3rd one will always be large enough) and that's it. You do not need 1e8 iterations to find that out, just math! ;-)

xdem/spatial_tools.py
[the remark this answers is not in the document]
adehecq: Use padding for odd values? 
Maybe skimage.transform.resize is very fast when using order=0 I don't know. You could test the performance of both in an even case. But possibly you're using a very expensive Swiss army knife when you only need a spoon ;-)

xdem/coreg.py
adehecq: A minute is quite long when you need to do it multiple times. And that's only for a small DEM...
It's always the drawback when splitting one code into smaller chunks, one might end up needing the same info twice and re-doing the same calculation...
So maybe the stats should be calculated by default and stored, so one does not need to redo this long calculation when one wants to output the stats?

erikmannerfelt:
Phew, adding the `copy` argument and testing for (almost) every case for it was a bit of work!! Now I have done that, with pytest parameterization to test 80 different cases, whether they can be returned as views or not, whether they are copied properly, and whether their shapes are acceptable for the `check_shapes` argument.

xdem/spatial_tools.py
erikmannerfelt: Added `copy` argument in 4f85f6b

xdem/coreg.py
erikmannerfelt: Oh wait, sorry I thought you meant the `fit` method!! I haven't benchmarked this part but I think it's more or less instantaneous. I can check after lunch.

xdem/spatial_tools.py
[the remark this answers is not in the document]
erikmannerfelt: I tried over and over to make np.repeat work... Could you please submit a PR to my branch or later when merged for your implementation? I'm afraid my matrix skills are not high enough for this!

adehecq:
> Phew, adding the `copy` argument and testing for (almost) every case for it was a bit of work!! Now I have done that, with pytest parameterization to test 80 different cases, whether they can be returned as views or not, whether they are copied properly, and whether their shapes are acceptable for the `check_shapes` argument.

Haha, thanks for the thorough job on testing!

xdem/spatial_tools.py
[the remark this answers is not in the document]
adehecq: Yes of course, no problem! Maybe resize with order=0 is doing the same thing under the hood so it's super fast.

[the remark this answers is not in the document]
adehecq: So we can leave it for later.

xdem/spatial_tools.py
erikmannerfelt: Implemented in 58a021b. I think the loop never iterated more than at most two times, so it's much better to remove it.

adehecq: Exactly, this is the first sentence I wrote! ;-)

erikmannerfelt:
Assuming the tests pass, I would consider this feature ready to merge! There are some possible improvements regarding performance that were suggested by @rhugonnet, but functionality-wise, I consider this mature.

I'll add some potential improvement issues if that sounds good for you.

rhugonnet:
> Assuming the tests pass, I would consider this feature ready to merge! There are some possible improvements regarding performance that were suggested by @rhugonnet, but functionality-wise, I consider this mature.
> 
> I'll add some potential improvement issues if that sounds good for you.

:+1: Great work

rhugonnet (approved):

adehecq:
> Assuming the tests pass, I would consider this feature ready to merge! There are some possible improvements regarding performance that were suggested by @rhugonnet, but functionality-wise, I consider this mature.
> 
> I'll add some potential improvement issues if that sounds good for you.

Sounds good. Well donne!

rhugonnet (approved):

erikmannerfelt:
Thanks, @rhugonnet !

Issue #144, opened 2021-06-07
Add slope map function


erikmannerfelt:
Like our trusty hillshade function, it would be nice to easily be able to generate slope maps. In fact, this is already done within the hillshade function, but there is no way to extract it!

erikmannerfelt:
Added in #147

PR #147, opened 2021-06-15
Add slope, aspect, planform/profile/normal curvature, and others

erikmannerfelt:
Solves #144.

Tests are included to validate that they are close to what GDAL would provide.

# TODO
- [X] Add tests for nan fill_method for quadric coefficient estimation
- [x] Expose `fill_method` and edge handling arguments to `get_terrain_attribute()`
- [ ] Somehow compare curvature with another tool, e.g. richdem (at least once, so we don't have to have it as a dependency)
- [X] Add edge handling arguments. Currently, it only does nearest fill.
- [X] Benchmark this tool with others, e.g. GDAL and RichDEM (@rhugonnet's suggestion). EDIT: It is between 4 and 10 times slower... More transparent, but much slower. Is this worth keeping?
- [x] Compare this tool with Savitzky-Golay (@adehecq's suggestion)

Things to add in the future (will be implemented as new issues):
* Add ruggedness index
* Add rugosity
* Add other attributes such as max slope/curvature
* Add more fill methods like bilinear and cubic (@rhugonnet's suggestion)
* Depending on its performance, replace the backend with Savitzky-Golay, keep it as an alternative, or not at all.


erikmannerfelt:
I suggested to @adehecq that the functions `hillshade`, `slope`, `aspect` (and other terrain attributes that may come) maybe fit better in a separate module. What do you think about making a new `terrain.py` module or similar, @rhugonnet and @adehecq?

rhugonnet:
> I suggested to @adehecq that the functions `hillshade`, `slope`, `aspect` (and other terrain attributes that may come) maybe fit better in a separate module. What do you think about making a new `terrain.py` module or similar, @rhugonnet and @adehecq?

100% agree! could even have (optional) richDEM functions in there ;)

adehecq:
Thanks for adding this!
My thought was that we should have a function like "get_terrain_attribute" which would take the same arguments as your functions + a list containing the terrain attributes that are needed, like "slope", "aspect" and possibly later "curvature" etc.
This way, when one wants to get the slope AND aspect at the same time, one doesn't need to run the gradient two times, which is very inefficient.
But if you don't have time to fix this, the code could be left as it is and an issue raised to merge the two.

It's also fine to have an implementation that uses np.gradient, but I'm not that sure that 1) it is very fast compared to approaches based on convolution 2) it's reliable with noisy DEMs.
We could have an optional Gaussian smoothing step for example, or use the Savitsky-Golay algorithm to be more robust to noise.

adehecq:
And +1 for the terrain.py module.

erikmannerfelt:
I can start on the `get_terrain_attribute` function that would accept multiple products if you think `np.gradient` is still the way to go, @adehecq. Otherwise, I can try to adapt the Savitsky-Golay implementation you sent a while ago (and maybe accelerate with Numba!). I also have issues with `np.gradient` as it only takes into account the left/right/bottom/top pixels (not the diagonals, nor a larger kernel), and it's not possible to accelerate with Numba (I spent hours trying to rewrite it to work!!).

What's your preference?
1. Move functionality as is (plus get_terrain_attribute) into terrain.py
2. I rewrite the functions with Savitsky-Golay (plus Numba maybe) into terrain.py
3. Close the PR and leave you in charge of the functions.

I am equally fine with either!

rhugonnet:
As you want! 
I'm not sure I completely see the scope as those functionalities already exist robustly in GDAL, richDEM, etc. Is there a reason you think it's worth investing additional time through numpy functions? Does it run a lot faster?

erikmannerfelt:
The main motivation (apart from transparency; GDAL is a bit of a black box...) is to have functions that are not dependent on files written on disk. Even GDAL's python bindings still require disk filepaths.

However..

I just learned of `tempfile.SpooledTemporaryFile`, a fascinating functionality that allows for a temporary file-like object in memory (to a certain size, then it falls back to disk). Instead of reinventing the wheel with aspect/slope/hillshade, one could:
1. "Save" the DEM to a file with a fixed resolution and an arbitrary CRS to memory (or disk if it's too big).
2. Run GDALProcessing and save it to a new `tempfile.SpooledTemporaryFile`.
3. "Read" the product from memory (or disk if it was too big) into a numpy ndarray.

I see two drawbacks:

1. The hillshade/slope/aspect maps cannot be calculated at once, like they could with a pure numpy implementation, but I don't know if it is that important anyway.
2. We would not be closer to a curvature algorithm, and would have to create code for that separately, which would probably repeat what is now written (which could be replaced by a GDAL wrapper). Thus, if curvature is on the table, it may lead to more code rather than less by sticking with GDAL!


I would prefer leaving RichDEM as it is not seemingly maintained anymore, is not on conda-forge, and populates the stdout without a good way of disabling it!!! 

The golden approach in my view is to have an efficient cell + neighbours approach (probably convolution?) that could calculate slope, aspect, (hillshade from the two), curvature, etc., or all at the same time, preferably in a multithreaded environment. 

What are your thoughts on this in relation to the suggestions above, @adehecq and @rhugonnet?

adehecq:
Regarding your first message:

> What's your preference?
> 
>     1. Move functionality as is (plus get_terrain_attribute) into terrain.py
>     2. I rewrite the functions with Savitsky-Golay (plus Numba maybe) into terrain.py
>     3. Close the PR and leave you in charge of the functions.
> 
I would go with 1 for now. Savitsky-Golay would need to be included in GeoUtils most likely.

Regarding your second point

> I just learned of tempfile.SpooledTemporaryFile, a fascinating functionality that allows for a temporary file-like object in memory (to a certain size, then it falls back to disk). Instead of reinventing the wheel with aspect/slope/hillshade, one could:
> 
>     1. "Save" the DEM to a file with a fixed resolution and an arbitrary CRS to memory (or disk if it's too big).
>     2. Run GDALProcessing and save it to a new tempfile.SpooledTemporaryFile.
>     3. "Read" the product from memory (or disk if it was too big) into a numpy ndarray.

I think it's a bit convoluted, and I'm pretty confident we can do the same thing with GDAL directly using Memory files (but I'm not 100% sure). So I'm not keen on that option.
The issue is that it's not recommended to import gdal and rasterio together... So ideally, all of this could be in rasterio, since it's all part of GDAL already. Maybe we should raise an issue?

At the moment, I would also favor an in-house option, because as @erikmannerfelt said, richDEM is not well maintained and is a burden to get xdem on conda-forge etc.
For now, we have a simple slope/aspect calculation based on the gradient. Then we could either use Savitsky-Golay, or see how this is done in richDEM and re-use that (if the license allows it).

erikmannerfelt:
I have now made `terrain.py` and I look forward to hearing your impressions!

The most important function is `get_terrain_attribute()`, from which one or multiple attributes can be calculated:
```python
import xdem

dem = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_ref_dem"])

slope, aspect = xdem.terrain.get_terrain_attribute(dem=dem.data, attribute=["slope", "aspect"], resolution=dem.res)
```
There are also convenience wrappers for the attributes that call `get_terrain_attribute()` but look nicer:
```python

aspect = xdem.terrain.aspect(dem=dem.data)  # This particular map is resolution-invariant
hillshade = xdem.terrain.hillshade(dem=dem.data, resolution=dem.res, azimuth=315., altitude=45., z_factor=1.0)
```

@adehecq, I also got function overloading working with `get_terrain_attribute()`! I don't remember now where we had issues with it, but this seems to work very well.

Finally, I've added a documentation chapter for terrain attributes

erikmannerfelt:
Here's the terrain chapter:
https://xdem-erikm.readthedocs.io/en/slope/terrain.html

docs/source/code/terrain_plot_aspect.py
```diff
@@ -0,0 +1,18 @@
+import matplotlib.pyplot as plt
+import xdem
+
+xdem.examples.download_longyearbyen_examples()
+
+dem = xdem.DEM(xdem.examples.FILEPATHS["longyearbyen_ref_dem"])
+
+aspect = xdem.terrain.aspect(dem.data)
+
+plt.imshow(aspect.squeeze(), cmap="jet")
```
adehecq, line 10: Just looking at the documentation, I thought jet is not the best colormap for aspect. It would be preferable to use a cyclic colormap, but I couldn't find a good one in matplotlib's defaults. If there is none, maybe we could make our own??
Then, I thought Raster.show() could be overwritten for the DEM class to have options to directly plot the hillshade, slope or aspect (with default colormap)! :-)

erikmannerfelt: I tried "hsv" first but that was absolutely terrible. "twiligt" is pretty nice in my opinion! What do you think?

Also, good idea with the plot override! Could also arguably fit in GeoUtils somehow.

image

xdem/terrain.py
```diff
@@ -0,0 +1,232 @@
...
+    resolution: tuple[float, float] | float,
+    degrees: bool = True,
+    hillshade_altitude: float = 45.0,
+    hillshade_azimuth: float = 315.0,
+    hillshade_z_factor: float = 1.0,
+) -> np.ndarray:
+    ...
+
+
+@overload
```
rhugonnet, line 24: I looked at the documentation, but I'm still not sure I understand completely what `@overload` does here exactly! I see the types of `attribute` and output are different, one is a list and the other a single value

rhugonnet (dismissed):
Not much to say, it's clear, consise and efficient, as always! I really like the tests of this one too :+1:

xdem/terrain.py
erikmannerfelt: It's a slightly clunky, but more or less functional way of describing return types that depend on the input arguments.

Let's say I have a function that provides word counts for one or multiple words:
```python

def word_count(words: str | list[str]) -> int | list[int]:
    if isinstance(words, str):
         words = [words]
    counts = list(map(len, words))

    return counts if len(counts) > 1 else counts[0]
```
The "problem" with this function is that linters and type checkers can't know whether the function will return an integer or a list of integers, so `@overload` is there to provide that information. If `@overload` is not used, the user has to treat the return type as either a list or an integer every time, which can get very tedious!

```python
from typing import overload

@overload
def word_count(words: str) -> int: ...

@overload
def word_count(words: list[str]) -> list[int]: ...

# here goes the function definition above
```
Does that make sense?

xdem/terrain.py
```diff
@@ -0,0 +1,232 @@
...
+                + np.cos(altitude_rad)
+                * np.cos(terrain_attributes["slope"])
+                * np.cos((azimuth_rad - np.pi / 2.0) - terrain_attributes["aspect"])
+            ),
+            0,
+            255,
+        ).astype("float32")
+
+    # Convert the unit if wanted.
+    if degrees:
```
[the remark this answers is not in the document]
erikmannerfelt: It is in radians, since aspect and slope are using `np.arctan2` and `np.arctan`, respectively. I haven't checked from where it begins, however (like if 0 is N, pi is south, and so on)!

xdem/terrain.py
```diff
@@ -0,0 +1,232 @@
...
+
+    Attributes:
+        * 'slope': The slope in degrees or radians (degs: 0=flat, 90=vertical).
+        * 'aspect': The slope aspect in degrees or radians (degs: 0=N, 90=E, 180=S, 270=W)
+        * 'hillshade': The shaded slope in relation to its aspect.
+
+    :param dem: The DEM to analyze.
+    :param attribute: The terrain attribute(s) to calculate.
+    :param resolution: The X/Y or (X, Y) resolution of the DEM.
+    :param degrees: Convert radians to degrees?
```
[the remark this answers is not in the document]
erikmannerfelt: Answered above.

erikmannerfelt:
Okay, I'm stoked for "my next trick"! I really really wanted a working curvature calculation, so I more or less had to reinvent the universe, like Carl Sagan would say. But it worked!

The underlying magic, which I believe we can make many more uses of, is `xdem.terrain.get_quadric_coefficients()` (may need a better name?). It uses parallelized numba to loop over every pixel, fit a quadric surface to its nearest 3x3 region, and return the associated coefficients. Thus, for every pixel, one can later use numpy to calculate any surface attribute that can be derived from this equation:

```
Z = Ax²y² + Bx²y + Cxy² + Dx² + Ey² + Fxy + Gx + Hy + I
```

for example, curvature is the second derivative of this equation:

```
Curvature = -2(D + E) * 100
```

I envision slope, aspect, planform curvature, profile curvature, and possibly others to be interesting products to robustly calculate from this function, instead of relying on `np.gradient` that only looks right, left, up and down.

If you can help me with the maths (how does one derive these products from the above equations???), I can implement them now. Otherwise, I can just do them in future PRs.

xdem/terrain.py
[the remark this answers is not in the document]
rhugonnet: Yes this is what I meant for "where it begins"! 
I think you can use the same calculation you did for degrees, but directly in radians. And then if `degrees=degrees` convert directly into degrees without any other calculation.

xdem/terrain.py
rhugonnet: Yes, perfect sense! Thanks a lot for the example :)

rhugonnet:
Went through your new commits, they are really great!!

**I don't think we are reinventing the universe, because there exists no implementation of this in Python (haven't found any)**. Only richDEM is in the picture, and it's based on C++ code. 

Apparently, with the formulation above, 
`SLOPE = -(G²+H²)**(1/2)`
`PROFC = -2(DH² + EG² - FGH)/(G²+H²)`
`PLANC = 2(DH² + EG² -FGH)/(G²+H²)`

References:
For the slope, the calculation is described here (p18, left side): https://ieeexplore.ieee.org/document/1456186
For the curvature (plan, profile, mean), but also slope, they are described here: https://onlinelibrary.wiley.com/doi/epdf/10.1002/esp.3290120107

rhugonnet:
While we're at it, we could also add:

**The Terrain Rugedness Index** 
- Riley formula is recommended for terrestrial (careful, there's an erratum to the original manuscript, described in the first page here): https://download.osgeo.org/qgis/doc/reference-docs/Terrain_Ruggedness_Index.pdf
- Wilson formula is recommended for bathymetry: https://www2.unil.ch/biomapper/Download/Wilson-MarGeo-2007.pdf.

**Rugosity**
- ***The Wilson paper is great***: summarizes all calculations, from slope/aspect/curvature to Terrain Ruggedness Index but also Rugosity(Roughness). 

I was also thinking we could add two common derivatives of aspect and slope: **Northness** and **Eastness**. Those quantify the exposition North/South and East/West between -1 and 1. Those are also mentioned in the Wilson, 2007.

rhugonnet:
For a great list of references/possibilities for morphometry calculations, see here: http://www.saga-gis.org/saga_tool_doc/2.2.3/ta_morphometry_0.html

I'm clearly in favor of implementing everything based of [6] 9 parameter 2nd order polynom (Zevenbergen & Thorne 1987), as discussed above. It is the default almost everywhere.

xdem/terrain.py
```diff
@@ -230,3 +259,156 @@ def hillshade(
...
+            for k in range(-1, 2):
+                row_indexer = min(max(row + k, 0), dem.shape[0] - 1)
+                col_indexer = min(max(col + j, 0), dem.shape[1] - 1)
+                Z[count] = dem[row_indexer, col_indexer]
+                count += 1
+
+        # Get a mask of all invalid (nan or inf) values.
+        invalids = ~np.isfinite(Z)
+
+        # Skip the pixel if it and all of its neighbours are invalid
```
rhugonnet, line 362: This is something we had no control over before, choosing the behaviour with NaNs will be really cool

xdem/terrain.py
```diff
@@ -230,3 +259,156 @@ def hillshade(
...
+    output = np.empty((9,) + dem.shape, dtype=dem.dtype) + np.nan
+
+    # Loop over every pixel concurrently.
+    for i in numba.prange(dem.size):
+        # Derive its associated row and column index.
+        col = i % dem.shape[1]
+        row = int(i / dem.shape[1])
+
+        # Extract the pixel and its 8 immediate neighbours.
+        # If the border is reached, just duplicate the closest neighbour to obtain 9 values.
```
rhugonnet, line 349: Same as comment below for the edge behaviour

xdem/terrain.py
erikmannerfelt: What kind of options would you like?

xdem/terrain.py
erikmannerfelt: How would one otherwise do it with edges? I'm very new to this!

erikmannerfelt:
> Went through your new commits, they are really great!!
> 
> **I don't think we are reinventing the universe, because there exists no implementation of this in Python (haven't found any)! :)**
> Only richDEM is in the picture, and it's based on C++ code.
> 
> Apparently, with the formulation above,
> `SLOPE = -(G²+H²)**(1/2)`
> `PROFC = -2(DH² + EG² - FGH)/(G²+H²)`
> `PLANC = 2(DH² + EG² -FGH)/(G²+H²)`
> 
> References:
> For the slope, the calculation is described here (p18, left side): https://ieeexplore.ieee.org/document/1456186
> For the curvature (plan, profile, mean), but also slope, they are described here: https://onlinelibrary.wiley.com/doi/epdf/10.1002/esp.3290120107

Wow, slope was just that easy, huh!? Below is a comparison: left is np.gradient, right is cubic magic

image


Do you have any idea of how to calculate aspect from the cubic magic, @rhugonnet? If so, we could remove np.gradient entirely!

erikmannerfelt:
> > Went through your new commits, they are really great!!
> > **I don't think we are reinventing the universe, because there exists no implementation of this in Python (haven't found any)! :)**
> > Only richDEM is in the picture, and it's based on C++ code.
> > Apparently, with the formulation above,
> > `SLOPE = -(G²+H²)**(1/2)`
> > `PROFC = -2(DH² + EG² - FGH)/(G²+H²)`
> > `PLANC = 2(DH² + EG² -FGH)/(G²+H²)`
> > References:
> > For the slope, the calculation is described here (p18, left side): https://ieeexplore.ieee.org/document/1456186
> > For the curvature (plan, profile, mean), but also slope, they are described here: https://onlinelibrary.wiley.com/doi/epdf/10.1002/esp.3290120107
> 
> Wow, slope was just that easy, huh!? Below is a comparison: left is np.gradient, right is cubic magic
> 
> image
> 
> Do you have any idea of how to calculate aspect from the cubic magic, @rhugonnet? If so, we could remove np.gradient entirely!

Or if the x/y gradient could be calculated. I see that the x and y gradient is still needed for the hillshade. "Worst case" I'll just make it so the `np.gradient` is only calculated if hillshades/aspects are requested.

rhugonnet:
> Do you have any idea of how to calculate aspect from the cubic magic, @rhugonnet? If so, we could remove np.gradient entirely!

Yep, it's `ASPECT = ARCTAN(-H/-G)`! (Zevenbergen, Equation 15).

erikmannerfelt:
> > Do you have any idea of how to calculate aspect from the cubic magic, @rhugonnet? If so, we could remove np.gradient entirely!
> 
> Yep, it's `ASPECT = ARCTAN(-H/-G)`! (Zevenbergen, Equation 15).

Beautiful!!

rhugonnet:
> > > Do you have any idea of how to calculate aspect from the cubic magic, @rhugonnet? If so, we could remove np.gradient entirely!
> > 
> > 
> > Yep, it's `ASPECT = ARCTAN(-H/-G)`! (Zevenbergen, Equation 15).
> 
> Beautiful!!

Yes! Don't know why he didn't write this as simply `ARCTAN(H/G)`, it's equivalent. It's written like this in Wilson, 2007 (that uses only 6 parameters instead of 9)

xdem/terrain.py
rhugonnet: I guess we could have:
`default` = return NaN if any of the 9 values are NaN (default in GDAL, SAGA)
`fill_mean` = return value using the mean of the 9 values to fill gaps, if center pixel is not a NaN
`fill_median` = return value using the median of the 9 values to fill gaps, if center pixel is not a NaN
`fill_bilinear` = return value after bilinear extrapolation of the 3x3 data (if possible)
`fill_cubic` = return value after cubic extrapolation of the 3x3 data (if possible)

Seems to me that the extrapolation will fail quite often if there is a lot of NaN, would it make sense to have a similar extrapolation but performed on a slightly larger kernel (5x5)? I'm also wondering if this would work better directly in the numba loop, or with a pre-filled DEM at specific locations (e.g. using a `binary.dilate(niter=2)` on a mask of valid data).
What do you think?

erikmannerfelt:
> Went through your new commits, they are really great!!
> 
> **I don't think we are reinventing the universe, because there exists no implementation of this in Python (haven't found any)! :)**
> Only richDEM is in the picture, and it's based on C++ code.
> 
> Apparently, with the formulation above,
> `SLOPE = -(G²+H²)**(1/2)`
> `PROFC = -2(DH² + EG² - FGH)/(G²+H²)`
> `PLANC = 2(DH² + EG² -FGH)/(G²+H²)`
> 
> References:
> For the slope, the calculation is described here (p18, left side): https://ieeexplore.ieee.org/document/1456186
> For the curvature (plan, profile, mean), but also slope, they are described here: https://onlinelibrary.wiley.com/doi/epdf/10.1002/esp.3290120107

Is there a typo in planform/profile curvature here, @rhugonnet? They seem to be the same only with an inverted sign.

xdem/terrain.py
erikmannerfelt: I can definitely implement the first three filling cases, and I'll try the other.

Changing kernel sizes is a bit harder, as the calculation for A, B, C... is based on 9 numbers and not more. I'm convinced there's a way to calculate them on more numbers, but that's way beyond my understanding of this!

xdem/terrain.py
rhugonnet: For consistency maybe the exact same as above, considering the edge as NaNs? Not sure how `fill_nearest` would work on an exact grid.
And maybe `default` should be more explicit `nofill` or `expand_nodata`... (not super inspired)?

xdem/terrain.py
rhugonnet: > Is there a typo in planform/profile curvature here, @rhugonnet? They seem to be the same only with an inverted sign.

Yes, my bad, corrected versions:
`PROFC = -2(D*G² + E*H² + FGH)/(G² + H²)`
`PLANC = 2(D*H² + E*G² - FGH)/(G² + H²)`

Equations 17 and 18 of Zevenbergen, 1987.

rhugonnet:
> Is there a typo in planform/profile curvature here, @rhugonnet? They seem to be the same only with an inverted sign.

Yes, my bad, corrected versions:
`PROFC = -2(D*G² + E*H² + FGH)/(G² + H²)`
`PLANC = 2(D*H² + E*G² - FGH)/(G² + H²)`

Equations 17 and 18 of Zevenbergen, 1987.

erikmannerfelt:
`ASPECT = ARCTAN(-H/-G)` didn't work, but `aspect = np.arctan2(-G, H)` did. Maybe just wrong signs somewhere..

rhugonnet:
> `ASPECT = ARCTAN(-H/-G)` didn't work, but `aspect = np.arctan2(-G, H)` did. Maybe just wrong signs somewhere..

Sounds good. You need `arctan2` to avoid the discontinuity in 0, and Zevenbergen notes: `The signs of the numerator and denominator of equation 15 determine in which quadrant theta (aspect) lies`

rhugonnet:
I'd thought it'd be `arctan2(-H,G)` though... 
We can check with richDEM also (https://github.com/GlacioHack/xdem/blob/main/xdem/coreg.py#L136), it's going to be time to move those functions away soon!

xdem/terrain.py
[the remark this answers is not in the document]
erikmannerfelt: Fixed in 9b9e3c3

xdem/terrain.py
erikmannerfelt: Added in e940f80. No tests yet though...

erikmannerfelt:
Now, all analyses use the quadric functionality, and `np.gradient` is no more!

The documentation of my branch is continuously updating with these features, so you can see what they look like:

https://xdem-erikm.readthedocs.io/en/slope/terrain.html

In the top, I've made a TODO list for things that are still left.

rhugonnet:
> Now, all analyses use the quadric functionality, and `np.gradient` is no more!
> 
> The documentation of my branch is continuously updating with these features, so you can see what they look like:
> 
> https://xdem-erikm.readthedocs.io/en/slope/terrain.html
> 
> In the top, I've made a TODO list for things that are still left.

Amazing!! What a progress made today :+1:  (my contribution was just reading papers, but it was fun as well :D)

Great TODO list! I'd add one more thing: 
- Benchmark the speed of the purely Python `numba` approach VS the C++ `richDEM`, which I quote "uses parallel processing and state of the art algorithms to quickly process even very large DEMs"

If we are slower, there's probably still some more margin for progress on the Python side! For example decorating with @transonic.jit through https://transonic.readthedocs.io/en/latest/ (as mentioned in the Nature paper we read the other day!)
Or even using Dask in the future, now it's all in Python so everything is possible :smile:

xdem/spatial_tools.py
```diff
@@ -176,7 +176,7 @@ def stack_rasters(rasters: list[gu.georaster.Raster], reference: Union[int, gu.R
     If use_ref_bounds is True, output will have the shape (N, height, width) where N is len(rasters) and \
 height and width is equal to reference's shape.
     If use_ref_bounds is False, output will have the shape (N, height2, width2) where N is len(rasters) and \
-height2 and width2 are set based on reference's resolution and the maximum extent of all rasters.
+height2 anp.gradient(dem_copy)nd width2 are set based on reference's resolution and the maximum extent of all rasters.
```
adehecq, line 179: typo?

erikmannerfelt: Oops! Fixed in bb4b70b

adehecq:
> Okay, I'm stoked for "my next trick"! I really really wanted a working curvature calculation, so I more or less had to reinvent the universe, like Carl Sagan would say. But it worked!
> 
> The underlying magic, which I believe we can make many more uses of, is `xdem.terrain.get_quadric_coefficients()` (may need a better name?). It uses parallelized numba to loop over every pixel, fit a quadric surface to its nearest 3x3 region, and return the associated coefficients for each pixel independently `(9, row, col)`. Thus, for every pixel, one can later use numpy to calculate any surface attribute that can be derived from this equation:
> 
> ```
> Z = Ax²y² + Bx²y + Cxy² + Dx² + Ey² + Fxy + Gx + Hy + I
> ```

This looks like a great in-house implementation! But I wonder how efficient it is to loop through each pixel, even if you use numba... Is it efficient if using only 1-2 cores?
The alternative that I've kept mentioning is the Savitsky Golay algorithm, that does... exactly what you did! It fits any polynomial to a subset window around your center pixel, but possibly much more efficiently as it uses convolution. And I think you can select the window size you want, so not limited to 3x3.
See the scipy's cookbook: https://scipy.github.io/old-wiki/pages/Cookbook/SavitzkyGolay
But I sent you a slightly tuned example I derived from this, some time ago. Have you looked at it?

This would not be a huge effort though to create a copy of get_quadratic_coeffs that uses SG instead of a loop. Let me know if you want me to show you again.

adehecq:
Should we try to merge that and add the Savitsky-Golay option to the issues?

xdem/terrain.py
erikmannerfelt: I added `"none"`, `"mean"` and `"median"` in e940f80. For the rest, I'd prefer to raise an issue for the future!

erikmannerfelt:
This PR is now in a state where the proper functionality is there.

Performance-wise, we have discussed using RichDEM or Savitzky-Golay magic.

- RichDEM is good because it's fast and memory efficient, but it's not easy to customize, for example how edges and nans should be treated. There is also a fixed set of terrain attributes that can be found, so TRI and rugosity would be off the table.
- Savitzky-Golay was more performant than this implementation when @adehecq tried it, but when I did, it was as fast/slow as my own implementation. Surely, there could be ways to improve this, but I would rather see it as a future improvement rather than something essential.

Things to add as issues:

- Add TRI
- Add rugosity
- Potentially add max/min slope/curvature etc.
- Add more nan fill methods, like bilinear/cubic interpolation.
- Potentially change the backend to Savitzky-Golay.

What are your thoughts of this idea, aka merging as soon as your potential comments are fixed, and then tackling each issue one-by-one?

rhugonnet:
Tackling one by one seems the best choice to me.
The functionalities/tests/documentation are there, so we should merge (and the PR is getting pretty big!). Great job :+1: 

What about putting the richDEM wrappers in `terrain.py` and integrating richDEM as dependency from conda-forge, do you think that would be best in a different PR also?

Additionally, I'm not sure exactly why the `subdivide_array` function and subfunctions changes are in this PR?

erikmannerfelt:
> Tackling one by one seems the best choice to me.
> The functionalities/tests/documentation are there, so we should merge (and the PR is getting pretty big!). Great job +1
> 
> What about putting the richDEM wrappers in `terrain.py` and integrating richDEM as dependency from conda-forge, do you think that would be best in a different PR also?
> 
> Additionally, I'm not sure exactly why the `subdivide_array` function and subfunctions changes are in this PR?

Thanks @rhugonnet!

I think a RichDEM integration would be a good PR, as there are some unsolved issues with that (when should RichDEM be preferred over in-house methods, what semi-black-box settings are RichDEM using, what about the NaN filling and edge cases).

Regarding `subdivide_array`, where have I changed it? I cannot see the differences!

rhugonnet:
> Regarding `subdivide_array`, where have I changed it? I cannot see the differences!

Can't see it anymore! Maybe it displayed only when you were merging

rhugonnet (approved):
Nice for the filling methods
I had reviewed all the rest already :)

adehecq:
> This PR is now in a state where the proper functionality is there.
> 
> Performance-wise, we have discussed using RichDEM or Savitzky-Golay magic.
> 
>     * RichDEM is good because it's fast and memory efficient, but it's not easy to customize, for example how edges and nans should be treated. There is also a fixed set of terrain attributes that can be found, so TRI and rugosity would be off the table.
> 
>     * Savitzky-Golay was more performant than this implementation when @adehecq tried it, but when I did, it was as fast/slow as my own implementation. Surely, there could be ways to improve this, but I would rather see it as a future improvement rather than something essential.
> 
> 
> Things to add as issues:
> 
>     * Add TRI
> 
>     * Add rugosity
> 
>     * Potentially add max/min slope/curvature etc.
> 
>     * Add more nan fill methods, like bilinear/cubic interpolation.
> 
>     * Potentially change the backend to Savitzky-Golay.
> 
> 
> What are your thoughts of this idea, aka merging as soon as your potential comments are fixed, and then tackling each issue one-by-one?

I also agree that this is time to merge what exists and see how to tackle the rest later one by one.
Good job!!

PR #148, opened 2021-06-16
Fix failing test and add handlers for warnings.

erikmannerfelt:
As mentioned in https://github.com/GlacioHack/GeoUtils/issues/211, a warning was triggered by `SatelliteImage` (and thus `DEM`) upon instantiation. This made some tests fail, as they have a "no warning" policy.

This PR fixes that, together with two warnings that have been bugging me for a while:
1. dDEM.interpolate("local_hypsometric") triggered an interpolation warning (which is supposed to be okay)
2. A matplotlib deprecation warning was triggered for a documentation code snippet, which I've fixed.

### DIFF
--- a/docs/source/code/coregistration_plot_nuth_kaab.py
+++ b/docs/source/code/coregistration_plot_nuth_kaab.py
@@ -33,7 +33,7 @@ plt.title(f"After coregistration. NMAD={nmad_post:.1f} m")
 img = plt.imshow(ddem_post, cmap="coolwarm_r", vmin=-vlim, vmax=vlim)
 plt.axis("off")
 plt.subplot2grid((1, 15), (0, 14), colspan=1)
-cbar = plt.colorbar(img, fraction=0.4)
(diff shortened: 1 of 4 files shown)


erikmannerfelt:
#108  will not pass the tests until this PR is merged into its history.

rhugonnet (approved):

rhugonnet:
Didn't know about the "no warning" policy!

adehecq:
I didn't realize that the DEM class was based upon SatelliteImage rather than Raster. I guess it's only to benefit from the metadata parsing?

adehecq (approved):

erikmannerfelt:
@rhugonnet, only a few tests have it enabled (by starting it with `warnings.simplefilter("error")`), since warnings are a sign of something going wrong in those tests! In the future, I would prefer for us to always fail on warnings (expected warnings should be caught accordingly), but that's for the future ;)

PR #149, opened 2021-06-17
Add N-dimensional binning function (for biases, non-stationarity quantification etc..)

rhugonnet:
## Binning simultaneously in several (N) dimensions based on scipy
Scipy base function on which I rely: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic_dd.html#scipy.stats.binned_statistic_dd
After some searching, the pure pandas solutions seemed a bit hard to extend in N-dimensions: https://stackoverflow.com/questions/60340107/pandas-dataframe-bin-on-multiple-columns-get-statistics-on-another-column
Suggestions of other implementations are welcome. Originally, my own functions were quite simply based on multiple `for` loops, so lots of room for improvement here  :exploding_head: (those can actually run quite fast, especially when using numba).

## Base usage (see full example in test)
```python
from xdem.spatial_tools import nmad
list_df_1d, list_df_2d, list_df_nd = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(), 
list_var=[slp.flatten(),ref.data.flatten(),asp.flatten()],list_var_names=['slope','elevation','aspect'], statistics=['count',np.nanmedian,nmad])
```
Binning in several dimension simultaneously lowers the amount of samples that can be used for estimation, so it's also very valuable to have the statistics for each 1D and 2D combinations.
Here by default the binning values are subdivided in 10 bins between their min and max. So subdivided into 10 subsamples in 1D, 100 in 2D and 1000 in 3D.
`list_df_1d` contains all possible binnings in 1D, for example with slope:
```
list_df_1d[0]
Out[15]: 
   count_slope  nanmedian_slope  nmad_slope  binmid_slope
0     487886.0        -0.092560    2.540733      1.784413
1     301319.0        -0.404083    2.913965      5.353238
2     265642.0         0.124084    3.007007      8.922064
3     206016.0         0.677917    3.451950     12.490889
4      46273.0         0.749176    4.222139     16.059715
5       3977.0         0.047241    5.870970     19.628540
6        695.0         0.736572    7.450644     23.197366
7        171.0        -0.417175   13.560306     26.766191
8         36.0         0.144989   14.016108     30.335017
9          5.0       -20.540710   16.012605     33.903843`
```
Same for `list_df_2d` for combinations of 2D binning.
```
list_df_2d[0]
Out[16]: 
    count_slope-elevation  ...  binmid_elevation
0                227518.0  ...         58.760407
1                 76479.0  ...        160.176258
2                 65977.0  ...        261.592110
3                 38701.0  ...        363.007962
4                 27911.0  ...        464.423814
..                    ...  ...               ...
95                    3.0  ...        565.839666
96                    0.0  ...        667.255517
97                    1.0  ...        768.671369
98                    1.0  ...        870.087221
99                    0.0  ...        971.503073
[100 rows x 5 columns]
```

## Limitations
1/ Currently, I output the results in several lists, each contain a pd.DataFrame for an easier manipulation. This is because binning variables can have different bin lengths, so the dimension cannot necessarily be concatenated into one dataframe.
An approach to circumvent this issue would be to have basic names for the statistics `count`, `nmad`, `binmid` and then add another column with a string containing the variable name `var`. Then they could all be concatenated in the same object, even the ones in multiple dimensions (now that I write it, I think that's highly preferable! it's an easy change :smile: )

2/ Right now only the middle value of the bin is returned. I could add 2 columns (`bin_min`, `bin_max`) to ensure we have this information as well?

3/ I'm not sure how well the `scipy.binned_statistic_dd` approach scales with millions of pixels... but I think it would be counterproductive to create our own binning functions that could become deprecated rapidly. Better to base ourselves on an existing pandas/scipy implementation! And I'm not sure `pandas.cut` is satisfactory for this kind of implementation right now (or maybe I missed something!)

### DIFF
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -45,6 +45,29 @@ def load_diff() -> tuple[gu.georaster.Raster, np.ndarray]:
 
     return diff, mask
 
+def load_ref_and_diff() -> tuple[gu.georaster.Raster, gu.georaster.Raster, np.ndarray]:
+    """Load example files to try coregistration methods with."""
+    examples.download_longyearbyen_examples(overwrite=False)
+
+    reference_raster = gu.georaster.Raster(examples.FILEPATHS["longyearbyen_ref_dem"])
+    to_be_aligned_raster = gu.georaster.Raster(examples.FILEPATHS["longyearbyen_tba_dem"])
+    glacier_mask = gu.geovector.Vector(examples.FILEPATHS["longyearbyen_glacier_outlines"])
+    inlier_mask = ~glacier_mask.create_mask(reference_raster)
+
+    metadata = {}
+    # aligned_raster, _ = xdem.coreg.coregister(reference_raster, to_be_aligned_raster, method="amaury", mask=glacier_mask,
+    #                                          metadata=metadata)
+    nuth_kaab = xdem.coreg.NuthKaab()
+    nuth_kaab.fit(reference_raster.data, to_be_aligned_raster.data,
+                  inlier_mask=inlier_mask, transform=reference_raster.transform)
+    aligned_raster = nuth_kaab.apply(to_be_aligned_raster.data, transform=reference_raster.transform)
+
+    diff = gu.Raster.from_array((reference_raster.data - aligned_raster),
+                                transform=reference_raster.transform, crs=reference_raster.crs)
+    mask = glacier_mask.create_mask(diff)
+
+    return reference_raster, diff, mask
+
 
 class TestVariogram:
 
@@ -171,3 +194,30 @@ class TestPatchesMethod:
             gsd=diff.res[0],
             area_size=10000
         )
+
+class TestBinning:
+
+    def test_nd_binning(self):
+
+        ref, diff, mask = load_ref_and_diff()
+
+        slp, asp = xdem.coreg.calculate_slope_and_aspect(ref.data.squeeze())
+
+        # 1d binning
+        output = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten()],list_var_names=['slope'])
+
+        # nmad goes up quite a bit with slope, we can expect a 10 m measurement error difference
+        assert output[0][0].nmad_slope.values[-1] - output[0][0].nmad_slope.values[0] > 10
+
+        # try custom stat
+        def percentile_80(a):
+            return np.nanpercentile(a, 80)
+
+        xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten()],list_var_names=['slope'], statistics=['count',percentile_80])
+
+        # 2d binning
+        output_2d = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten(),ref.data.flatten()],list_var_names=['slope','elevation'])
+
+        # nd binning
+        output_3d = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten(),ref.data.flatten(),asp.flatten()],list_var_names=['slope','elevation','aspect'])
+
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -21,30 +21,6 @@ with warnings.catch_warnings():
 
 PLOT = False
 
-
-def load_diff() -> tuple[gu.georaster.Raster, np.ndarray]:
-    """Load example files to try coregistration methods with."""
-    examples.download_longyearbyen_examples(overwrite=False)
-
-    reference_raster = gu.georaster.Raster(examples.FILEPATHS["longyearbyen_ref_dem"])
-    to_be_aligned_raster = gu.georaster.Raster(examples.FILEPATHS["longyearbyen_tba_dem"])
-    glacier_mask = gu.geovector.Vector(examples.FILEPATHS["longyearbyen_glacier_outlines"])
-    inlier_mask = ~glacier_mask.create_mask(reference_raster)
-
-    metadata = {}
-    # aligned_raster, _ = xdem.coreg.coregister(reference_raster, to_be_aligned_raster, method="amaury", mask=glacier_mask,
-    #                                          metadata=metadata)
-    nuth_kaab = xdem.coreg.NuthKaab()
-    nuth_kaab.fit(reference_raster.data, to_be_aligned_raster.data,
-                  inlier_mask=inlier_mask, transform=reference_raster.transform)
-    aligned_raster = nuth_kaab.apply(to_be_aligned_raster.data, transform=reference_raster.transform)
-
-    diff = gu.Raster.from_array((reference_raster.data - aligned_raster),
-                                transform=reference_raster.transform, crs=reference_raster.crs)
-    mask = glacier_mask.create_mask(diff)
-
-    return diff, mask
-
 def load_ref_and_diff() -> tuple[gu.georaster.Raster, gu.georaster.Raster, np.ndarray]:
     """Load example files to try coregistration methods with."""
     examples.download_longyearbyen_examples(overwrite=False)
@@ -76,7 +52,7 @@ class TestVariogram:
     def test_empirical_fit_variogram_running(self):
 
         # get some data
-        diff, mask = load_diff()
+        diff, mask = load_ref_and_diff()[1:3]
 
         x, y = diff.coords(offset='center')
         coords = np.dstack((x.flatten(), y.flatten())).squeeze()
@@ -184,7 +160,7 @@ class TestPatchesMethod:
 
     def test_patches_method(self):
 
-        diff, mask = load_diff()
+        diff, mask = load_ref_and_diff()[1:3]
 
         warnings.filterwarnings("error")
         # check the patches method runs
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -177,10 +177,10 @@ class TestBinning:
 
         ref, diff, mask = load_ref_and_diff()
 
-        slp, asp = xdem.coreg.calculate_slope_and_aspect(ref.data.squeeze())
+        slope, aspect = xdem.coreg.calculate_slope_and_aspect(ref.data.squeeze())
 
         # 1d binning
-        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten()],list_var_names=['slope'])
+        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten()],list_var_names=['slope'])
 
         # nmad goes up quite a bit with slope, we can expect a 10 m measurement error difference
         assert df.nmad.values[-1] - df.nmad.values[0] > 10
@@ -189,11 +189,11 @@ class TestBinning:
         def percentile_80(a):
             return np.nanpercentile(a, 80)
 
-        xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten()],list_var_names=['slope'], statistics=['count',percentile_80])
+        xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten()],list_var_names=['slope'], statistics=['count',percentile_80])
 
         # 2d binning
-        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten(),ref.data.flatten()],list_var_names=['slope','elevation'])
+        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten(),ref.data.flatten()],list_var_names=['slope','elevation'])
 
         # nd binning
-        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten(),ref.data.flatten(),asp.flatten()],list_var_names=['slope','elevation','aspect'])
+        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten(),ref.data.flatten(),aspect.flatten()],list_var_names=['slope','elevation','aspect'])
 
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -185,8 +185,8 @@ class TestBinning:
         # check length matches
         assert df.shape[0] == 10
         # check bin edges match the minimum and maximum of binning variable
-        assert np.nanmin(slope) == np.min(df.slope)
-        assert np.nanmax(slope) == np.max(df.slope)
+        assert np.nanmin(slope) == np.min(pd.IntervalIndex(df.slope).left)
+        assert np.nanmax(slope) == np.max(pd.IntervalIndex(df.slope).right)
 
         # 1d binning with 20 bins
         df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(), list_var=[slope.flatten()], list_var_names=['slope'],
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -180,7 +180,7 @@ class TestBinning:
         slope, aspect = xdem.coreg.calculate_slope_and_aspect(ref.data.squeeze())
 
         # 1d binning, by default will create 10 bins
-        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten()],list_var_names=['slope'])
+        df = xdem.spstats.nd_binning(values=diff.data.flatten(),list_var=[slope.flatten()],list_var_names=['slope'])
 
         # check length matches
         assert df.shape[0] == 10
@@ -189,7 +189,7 @@ class TestBinning:
         assert np.nanmax(slope) == np.max(pd.IntervalIndex(df.slope).right)
 
         # 1d binning with 20 bins
-        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(), list_var=[slope.flatten()], list_var_names=['slope'],
+        df = xdem.spstats.nd_binning(values=diff.data.flatten(), list_var=[slope.flatten()], list_var_names=['slope'],
                                            list_var_bins=[[20]])
         # check length matches
         assert df.shape[0] == 20
@@ -202,16 +202,16 @@ class TestBinning:
             return np.nanpercentile(a, 80)
 
         # check the function runs with custom functions
-        xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten()],list_var_names=['slope'], statistics=['count',percentile_80])
+        xdem.spstats.nd_binning(values=diff.data.flatten(),list_var=[slope.flatten()],list_var_names=['slope'], statistics=['count',percentile_80])
 
         # 2d binning
-        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten(),ref.data.flatten()],list_var_names=['slope','elevation'])
+        df = xdem.spstats.nd_binning(values=diff.data.flatten(),list_var=[slope.flatten(),ref.data.flatten()],list_var_names=['slope','elevation'])
 
         # dataframe should contain two 1D binning of length 10 and one 2D binning of length 100
         assert df.shape[0] == (10 + 10 + 100)
 
         # nd binning
-        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten(),ref.data.flatten(),aspect.flatten()],list_var_names=['slope','elevation','aspect'])
+        df = xdem.spstats.nd_binning(values=diff.data.flatten(),list_var=[slope.flatten(),ref.data.flatten(),aspect.flatten()],list_var_names=['slope','elevation','aspect'])
 
         # dataframe should contain three 1D binning of length 10 and three 2D binning of length 100 and one 2D binning of length 1000
         assert df.shape[0] == (1000 + 3 * 100 + 3 * 10)
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -204,10 +204,10 @@ class TestBinning:
         slp, asp = xdem.coreg.calculate_slope_and_aspect(ref.data.squeeze())
 
         # 1d binning
-        output = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten()],list_var_names=['slope'])
+        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten()],list_var_names=['slope'])
 
         # nmad goes up quite a bit with slope, we can expect a 10 m measurement error difference
-        assert output[0][0].nmad_slope.values[-1] - output[0][0].nmad_slope.values[0] > 10
+        assert df.nmad.values[-1] - df.nmad.values[0] > 10
 
         # try custom stat
         def percentile_80(a):
@@ -216,8 +216,8 @@ class TestBinning:
         xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten()],list_var_names=['slope'], statistics=['count',percentile_80])
 
         # 2d binning
-        output_2d = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten(),ref.data.flatten()],list_var_names=['slope','elevation'])
+        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten(),ref.data.flatten()],list_var_names=['slope','elevation'])
 
         # nd binning
-        output_3d = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten(),ref.data.flatten(),asp.flatten()],list_var_names=['slope','elevation','aspect'])
+        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten(),ref.data.flatten(),asp.flatten()],list_var_names=['slope','elevation','aspect'])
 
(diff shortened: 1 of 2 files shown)


erikmannerfelt:
This looks really cool!!! I didn't know that existed, and I see so many uses for it! I need some more time to understand this haha, but I'll just quickly chime in with varying row lengths as pandas multiindices:

```python
>>> import pandas as pd
>>> import numpy as np
# Create three random rows of data
>>> bin1_data = np.random.randn(3 * 3).reshape(3, 3)
# Create six random rows of data
>>> bin2_data = np.random.randn(6 * 3).reshape(3, 3)
>>> bin2_data
array([[-0.99161969, -1.82319573, -1.05818692],
       [-0.76215282,  1.23210638, -0.13424905],
       [ 0.8560318 , -0.11520882, -0.85442712],
       [-1.4066068 , -0.34847478,  0.63734095],
       [-0.33961168,  1.28699075, -0.68437763],
       [-1.82674464, -0.1591328 , -0.60634219]])
>>> bins = [bin1_data, bin2_data]
# Concatenate the data on a multi-index level with the keys [0, 1, ...]
>>> bins_df = pd.concat([pd.DataFrame(data=bin_data, columns=["slope", "aspect", "dh"]) for bin_data in bins], keys=range(len(bins)))
>>> bins_df
        slope    aspect        dh
0 0  0.460147  1.358859  0.615630
  1 -0.762801  0.019033 -0.161678
  2 -0.314703  0.081407  0.280405
1 0 -0.991620 -1.823196 -1.058187
  1 -0.762153  1.232106 -0.134249
  2  0.856032 -0.115209 -0.854427
  3 -1.406607 -0.348475  0.637341
  4 -0.339612  1.286991 -0.684378
  5 -1.826745 -0.159133 -0.606342
# The bin-wise mean can easily be calculated using the `level` argument
>>>>>> bins_df["aspect"].mean(level=0)
0    0.486433
1    0.012181
Name: aspect, dtype: float64
```
Does that help your nr. 1?

rhugonnet:
> Does that help your nr. 1?

I didn't know about this level argument! Thanks for the example.
I was thinking of doing something similar: instead of relying on the level, add an additional column with the `var` name to be able to sort/groupby quickly which binning we are interested in.

I'll implement this and show an example, you can tell me if it's clearer!

rhugonnet:
## Update/fixes

(numbers correspond to the points listed above)
1/ Everything is concatenated into a single dataset: all 1D binnings with different variables (added `name_var1`, `name_var2` column names for quick selection) which in turn correspond to `bin_mid_var1`, etc. Binning in all dimensions are also concatenated and a new column `nd` (avoiding `ndim`, that already exists as a pd.DataFrame attribute) reports the number of dimensions used to enable selection/grouping in the DataFrame.
2/ Lower and upper bin values are added to the user output.
3/ Scipy solution runs fast, so efficiency can probably be put aside for now (and added later as an issue if it arises).

## Example
```python
df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=
slp.flatten(),ref.data.flatten(),asp.flatten()],list_var_names=['slope','elevation','aspect'])
```
Looking at the 1D binning outputs:
```
df[df.nd==1][['count','nmad','bin_mid_var1','name_var1']]
Out[9]: 
      count       nmad  bin_mid_var1  name_var1
0  487886.0   2.540733      1.784413      slope
1  301319.0   2.913965      5.353238      slope
2  265642.0   3.007007      8.922064      slope
3  206016.0   3.451950     12.490889      slope
4   46273.0   4.222139     16.059715      slope
5    3977.0   5.870970     19.628540      slope
6     695.0   7.450644     23.197366      slope
7     171.0  13.560306     26.766191      slope
8      36.0  14.016108     30.335017      slope
9       5.0  16.012605     33.903843      slope
0  243045.0   2.171478     58.760407  elevation
1  143183.0   2.164787    160.176258  elevation
2  184779.0   2.311427    261.592110  elevation
3  178005.0   3.127676    363.007962  elevation
4  160431.0   3.851149    464.423814  elevation
5  143249.0   3.953540    565.839666  elevation
6  120596.0   3.684873    667.255517  elevation
7   89779.0   3.557733    768.671369  elevation
8   40698.0   3.001306    870.087221  elevation
9    8255.0   2.926922    971.503073  elevation
0  152706.0   2.222526      0.314159     aspect
1  141249.0   2.410854      0.942476     aspect
2  159072.0   2.560050      1.570794     aspect
3  133223.0   2.862736      2.199111     aspect
4  116850.0   3.184911      2.827428     aspect
5   99464.0   3.669037      3.455746     aspect
6  112596.0   3.940283      4.084063     aspect
7  119228.0   3.481633      4.712381     aspect
8  129925.0   2.768360      5.340698     aspect
9  147707.0   2.595048      5.969015     aspect
```
The 2D outputs:
```
df[df.nd==2][['count','nmad','bin_mid_var1','name_var1','bin_mid_var2','name_var2']]
Out[10]: 
       count       nmad  bin_mid_var1  name_var1  bin_mid_var2  name_var2
0   227518.0   2.119683      1.784413      slope     58.760407  elevation
1    76479.0   1.850999      1.784413      slope    160.176258  elevation
2    65977.0   1.886969      1.784413      slope    261.592110  elevation
3    38701.0   4.202140      1.784413      slope    363.007962  elevation
4    27911.0  10.681344      1.784413      slope    464.423814  elevation
..       ...        ...           ...        ...           ...        ...
95     515.0   3.263367    971.503073  elevation      3.455746     aspect
96     557.0   2.940225    971.503073  elevation      4.084063     aspect
97     875.0   2.771731    971.503073  elevation      4.712381     aspect
98    1130.0   2.053958    971.503073  elevation      5.340698     aspect
99     450.0   2.043642    971.503073  elevation      5.969015     aspect
[300 rows x 6 columns]
```
And in 3D:
```
df[df.nd==3][['count','bin_mid_var1','name_var1','bin_mid_var2','name_var2','bin_mid_var3','name_var3']]
Out[11]: 
       count  bin_mid_var1 name_var1  ...  name_var2 bin_mid_var3  name_var3
0    28263.0      1.784413     slope  ...  elevation     0.314159     aspect
1    25710.0      1.784413     slope  ...  elevation     0.942476     aspect
2    28277.0      1.784413     slope  ...  elevation     1.570794     aspect
3    30895.0      1.784413     slope  ...  elevation     2.199111     aspect
4    24232.0      1.784413     slope  ...  elevation     2.827428     aspect
..       ...           ...       ...  ...        ...          ...        ...
995      0.0     33.903843     slope  ...  elevation     3.455746     aspect
996      0.0     33.903843     slope  ...  elevation     4.084063     aspect
997      0.0     33.903843     slope  ...  elevation     4.712381     aspect
998      0.0     33.903843     slope  ...  elevation     5.340698     aspect
999      0.0     33.903843     slope  ...  elevation     5.969015     aspect
[1000 rows x 7 columns]
```

Voilà !

tests/test_spstats.py
```diff
@@ -45,6 +45,29 @@ def load_diff() -> tuple[gu.georaster.Raster, np.ndarray]:
 
     return diff, mask
 
+def load_ref_and_diff() -> tuple[gu.georaster.Raster, gu.georaster.Raster, np.ndarray]:
```
adehecq, line 48: So far, we've always put the test data in the definition of the class, see e.g. https://github.com/GlacioHack/xdem/blob/2a389a09ae4d68f7e765e576b1c9afd99021dc92/tests/test_volume.py#L13
I don't know which one is better. Maybe @erikmannerfelt has an idea?
I guess it makes sense to have a function that loads this sample data if often used... On the other hand, it's weird to have a function in the tests that is never tested! ^_^

tests/test_spstats.py
```diff
@@ -171,3 +194,30 @@ def test_patches_method(self):
...
             area_size=10000
         )
+
+class TestBinning:
+
+    def test_nd_binning(self):
+
+        ref, diff, mask = load_ref_and_diff()
+
+        slp, asp = xdem.coreg.calculate_slope_and_aspect(ref.data.squeeze())
```
adehecq, line 204: I think now we should try to have variable names that are explicit rather than short. So using 'slope' and 'aspect' might be clearer for reading.

tests/test_spstats.py
```diff
@@ -171,3 +194,30 @@ def test_patches_method(self):
...
+
+        # 1d binning
+        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten()],list_var_names=['slope'])
+
+        # nmad goes up quite a bit with slope, we can expect a 10 m measurement error difference
```
adehecq, line 209: I'm not really sure that is the kind of core functionality of the function you want to test.
What you should check is that the output has the right dimensions and range (let's say you have a binning of 10, the output should have a length 10, and the min/max values are known.

tests/test_spstats.py
```diff
@@ -171,3 +194,30 @@ def test_patches_method(self):
...
+        assert df.nmad.values[-1] - df.nmad.values[0] > 10
+
+        # try custom stat
+        def percentile_80(a):
+            return np.nanpercentile(a, 80)
+
+        xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slp.flatten()],list_var_names=['slope'], statistics=['count',percentile_80])
```
adehecq, line 216: Here and elsewhere, you should have some assert statements.

adehecq: You can see e.g. how it is done with a relatively similar function in volume.py: https://github.com/GlacioHack/xdem/blob/2a389a09ae4d68f7e765e576b1c9afd99021dc92/tests/test_volume.py#L129

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
     from skgstat import models
 
+def nd_binning_scipy(dh: np.ndarray, list_var: List[np.ndarray], list_var_names=List[str], list_var_bins: Optional[List[Iterable]] = None,
+                     statistics: list[Union[str, Callable, None]] = ['count', np.nanmedian ,nmad], list_ranges : Optional[List[Sequence]] = None) \
+        -> tuple[list[pd.DataFrame],list[pd.DataFrame],pd.DataFrame]:
+    """
+    Quantify non-stationarities in elevation measurement errors with one or several, terrain or instrument-related,
+    explanatory variables. See documentation of scipy.stats.binned_statistic_dd for more details.
+
+    :param dh: elevation differences
```
adehecq, line 36: It's good to mention if there are some expectations regarding the shape of the arrays (1D, 2D, 3D) and if the shape should be the same for all inputs (presumably yes).

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
+    :param list_var: list of explanatory variables
+    :param list_var_names: list of names of the explanatory variables
+    :param list_var_bins: list of bins for the explanatory variables
```
adehecq, line 39: You should state what the default is (10 bins from min to max), as it is not clear from the docstring.

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
+        list_var_bins = (10,) * len(list_var_names)
+    statistics_name = [f if isinstance(f,str) else f.__name__ for f in statistics]
+
+    # get binned statistics in 1d: a simple loop is sufficient
+    list_df_1d = []
+    for i, var in enumerate(list_var):
+        df_stats_1d = pd.DataFrame()
+        # get statistics
+        for j, statistic in enumerate(statistics):
+            stats_binned_1d, bedges_1d = binned_statistic(var,dh,statistic=statistic,bins=list_var_bins[i],range=list_ranges)[:2]
```
adehecq, line 57: I see some missing spaces that won't make the linters happy! @erikmannerfelt close your eyes!! ;-)

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
+            # save in a dataframe
+            df_stats_1d[statistics_name[j]] = stats_binned_1d
+        # we need to get the middle of the bins from the edges, to get the same dimension length
+        df_stats_1d['bin_mid_var1'] = (bedges_1d[1:] + bedges_1d[:-1]) / 2
```
adehecq, line 61: I think Erik uses a magic panda function to do this autmoatically in the hypsometric_binning. See https://github.com/GlacioHack/xdem/blob/2a389a09ae4d68f7e765e576b1c9afd99021dc92/xdem/volume.py#L88

adehecq: So the index will contain the min, max and mid value if I'm correct.

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
+            df_stats_2d['nd'] = 2
+
+            list_df_2d.append(df_stats_2d)
+
+
+    # get binned statistics in nd, without redoing the same stats
+    df_stats_nd = pd.DataFrame()
+    if len(list_var)>2:
+        for j, statistic in enumerate(statistics):
+            stats_binned_2d, list_bedges = binned_statistic_dd(list_var,dh,statistic=statistic,bins=list_var_bins,range=list_ranges)[0:2]
```
adehecq, line 103: Can this nd case not be also applied to the 1D and 2D cases? This would make the code simpler...

adehecq:
This will be super useful for analyzing possible biases in the dDEMs and quantifying uncertainties, thanks!!
I made some comments that I think could help making the code simpler and more readable.
And the tests should be a bit more thorough than this! ;-)

tests/test_spstats.py
rhugonnet: Actuallty @erikmannerfelt added this a couple months ago, so should be OK :wink: 
I will collapse the two functions into one though, it will save space

tests/test_spstats.py
rhugonnet: done

tests/test_spstats.py
rhugonnet: Added

xdem/spstats.py
rhugonnet: Added

xdem/spstats.py
rhugonnet: Added

xdem/spstats.py
rhugonnet: Running black & co only before merging will save time, I don't want to format at every review step

xdem/spstats.py
rhugonnet: I could use the same function, yes, but I'd still need to do 3 separate loops to get the binning in all dimensions.
So nothing would change in the code except the function name of 1d, 2d and the format of their output variables, which would take a few more line to reshape in the right format.
 
The 1d, 2d functions were made so it would be easier to manipulate their outputs, so I think we shoud leave them as is

tests/test_spstats.py
rhugonnet: Done in 3fd6d83

xdem/spstats.py
rhugonnet: Modified the bins using Interval Indexes, and extending those to N-D in 985722f

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
 
+def nd_binning_scipy(dh: np.ndarray, list_var: List[np.ndarray], list_var_names=List[str], list_var_bins: Optional[List[Iterable]] = None,
+                     statistics: list[Union[str, Callable, None]] = ['count', np.nanmedian ,nmad], list_ranges : Optional[List[Sequence]] = None) \
+        -> tuple[list[pd.DataFrame],list[pd.DataFrame],pd.DataFrame]:
+    """
+    Quantify non-stationarities in elevation measurement errors with one or several, terrain or instrument-related,
+    explanatory variables. See documentation of scipy.stats.binned_statistic_dd for more details.
+
+    :param dh: elevation differences
+    :param list_var: list of explanatory variables
```
erikmannerfelt: I'm afraid I had to refer to the tests to understand what this was (and now I do). Could this be elaborated upon?

```
List of explanatory variables to compare in relation to 'dh'   # Or 'arr' / 'values' (see my comment above)
```
 is maybe slightly more descriptive?

xdem/spstats.py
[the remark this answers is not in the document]
erikmannerfelt: Also, as I mentioned in the tests; if one wants e.g. 20 bins, could this not be supplied as an integer instead of a nested list `[[20]]`?

xdem/spstats.py
erikmannerfelt: Yes, IntervalIndex is great! It has `left`, `right` and `mid` properties. I guess it's not called min/max because they might be either ascending or descending.

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
+        df_stats_1d['name_var1'] = list_var_names[i]
+        df_stats_1d['nd'] = 1
+
+        list_df_1d.append(df_stats_1d)
+
+    # get binned statistics in 2d: all possible 2d combinations
+    list_df_2d = []
+    if len(list_var)>1:
+        combs = list(itertools.combinations(list_var_names, 2))
+        for i, comb in enumerate(combs):
```
erikmannerfelt, line 72: Would this work in parallel (maybe in a future PR)?

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
+
+
+    # get binned statistics in nd, without redoing the same stats
+    df_stats_nd = pd.DataFrame()
+    if len(list_var)>2:
+        for j, statistic in enumerate(statistics):
+            stats_binned_2d, list_bedges = binned_statistic_dd(list_var,dh,statistic=statistic,bins=list_var_bins,range=list_ranges)[0:2]
+            df_stats_nd[statistics_name[j]] = stats_binned_2d.flatten()
+        list_bin_mid, list_bin_low, list_bin_upp = ([] for i in range(3))
+        for bedges in list_bedges:
```
erikmannerfelt, line 99: The documentation is a little sparse here. Could some descriptive comments be added? :)

erikmannerfelt (commented):
I now had a thorough look and this looks super useful! Great job!!

I have a few comments, mostly not about your intended use, and I envision many to be able to use it for completely different purposes. Perhaps this could be made more generic to be a multi-purpose function (basically just rename the dh argument to something generic, modify the docstring a bit, and it's done!)?

Also, now that I've read the entire documentation and the tests, I more or less understand the functionality. I can unfortunately not say that I understand it completely, and I worry that this might be reflective of other users' experience as well. Could you please elaborate a bit on the docstrings, perhaps add a few minimal examples, and optimally also add documentation to readthedocs (this can of course be done in a later stage)?

tests/test_spstats.py
```diff
@@ -171,3 +170,49 @@ def test_patches_method(self):
...
+
+        # nmad goes up quite a bit with slope, we can expect a 10 m measurement error difference
+        assert df.nmad.values[-1] - df.nmad.values[0] > 10
+
+        # try custom stat
+        def percentile_80(a):
+            return np.nanpercentile(a, 80)
+
+        # check the function runs with custom functions
+        xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten()],list_var_names=['slope'], statistics=['count',percentile_80])
```
erikmannerfelt, line 205: I assume that `count` is a statistic that is built into scipy? What happens if I want the statistic `blasjdhakwjhd`? Does it fail and raise a descriptive error? That would be nice to test!

tests/test_spstats.py
```diff
@@ -171,3 +170,49 @@ def test_patches_method(self):
             gsd=diff.res[0],
             area_size=10000
         )
+
+class TestBinning:
```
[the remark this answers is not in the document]
rhugonnet: The next functions to come on non-stationarity will need the same data loading, and should thus fit in that same class

tests/test_spstats.py
```diff
@@ -171,3 +170,49 @@ def test_patches_method(self):
...
+
+    def test_nd_binning(self):
+
+        ref, diff, mask = load_ref_and_diff()
+
+        slope, aspect = xdem.coreg.calculate_slope_and_aspect(ref.data.squeeze())
+
+        # 1d binning, by default will create 10 bins
+        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(),list_var=[slope.flatten()],list_var_names=['slope'])
```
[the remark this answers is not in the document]
rhugonnet: > Ah, I guess not flattening the array will do 2D/ND binning? There are no tests for that!!

Not supported by scipy, scipy will return the appropriate error.

tests/test_spstats.py
```diff
@@ -171,3 +170,49 @@ def test_patches_method(self):
...
+
+        # check length matches
+        assert df.shape[0] == 10
+        # check bin edges match the minimum and maximum of binning variable
+        assert np.nanmin(slope) == np.min(df.bin_low_var1)
+        assert np.nanmax(slope) == np.max(df.bin_upp_var1)
+
+        # 1d binning with 20 bins
+        df = xdem.spstats.nd_binning_scipy(dh=diff.data.flatten(), list_var=[slope.flatten()], list_var_names=['slope'],
```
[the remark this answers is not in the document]
rhugonnet: Added that option

tests/test_spstats.py
```diff
@@ -171,3 +170,49 @@ def test_patches_method(self):
...
+                                           list_var_bins=[[20]])
+        # check length matches
+        assert df.shape[0] == 20
+
+        # nmad goes up quite a bit with slope, we can expect a 10 m measurement error difference
+        assert df.nmad.values[-1] - df.nmad.values[0] > 10
+
+        # try custom stat
+        def percentile_80(a):
```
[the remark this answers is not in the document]
rhugonnet: This functionality relates to the scipy function and is documented, it's not really our job to test it

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
 from scipy.optimize import curve_fit
+from scipy.stats import binned_statistic, binned_statistic_2d, binned_statistic_dd
+from xdem.spatial_tools import nmad
 
 with warnings.catch_warnings():
     warnings.filterwarnings("ignore", category=DeprecationWarning)
     import skgstat as skg
     from skgstat import models
 
+def nd_binning_scipy(dh: np.ndarray, list_var: List[np.ndarray], list_var_names=List[str], list_var_bins: Optional[List[Iterable]] = None,
```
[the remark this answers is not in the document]
rhugonnet: Forgot to change! (I tried another one based on pandas at first)
Updated

xdem/spstats.py
```diff
@@ -9,19 +9,123 @@
...
+                     statistics: list[Union[str, Callable, None]] = ['count', np.nanmedian ,nmad], list_ranges : Optional[List[Sequence]] = None) \
+        -> tuple[list[pd.DataFrame],list[pd.DataFrame],pd.DataFrame]:
+    """
+    Quantify non-stationarities in elevation measurement errors with one or several, terrain or instrument-related,
```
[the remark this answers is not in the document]
rhugonnet: Very true :+1: 
Updated the doc, changed `dh` in `values` (to match the scipy function).

Yes you did! (I think)

xdem/spstats.py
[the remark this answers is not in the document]
rhugonnet: Added the option of providing an integer for list_var_bins :)
I pushed the update (based on Amaury's comments) for improving the dimension/description of the variable size during your review :grimacing:

xdem/spstats.py
rhugonnet: This is tricky memory-wise, as binning in parallel for different combinations/statistics requires duplication of the data.

tests/test_spstats.py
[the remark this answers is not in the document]
erikmannerfelt: Cool!

tests/test_spstats.py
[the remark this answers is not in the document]
erikmannerfelt: Maybe it's worth either running `np.ravel(array)` on the input or raising a `ValueError` if the wrong shape is given then?

tests/test_spstats.py
rhugonnet: scipy raises an error indeed: `ValueError: invalid statistic blasjdhakwjhd`
This is checked within their tests, I don't think we need to add something ourselves when it's simply passing a parameter directly? Otherwise the tests would become very long.

tests/test_spstats.py
```diff
@@ -171,3 +170,49 @@ def test_patches_method(self):
...
             area_size=10000
         )
+
+class TestBinning:
+
+    def test_nd_binning(self):
+
+        ref, diff, mask = load_ref_and_diff()
+
+        slope, aspect = xdem.coreg.calculate_slope_and_aspect(ref.data.squeeze())
```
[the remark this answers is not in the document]
rhugonnet: > I now had a thorough look and this looks super useful! Great job!!
> 
> I have a few comments, mostly not about your intended use, and I envision many to be able to use it for completely different purposes. Perhaps this could be made more generic to be a multi-purpose function (basically just rename the dh argument to something generic, modify the docstring a bit, and it's done!)?
> 
> Also, now that I've read the entire documentation and the tests, I more or less understand the functionality. I can unfortunately not say that I understand it completely, and I worry that this might be reflective of other users' experience as well. Could you please elaborate a bit on the docstrings, perhaps add a few minimal examples, and optimally also add documentation to readthedocs (this can of course be done in a later stage)?

Thanks a lot for the comments!! I was updating the function + parameters description with Amaury's comments when you reviewed, so hopefully it's already a lot better :)

rhugonnet:
> I now had a thorough look and this looks super useful! Great job!!
> 
> I have a few comments, mostly not about your intended use, and I envision many to be able to use it for completely different purposes. Perhaps this could be made more generic to be a multi-purpose function (basically just rename the dh argument to something generic, modify the docstring a bit, and it's done!)?
> 
> Also, now that I've read the entire documentation and the tests, I more or less understand the functionality. I can unfortunately not say that I understand it completely, and I worry that this might be reflective of other users' experience as well. Could you please elaborate a bit on the docstrings, perhaps add a few minimal examples, and optimally also add documentation to readthedocs (this can of course be done in a later stage)?

Thanks a lot for the comments!! I was updating the function + parameters description with Amaury's comments when you reviewed, so hopefully it's already a lot better :)

tests/test_spstats.py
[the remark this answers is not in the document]
rhugonnet: Added the first in 94bcfc9
The second will be done by scipy if that's the case, no need to duplicate!

rhugonnet:
@erikmannerfelt @adehecq I think this is ready to merge, so that we can build upon it :)

erikmannerfelt (approved):
Great job!!

Weird with the type annotations (I saw how it failed in CI and I can't explain why...), but let's just fix that later, whatever it was.

PR #151, opened 2021-06-22
Add robust polynomial, sum of sinusoids fitting

### LINKED ISSUE #50: Add polynomial fit, sinusoidal fit, along/across-track sampling
We need:
- Sampling along across-track,
- Sampling along along-track,
- Polynomial fit based on minimized RMSE for multiple degrees,
- Sinusoidal fit based on priors and basin-hopping optimization

All available here:  https://github.com/luc-girod/MMASTER-workflows/blob/master/pymmaster/mmaster_tools.py

rhugonnet:
# Polynomial fitting + Sum of sin fitting + Across/Along-track sampling

Resolves #50
 
```python
coefs, deg = xdem.spatial_tools.robust_polynomial_fit(x, y, estimator='Theil-Sen')
```

### 2 robust polynomial fitting solutions
1/ One combining `sklearn.linear_model` and `sklearn.preprocessing.PolynomialFeatures`: solves a polynomial with robust estimators **Linear Regression**, **Theil-Sen** (median approach), **RANSAC** or **Huber**.
2/ Using `scipy.optimize.least_squares` and specific loss function, some are quite robust to outliers (see example in tests).

### Both implemented in the same function
**Linear** can be solved with `scipy` or `sklearn`. While **Theil-Sen**, **RANSAC** and **Huber** only with `sklearn`.

### Input/Output
Simple input: `x`, `y` as input, choice of `estimator`, cost function `cost_func` (by default, median absolute error), and a few other options described in the docs.
Simple output: polynomial degree as integer, and coefficients in a vector.

### Choosing the best polynomial
Simply using the polynomial that has the lowest cost (less spread between true and predicted values) is known to not be a good approach for choosing the optimal degree, as it can lead to overfitting. Here I wrote a simple function that selects the polynomial of smallest degree within a percentage margin of the "best cost" found by the fit.

So, for instance, if degree 1 fit has a cost of **100**, degree 2 fit a cost of **20**, degree 3 fit a cost of **5** and then degrees 4 to 6 fits have a cost **between 4 and 5**, the function (with a margin of 20% by default) will select degree 3 as the optimal solution.

EDIT:

### TODOLIST to finalize PR

- [x] Fix tests
- [x] Re-organize estimator subfunction wrappers to be more generic
- [x] Refactor `poly` and `sumofsin` functions to call new wrapper functions
- [x] Add `kwargs` argument that can be identified to any subfunction call (same logic as in `spatialstats.py`)
- [x] Fix circular import
- [x] Move functions to new module `fit.py` following #157


xdem/spatial_tools.py
```diff
@@ -396,3 +402,147 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+        raise ValueError('Attribute estimator must be one of "Linear", "Theil-Sen", "RANSAC" or "Huber".')
+    if not isinstance(linear_pkg, str) or linear_pkg not in ['sklearn','scipy']:
+        raise ValueError('Attribute linear_pkg must be one of "scipy" or "sklearn".')
+
+    # select sklearn estimator
+    dict_estimators = {'Linear': LinearRegression(), 'Theil-Sen':TheilSenRegressor(random_state=random_state)
+        , 'RANSAC': RANSACRegressor(random_state=random_state), 'Huber': HuberRegressor()}
+    est = dict_estimators[estimator]
+
+    # TODO: this should be a function outside (waiting for Amaury's PR)
```
rhugonnet, line 493: this will be replaced by Amaury's subsample function once it's merged

erikmannerfelt:
Nice! I presume this is the first step toward your bias corrections?

Is scikit-learn already installed due to some other dependency? It is not a direct dependency in the environment file.

rhugonnet:
> Nice! I presume this is the first step toward your bias corrections?
> 
> Is scikit-learn already installed due to some other dependency? It is not a direct dependency in the environment file.

Yes, first towards a series of `BiasCorr` classes. I thought it would make more sense to partition it, especially as those 1D fitting functions can be used for other applications!

Ups! Didn't check, it was in my local xdem environment for some reason.
I guess that then the main question is: should we have scikit-learn as a dependency @adehecq @erikmannerfelt ? 
It's becoming very popular (linked to Dask for example), and it has a lot of great statistic tools (including Gaussian Processes) so I'm clearly for it :smile:

adehecq:
Looks very promising!! :-)

Regarding the dependency on scikit-learn (and of other packages in general), I believe that unless not importing the module makes xdem useless, e.g. rasterio, we should not make it a hard dependency. So basically, it should only be imported within the functions where it is needed.
We could have two requirement files, one with only the hard dependencies, and one with all, so that it's easier to install if one wants to use all the xdem functionalities.

rhugonnet:
For now, I have added an `_has_sklearn` flag triggered by the sklearn import failure in 
78e2ef7, and checked in the function, as @erikmannerfelt did for `csv2`, `richDEM`.
Is this what you meant by "importing within the function", @adehecq ? I'm not sure of the best practice here.

Should we open an issue (improvement) for creating a file for "full environment" and one for "minimal environment"?

tests/test_spatial_tools.py
```diff
@@ -2,6 +2,7 @@
 
 Author(s):
     Erik S. Holmlund
```
erikmannerfelt, line 4: @adehecq  and I discussed that we should probably remove the author part of each script, as many others may contribute in the future. For figuring out who did what, there's always `git blame` ;)

xdem/spatial_tools.py
```diff
@@ -8,10 +8,20 @@
...
 from tqdm import tqdm
 
+try:
+    from sklearn.metrics import mean_squared_error, median_absolute_error
+    from sklearn.linear_model import (
+        LinearRegression, TheilSenRegressor, RANSACRegressor, HuberRegressor)
+    from sklearn.pipeline import make_pipeline
+    from sklearn.preprocessing import PolynomialFeatures
+    _has_sklearn = True
+except ImportError:
```
erikmannerfelt, line 23: Maybe too pedantic (you can ignore if you want) but this will be caught if a sub-dependency of sklearn is missing but sklearn exists. This should not happen with modern package managers, but it just caught my attention. Technically, one should also match the error message.

xdem/spatial_tools.py
```diff
@@ -398,6 +408,149 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
     # The output is scaled by "(x + 0.6) / 1.84" to make it more similar to GDAL.
     return np.clip(255 * (shaded + 0.6) / 1.84, 0, 255).astype("float32")
 
+def get_xy_rotated(raster: gu.georaster.Raster, myang: float) -> tuple[np.ndarray, np.ndarray]:
+    """
+    Rotate x, y axes of image to get along- and cross-track distances.
```
erikmannerfelt, line 413: This returns pixel coordinates rotated around the lower left corner, right? Where do the "cross-track distances" come in?

xdem/spatial_tools.py
```diff
@@ -398,6 +408,149 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+    :param myang: angle by which to rotate axes (degrees)
+    :returns xxr, yyr: arrays corresponding to along (x) and cross (y) track distances.
+    """
+    # creates matrices for along and across track distances from a reference dem and a raster angle map (in radians)
+
+    myang = np.deg2rad(myang)
+
+    # get grid coordinates
+    xx, yy = raster.coords(grid=True)
+    xx = xx - np.min(xx)
```
erikmannerfelt, line 424: `xx -= xx.min()`, or does this do something more specific?

xdem/spatial_tools.py
```diff
@@ -398,6 +408,149 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
     # The output is scaled by "(x + 0.6) / 1.84" to make it more similar to GDAL.
     return np.clip(255 * (shaded + 0.6) / 1.84, 0, 255).astype("float32")
 
+def get_xy_rotated(raster: gu.georaster.Raster, myang: float) -> tuple[np.ndarray, np.ndarray]:
+    """
+    Rotate x, y axes of image to get along- and cross-track distances.
+    :param raster: raster to get x,y positions from.
+    :param myang: angle by which to rotate axes (degrees)
```
erikmannerfelt, line 415: Why did you choose `myang` instead of something more usual, like `angle`?

xdem/spatial_tools.py
erikmannerfelt: It could also forego the raster class (if we want) by using `xdem.coreg._get_x_and_y_coords` on a transform and the shape of the array. I don't know if that is necessary or better; just a suggestion.

xdem/spatial_tools.py
```diff
@@ -398,6 +408,149 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+    xxr = np.multiply(xx, np.cos(myang)) + np.multiply(-1 * yy, np.sin(myang))
+    yyr = np.multiply(xx, np.sin(myang)) + np.multiply(yy, np.cos(myang))
+
+    # re-initialize coordinate at zero
+    xxr = xxr - np.nanmin(xxr)
+    yyr = yyr - np.nanmin(yyr)
+
+    return xxr, yyr
+
+def choice_best_polynomial(cost: np.ndarray, margin_improvement : float = 20., verbose: bool = False) -> int:
```
erikmannerfelt, line 437: The use-case of this is very specific to `robust_polynomial_fit()`, right? If so, perhaps it should be prefixed with an underscore, as it may not be useful for the public interface.

xdem/spatial_tools.py
```diff
@@ -398,6 +408,149 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+        raise ValueError('Attribute linear_pkg must be one of "scipy" or "sklearn".')
+
+    # select sklearn estimator
+    dict_estimators = {'Linear': LinearRegression(), 'Theil-Sen':TheilSenRegressor(random_state=random_state)
+        , 'RANSAC': RANSACRegressor(random_state=random_state), 'Huber': HuberRegressor()}
+    est = dict_estimators[estimator]
+
+    # TODO: this should be a function outside (waiting for Amaury's PR)
+    # remove NaNs
+    mykeep = np.logical_and.reduce((np.isfinite(y), np.isfinite(x)))
```
erikmannerfelt, line 500: What does `np.logical_and.reduce` do? I've never seen it before!

xdem/spatial_tools.py
```diff
@@ -398,6 +408,149 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+    x = x[mykeep]
+    y = y[mykeep]
+
+    # subsample
+    index_sub = subsample_raster(x, 25000, return_indices=True)
```
erikmannerfelt, line 505: Is 25000 always a good value or should this be either dependent on the input size (`int(x.size * 0.1)` or whatever) or an argument?

xdem/spatial_tools.py
```diff
@@ -398,6 +408,149 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+            mycost[deg - 1] = myresults.cost
+            coeffs[deg - 1, 0:myresults.x.size] = myresults.x
+        # otherwise, it's from sklearn
+        else:
+            if not _has_sklearn:
+                raise ValueError("Optional dependency needed. Install 'scikit-learn'")
+            p = PolynomialFeatures(degree=deg)
+            model = make_pipeline(p, est)
+            model.fit(x.reshape(-1,1), y)
+            mad = cost_func(model.predict(x.reshape(-1,1)), y)
```
erikmannerfelt, line 538: This is not necessarily MAD, right? Perhaps it could be called `error`?

xdem/spatial_tools.py
```diff
@@ -398,6 +408,149 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+            elif estimator == 'RANSAC':
+                c = est.estimator_.coef_
+            coeffs[deg - 1, 0:deg+1] = c
+
+    # selecting the minimum (not robust)
+    # fidx = mycost.argmin()
+    # choosing the best polynomial with a margin of improvement on the cost
+    fidx = choice_best_polynomial(cost=mycost, margin_improvement=margin_improvement, verbose=verbose)
+
+    return np.trim_zeros(coeffs[fidx], 'b'), fidx+1
```
erikmannerfelt, line 553: Why is it `fixd + 1` and not `fixd`? Also, what does the "fix" part mean of the variable name?

adehecq:
> For now, I have added an `_has_sklearn` flag triggered by the sklearn import failure in
> [78e2ef7](https://github.com/GlacioHack/xdem/commit/78e2ef7be2d862ae8c017fb41fef2fc1b0129b95), and checked in the function, as @erikmannerfelt did for `csv2`, `richDEM`.
> Is this what you meant by "importing within the function", @adehecq ? I'm not sure of the best practice here.
> 
> Should we open an issue (improvement) for creating a file for "full environment" and one for "minimal environment"?

My idea was to have the import statement within the function directly. I was discussing with Fabien about it and there are pro/cons to each approach:
- with the current approach, if you don't catch and raise a specific error (as you did perfectly here!) then the error becomes difficult to trace for the user. So we need to make sure that the error is caught everywhere in the code.
- with the import within the function, it's more transparent and an ImportError is raised when calling the function, which is then easy to understand. But if using the function in multiple processes, then the module will be loaded each time...
Any opinion on this @erikmannerfelt ?

tests/test_spatial_tools.py
```diff
@@ -178,6 +180,64 @@ def make_gdal_hillshade(filepath) -> np.ndarray:
...
+        # and some outliers
+        y[50:75] = 0
+        y[900:925] = 1000
+
+        # test linear estimators
+        coefs, deg = xdem.spatial_tools.robust_polynomial_fit(x,y, estimator='Linear', linear_pkg='scipy', loss='soft_l1', f_scale=0.5)
+
+        # scipy solution should be quite robust to outliers/noise (with the soft_l1 method and f_scale parameter)
+        # however, it is subject to random processes inside the scipy function (couldn't find how to fix those...)
+        assert deg == 3 or deg == 4 # can find degree 3, or 4 with coefficient close to 0
```
erikmannerfelt, line 205: Here and below, the approximately same calculation is done with different estimators. Could this be simplified using `pytest.mark.parametrize`?

erikmannerfelt (commented):
Nice functionality! I have some quite small remarks but generally I like it!

xdem/spatial_tools.py
adehecq: Yes, I thought about the same thing (actually Fabien did!). I was quickly browsing to see if there's a good practice on how to handle optional imports. The only improvement I could find is to use importlib to check if a module exists or not: https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder.find_spec

adehecq: So you could first check that sklearn was found. If yes, then import the submodules, if not, set _has_sklearn to False.
This could be made consistent in all the code.

erikmannerfelt:
> > For now, I have added an `_has_sklearn` flag triggered by the sklearn import failure in
> > [78e2ef7](https://github.com/GlacioHack/xdem/commit/78e2ef7be2d862ae8c017fb41fef2fc1b0129b95), and checked in the function, as @erikmannerfelt did for `csv2`, `richDEM`.
> > Is this what you meant by "importing within the function", @adehecq ? I'm not sure of the best practice here.
> > Should we open an issue (improvement) for creating a file for "full environment" and one for "minimal environment"?
> 
> My idea was to have the import statement within the function directly. I was discussing with Fabien about it and there are pro/cons to each approach:
> 
>     * with the current approach, if you don't catch and raise a specific error (as you did perfectly here!) then the error becomes difficult to trace for the user. So we need to make sure that the error is caught everywhere in the code.
> 
>     * with the import within the function, it's more transparent and an ImportError is raised when calling the function, which is then easy to understand. But if using the function in multiple processes, then the module will be loaded each time...
>       Any opinion on this @erikmannerfelt ?

I think generally it is advised to import at the top-level because running the function multiple times would otherwise run the imports multiple times. This is not very slow, but it's also not instantaneous. [See StackOverflow for reference](https://stackoverflow.com/questions/128478/should-import-statements-always-be-at-the-top-of-a-module). I think both approaches are fine; as you mention, there are pros and cons to both. I think we should just be consistent, and for now, we have the `_has_blabla` syntax in other places.

rhugonnet:
> My two cents are that `environment.yml` should contain all optional dependencies as well. Most people (that are not developers) will either use pip, whereby the dependencies/optionals are parsed from `setup.py`, or from conda-forge, whereby it will be read from the conda recipe (which only specifies the mandatory dependencies, thus skipping the optionals). Thus, `environment.yml` has a weird niche where it is mostly only used by people that either want to develop, those who want main instead of the latest release, or those that want to modify the code. In these cases, I think there's no real drawback to have them install optional dependencies as well.
> 
> Thoughts, @adehecq and @rhugonnet ?

Fully agree, `setup.py` will do the job for the mandatory ones (almost forgot about it)

adehecq:
> > For now, I have added an `_has_sklearn` flag triggered by the sklearn import failure in
> > [78e2ef7](https://github.com/GlacioHack/xdem/commit/78e2ef7be2d862ae8c017fb41fef2fc1b0129b95), and checked in the function, as @erikmannerfelt did for `csv2`, `richDEM`.
> > Is this what you meant by "importing within the function", @adehecq ? I'm not sure of the best practice here.
> > Should we open an issue (improvement) for creating a file for "full environment" and one for "minimal environment"?
> 
> My two cents are that `environment.yml` should contain all optional dependencies as well. Most people (that are not developers) will either use pip, whereby the dependencies/optionals are parsed from `setup.py`, or from conda-forge, whereby it will be read from the conda recipe (which only specifies the mandatory dependencies, thus skipping the optionals). Thus, `environment.yml` has a weird niche where it is mostly only used by people that either want to develop, those who want main instead of the latest release, or those that want to modify the code. In these cases, I think there's no real drawback to have them install optional dependencies as well.
> 
> Thoughts, @adehecq and @rhugonnet ?

Just to clarify, where is the information on the dependencies for conda-forge stored? I thought it was in the `environment.yml`. If not, then I agree that this file is a good place to put all optional requirements. But in that case, we could probably merge it with the `dev-environment.yml`, if it's only to be used by people developing/testing the code.
And what is the point of the `dev-requirements.txt`?
In all cases, it would be great to have a place where we explain where dependencies should be added. For now there are at least 3-4 files where this should be updated no?

xdem/spatial_tools.py
rhugonnet: Done in 20e8e5e

tests/test_spatial_tools.py
rhugonnet: Removed authors!

rhugonnet: Or wait, I'll leave that for a separate PR

rhugonnet: Opened #153

xdem/spatial_tools.py
rhugonnet: Opened #154

xdem/spatial_tools.py
rhugonnet: Remnants from a previous version of the code

xdem/spatial_tools.py
rhugonnet: Again, remnants from a previous version of the code

xdem/spatial_tools.py
rhugonnet: Good idea! Added

xdem/spatial_tools.py
rhugonnet: Here it is useless (again, remnant of a previous version of the code).

Basically, it allows you to concatenate as many conditions as you desire:
`np.logical_and.reduce((cond1,cond2,cond3))` is equivalent to `np.logical_and(np.logical_and(cond1,cond2),cond2)` 
Quite useful :wink:

xdem/spatial_tools.py
rhugonnet: Added as argument

xdem/spatial_tools.py
rhugonnet: Indeed, changed into `cost` for consistency with the rest

xdem/spatial_tools.py
rhugonnet: Added a comment for clarification, and renamed into `final_index`

tests/test_spatial_tools.py
rhugonnet: Unfortunately the assert tests are all different, because it does not perform so well, so I think it will get harder to read with `pytest.mark.parametrize`.
But I should add a test case without outliers (where they all perform well) using `pytest.mark.parametrize` :)

xdem/spatial_tools.py
erikmannerfelt: Aah, nice! Thanks!!

tests/test_spatial_tools.py
rhugonnet: Added in 86d565b

tests/test_spatial_tools.py
```diff
@@ -182,7 +210,27 @@ def make_gdal_hillshade(filepath) -> np.ndarray:
...
+        x = np.linspace(1, 10, 1000)
+        # exact polynomial
+        true_coefs = [-100, 5, 3, 2]
+        y = true_coefs[0] + true_coefs[1] * x + true_coefs[2] * x ** 2 + true_coefs[3] * x ** 3
+
+        coefs, deg = xdem.spatial_tools.robust_polynomial_fit(x, y, linear_pkg=pkg_estimator[0], estimator=pkg_estimator[1], random_state=42)
+
+        assert deg == 3 or deg == 4
+        assert np.abs(coefs[0] - true_coefs[0]) <= 100
+        assert np.abs(coefs[1] - true_coefs[1]) < 5
```
erikmannerfelt, line 229: Could also do `assert coefs[0] == pytest.approx(true_coeffs[0], abs=100)` 

Don't know if it's actually more readable now that I wrote it...

erikmannerfelt (dismissed):
Nice! I just have one completely optional suggestion.

tests/test_spatial_tools.py
```diff
@@ -178,6 +208,136 @@ def make_gdal_hillshade(filepath) -> np.ndarray:
...
+                                           ('sklearn','RANSAC'),('sklearn','Huber')])
+    def test_robust_polynomial_fit(self, pkg_estimator: str) -> None:
+
+        np.random.seed(42)
+
+        # x vector
+        x = np.linspace(1, 10, 1000)
+        # exact polynomial
+        true_coefs = [-100, 5, 3, 2]
+        y = true_coefs[0] + true_coefs[1] * x + true_coefs[2] * x ** 2 + true_coefs[3] * x ** 3
```
adehecq, line 223: You could even use [np.polyval](https://numpy.org/doc/stable/reference/generated/numpy.polyval.html) for that

tests/test_spatial_tools.py
```diff
@@ -178,6 +208,136 @@ def make_gdal_hillshade(filepath) -> np.ndarray:
...
+
+        coefs, deg = xdem.spatial_tools.robust_polynomial_fit(x, y, linear_pkg=pkg_estimator[0], estimator=pkg_estimator[1], random_state=42)
+
+        assert deg == 3 or deg == 4
+        assert np.abs(coefs[0] - true_coefs[0]) <= 100
```
adehecq, line 228: Why do you allow such a large error? Should it not be much closer to the actual value than that??

tests/test_spatial_tools.py
```diff
@@ -178,6 +208,136 @@ def make_gdal_hillshade(filepath) -> np.ndarray:
...
+        assert np.abs(coefs[2] - true_coefs[2]) < 1
+        assert np.abs(coefs[3] - true_coefs[3]) < 1
+
+        # the sklearn Linear solution with MSE cost function will not be robust
+        coefs2, deg2 = xdem.spatial_tools.robust_polynomial_fit(x,y, estimator='Linear', linear_pkg='sklearn', cost_func=mean_squared_error, margin_improvement=50)
+        assert deg2 != 3
+        # using the median absolute error should improve the fit, but the parameters will still be hard to constrain
+        coefs3, deg3 = xdem.spatial_tools.robust_polynomial_fit(x,y, estimator='Linear', linear_pkg='sklearn', cost_func=median_absolute_error, margin_improvement=50)
+        assert deg3 == 3
+        assert np.abs(coefs3[0] - true_coefs[0]) > 50
```
adehecq, line 265: It's a bit weird to test that a method will be "at least as bad as..." !
Or did you mean the comparison sign to be the other way around?

xdem/spatial_tools.py
```diff
@@ -398,6 +509,336 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+
+    return xxr, yyr
+
+def rmse(z: np.ndarray) -> float:
+    """
+    Return root mean square error
+    :param z: Residuals between predicted and true value
+    :return: Root Mean Square Error
+    """
+    return np.sqrt(np.nanmean(np.square(np.asarray(z))))
```
adehecq, line 546: I don't think you need the np.asarray. If you provide a list to a numpy function, it will do the conversion automatically I believe.

xdem/spatial_tools.py
```diff
@@ -398,6 +509,336 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+
+def huber_loss(z: np.ndarray) -> float:
+    """
+    Huber loss cost (reduces the weight of outliers)
+    :param z: Residuals between predicted and true values
+    :return: Huber cost
+    """
+    out = np.asarray(np.square(z) * 1.000)
```
adehecq, line 554: What is the point of multiplying by 1.000? If you need to change the dtype, then use self.astype().

adehecq: And same comment about the use of np.asarray. np.square will certainly output a numpy array.

xdem/spatial_tools.py
```diff
@@ -398,6 +509,336 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+    out[np.where(z > 1)] = 2 * np.sqrt(z[np.where(z > 1)]) - 1
```
adehecq, line 555: Could it not be just one line:
`out = np.where(z > 1, 2 * np.sqrt(z[np.where(z > 1)]) - 1, np.square(z))`
Would be clearer, because it's hard to understand the metric as is written.

xdem/spatial_tools.py
```diff
@@ -398,6 +509,336 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+    # selecting the minimum (not robust)
+    # final_index = mycost.argmin()
+    # choosing the best polynomial with a margin of improvement on the cost
+    final_index = _choice_best_order(cost=costs, margin_improvement=margin_improvement, verbose=verbose)
+
+    # the degree of the polynom correspond to the index plus one
+    return np.trim_zeros(coeffs[final_index], 'b'), final_index + 1
+
+
+def _fitfun_sumofsin(x: np.array, params: np.ndarray) -> np.ndarray:
```
adehecq, line 712: The name is not very adequate. You are not fitting a function, but calculating its values.
Maybe sumofsin_val? (to be somewhat similar to np.polyval)

xdem/spatial_tools.py
```diff
@@ -398,6 +509,336 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+    """
+    Function for a sum of N frequency sinusoids
+    :param x: array of coordinates (N,)
+    :param p: list of tuples with amplitude, frequency and phase parameters
+    """
+    aix = np.arange(0, params.size, 3)
+    bix = np.arange(1, params.size, 3)
+    cix = np.arange(2, params.size, 3)
+
+    val = np.sum(params[aix] * np.sin(np.divide(2 * np.pi, params[bix]) * x[:, np.newaxis] + params[cix]), axis=1)
```
adehecq, line 722: What's the benefit of np.divide over the normal / ?

xdem/spatial_tools.py
```diff
@@ -398,6 +509,336 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+
+    return val
+
+def robust_sumsin_fit(x: np.ndarray, y: np.ndarray, nb_frequency_max: int = 3,
```
adehecq, line 726: This makes me think. It's great the degree/number of sinusoids is calculated automatically. But if one wants a specific "degree", would that be possible?
An option I see is to not only have a max, but also a min "degree", here and for the polynomial fit. SO if min/max are equal, the degree is constrained.

xdem/spatial_tools.py
```diff
@@ -398,6 +509,336 @@ def hillshade(dem: Union[np.ndarray, np.ma.masked_array], resolution: Union[floa
...
+                      random_state: Optional[Union[int,np.random.Generator,np.random.RandomState]] = None, verbose: bool = False) -> tuple[np.ndarray,int]:
+    """
+    Given 1D data x, y, compute a robust sum of sinusoid fit to the data. The number of frequency is chosen
+    automatically by comparing residuals for multiple fit orders of a given estimator.
+    :param x: input x data (N,)
+    :param y: input y data (N,)
+    :param nb_frequency_max: maximum number of phases
+    :param bounds_amp_freq_phase: bounds for amplitude, frequency and phase (L, 3, 2) and
+    with mean value used for initialization
+    :param significant_res: significant resolution of X data to optimize algorithm search
```
adehecq, line 738: I don't understand this parameter.

adehecq: At least without looking at the code.

adehecq:
Looks great! This makes me think we could use your robust polynomial fit in coreg deramp.
I have some minor comments, but after that it's good for me.

adehecq:
@rhugonnet what's the status of this PR?

tests/test_spatial_tools.py
rhugonnet: It is clearer! Done in 5cfe2e7

tests/test_spatial_tools.py
rhugonnet: Done :+1:

tests/test_spatial_tools.py
rhugonnet: Slight variations in high-order coefficients result in good fits with very different low-order coefficients, hence the variability...
Don't know if there is a smartest way to check a polynomial accounting for this effect

tests/test_spatial_tools.py
rhugonnet: Well adding outliers mean the non-robust, "classic" estimators don't perform well. So yes, this is actually what I check :sweat_smile: 
I added some comments to make the meaning of the comparison clearer

xdem/spatial_tools.py
rhugonnet: Remains from previous code, removed

xdem/spatial_tools.py
rhugonnet: Corrected

xdem/spatial_tools.py
rhugonnet: Modified this way

xdem/spatial_tools.py
rhugonnet: Refactored into `sumofsinval` :)

xdem/spatial_tools.py
rhugonnet: None apparently, rewrote into / (not my code originally)

xdem/spatial_tools.py
rhugonnet: > @rhugonnet what's the status of this PR?

On it, trying to homogenize things and push a final version!

rhugonnet:
> @rhugonnet what's the status of this PR?

On it, trying to homogenize things and push a final version!

xdem/spatial_tools.py
rhugonnet: Added a better description!

rhugonnet:
Again, I have an assertion error that happens only in CI, while everything passes locally... and this is for a calculation with a fixed `random_state` 
So glad to be back :sob:

rhugonnet:
@adehecq @erikmannerfelt All ready for approval to be merged, except for that one test that seems to fail randomly in CI (while it never does locally, the `random_state` is well fixed and gives a constant result...).

rhugonnet:
I'm still desperately trying to understand just the base: why :pray:

rhugonnet:
> I'm still desperately trying to understand just the base: why

After a few hours, still can't trace it back. :cry: 
I'm giving up for now, let's just open an issue that I'll reference to this comment.
Description here:

`test_robust_sumsin_fit` randomly fails, in CI only, for the following assertion:
```python
        for i in range(2):
            assert coefs[3*i] == pytest.approx(true_coefs[3*i], abs=0.02)
```
The function call is based on `scipy.optimize.basinhopping`. I cannot reproduce the issue locally, even when setting up `xdem-dev` environment the same way as in CI.
This is very strange because the test uses a random state `random_state=42` that is used both by `xdem.spatial_tools.subsample_raster` and by `basinhopping` (seed argument). And locally the output is indeed not subject to any random variation (constant).
The error seems to occur 75% of the time in CI. When it does, the `scipy` version (1.7.1) and `numpy` version (1.20.3) are exactly the same as the ones I have locally without error after running 15+ times. So the issue doesn't seem to be from there either...
The mystery remains!

xdem/spatial_tools.py
adehecq: Yes, could be a nice idea to combine both with an optional rotation argument.

adehecq (approved):
Great job !
I haven't looked in detail at your latest changes, but if you took into account our comments, it should be alright.
It's annoying you couldn't find the issue behind the test randomly failing...

PR #158, opened 2021-06-26
Add `BiasCorr` classes and rename previous `coreg.BiasCorr` in `coreg.VerticalShift`

rhugonnet:
## Summary

This PR adds non-rigid co-registration "bias-correction" methods through the class `BiasCorr` and subclasses located in `biascorr.py`! :smile: 

In details, this PR:
- Refactors the old vertical shift method `coreg.BiasCorr` into `coreg.VerticalShift`. 
- Renames the old `Coreg` parent class into `Rigid`, as all of its methods (`to_matrix`, etc) and its subclasses are only relevant for rigid transformations,
- Creates a new "elder" parent class named `Coreg` that contains only the generic methods to any processing step (e.g., `residuals()`) and is subclassed by `Rigid` for rigid transforms, `BiasCorr` for non-rigid transforms, and `CoregPipeline` and  `BlockwiseCoreg` to be able to use those with any subclass of `Coreg` (can now do a pipeline with a `Coreg` method and a `BiasCorr` method, for example, and any other class we choose to add below `Coreg`),
- Rewrite the content of `Coreg.fit()` into a separate `_preprocess` function, and uses `subsample_raster` from GeoUtils to consistently subsample the inputs,
- Streamlines the optimization pipelines in `fit.py` to be consistent in the way they return optimized parameters and accept user `kwargs`,
- Adds a fully modular framework based on the choice "fit_or_bin" to run any `BiasCorr` class: the user can either pass any `fit_func` (e.g., `np.polyval`) and `fit_optimizer` (e.g., `scipy.optimize.curve_fit`) functions to determine a *parametric solution* to their bias, or any `bin_sizes` (e.g., `10`), `bin_statistic` (e.g., `np.nanmedian`) and `bin_apply_method` (e.g., "linear" interpolation) to determine an *empirical solution* to their bias, or **to do both (recommended option)**!
- Defines robust and case-specific default parameters of "fit_or_bin" for subclasses `Deramp`, `DirectionalBias` and `TerrainBias` to allow user to easily run the classes without customization,
- Fixes several bugs that were in `fit.py` and `coreg.py`,
- Adds `test_biascorr.py` with many tests using synthetic biases,
- Adds a basic documentation page (to be re-worked once `coreg.py` is updated, **see thoughts below**),
- Adds a basic gallery example (also to be re-worked once `coreg.py` is updated).

## TO-DO:

- [x] Rename `Deramp` into `Tilt` in `coreg.py`, and fix degree to `1` (pointing to `Deramp` for degree > 1 ?),
- [x] Add the option for users to do `bin_and_fit` to simplify big data problems,
- [x] Write documentation page for `BiasCorr` (to streamline later with updated `Coreg`?),
- [x] Add gallery examples for `BiasCorr` (to streamline later with updated `Coreg`?).

## Thoughts on `Coreg` moving forward

We should probably copy the same logic in `Rigid` classes: 
- Being able to pass any optimizer or binning parameters to resolve a minimization problem parametrically or empirically,
- Reflect these changes and other recent changes in the documentation!

But this should probably live in a separate PR! :sweat_smile:

## Old draft text below

> 
> Draft PR: will need PR #151 to be merged to become functional for testing
> 
> Mostly trying to organize things logically in `biascorr.py` right now. It's a bit complex! @erikmannerfelt @adehecq Tell me what you think
> 
> Reminder of how `biascorr.py` differs from `coreg.py:` integrates all transformations that are not "affine". You can pass any "bias_fun" function to fit the bias along the given dimension. Right now: `robust_polynomial` or `robust_sumsin` in 1D.
> 
> ### Structural draft: DirectionalBias and TerrainBias
> **DirectionalBias** can be used to correct in 1D along any angle, notably including along-track and across-track corrections.
> 
> **TerrainBias** can be used to correct in 1D along any variable, notably terrain (elevation, curvature). Because those are inter-connected, I thought it'd be good to have them all in the same module. But we could expand this to any kind of variable really.
> 
> But in practice, those are part of a same category of a **1D bias correction with an external variable**. It's the exact same operations, except that DirectionalBias is able to derive the binned direction from the np.ndarray directly (no need to pass it to the function) and could possibly find the along/across angle automatically from the shape of the data (would be an amazing functionality :exploding_head:), and TerrainBias can derive the terrain attribute of interest directly from the provided DEMs so we wouldn't need the extra lines to pass it to the function (except if the user wants it).
> 
> So... maybe we should make those subclasses of a `BiasCorr1D` itself subclass or `Coreg`? So that we don't copy/paste the input/outputs?
> 
> ### Previous BiasCorr -> VerticalShift and previous ZScaleCorr -> TerrainBias
> Refactored `BiasCorr` into `VerticalShift` in all functions and documentations, and also comments.
> Removed `ZScaleCorr` that'll be replaced by `TerrainBias` (or the like)
> 
> 
> ### EDIT 09/09/2021:
> 
> Regarding the structure, I'm thinking that this is the most logical approach:
> 
> - Bias1D:
> including "convenience classes" (that have methods/1D sampling pre-parametrized):
> -> DirectionalBias
> -> TerrainBias
> 
> - Bias2D:
> including convenience class:
> -> Deramp
> 
> - BiasND (for 3+ dimensions)
> 
> Each method can work with a different method of bias prediction:
> - Parametric:
> -> Polynomial fit (available for 1D, 2D)
> -> Sum of sinusoid fit (1D only)
> - Numeric approximation:
> -> Binning + bilinear interpolation (1D, 2D, ND)
> - (TO BE IMPLEMENTED) Non-parametric:
> -> Gaussian Process fit
> 
> 
> ### TODO to finalize PR:
> 
> - [x] Homogenize functions in #151 to facilitate coefficient retrieval/apply functions
> - [x] Structure in: `BiasCorr1D`, `BiasCorr2D`, `BiasCorrND`
> - [x] Create convenience subclasses `DirectionalBias`, `TerrainBias`, `Deramp`, `Tilt` (that last one in `coreg.py`)
> - [x] Add tests on the implementation


xdem/biascorr.py
```diff
@@ -0,0 +1,113 @@
...
+            print("Directional bias estimated")
+
+        self._meta['angle'] = angle
+        self._meta['degree'] = deg
+        self._meta["coefs"] = coefs
+
+    def _to_matrix_func(self) -> np.ndarray:
+        """Convert the bias to a transform matrix."""
+
+        raise ValueError(
```
erikmannerfelt, line 49: This is necessary to not raise a `NotImplementedError`. However, it would probably be better if there was a generic way to do this, no?

If you added this to line 747 in main:
```python
if self._is_affine == False:
     raise ValueError(f"Non-affine coreg class ({type(self)}) cannot be represented by transformation matrices.")
```

then this would work better for all non-affine coreg subclasses. The only thing one has to specify in the `__init__` is `self._is_affine = False`.

xdem/biascorr.py
```diff
@@ -0,0 +1,113 @@
...
+
+import xdem
+
+
+class DirectionalBias(xdem.coreg.Coreg):
+    """
+    For example for DEM along- or across-track bias correction.
+    """
+
+    def __init__(self, bias_func : Callable = xdem.robust_stats.robust_polynomial_fit):  # pylint: disable=super-init-not-called
```
erikmannerfelt, line 17: Callables need to be subscripted to signify their inputs and outputs, no? They need to take two ndarrays as inputs (plus kwargs) and will return an int and an ndarray, right? If so, it should be:
```python
Callable[[np.ndarray, np.ndarray], tuple[int, np.ndarray]]
```

erikmannerfelt: And if it accepts kwargs, there's no current great way of doing it. One has to do:
```python
Callable[..., tuple[int, np.ndarray]]
```
where the ellipsis signifies "any number of arguments"

xdem/biascorr.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        Instantiate an directional bias correction object.
+
+        :param bias_func: The function to fit the bias. Default: robust polynomial of degree 1 to 6.
+        """
+        super().__init__(meta={"bias_func": bias_func})
```
erikmannerfelt, line 23: After this, `self._is_affine` should be set to False to work optimally with all parts of `Coreg`!

xdem/biascorr.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, transform: Optional[rio.transform.Affine],
+                  weights: Optional[np.ndarray], angle: Optional[float] = None, verbose: bool = False,**kwargs):
```
erikmannerfelt, line 26: `**kwargs` does not exist for `Coreg.fit` which needs to be added. Feel free to do so! Additionally, it needs to be supplied to line 507 in main:
https://github.com/GlacioHack/xdem/blob/21e196ab721cfb1868359ee499e2677810a8e112/xdem/coreg.py#L507

xdem/biascorr.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if verbose:
+            print('Getting directional coordinates')
+
+        diff = ref_dem - tba_dem
+        x, _ = xdem.spatial_tools.get_rotated_xy(ref_dem,angle=angle)
+
+        if verbose:
+            print("Estimating directional bias correction with function "+ self.meta['bias_func'].__name__)
+        deg, coefs = self.meta["bias_func"](x,diff,**kwargs)
```
erikmannerfelt, line 37: Self dot underscore meta! Interesting that this doesn't fail in the tests.

xdem/biascorr.py
```diff
@@ -0,0 +1,113 @@
...
+
+    With elevation: often useful for nadir image DEM correction, where the focal length is slightly miscalculated.
+    With curvature: often useful for a difference of DEMs with different effective resolution.
+
+    DISCLAIMER: An elevation correction may introduce error when correcting non-photogrammetric biases, as generally
+    elevation biases are interlinked with curvature biases.
+    See Gardelle et al. (2012) (Figure 2), http://dx.doi.org/10.3189/2012jog11j175, for curvature-related biases.
+    """
+
+    def __init__(self, bias_func = xdem.robust_stats.robust_polynomial_fit):
```
erikmannerfelt, line 65: Missing type hint. See the above suggestions.

xdem/biascorr.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        Instantiate an terrain bias correction object
+
+        :param bias_func: The function to fit the bias. Default: robust polynomial of degree 1 to 6.
+        """
+        super().__init__(meta={"bias_func": bias_func})
```
erikmannerfelt, line 71: Add `self._is_affine = False`

xdem/biascorr.py
```diff
@@ -0,0 +1,113 @@
...
+    def _fit_func(self, ref_dem: np.ndarray, tba_dem: np.ndarray, attribute: np.ndarray,
+                  transform: Optional[rio.transform.Affine], weights: Optional[np.ndarray], verbose: bool = False,
+                  **kwargs):
+        """Estimate the bias using the bias_func."""
+
+        diff = ref_dem - tba_dem
+
+        if verbose:
+            print("Estimating terrain bias correction with function " + self.meta['bias_func'].__name__)
+        deg, coefs = self.meta["bias_func"](attribute, diff, **kwargs)
```
erikmannerfelt, line 82: Again, `self.meta` -> `self._meta`

xdem/coreg.py
```diff
@@ -756,43 +756,43 @@ def _apply_pts_func(self, coords: np.ndarray) -> np.ndarray:
         raise NotImplementedError("This should have been implemented by subclassing")
 
 
-class BiasCorr(Coreg):
+class VerticalShift(Coreg):
```
erikmannerfelt, line 759: Should this not be moved to `biascorr.py`?

tests/test_coreg.py
```diff
@@ -218,20 +218,20 @@ def test_deramping(self):
         # Check that the mean periglacial offset is low
         assert np.abs(np.mean(periglacial_offset)) < 1
 
-        # Try a 0 degree deramp (basically bias correction)
+        # Try a 0 degree deramp (basically vertical shift correction)
         deramp0 = coreg.Deramp(degree=0)
```
erikmannerfelt, line 222: Deramp should maybe also be in `biascorr.py`?

docs/source/coregistration.rst
```diff
@@ -106,16 +106,16 @@ Example
         :lines: 44-50
 
 
-Bias correction
+Vertical shift
 ^^^^^^^^^^^^^^^
```
erikmannerfelt, line 110: Remove one `^` to be equally long with the title.

erikmannerfelt (commented):
Very nice!! I think some existing Coreg classes should be in `biascorr.py` as well. What do you think, @adehecq?

erikmannerfelt:
Oh, and there's no `tests/test_biascorr.py`. Should this not be added?

tests/test_coreg.py
```diff
@@ -73,84 +73,84 @@ def test_from_classmethods(self):
...
-    def test_bias(self):
+    def test_vertical_shift(self):
         warnings.simplefilter("error")
 
-        # Create a bias correction instance
-        biascorr = coreg.BiasCorr()
-        # Fit the bias model to the data
-        biascorr.fit(**self.fit_params)
+        # Create a vertical shift correction instance
+        vshiftcorr = coreg.VerticalShift()
```
adehecq, line 80: Technically, this should have been renamed vshift and not vshiftcorr, which sounds a bit redundant. But that's detail...

adehecq: Oh no, vshift is already used for the shift value.

xdem/coreg.py
adehecq: On a logical sense, it should yes. On the other hand, I wonder if people won't get lost in having two different modules that deal with relatively similar issues. The along/across-track corrections have a very different use case, but a vertical shift or deramping is most often used during coregistration...
I don't have a strict opinion on this, just bringing up the issue!

adehecq:
Seems to be in good shape! I guess you're still waiting to merge #151 before moving further?

tests/test_coreg.py
rhugonnet: It's delicate, as degree 1 will be affine, but higher degree won't be.
Should we split those somehow?
We could have a `Tilt` class in `coreg.py` (i.e. Deramp with degree=1) and a `Deramp` class in `biascorr.py` which, if called with a degree=1, just passes to `Tilt`.
@adehecq @erikmannerfelt  What do you think?

xdem/biascorr.py
rhugonnet: Thanks a lot for the details, added!

xdem/biascorr.py
rhugonnet: Added

xdem/biascorr.py
rhugonnet: That's because there is no test on this function yet ;)

xdem/biascorr.py
rhugonnet: Zgreed. All of this is added!

xdem/coreg.py
rhugonnet: A `VerticalShift` is an affine transformation, so we should keep it in `coreg.py`... no?

tests/test_coreg.py
adehecq: I think it would be confusing to have both Tild and Deramp (plus deramp has basically the same meaning as tilt, as it involves a plane surface).
I'm happy to have Deramp in biascorr, if it can be called the exact same way as the other coreg functions, and we clearly state the difference between the two categories somewhere and summarize all possible options.

rhugonnet:
@adehecq @erikmannerfelt 
Two years later... this PR is finally ready!! :sweat_smile: 

It was a tough one but I'm quite happy with the result, which keeps all of @erikmannerfelt's original `Coreg` implementation, moves up the core `Coreg` class (and replaces it, at the position of the original one, into a `Rigid` class for rigid transforms), which allows to combine any `Coreg` subclass such as `BiasCorr`, `Rigid` (and potentially soon `Filter`) in a pipeline or a block-wise manner!

**Moreover, solving a method is now fully modular!** We can perform a "bin" estimate, a "fit" estimate or a "bin + fit" estimate, passing any function/optimizer and any binning parameters.
More details in the new PR description above. Also a temporary documentation page here: https://xdem-rhugonnet.readthedocs.io/en/biascorr/biascorr.html (I didn't polish it because we'll likely have to update it soon, if we wish to incorporate the same changes to `Coreg`).
For the modularity passing a fit function and optimizer, everything is done consistently with `scipy.optimize.curve_fit`, so it will be easy to document and give examples! :smile: 

Other than the technical details, I also identified three important points to discuss:

1. **Class structure:** What do you think of the current class structure (see inheritance diagram on the doc page, link above)? In practice, the structure adds a lot of consistency in applying correction/transforms and combining methods, but we might have to make some classes non-public for user simplicity (`Coreg`, `Rigid` or `BiasCorr` for example, although technically some can be used directly). Or, alternatively, make "generic" classes very well documented.

2. **Naming**: Right now a lot of names are OK to understand for us, but probably not the wider community. The `BiasCorr` parent class could also be `Nonrigid`, the `Coreg` class could become quickly a bit too broad (if we add a `Filter` class, for example)...

3. **Relevance of fit/apply:** Right now, we use `fit()` and `apply()` every time. However, many users will simply want to re-use the same data and do some kind of `fit_and_apply()`. On top of that, in the case of certain `BiasCorr` classes, using `apply` is always going to be irrelevant (for example for `Deramp` or `DirectionalBias`, where the coordinates will be very different for two DEMs). The same can be said for a `Filter`, that only really has an `apply()` once user-parameter are passed. So what should we do moving forward? Maybe introduce a method that runs all the step for any method, such as `fit_and_apply()` or simply `run()`?

I also have some ideas moving forward for plotting tools! Thanks to the consistent "bin" or "fit" methods and associated parameters (how the fit function handles arguments, for instance), we can fairly easily introduce generic plotting tools that will work for any scenario :blush: ! (just a generic `plot()` in `Coreg`, or maybe in `BiasCorr` and `Rigid`) 

Curious to hear what you think! :grin:

rhugonnet:
Upon some more thinking, I think keeping `BiasCorr` could be fine, because non-rigid registration methods typically involve a certain type of deformation. For instance, Coherent Point Drift involves bending the points following "Motion Coherence Theory", which is very specific (and not necessarily relevant for elevation applications).

rhugonnet:
Merged with #346

erikmannerfelt:
Hi @rhugonnet,

This is really the right way forward and I'm grateful that you've spent so much time improving this! While I do have questions on the current setup, I think splitting the Coreg class into rigid/nonrigid methods may be very useful.

1. Could the name `Rigid` be `RigidCoreg` or similar? Calling it only `Rigid` to me sounds like it's just a transform, but it's in reality a co-registration class which estimates a rigid transform.
2. `Tilt` does not result in a rigid transform, last I checked. Since there's no horizontal transformation, applying it is akin to rotating and scaling at the same time. I can elaborate if that was confusing but we've already discussed this at length in one of the PRs. I think a name like `Deramp` (like it was) or `Detrend` is more suitable, as it's not just a rotation.
3. I think `RigidCoreg` and `NonRigidCoreg` would make more sense instead of `Rigid` and `BiasCorr`. I think it would make it easier to understand how they work for new people getting in to the code. Technically, a rigid transformation corrects a 3D bias, right? So technically, all `Coreg` classes correct biases, no?
4. If `Filter` will be subclassed from `Coreg` in the future, should `Coreg` change name? Something like just `Corr`, `Correction`, or `Filter` actually, would make more sense, and rigid /non-rigid coreg is subclassed from it. I personally don't think it makes sense to subclass a filter from a co-registration object, but just changing the name of the latter would solve that confusion. Bonus idea, the `Corr` or whatever class could be in its separate module to make the structure even simpler and to reduce on the already very lengthy `coreg.py` module.
5. Small thing but `coreg.CoregDict` is hanging in the air in the inheritance diagram. 
6. `GradientDescending` is missing from the inheritance diagram!


Adding a `fit_and_apply()` is a great idea I think. It's similar to how OpenCV's feature detection algorithms inherit, `detect()`, `compute()` and `detectAndCompute()` from `cv::Feature2D`. See [ORB for example](https://docs.opencv.org/4.x/db/d95/classcv_1_1ORB.html).

I'll do a code review as well soon (hopefully today). These were just ideas from reading the documentation and your comments! 

Again, I'm really excited for this improvement!

xdem/coreg.py
```diff
@@ -729,9 +876,7 @@ def fit_pts(
         ref_dem = reference_dem[ref_valid]
 
         if mask_high_curv:
-            planc, profc = xdem.terrain.get_terrain_attribute(
-                tba_dem, attribute=["planform_curvature", "profile_curvature"]
-            )
+            planc, profc = get_terrain_attribute(tba_dem, attribute=["planform_curvature", "profile_curvature"])
```
erikmannerfelt, line 879: This (together with `maxc`) will create three rasters, and then a fourth is created at `mask_hc`. We should probably have a "max_curvature" attribute for `get_terrain_attribute()`. But in the meantime, if you change this and the line below, this line will have three rasters in memory like before, but then two are garbage collected. Therefore, when `mask_hx` is defined, only two are in memory instead of four.

In code, could you change this to:

```python
maxc = np.maximum(np.abs(get_terrain_attribute(tba_dem, attribute=["planform_curvature", "profile_curvature"])), axis=0)
```
or add
```python
del planc
del profc
``` 
after `maxc` is created?

xdem/coreg.py
```diff
@@ -1211,296 +1234,39 @@ def _fit_func(
...
-
-        # If degree==0, it's just a bias correction
-        empty_matrix = np.diag(np.ones(4, dtype=float))
-
-        empty_matrix[2, 3] += self._meta["coefficients"][0]
-
-        return empty_matrix
-
-
 class CoregPipeline(Coreg):
```
erikmannerfelt, line 1253: This should be renamed to just `Pipeline` or `CorrPipeline` to reflect that it may not only be for co-registration, right? It may also be for bias correction (and filters in the future)

xdem/coreg.py
```diff
@@ -1568,627 +1354,592 @@ def _to_matrix_func(self) -> NDArrayf:
...
-        """Append Coreg(s) or a CoregPipeline to the pipeline."""
-        if not isinstance(other, Coreg):
-            other = list(other)
-        else:
-            other = [other]
-
-        pipelines = self.pipeline + other
-
-        return CoregPipeline(pipelines)
+class BlockwiseCoreg(Coreg):
```
erikmannerfelt, line 1358: Same as `Pipeline`, this should probably be renamed to something without "Coreg" in it.

tests/test_biascorr.py
```diff
@@ -0,0 +1,518 @@
...
+        # Check all parameters are the same within 10%
+        fit_params = dirbias._meta["fit_params"]
+        assert np.shape(fit_params) == np.shape(params)
+        assert np.allclose(params, fit_params, rtol=0.1)
+
+        # Run apply and check that 99% of the variance was corrected
+        corrected_dem = dirbias.apply(bias_dem)
+        assert np.nanvar(corrected_dem - self.ref) < 0.01 * np.nanvar(synthetic_bias)
+
+    def test_deramp(self) -> None:
```
erikmannerfelt, line 426: This test doesn't really seem to do anything other than making sure it was instantiated correctly. If that's intentional, perhaps call it `test_deramp_instantiation()`

tests/test_biascorr.py
```diff
@@ -0,0 +1,518 @@
...
+        assert np.shape(fit_params) == np.shape(params)
+        assert np.allclose(
+            params.reshape(order + 1, order + 1)[-1:, -1:], fit_params.reshape(order + 1, order + 1)[-1:, -1:], rtol=0.1
+        )
+
+        # Run apply and check that 99% of the variance was corrected
+        corrected_dem = deramp.apply(bias_dem)
+        assert np.nanvar(corrected_dem - self.ref) < 0.01 * np.nanvar(synthetic_bias)
+
+    def test_terrainbias(self) -> None:
```
erikmannerfelt, line 470: Same as wiht the `test_deramp()`; it only tests instantiation.

tests/test_biascorr.py
```diff
@@ -0,0 +1,518 @@
...
+
+        # Try default "fit" parameters instantiation
+        tb = biascorr.TerrainBias()
+
+        assert tb._fit_or_bin == "bin"
+        assert tb._meta["bin_sizes"] == 100
+        assert tb._meta["bin_statistic"] == np.nanmedian
+        assert tb._meta["terrain_attribute"] == "maximum_curvature"
+
+    def test_terrainbias__synthetic(self) -> None:
```
erikmannerfelt, line 481: Double underscore again. Intentional?

tests/test_biascorr.py
```diff
@@ -0,0 +1,518 @@
...
+        # Get maximum curvature
+        maxc = xdem.terrain.get_terrain_attribute(self.ref, attribute="maximum_curvature")
+
+        # Create a bias depending on bins
+        synthetic_bias = np.zeros(np.shape(self.ref.data))
+
+        bin_edges = np.array((-1, 0, 0.1, 0.5, 2, 5))
+        bias_per_bin = np.array((-5, 10, -2, 25, 5))
+        for i in range(len(bin_edges) - 1):
+            synthetic_bias[np.logical_and(maxc.data >= bin_edges[i], maxc.data < bin_edges[i + 1])] = bias_per_bin[i]
```
erikmannerfelt, line 493: Could you comment on what this does? My soupy brain can't follow

tests/test_biascorr.py
```diff
@@ -0,0 +1,518 @@
...
+
+        # Add bias to the second DEM
+        bias_dem = self.ref - synthetic_bias
+
+        # Run the binning
+        deramp = biascorr.TerrainBias(
```
erikmannerfelt, line 499: Perhaps change the variable name to something else than `deramp`

xdem/biascorr.py
```diff
@@ -0,0 +1,830 @@
...
+        bias_vars: dict[str, NDArrayf],  # Never None thanks to BiasCorr.fit() pre-process
+        transform: rio.transform.Affine,  # Never None thanks to Coreg.fit() pre-process
+        crs: rio.crs.CRS,  # Never None thanks to Coreg.fit() pre-process
+        weights: None | NDArrayf = None,
+        verbose: bool = False,
+        **kwargs,
+    ) -> None:
+
+        # Check bias variable
+        if bias_vars is None or len(bias_vars) <= 2:
```
erikmannerfelt, line 561: This says less than or equal to 2, but the exception says more than two. Which one is correct?

erikmannerfelt (commented):
Is this the largest PR to date? :P 

The coreg.py diff has become messed up since things have moved up and down. Therefore it was hard for me to evaluate the last 1/3 of the changes. I guess I'll just have to trust you! If tests pass, however, it can't be that wrong ;)

erikmannerfelt:
Oh yeah. What if instead of `Rigid`(-`Coreg`), we call it `Affine`(-`Coreg`)? Then, any process that can be explained as an affine matrix would fit into this. It would solve the problem with `Tilt` not being rigid, but being affine.

tests/test_spatialstats.py
```diff
@@ -277,6 +277,70 @@ def test_interp_nd_binning_realdata(self) -> None:
         # Check a value is returned outside the grid
         assert all(np.isfinite(fun(([-5, 50], [-500, 3000], [-2 * np.pi, 4 * np.pi]))))
 
+    def test_get_perbin_nd_binning(self) -> None:
+        """Test the get per-bin function."""
+
+        # Read nd_binning output
+        df = pd.read_csv(
+            os.path.join(examples._EXAMPLES_DIRECTORY, "df_3d_binning_slope_elevation_aspect.csv"), index_col=None
```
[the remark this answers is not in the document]
rhugonnet: Yes!

xdem/coreg.py
```diff
@@ -500,53 +513,195 @@ def _get_x_and_y_coords(shape: tuple[int, ...], transform: rio.transform.Affine)
     return x_coords, y_coords
 
 
-CoregType = TypeVar("CoregType", bound="Coreg")
+def _preprocess_coreg_input(
```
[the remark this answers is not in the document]
rhugonnet: Good idea, changed!

doc/source/coregistration.md
```diff
@@ -131,47 +134,45 @@ aligned_dem = nuth_kaab.apply(tba_dem)
...
 ```
 
-## Deramping
+## Tilt
 
-{class}`xdem.coreg.Deramp`
+{class}`xdem.coreg.Tilt`
 
-- **Performs:** Bias, linear or nonlinear vertical corrections.
+- **Performs:** A 2D plan tilt correction.
```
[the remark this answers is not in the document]
rhugonnet: Corrected

doc/source/biascorr.md
```diff
@@ -1,13 +1,192 @@
...
+glacier_outlines = gu.Vector(xdem.examples.get_path("longyearbyen_glacier_outlines"))
+# Create an inlier mask of terrain outside the glacier polygons
+inlier_mask = glacier_outlines.create_mask(ref_dem)
+```
+
+(biascorr-deramp)=
+
+## Deramping
+
+{class}`xdem.biascorr.Deramp`
```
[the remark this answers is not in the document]
rhugonnet: Fixed

xdem/coreg.py
```diff
@@ -577,82 +733,73 @@ def fit(
...
+        # Flag that the fitting function has been called.
+        self._fit_called = True
 
-        if crs is None:
-            raise ValueError("'crs' must be given if both DEMs are array-like.")
+        return self
 
-        ref_dem, ref_mask = get_array_and_mask(reference_dem)
-        tba_dem, tba_mask = get_array_and_mask(dem_to_be_aligned)
+    def residuals(
```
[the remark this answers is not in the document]
rhugonnet: Great idea, done!

xdem/coreg.py
rhugonnet: Done :)

xdem/coreg.py
rhugonnet: Agreed, will bring it up in summary reply below for us to choose names! :slightly_smiling_face:

xdem/coreg.py
rhugonnet: Same as above, will bring it up in summary reply below!

xdem/spatialstats.py
```diff
@@ -189,6 +189,24 @@ def nd_binning(
     return df_concat
 
 
+# Function to convert IntervalIndex written to str in csv back to pd.Interval
```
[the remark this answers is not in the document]
rhugonnet: Added to #378

tests/test_biascorr.py
```diff
@@ -0,0 +1,518 @@
...
+        # Try default "fit" parameters instantiation
+        deramp = biascorr.Deramp()
+
+        assert deramp._fit_or_bin == "fit"
+        assert deramp._meta["fit_func"] == polynomial_2d
+        assert deramp._meta["fit_optimizer"] == scipy.optimize.curve_fit
+        assert deramp._meta["poly_order"] == 2
+
+    @pytest.mark.parametrize("order", [1, 2, 3, 4])  # type: ignore
+    def test_deramp__synthetic(self, order: int) -> None:
```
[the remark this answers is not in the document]
rhugonnet: Actually a new habit I got from contributing to `pyproj` and some other packages: they classify "sub-tests" of the same function using a double underscore, if it's just about a different input/option.
For example: `test_nuthkaab`, `test_nuthkaab__subsample`, `test_nuthkaab__weights`, ...

I realized this distinction was a bit disorganised in some of our tests, we could adopt this as well? Do you know of any further rules or common practice on this @erikmannerfelt? (we could add a short summary to the xDEM Wiki to remember good practices on tests)

tests/test_biascorr.py
rhugonnet: Following the logic above, no option is passed here so the test implicitly only calls the function/class with default arguments.

tests/test_biascorr.py
rhugonnet: See above.

tests/test_biascorr.py
rhugonnet: See above.

tests/test_biascorr.py
rhugonnet: Done!

tests/test_biascorr.py
rhugonnet: Good catch, done

xdem/biascorr.py
rhugonnet: Fixed

rhugonnet:
Thanks a lot @erikmannerfelt for the thorough read and comments! All accounted for :slightly_smiling_face: 

We just need to decide on names! I agree with all of your replies above.



- For the parent `Coreg` class: I like `Corr` or `Correction` for affine/non-affine, but not sure it'd work so well with `Filter`... maybe `Processing` or `ProcStep` but it's a bit generic.  So we could keep the name `Coreg` for affine/nonaffine, and maybe `CoregStep` or `CoregProc` would qualify better for including `Filter`?
- For `Rigid`: I like `AffineCoreg` proposed by @erikmannerfelt, as affine transformations can always be written in a 4x4 matrix, `Tilt` would fit in it, and it encompasses rigid methods!
- For `BiasCorr`: I guess `NonAffineCoreg`


 I agree that "bias corrections" is a bit generic, it technically applies to all... But in our IEEE paper we did call those "specific biases" (which didn't mean much except "anything but affine").

adehecq:
Hey,
Wow, that feels so weird to look back at comments that date back as far as 2 years ago !
Thanks a lot for getting back to this @rhugonnet and the huge effort of the past weeks !! 🙌  I think it's heading in a very nice direction. Let me make some generic comments first so we can agree on the structure/names. 

1. Regarding the class structure and naming (it's difficult to separate both).
- I think we should have a `Pipeline` (or `ProcPipeline`) and a `ProcStep` living in a separate submodule, that are designed to be generic and be parent classes for coreg, filters and why not interpolation or volume calculation. These should probably be hidden classes, because the user has no reason to modify their behavior.
- I would have a coreg.py module with a `CoregAffine` and `CoregNonAffine` subclasses. Note: shouldn't "Coreg" lie before "Affine/NonAffine" because the main characteristic is coreg, then the subtype? I would not have a 3rd BiasCorr subclass as it makes things a bit more complex, but I'm not strongly opposed to it if you're both in favor. I think with a nice diagram, the differences between them should be quite clear.
- I would have a separate filters.py module, with the two suggested `FilterGlobal` and `FilterKernel` (again switching the items for consistency) and more if appropriate.
For me the diagram would look something like this (not sure about how to represent the link between `ProcStep` and `Pipeline` and also `BlockWise` with the other coreg methods). This also made me think. Probably the BlockWise approach does not make sense for filters, because it is the same idea that the FilterKernel, so maybe it should stay in coreg.py?
xdem_diagram

2. Regarding fit/apply, I totally agree that in most cases both steps should be run at the same time. I prefer `run` rather than `fit_and_apply` which does not read so well, but if this is consistent with what is done in other libraries like OpenCV, I'm open to using the latest.

3. I had a quick look at the documentation and I really like it! I like that you have the bullet points briefly describing each method at the top. I had a few minor comments on this (e.g. Deramping is also very frequently used for SAR-derived DEMs), but maybe it's easier if I go through the documentation in detail once we have the final terminology/structure.

I'm happy to have a more detailed look at the code later, but again, given that it's a very large PR and will take a significant amount of time, I'd rather do it on a close-to-final version. Plus, @erikmannerfelt already made a first round and I don't think I can compete in term of Python rigorousness 😄

erikmannerfelt:
I like @adehecq's suggestion on the structure! Depending on the new file length (since the `Coreg` functionality is now in a `proc_step.py` file or something)  they could live in different python files maybe:

```
-- xdem
  | - proc_step.py
  | filters/
      | - __init__.py
      | - global_filters.py
      | - kernel_filters.py
  | coreg/
      | - __init__.py
      | - base.py  # Maybe not needed, but `Coreg` could live here if we need extra inheritance from `ProcStep`
      | - affine_coreg.py
      | - non_affine_coreg.py
  | ...
```
where the new `__init__.py` would import all classes in the `coreg/` and `filter/` submodules. `Blockwise` and `Pipeline` could live in `proc_step.py`.

Thoughts?

I'm fine with either `run()` or `fit_and_apply()`. If you think `run()` is better, then lets take that.


@rhugonnet, when this is figured out, I think we should finally merge!!

rhugonnet:
Thanks both for the great inputs! :smile:

I initially agreed with everything and tried to move forward, but then realized there's an issue with this structure :sweat_smile:: our processing steps and pipelines as they stand are not generic, but specific to two DEMs later differenced, or a DEM and point data, and expect these inputs. So, for instance, we wouldn't be able to apply the same `GlobalFilter` for a co-registration pipeline, or for filtering a single DEM...
And, if we tried to do so, it could rapidly turn into a mess in terms of consistency of method arguments in the classes, or knowing what object is considered by the filter (or processing step in general).

So we'd need to clearly define to what objects the "steps" and "pipelines" refer to. As of now, we just have things that work on `dh`, so maybe we should just keep it that way and have `dh`-centered classes in the pipeline? 
I'm not sure it would be very useful to use a filter class on a single DEM anyway, as there isn't many other processing steps to combine them with right now... 
**Actually, I think the justification we have for building pipelines is specific to co-registration, because of the need to `fit` a model, and then `apply` it to one or several outputs independently, and being able to intelligently combine several transformations through associated metadata, and access all of that metadata in the class. I don't think that this is really needed for other processing steps (filtering, terrain, interpolation, volume change)?**
In the case of filters for DEMs, maybe it's better to have them as a class method `DEM.filter()` and a corresponding generic function in `filters/`? (like the ones we have in `terrain.py`, we should wrap `terrain` methods in the `DEM` class too!)

In short:
- **For allowing filters in coregistration pipelines:** we could have a specific convenience `CoregFilter` class that lives in `coreg/` and would wrap the functionalities in `filters.py`, and could apply them to the `ref_dem`, or the `tba_dem`, or both, or the `dh`?
- **For the file structure:** building on what @erikmannerfelt laid out, it would keep all the same, and just remove the `proc_step.py`, add `coreg_filters.py` in `coreg/` and imply that the `filters/` module is function-based:
```
-- xdem
  | filters/
      | - __init__.py
      | - global_filters.py # Only functions
      | - kernel_filters.py # Only functions
  | coreg/
      | - __init__.py
      | - base.py  # To have `CoregStep`, `CoregPipeline` and `BlockwiseCoreg`
      | - affine_coreg.py
      | - non_affine_coreg.py
      | - coreg_filters.py # Classes wrapping filters for coreg
  | ...
```
- For the naming, as everything would be related to co-registration, we could keep `Coreg` everywhere. For the order @adehecq mentioned, the adjective seem to usually come first in English (e.g., in `pyproj`: `VerticalCRS` or `GeographicalCRS` are subtypes of `CRS`), so `AffineCoreg` and `NonAffineCoreg` sound better to me:
```
All inherited from CoregBase:
-- BlockwiseCoreg
-- CoregPipeline
-- CoregStep
  | - AffineCoreg
      | - ICP
      | - NuthKaab
  | - NonAffineCoreg
      | - CPD
      | - Deramp
  | - CoregFilter
      | - MagicFilter
      | - TheBestFilterInTheWorld
```

What do you think? I hope I didn't miss anything.

If you are both fine with the naming, you can just send a :+1: and I'll start making the changes related to this PR. I agree @erikmannerfelt, we should merge this soon!

adehecq:
Let's do that then !
Indeed, the Pipeline makes most sense for coregistration and filtering during coregistration. I would argue that it could be interesting for other aspects too, but with some difficult differences in the inputs as you highlighted. Wouldn't it be cool to have a pipeline like:
`coreg.NuthKaab() + filters.stdfilter() + volume.hypsometric_filling + volume.volume_change()`
but I think we can live with regular line by line processing pipelines 😉 
I believe we could get around the issue in the number of inputs by providing a list of rasters rather than 1 or 2 rasters, but it would indeed be a hassle to change all of that and may make things overly complex.

Ok for the order. It also sounded strange to me, but it made more sense to me if we had for example coreg and filters living in the same submodule, because then it would be easier to identify each category (e.g. if we had xdem.AffineCoreg, xdem.GlobalFilter, xdem.NonAffineCoreg etc it would be puzzling). Because they live in two different submodules, it's fine.

rhugonnet:
Re-structuration done according to comments!

Two points:
- I kept `biascorr.py` for non-affine/bias-correction methods, because we barely have any common "non-affine" registration methods (Coherent Point Drift), rather they're all corrections. The documentation that will come in the next PR will make the definitions clear.
- I added a `pipelines.py` module in `xdem/coreg/` for the `dem_coregistration` function (which has to be in another file than `base.py`, `affine.py` or `biascorr.py`, otherwise there are circular import issues), and where we'll also be able to define future "pre-defined" pipelines! :smile:

After more than 2 years... merging! :partying_face:

Issue #157, opened 2021-06-26
Sort the mess in spatial_tools.py


rhugonnet:
Right now it's a bit our "dumper" module.

Looking at what's inside, I think we could split into:
- a `robust_stats.py` module, that includes `nmad`, `nd_binning`, `robust_polynomial_fit`, `robust_sumsin_fit`, `subsample`, and all related subfunctions
- move `stack_rasters`, `merge_rasters`, `merge_bounding_boxes`, `get_array_and_mask`, etc... to **geoutils**
- rename `spstats` into `spatial_stats.py` for consistency with `robust_stats.py`

rhugonnet:
@erikmannerfelt @adehecq An opinion on this?

erikmannerfelt:
I agree on your division! I would like to discuss the `robust_stats.py`, however, as I interpret "robust" as just another synonym for "good". It would (for me) sound like calling it `good_stats.py` (as opposed to the `bad_stats.py`? ;) ). Maybe I'm alone on that opinion? What do you think, @adehecq?

I especially agree on the -> geoutils suggestions.

rhugonnet:
Maybe there's a better subdivision for that module?
For the naming `robust`, it is the term widely used in statistics for this kind of methods, from a "simple" NMAD to more complex M-L methods like RANSAC (https://en.wikipedia.org/wiki/Robust_statistics).

adehecq:
Ok to move some of the functionalities to geoutils. 
Should it be in georaster directly? Right now, it's mostly just one class and not really a lot of other functions, but this can change.

erikmannerfelt:
CLosed with 226

PR #159, opened 2021-06-27
Spatial and robust stats, intro documentation + Update of `spatialstats.py` with new `scikit-gstat` features

### LINKED ISSUE #152: Make use of subsampling function
In #141, a function for subsampling a raster has been written, but is not used yet. This should be fixed.
Especially useful at:
https://github.com/GlacioHack/xdem/blob/e34e6c8502bfc3453fa044664c34c72682dfdf12/xdem/coreg.py#L285
https://github.com/GlacioHack/xdem/blob/e34e6c8502bfc3453fa044664c34c72682dfdf12/xdem/coreg.py#L489

rhugonnet:
EDIT PR ALMOST READY:

Partly resolves #152 

## Refactors
- Refactored `spstats` into `spatialstats` for clarity and consistency with other module names (`spatial_tools` will disappear soon, as discussed in #157)

## Added documentation
- Added `Introduction`
- Added `Spatial statistics` page
- Added `Robust statistics` page
- Added gallery example `Non-stationarity in elevation measurement errors`
- Added gallery example: `Spatial correlation in elevation measurement errors`
- Added structure for other pages: `Bias corrections`, `Filters`, all added to API

## Added/Improved plotting functionalities
- Added `plot_1d_binning`, `plot_2d_binning` for results of `nd_binning`
- Improved `plot_vgm` for results of `sample_multirange_variogram`

## Refined variogram sampling and implementation of random methods
- Modified `sample_multirange_variogram` to wrap methods pre-sampling of Raster point (functions not available for users such as `_get_pdist_variogram`, `_subsample_wrapper`) with different methods (circular, ring, random).
- Added new `RasterEquidistantMetricSpace` method of `scikit-gstat` into a wrapper function `_get_cdist_variogram`. 
Fits in one-line summary, but actually a lot of work:
https://github.com/mmaelicke/scikit-gstat/issues/111
https://github.com/mmaelicke/scikit-gstat/pull/112
https://github.com/mmaelicke/scikit-gstat/pull/113
https://github.com/mmaelicke/scikit-gstat/pull/114
- All `kwargs` of `skgstat.Variogram` and `skgstat.MetricSpace` can be passed, those are automatically sorted by the `sample_multirange_variogram` function.
- Added support for `random_state` to pass seeds of random sequences, in order for tests and examples to have fixed results (added `random_state` argument to many functions, including the `subsample_raster` function written by Amaury). The random seeds are also passed down to `scikit-gstat` functions, when those use random sequences.
- Added better description of functions, and improved line-by-line comments

## Tests and fixes
- Added more robust tests for all functions related to `sample_multirange_variogram`
- Refined tests of `fit_sum_variogram`
- Fixed some issues with `patches_method`, for it to run faster (@adehecq your convolution approach is probably still better, we should implement it!)

## TODO (just thought about it):

- [x]  Add random_state for `patches_method`
- [x]  Improve tests for `patches_method`
- [x]  Fix bug in histogram of `plot_vgm` when log scale is used.
- [x] Fix spatialstats.py for referencing code in `Spatial statistics` page.
- [x] Clarify `run` and `nruns` arguments of sample_multirange_variogram


docs/source/intro.rst
```diff
@@ -0,0 +1,72 @@
+.. _intro:
+
+Introduction: why is it complex to assess DEM accuracy and precision?
+=====================================================================
+
+Digital Elevation Models are numerical representations of elevation. They are generated from different instruments (e.g., radiometer, radar, lidar), acquired in different conditions (e.g., ground, airborne, satellite), and using different post-processing techniques (e.g., stereophotogrammetry, interferometry).
```
erikmannerfelt, line 6: Technically, a DEM only one type of numerical representation of elevation. Others are contour lines, meshes and point clouds. Perhaps it is worth adding something along this on the next line (or refer to a more technical description somewhere else):
`Specifically, a DEM is a georeferenced image, where each pixel represents the elevation of that point.`

docs/source/intro.rst
```diff
@@ -0,0 +1,72 @@
...
+
+While some complexities are specific to certain instruments and methods, all DEMs generally have:
+
+- an **arbitrary Ground Sampling Distance (GSD)** that does not necessarily represent their underlying spatial resolution,
```
erikmannerfelt, line 10: To an unexperienced reader (for which this is targeted, right?) GSD and spatial resolution may be interpreted to be the same. Perhaps add a link to a source that differs the two?

docs/source/intro.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Accuracy and precision
+**********************
+
+Both accuracy and precision are important factors to account for when analyzing DEMs:
+
+- the **accuracy** (systematic error) of a DEM describes how close a DEM is to the true location of measured elevations on the Earth's surface,
+- the **precision** (random error) of a DEM describes the typical spread of its error in measurement, independently of a possible bias from the true positioning.
+
+TODO: Add a little schematic!
```
erikmannerfelt, line 26: I like this (type of) schematic!

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,108 @@
...
+
+The standard error (SE) of a statistic is the standard deviation of the distribution of this statistic.
+For spatially distributed samples, the standard error of the mean (SEM) is of great interest as it allows quantification of the error of a mean (or sum) of samples in space.
+
+The standard error  :math:`\sigma_{\overline{dh}}` of the mean :math:`\overline{dh}` of elevation changes samples :math:`dh` is typically derived as:
+
+.. math::
+
+        \sigma_{\overline{dh}} = \frac{\sigma_{dh}}{\sqrt{N}},
+
```
erikmannerfelt, line 70: Could this be suffixed by a python version of the equation? Algebra is just another programming language, right? ;)

Something like:
```python
mean_stderr = dispersion / np.sqrt(n_observations)
```

I think it would be nice to consistently have both algebra and python representations. The conciseness of algebra can often be quite daunting, especially for newcomers. But its universal nature makes it great to have, so a compromise would be to always do both. What do you think?

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,108 @@
...
+where :math:`\sigma_{dh}` is the dispersion of the samples, and :math:`N` is the number of **independent** observations.
```
erikmannerfelt, line 71: Here is the first time that the word "dispersion" is used. Would be nice with an explanation of its meaning and/or a link to more information (such as wikipedia?)

docs/source/code/spatialstats.py
```diff
@@ -0,0 +1,40 @@
+import xdem
```
erikmannerfelt, line 1: I usually prepend the script with a one-liner docstring to shortly describe the aim with this particular code example.

Also, it doesn't seem to be used yet in the text, right? Same with the other code example.

erikmannerfelt (commented):
This will be a fantastic resource!! I look forward to seeing the text be more integrated with figures, schematics and the xdem API. Let me know if there's something I can do!

docs/source/intro.rst
adehecq: Or I often say a DEM is a **gridded** representation of elevation.

docs/source/intro.rst
adehecq: Yes, that's pretty good! Where does the figure come from? Any copyright?

docs/source/spatialstats.rst
adehecq: I prefer writing the algebra formulation which is universal. 
But an option could be to point out to the line of code where it is implemented. I don't know how easy it is to track the code line when the code changes though.

docs/source/spatialstats.rst
adehecq: We should call it by its name, i.e. standard deviation!

adehecq: Ok, I guess this is to be more generic in case we use NMAD etc.
Then I think the term is statistical dispersion, but dispersion or spread are synonyms: https://en.wikipedia.org/wiki/Statistical_dispersion

docs/source/code/spatialstats.py
```diff
@@ -0,0 +1,40 @@
...
+to_be_aligned_raster = gu.georaster.Raster(xdem.examples.FILEPATHS["longyearbyen_tba_dem"])
+glacier_mask = gu.geovector.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines"])
+inlier_mask = ~glacier_mask.create_mask(reference_raster)
+
+nuth_kaab = xdem.coreg.NuthKaab()
+nuth_kaab.fit(reference_raster.data, to_be_aligned_raster.data,
+              inlier_mask=inlier_mask, transform=reference_raster.transform)
+aligned_raster = nuth_kaab.apply(to_be_aligned_raster.data, transform=reference_raster.transform)
+
+ddem = gu.Raster.from_array((reference_raster.data - aligned_raster),
```
adehecq, line 18: Now you can just take the difference of both rasters, without using this convoluted method!

adehecq: Right, but aligned_raster is not a Raster, just an array. Never mind.

docs/source/code/spatialstats.py
```diff
@@ -0,0 +1,40 @@
...
+
+# load diff and mask
+xdem.examples.download_longyearbyen_examples(overwrite=False)
+
+reference_raster = gu.georaster.Raster(xdem.examples.FILEPATHS["longyearbyen_ref_dem"])
+to_be_aligned_raster = gu.georaster.Raster(xdem.examples.FILEPATHS["longyearbyen_tba_dem"])
+glacier_mask = gu.geovector.Vector(xdem.examples.FILEPATHS["longyearbyen_glacier_outlines"])
+inlier_mask = ~glacier_mask.create_mask(reference_raster)
+
+nuth_kaab = xdem.coreg.NuthKaab()
```
adehecq, line 13: I think we should have a test case of two aligned DEMs, to avoid running the Nuth & Kaab each time...

docs/source/code/spatialstats_model_vgm.py
```diff
@@ -0,0 +1,38 @@
+import matplotlib.pyplot as plt
```
adehecq, line 1: Do we actually need both examples? Thery're almost the same, so we should just keep the second one, no?

docs/source/coregistration.rst
```diff
@@ -1,7 +1,16 @@
...
+===============
+
+Coregistration between DEMs correspond to aligning the digital elevation models in three dimension.
+
+Transformations that can be described by a 3-dimensional affine function are included in coregistration methods.
+Those transformations include :
+
+- vertical and horizontal shifts,
+- rotations,
+- stretching and squeezing.
```
adehecq, line 13: I would call it "scale change". For me squeezing sounds like np.squeeze, that we're seeing way to often in xdem! ;-)

docs/source/intro.rst
adehecq: Are DEM really generated from radiometers?? That's at least not the first one that comes to my mind! I would just say, "optical sensor" maybe.

adehecq: Also, I would just say "photogrammetry" (without the stereo)

docs/source/intro.rst
adehecq: I would also remove the term "arbitrary", as the choice of the resolution is not fully arbitrary. If it's a standard product, the resolution still gives an idea on what the actual resolution is (although these days resolution mostly depends on the access plan, e.g. for TDX, 90 m is free, 12 m is for purchase).

docs/source/intro.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Absolute or relative accuracy
+*****************************
+
+The measure of accuracy can be further divided into two aspects:
+
+- the **absolute accuracy** of a DEM describes the average shift to the true positioning. Studies interested in analyzing features of a single DEM in relation to other georeferenced data might give great importance to this potential bias.
+- the **relative accuracy** of a DEM is the potential shifts, tilts, and deformations in relation to other elevation data that does not necessarily matches a given referencing. Studies interested in comparing DEMs between themselves might be only interested in this accuracy.
```
adehecq, line 34: it's weird to have a singular then plural, so something's wrong. Maybe "the **relative accuracy** of a DEM is related to the potential shifts, tilts, and deformations with reference to other elevation..."

docs/source/intro.rst
```diff
@@ -0,0 +1,72 @@
...
+
+TODO: Add a point data - DEM co-registration plot
+
+Optimizing DEM relative accuracy
+**********************************
+
+As the **absolute accuracy** can be corrected a posteriori using reference elevation datasets, many analyses only focus on **relative accuracy**, i.e. the remaining biases between several DEMs co-registered relative one to another.
+By harnessing the denser, nearly continuous sampling of raster DEMs (in opposition to the sparser sampling of higher-accuracy point elevation data), one can identify and correct other types of biases:
+
+- Terrain-related biases that can originate from the difference of resolution of DEMs, or instrument processing deformations.
```
adehecq, line 54: You could add some scientific references here, like you do for the variograms. For example the Gardelle paper here, and Luc's paper for the line below.

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,108 @@
...
+However, several issues arise to estimate the standard error of a mean of elevation observations samples:
+
+1. The dispersion :math:`\sigma_{dh}` cannot be estimated directly on changing terrain that we are usually interested in measuring (e.g., glacier, snow, forest).
+2. The dispersion :math:`\sigma_{dh}` typically shows important non-stationarities (e.g., an error 10 times as large on steep slopes than flat slopes).
+3. The number of samples :math:`N` is generally not equal to the number of sampled DEM pixels, as those are not independent in space and the Ground Sampling Distance of the DEM does not necessarily correspond to its effective resolution.
+
+Note that the SE represents completely stochastic (random) errors, and is therefore not accounting for possible remaining systematic errors have been accounted for, e.g. using one or multiple :ref:`coregistration` approaches.
+
+
+Methods for DEM precision estimation
```
adehecq, line 82: This will turn into a whole book ! ;-)

docs/source/spatialstats.rst
rhugonnet: Amaury's solution seems good, maybe we could have a file in /source/code that has all the equations in Python we can point to? 
Not sure it should be tested by `test_docs.py`, though. We could place it in another folder: source/equations or similar?

docs/source/intro.rst
erikmannerfelt: If you just search for "accuracy vs precision" on google images, tons of these will show up. I'm sure one is copyright-easy. Worst case, we can make our own in matplotlib!

docs/source/spatialstats.rst
rhugonnet: Indeed, the standard deviation is only a measure of the dispersion, and the mean a measure of central tendency.
Because we use robust statistics, we'll have to specify that we tend to use the median/50th percentile or the NMAD/68th absolute percentile as measures of central tendency and dispersion, rather than the typical mean/std.

This should be clearly introducted in a few words. Maybe in the robust statistics documentation, and point there from the spatial statistics documentation when a the term "dispersion" is brought up?

docs/source/coregistration.rst
erikmannerfelt: But it's not always just scaling! It's warping too with Deramp and the upcoming BlockwiseCoreg

docs/source/code/spatialstats.py
rhugonnet: Coming soon!

docs/source/code/spatialstats.py
rhugonnet: +1
@erikmannerfelt What do you think is the best approach here? Should we run the N&K once in `xdem/example.py` and store it very much like the raw data, maybe?

docs/source/code/spatialstats_model_vgm.py
rhugonnet: Yes, work in progress!

docs/source/coregistration.rst
rhugonnet: And scaling is the commonly used term: https://en.wikipedia.org/wiki/Affine_transformation

docs/source/spatialstats.rst
rhugonnet: For sure it will :D

docs/source/code/spatialstats.py
erikmannerfelt: Hmm. That's good for performance, but I don't see a good way to do it. Ideas?

docs/source/intro.rst
rhugonnet: All good for that one, it's hosted on AntarcticGlaciers.org with a CC-BY-NC-SA license and no credit ;)

docs/source/intro.rst
erikmannerfelt: I know this is maybe overkill, but you can have  a gridded mesh of elevation too! When people say DEMs, they always mean an array with georeferencing, no?

Anyway, the way you've put it is great @rhugonnet. I'm just doing my best to be annoying!

docs/source/code/spatialstats.py
rhugonnet: If we encapsulate the `download_longyearbyen_example` function into a `prepare_longyearbyen_example` functions which runs the download + runs N&K and writes to disk, would that work?

We would need to import xdem into `examples.py` which... shouldn't raise a circular import? (I'm not sure really)

docs/source/intro.rst
erikmannerfelt: It fits very well in the rendered documentation! Perhaps add a source or license statement below to be on the safe side?

docs/source/spatialstats.rst
adehecq: Sounds good!

docs/source/code/spatialstats.py
adehecq: My idea was to just align it once and add it to xdem-data.

docs/source/coregistration.rst
adehecq: Good point. Maybe we should have scaling, and then another point that is e.g. "non-linear shifts" or so

adehecq: Or "rubber-sheet" deformation (or whatever you called it @erikmannerfelt at some point.

docs/source/intro.rst
adehecq: I still think we should add the "gridded" term, as a DEM is always gridded (even though this may apply to other objects too).
Well, actually Wikipedia says:"A DEM can be represented as a raster (a grid of squares, also known as a heightmap when representing elevation) or as a vector-based triangular irregular network (TIN). "

rhugonnet: It's been added ;)

docs/source/code/spatialstats.py
erikmannerfelt: Otherwise, there could be an examples function that coregisters and saves the DEM. Then all functions can just "get-or-generate" the DEM (so it only has to be generated once).

Perhaps this fits as a PR, as a large part of the documentation and test suite will need changing?

docs/source/intro.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Optimizing DEM absolute accuracy
+**********************************
+
+Shifts due to poor absolute accuracy are common in elevation datasets, and can be easily corrected by performing a DEM co-registration to precise and accurate, quality-controlled elevation data such as ICESat and ICESat-2.
+Quality-controlled DEMs aligned on high-accuracy data also exists, such as TanDEM-X global DEM (see `Rizzoli et al. (2017) <https://doi.org/10.1016/j.isprsjprs.2017.08.008>`_)
+
+Those biases can be corrected using the methods described in :ref:`coregistration`.
+
+TODO: Add a point data - DEM co-registration plot
```
erikmannerfelt, line 49: You can either add a figure here or a gallery with DEM coregistration examples!!

docs/source/intro.rst
```diff
@@ -0,0 +1,75 @@
...
+
+TODO: Add a plot on co-registration + bias correction between two DEMs
+
+Quantifying DEM precision
+**************************
+
+While dealing with **accuracy** is quite straightforward as it consists of minimizing the differences (biases) between several datasets, assessing the **precision** of DEMs can be much more complex.
+Measurement errors of a DEM cannot be quantified by a simple difference and require statistical inference.
+
+The **precision** of DEMs has historically been reported by a single metric (e.g., precision of :math:`\pm` 2 m), but recent studies have shown the limitations of such simple metrics and provide more statistically-advanced methods to account for potential variabilities in precision and related correlations in space.
```
erikmannerfelt, line 70: Typical academic review comment; which recent studies!? :)

docs/source/intro.rst
```diff
@@ -0,0 +1,75 @@
...
+- a `ground sampling distance <https://en.wikipedia.org/wiki/Ground_sample_distance>`_ (GSD), or pixel size, **that does not necessarily represent the underlying spatial resolution of the observations**,
+- a `georeferencing <https://en.wikipedia.org/wiki/Georeferencing>`_ **that can subject to shifts, tilts or other deformations** due to inherent instrument errors, noise, or associated post-processing schemes,
+- a large number of `outliers <https://en.wikipedia.org/wiki/Outlier>`_ **that remain difficult to filter** as they can originate from various sources (e.g., photogrammetric blunders, clouds).
+
+These factors lead to difficulties in assessing the accuracy and precision of DEMs, which are necessary to perform further analysis.
+
+In ``xdem``, we provide a framework with state-of-the-art methods published in the scientific literature to make DEM calculations consistent, reproducible, and easy.
+
+Accuracy and precision
+**********************
```
erikmannerfelt, line 19: `rST` is weird with section headers. Any order seems to work as long as you're consistent, but in some cases, that also doesn't seem to be the case!

The official order is:
-  \# with overline, for parts
-  \* with overline, for chapters
-  =, for sections
-  -, for subsections
- ^, for subsubsections
-  ", for paragraphs

I think it's best if we stay consistent in the documentation, so I recommend we use this order!

Reference:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections

erikmannerfelt: For this particular header, I think the recommended character would be:

```rst
Accuracy and precision
----------------------
```

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,143 @@
...
+However, several issues arise to estimate the standard error of a mean of elevation observations samples:
+
+1. The dispersion :math:`\sigma_{dh}` cannot be estimated directly on changing terrain that we are usually interested in measuring (e.g., glacier, snow, forest).
+2. The dispersion :math:`\sigma_{dh}` typically shows important non-stationarities (e.g., an error 10 times as large on steep slopes than flat slopes).
+3. The number of samples :math:`N` is generally not equal to the number of sampled DEM pixels, as those are not independent in space and the Ground Sampling Distance of the DEM does not necessarily correspond to its effective resolution.
+
+Note that the SE represents completely stochastic (random) errors, and is therefore not accounting for possible remaining systematic errors have been accounted for, e.g. using one or multiple :ref:`coregistration` approaches.
+
+
+Workflow for DEM precision estimation
```
erikmannerfelt, line 94: This could also refer to an example.

I thought it was great to have these `.. literalinclude: \n :lines:: ...` directives at first.
A large problem that I've found is that they need maintenance whenever the code changes; something which is easy to miss.
I just fixed most of these blunders in the latest PRs, but there's one (tiny) blunder still in the Coregistration chapter:

https://xdem.readthedocs.io/en/latest/coregistration.html#introduction

In the bottom of the **Example data** code block, there's a "##################################" line which should not be visible.

I know that it was I that suggested this way of including code, so I'm sorry that I now recommend slightly against it... With the new examples structure, however, I think this issue can quite neatly be circumvented!

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,143 @@
...
+""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+
+Spatial correlation of elevation measurement errors
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+TODO: Add this section based Rolstad et al. (2009), Dehecq et al. (2020), Hugonnet et al. (in prep)
+
+Quantify and model spatial correlations
+"""""""""""""""""""""""""""""""""""""""
```
erikmannerfelt, line 118: This would also make a nice example!

xdem/spstats.py
```diff
@@ -885,22 +886,29 @@ def patches_method(dh : np.ndarray, mask: np.ndarray[bool], gsd : float, area_si
     return df
 
 
-def plot_vgm(df: pd.DataFrame, fit_fun: Callable = None):
+def plot_vgm(df: pd.DataFrame, list_fit_fun: Optional[list[Callable]] = None, list_fit_fun_label: Optional[list[str]] = None):
```
erikmannerfelt, line 889: As mentioned in one of your 805920 other PRs ;), `Callable` needs an argument and return type annotation.

xdem/spstats.py
```diff
@@ -885,22 +886,29 @@ def patches_method(dh : np.ndarray, mask: np.ndarray[bool], gsd : float, area_si
...
 
     fig, ax = plt.subplots(1)
```
erikmannerfelt, line 891: Completely up to you if you want to include it, but I've added an `ax: plt.Axes = plt.gca()` argument to many of my personal plotting projects, this allows multiple types of calls, as it either plots on a given axis, or defaults to the current one:

```python
plt.subplot(131)
plot_vgm(...)  # This will plot on subplot 121
ax2 = plt.subplot(132)
plot_vgm(...)  # This will plot on subplot 122
ax3 = plt.subplot(133)
plot_vgm(..., ax=ax2)  # This will also plot on 122. Madness!

```

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,143 @@
...
+Pixel-wise elevation measurement error
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+TODO
+
+
+Spatially-integrated elevation measurement error
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The standard error (SE) of a statistic is the standard deviation of the distribution of this statistic.
```
erikmannerfelt, line 74: This is also described here: https://xdem.readthedocs.io/en/latest/spatial_stats.html#standard-error

xdem/spstats.py
erikmannerfelt, line 889: Missing return type annotation.

erikmannerfelt, line 889: Missing docstring.

xdem/spstats.py
rhugonnet: Yes, quite familiar with massive plotting exercises... é_è

Correct me if I'm wrong: doing this would mean that we'd have to always initatilizing `ax` outside the function when we don't want to plot on an existing one, though?
Or could we add a loop at the start of each function to check if `ax` exists, and create it if it doesn't?

xdem/spstats.py
rhugonnet: Double-checking:
If the function is vectorized, should I use `Callable[[np.ndarray],np.ndarray]` to describe this, or rather `Callable[[float],float]` to describe the type of input needed (and forgetting about the fact that it's vectorized)?

xdem/spstats.py
erikmannerfelt: `plt.gca()` means "Get the Current Axis (gca), or make one if there is not current axis", so there's no need for any other check!

docs/source/intro.rst
```diff
@@ -0,0 +1,75 @@
+.. _intro:
+
+Introduction: why is it complex to assess DEM accuracy and precision?
+=====================================================================
+
+Digital Elevation Models are numerical, gridded representations of elevation. They are generated from different instruments (e.g., optical sensors, radar, lidar), acquired in different conditions (e.g., ground, airborne, satellite), and using different post-processing techniques (e.g., photogrammetry, interferometry).
```
adehecq, line 6: post-processing should be processing

docs/source/intro.rst
```diff
@@ -0,0 +1,75 @@
...
+
+While some complexities are specific to certain instruments and methods, all DEMs generally possess:
+
+- a `ground sampling distance <https://en.wikipedia.org/wiki/Ground_sample_distance>`_ (GSD), or pixel size, **that does not necessarily represent the underlying spatial resolution of the observations**,
+- a `georeferencing <https://en.wikipedia.org/wiki/Georeferencing>`_ **that can subject to shifts, tilts or other deformations** due to inherent instrument errors, noise, or associated post-processing schemes,
```
adehecq, line 11: that can be

docs/source/intro.rst
```diff
@@ -0,0 +1,75 @@
...
+- a large number of `outliers <https://en.wikipedia.org/wiki/Outlier>`_ **that remain difficult to filter** as they can originate from various sources (e.g., photogrammetric blunders, clouds).
+
+These factors lead to difficulties in assessing the accuracy and precision of DEMs, which are necessary to perform further analysis.
+
+In ``xdem``, we provide a framework with state-of-the-art methods published in the scientific literature to make DEM calculations consistent, reproducible, and easy.
```
adehecq, line 16: state-of-the-art as of July 2021. Might be outdated in 1 year if we don't pay attention! ;-)

docs/source/intro.rst
```diff
@@ -0,0 +1,75 @@
...
+
+- the **absolute accuracy** of a DEM describes the average shift to the true positioning. Studies interested in analyzing features of a single DEM in relation to other georeferenced data might give great importance to this potential bias.
+- the **relative accuracy** of a DEM is related to the potential shifts, tilts, and deformations with reference to other elevation data that does not necessarily matches the true positioning. Studies interested in comparing DEMs between themselves might be only interested in this accuracy.
+
+TODO: Add another little schematic!
+
+Optimizing DEM absolute accuracy
+**********************************
+
+Shifts due to poor absolute accuracy are common in elevation datasets, and can be easily corrected by performing a DEM co-registration to precise and accurate, quality-controlled elevation data such as ICESat and ICESat-2.
```
adehecq, line 44: https://icesat.gsfc.nasa.gov/icesat/ and https://icesat-2.gsfc.nasa.gov/

docs/source/robuststats.rst
```diff
@@ -1,10 +1,9 @@
...
+Robust statistics
 ==================
 
-How does one calculate the error between two DEMs?
-This question is the basis of numerous academic papers, but their conclusions are often hard to grasp as the mathematics behind them can be quite daunting.
-In addition, the lack of a simple implementation in a modern programming language makes these methods obscure and used only by those who can program it themselves.
-One of the goals of ``xdem`` is to simplify state-of-the-art statistical measures, to allow accurate DEM comparisons for everyone, regardless of one's statistical talent.
+Digital Elevation Models often contain outliers that hamper further analysis.
+In order to deal with outliers, ``xdem`` integrates statistical measures robust to outliers to be used for estimation of the
+mean or dispersion of a sample, or more complex function fitting.
```
adehecq, line 6: I cannot comment unchanged lines, but the definition of median (l 28) is wrong. Worth changing in this PR!

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,157 @@
...
+1. The mean of the variable of interest is stationary in space, i.e. constant over sufficiently large areas,
+2. The variance of the variable of interest is stationary in space, i.e. constant over sufficiently large areas.
+3. The covariance between two observations only depends on the spatial distance between them, i.e. no other factor than this distance plays a role in the spatial correlation of measurement errors.
+
+A sufficiently large averaging area is an area expected to fit within the spatial domain studied.
+
+In other words, for a reliable analysis, the DEM should:
+
+1. Not contain systematic biases that do not average out over sufficiently large distances (e.g., shifts, tilts), but can contain pseudo-periodic biases (e.g., along-track undulations),
+2. Not contain measurement errors that vary significantly in space.
```
adehecq, line 43: This makes me think, it would be great to have a tool to test this in xdem. I guess a large mean and std filter that accounts for NaNs would do the job... Mmmmh more work for me!

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,157 @@
...
+3. Not contain factors that significantly affect the distribution of measurement errors, except for the spatial distance.
+
+Quantifying the precision of a single DEM, or of a difference of DEMs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To statistically infer the precision of a DEM, the DEM has to be compared against independent elevation observations.
+
+If the other elevation data is known to be of higher-precision, one can assume that the analysis of differences will represent the precision of the rougher DEM.
+
+.. math::
```
adehecq, line 53: I would present this the other way around. Put the equation below (sum of variance), then say that if one variance is much smaller than the other, we can approximate.

docs/source/spatialstats.rst
adehecq: +1, for all these examples, it would be great to have a separate code that shows in the example gallery!

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,157 @@
...
+
+.. literalinclude:: code/spatialstats.py
+        :lines: 19-20
+
+For a single range model:
+
+.. literalinclude:: code/spatialstats.py
+        :lines: 22-23
+
+For multiple range model:
```
adehecq, line 104: Multi-range variograms, that's the dream !! <3

docs/source/spatialstats.rst
```diff
@@ -0,0 +1,157 @@
...
+TODO: Add this section based on Krige's relation (Webster & Oliver, 2007), Hugonnet et al. (in prep)
+
+
+Metrics for DEM precision
+*************************
+
+Historically, the precision of DEMs has been reported as a single value indicating the random error at the scale of a single pixel, for example :math:`\pm 2` meters.
+
+However, there is several limitations to this metric:
+- studies have shown significant variability of elevation measurement errors with terrain attributes, such as the slope, but also with the type of terrain
```
adehecq, line 128: A breakline is missing. Also there is only limitation written for now!

examples/plot_nonstationary_error.py
```diff
@@ -0,0 +1,205 @@
...
+Non-stationarity of elevation measurement errors
+================================================
+
+Digital elevation models have a precision that can vary with terrain and instrument-related variables. However, quantifying
+this precision is complex and non-stationarities, i.e. variability of the measurement error, has rarely been
+accounted for, with only some studies that used arbitrary filtering thresholds on the slope or other variables (see :ref:`intro`).
+
+Quantifying the non-stationarities in elevation measurement errors is essential to use stable terrain as a proxy for
+assessing the precision on other types of terrain (Hugonnet et al., in prep) and allows to standardize the measurement
+errors to reach a stationary variance, an assumption necessary for spatial statistics (see :ref`spatialstats`).
```
adehecq, line 11: typo in the ref.

examples/plot_nonstationary_error.py
```diff
@@ -0,0 +1,205 @@
...
+Here, we show an example in which we identify terrain-related non-stationarities for a DEM difference of Longyearbyen glacier.
+We quantify those non-stationarities by `binning <https://en.wikipedia.org/wiki/Data_binning>`_ robustly
+in N-dimension using :func:`xdem.spstats.nd_binning` and applying a N-dimensional interpolation
+:func:`xdem.spstats.interp_nd_binning` to estimate a numerical function of the measurement error and derive the spatial
+distribution of elevation measurement errors of the difference of DEMs.
+
+**Reference**: `Hugonnet et al. (2021) <https://doi.org/10.1038/s41586-021-03436-z>`_, applied to the terrain slope
+and quality of stereo-correlation (Equation 1, Extended Data Fig. 3a).
+"""
+# sphinx_gallery_thumbnail_number = 8
```
adehecq, line 22: what does this mean?

examples/plot_nonstationary_error.py
```diff
@@ -0,0 +1,205 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+import xdem
+import geoutils as gu
+
+# %%
+# We start by loading example files including a difference of DEMs at Longyearbyen glacier, the reference DEM later used to derive
+# several terrain attributes, and the outlines to rasterize a glacier mask.
+# Prior to differencing, the DEMs were aligned using :class:`xdem.coreg.NuthKaab` as shown in
+# the :ref:`sphx_glr_auto_examples_plot_nuth_kaab.py` example. We later refer to those elevation differences as *dh*.
```
adehecq, line 32: The code uses ddem, the terminology should be consistent...

examples/plot_nonstationary_error.py
```diff
@@ -0,0 +1,205 @@
...
+
+ref_dem = xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
+ddem = xdem.DEM(xdem.examples.get_path("longyearbyen_ddem"))
+glacier_outlines = gu.Vector(xdem.examples.get_path("longyearbyen_glacier_outlines"))
+mask_glacier = glacier_outlines.create_mask(ddem)
+
+# %%
+# We remove values on unstable terrain
+ddem.data[mask_glacier] = np.nan
```
adehecq, line 41: This is not the way we should do it, as this will conflict with the actual nodata value used. But currently, masking pixels require changing both self.data.data and self.data.mask... I need to find a solution for this.

examples/plot_nonstationary_error.py
```diff
@@ -0,0 +1,205 @@
...
+# We use :func:`xdem.spstats.nd_binning` to perform N-dimensional binning on all those terrain variables, with uniform
+# bin length divided by 30. We use the :ref:`spatial_stats_nmad` as a robust measure of `statistical dispersion <https://en.wikipedia.org/wiki/Statistical_dispersion>`_.
+
+df = xdem.spstats.nd_binning(values=ddem.data,list_var=[slope, aspect, planc, profc],
+                             list_var_names=['slope','aspect','planc','profc'],
+                             statistics=['count',np.nanmedian,xdem.spstats.nmad],
+                             list_var_bins=30)
+
+# %%
+# We obtain a dataframe with the 1D binning results for each variable, the 2D binning results for all combinations of
```
adehecq, line 63: Panda's dataframe.
Is it possible to display the dataframe in the doc the same way it would show up in a notebook? Just try adding a print statement in the code.

examples/plot_nonstationary_error.py
```diff
@@ -0,0 +1,205 @@
...
+# differences due to a non-stationarity of variance with the terrain slope.
+#
+# What about the aspect?
+
+xdem.spstats.plot_1d_binning(df, 'aspect', 'nmad', 'Aspect (degrees)', 'NMAD of dh (m)')
+
+# %%
+# There is no variability with the aspect which shows a dispersion averaging 2-3 meters, i.e. that of the complete sample.
+#
+# What about the plan curvature?
```
adehecq, line 85: Damn, the suspense is at its peak!!

examples/plot_nonstationary_error.py
```diff
@@ -0,0 +1,205 @@
+"""
+Non-stationarity of elevation measurement errors
```
adehecq, line 2: After reading through, I believe this should be a section in the xdem documentation, rather than hidden in the example gallery?

xdem/spstats.py
adehecq: You can do the same way as I do it in georaster: https://github.com/GlacioHack/GeoUtils/blob/12bd607fe8024720ff264545abf0cc5ba0156ea6/geoutils/georaster.py#L1242
So, if none specified, a new figure is created, otherwise uses the ax that is passed.

xdem/spstats.py
```diff
@@ -1003,22 +1009,273 @@ def patches_method(dh : np.ndarray, mask: np.ndarray[bool], gsd : float, area_si
...
 
 
-def plot_vgm(df: pd.DataFrame, fit_fun: Callable = None):
-
+def plot_vgm(df: pd.DataFrame, list_fit_fun: Optional[list[Callable[[float],float]]] = None,
+             list_fit_fun_label: Optional[list[str]] = None):
+    """
+    Plot empirical variogram, with optionally one or several model fits
+    :param df: dataframe of empirical variogram
+    :param list_fit_fun: list of function fits
```
adehecq, line 1017: Just reading at the docstring, I'm not sure where those functions come from. Maybe just specify that these functions can be obtain e.g. with the function XXX.

xdem/spstats.py
```diff
@@ -1003,22 +1009,273 @@ def patches_method(dh : np.ndarray, mask: np.ndarray[bool], gsd : float, area_si
...
 
     ax.set_xlabel('Lag (m)')
     ax.set_ylabel(r'Variance [$\mu$ $\pm \sigma$]')
     ax.legend(loc='best')
-    ax.grid()
-    plt.show()
+
+    return ax
+
+def plot_1d_binning(df: pd.DataFrame, var_name: str, statistic_name: str, label_var: Optional[str] = None,
```
adehecq, line 1047: Same comment regarding the use of ax, to allow more flexibility.

adehecq: Although this will be a bit more complex since you're using subplots here already. But I wonder if you can generate subplots from an ax instance.

xdem/spstats.py
```diff
@@ -1003,22 +1009,273 @@ def patches_method(dh : np.ndarray, mask: np.ndarray[bool], gsd : float, area_si
...
+
+    # Need a grid plot to show the sample count and the statistic
+    fig = plt.figure()
+    grid = plt.GridSpec(10, 10, wspace=0.5, hspace=0.5)
+
+    # First, an axis to plot the sample histogram
+    ax0 = fig.add_subplot(grid[:3, :])
+    ax0.set_xticks([])
+
+    # Plot the histogram manually with fill_between
```
adehecq, line 1079: Why is that needed? Can't you either use Panda's functionality, or plt.bar?

docs/source/intro.rst
erikmannerfelt: But then we have to make sure to keep this updated all the time! State-of-the-art as of June 2018 doesn't sound that nice, but RichDEM is still leading in performance compared to other tools as far as I'm aware!

examples/plot_nonstationary_error.py
erikmannerfelt: The comment signifies which of the plots will make the gallery thumbnail. Unfortunately, it has to show up in the rendered code as well!

examples/plot_nonstationary_error.py
erikmannerfelt: Yep! There's no super-pretty-print like in notebooks, but maybe the `df.to_markdown()` could render similarly?

Otherwise, just printing works.

xdem/spstats.py
```diff
@@ -1003,22 +1009,273 @@ def patches_method(dh : np.ndarray, mask: np.ndarray[bool], gsd : float, area_si
...
+
+    # Third, an axis to plot the data as a colored grid
+    ax = fig.add_subplot(grid[3:, :-3])
+
+    # Define limits of colormap is none are provided, robust max and min using percentiles
+    if vmin is None and vmax is None:
+        vmax = np.nanpercentile(df_sub[statistic_name].values, 99)
+        vmin = np.nanpercentile(df_sub[statistic_name].values, 1)
+
+    # Create custom colormap
```
adehecq, line 1216: We discussed in the "terrain attribute" discussions that maybe we should have a set of colormaps somewhere in xdem. Should we raise an issue about that?

xdem/spstats.py
```diff
@@ -1003,22 +1009,273 @@ def patches_method(dh : np.ndarray, mask: np.ndarray[bool], gsd : float, area_si
...
+    # In case the axis value does not agree with the scale (e.g., 0 for log scale)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        ax.set_xlim((np.min(interval_var_1.left),np.max(interval_var_1.right)))
+        ax.set_ylim((np.min(interval_var_2.left),np.max(interval_var_2.right)))
+
+    # Fourth and finally, add a colormap and nodata color to the legend
+    axcmap = fig.add_subplot(grid[:3, -3:])
+
+    # Remove ticks, labels, frame
```
adehecq, line 1253: This kind of function that can be re-used elsewhere could go in a dedicated module for plotting (same as for the colormaps)

examples/plot_nonstationary_error.py
erikmannerfelt: Turns out, `to_markdown()` needs an optional dependency. Maybe it could render html? `df.to_html()`

xdem/spstats.py
erikmannerfelt: Perhaps https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.inset_axes.html?

examples/plot_nonstationary_error.py
adehecq: Thanks for the explanation, good to know!

xdem/spstats.py
adehecq: Yes, could work maybe. Just be aware of the warning "This method is experimental as of 3.0, and the API may change."

examples/plot_nonstationary_error.py
erikmannerfelt: Ah, you can just type the dataframe in the bottom of the cell, and it will show the html version (more specifically, it will call `__repl_html__()` if it exists.).

Reference: https://sphinx-gallery.github.io/stable/configuration.html?highlight=html#controlling-what-output-is-captured

Example:
image

erikmannerfelt: Leads me to think the `Raster` and `DEM` classes should have a `__repr_html__()` method!!

erikmannerfelt: Added https://github.com/GlacioHack/GeoUtils/issues/225

docs/source/spatialstats.rst
rhugonnet: You are right!

examples/plot_nonstationary_error.py
rhugonnet: > Unfortunately, it has to show up in the rendered code as well!
I was wondering if it could be hidden! Good to know, it's not taking too much space anyway :)

examples/plot_nonstationary_error.py
rhugonnet: You are right, replaced!

examples/plot_nonstationary_error.py
rhugonnet: Yes, a function very much like `get_array_and_mask` that works for all types.
Maybe `update_mask`?

examples/plot_nonstationary_error.py
rhugonnet: Used this. With this method, the table will display as large as it needs to be, though:

Screenshot from 2021-07-16 12-05-12

xdem/spstats.py
rhugonnet: It's a bit annoying to deal with `plt.bar` arguments when the width varies for each bar, and other histogram methods need the full distribution (which here has already been binned).
I think this is the simplest approach!

xdem/spstats.py
rhugonnet: Yes, I guess this module will be created with all the plots we need to do for the `Coreg` functions!

examples/plot_nonstationary_error.py
rhugonnet: Right now, I've parted things this way:
**Section** = Background + Equations + One-liner example of functionality pointing to example
**Example** = Fully commented workflow using the one-liner
With one example per one-liner command referenced in the documentation sections.

We can move things around later, it should be quite easy once the content is there!

xdem/spstats.py
rhugonnet: Raised in #191

xdem/spstats.py
rhugonnet: Raised in #190

xdem/spstats.py
rhugonnet: Clarified

docs/source/intro.rst
```diff
@@ -0,0 +1,75 @@
...
+Accuracy and precision
+**********************
+
+Both `accuracy and precision <https://en.wikipedia.org/wiki/Accuracy_and_precision>`_ are important factors to account for when analyzing DEMs:
+
+- the **accuracy** (systematic error) of a DEM describes how close a DEM is to the true location of measured elevations on the Earth's surface,
+- the **precision** (random error) of a DEM describes the typical spread of its error in measurement, independently of a possible bias from the true positioning.
+
+.. figure:: imgs/precision_accuracy.png
+    :width: 800
```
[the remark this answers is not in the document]
rhugonnet: Changed to 80%, works well :)

docs/source/intro.rst
rhugonnet: Changed with the official order!

xdem/spstats.py
rhugonnet: Added

examples/plot_nonstationary_error.py
```diff
@@ -0,0 +1,213 @@
...
+# We should better define our bins to avoid sampling bins with too many or too few samples. For this, we can partition
+# the data in quantiles in :func:`xdem.spatialstats.nd_binning`.
+# Note: we need a higher number of bins to work with quantiles and still resolve the edges of the distribution. Thus, as
+# with many dimensions the N dimensional bin size increases exponentially, we avoid binning all variables at the same
+# time and instead bin one at a time.
+# We define 1000 quantile bins of size 0.001 (equivalent to 0.1% percentile bins) for the profile curvature:
+
+df = xdem.spatialstats.nd_binning(values=dh.data, list_var=[profc], list_var_names=['profc'],
+                                  statistics=['count', np.nanmedian, xdem.spatialstats.nmad],
+                                  list_var_bins=[[np.nanquantile(profc,0.001*i) for i in range(1001)]])
```
erikmannerfelt, line 108: `np.nanquantile(profc, np.linspace(0, 1, 1000))` is slightly easier syntax!

rhugonnet: you're right, I changed all of those!

rhugonnet: in fa471c9

adehecq:
@rhugonnet is there anything left to be done for this PR or is it ready for approval?

rhugonnet:
> @rhugonnet is there anything left to be done for this PR or is it ready for approval?

Almost done here :smile:! @adehecq 
I have edited the initial comment with a summary of everything contained by the PR. Most things were highly dependent on one another (plotting/documentation, mostly), so it was hard to split down into several PRs.

I have a very short to-do-list of minor fixes to add, should be done in a couple hours. If you are happy with the rest, it's ready for approval!

docs/source/code/spatialstats.py
```diff
@@ -0,0 +1,34 @@
+"""Code example for spatial statistics"""
+import xdem
+import geoutils as gu
+import numpy as np
+#
+# # Load data
+# ddem = gu.georaster.Raster(xdem.examples.get_path("longyearbyen_ddem"))
+# glacier_mask = gu.geovector.Vector(xdem.examples.get_path("longyearbyen_glacier_outlines"))
+# mask = glacier_mask.create_mask(ddem)
+#
```
adehecq, line 10: Is it normal that it's all commented?

examples/plot_vgm_error.py
```diff
@@ -0,0 +1,256 @@
...
+# Prior to differencing, the DEMs were aligned using :class:`xdem.coreg.NuthKaab` as shown in
+# the :ref:`sphx_glr_auto_examples_plot_nuth_kaab.py` example. We later refer to those elevation differences as *dh*.
+
+dh = xdem.DEM(xdem.examples.get_path("longyearbyen_ddem"))
+glacier_outlines = gu.Vector(xdem.examples.get_path("longyearbyen_glacier_outlines"))
+mask_glacier = glacier_outlines.create_mask(dh)
+
+# %%
+# We remove values on glacier terrain in order to isolate stable terrain, our proxy for elevation measurement errors.
+dh.data[mask_glacier] = np.nan
```
adehecq, line 47: It's bad practice to use NaNs with the masked arrays. The values should be masked instead. I just read on the numpy.ma module that the best way to do it is
```
dh.data[mask_glacier] = np.ma.masked
```
The only issue is that dh.data.data is unchanged, so if one were e.g. to save the new array, the mask would not be preserved. To ensure that, one must also change the data attribute:
```
dh.data.data[mask_glacier] = dh.nodata
```
Which is not very intuitive I must admit.

examples/plot_vgm_error.py
```diff
@@ -0,0 +1,256 @@
...
+
+# %%
+# We estimate the average per-pixel elevation measurement error on stable terrain, using both the standard deviation
+# and normalized median absolute deviation. For this example, we do not account for the non-stationarity in elevation
+# measurement errors quantified in :ref:`sphx_glr_auto_examples_plot_nonstationary_error.py`.
+print('STD: {:.2f}'.format(np.nanstd(dh.data))+' meters.')
```
adehecq, line 53: It's clearer if you do it in one string:
`'STD: {:.2f} meters'.format(np.nanstd(dh.data))`

adehecq: Same below

examples/plot_vgm_error.py
```diff
@@ -0,0 +1,256 @@
...
+print('NMAD: {:.2f}'.format(xdem.spatialstats.nmad(dh.data))+' meters.')
+
+# %%
+# The two measures of dispersion are quite similar showing that, on average, there is a small influence of outliers on the
+# elevation differences. The per-pixel precision is about :math:`\pm` 2.5 meters.
+# **Does this mean that every pixel has an independent measurement error of** :math:`\pm` **2.5 meters?**
+# Let's plot the elevation differences to visually check the quality of the data.
+plt.figure(figsize=(8, 5))
+plt_extent = [
```
adehecq, line 62: You can do it all in one-line with:
```
plt.figure(figsize=(8, 5))
dh.show(ax=plt.gca(), cmap='RdYlBu', vmin=-4, vmax=4, cb_title='Elevation differences (m)')
```
(ok two lines are needed if you want to set the figure size...)

examples/plot_vgm_error.py
```diff
@@ -0,0 +1,256 @@
...
+# differences (blue and red, respectively) appear correlated over large distances.
+#
+# Conclusion: **These correlated errors need to be quantified to reliably estimate elevation measurement errors for a
+# sum, or average of elevation differences samples for a specific surface area**.
+
+# %%
+# Additionally, we notice that the elevation differences are still polluted by unrealistically large elevation
+# differences near glaciers, probably because the glacier inventory is more recent than the data, and the outlines are too small.
+# To remedy this, we filter large elevation differences outside 4 NMAD.
+dh.data[np.abs(dh.data) > 4 * xdem.spatialstats.nmad(dh.data)] = np.nan
```
adehecq, line 85: Same comment as before regarding the use of NaNs. You could also possibly use np.ma functions directly like [masked_outside](https://numpy.org/doc/stable/reference/generated/numpy.ma.masked_outside.html#numpy.ma.masked_outside)

adehecq: Or you can also use the newly added "buffer" option in create_mask! :smiley:

examples/plot_vgm_error.py
```diff
@@ -0,0 +1,256 @@
...
+plt.show()
+
+# %%
+# To quantify the spatial correlation of the data, we sample an empirical variogram.
+# The empirical variogram describes the variance between the elevation differences of pairs of pixels depending on their
+# distance. This distance between pairs of pixels if referred to as spatial lag.
+# To perform this procedure effectively, we use improved methods that provide efficient pairwise sampling methods for
+# large grid data in `scikit-gstat <https://mmaelicke.github.io/scikit-gstat/index.html>`_, which are encapsulated
+# conveniently by :func:`xdem.spatialstats.sample_multirange_variogram`:
+df = xdem.spatialstats.sample_multirange_variogram(
```
adehecq, line 108: As discussed in PR #175, the functions in spatial_stats could be modified to accept either an array or a Raster.

examples/plot_vgm_error.py
```diff
@@ -0,0 +1,256 @@
...
+# to later compare to variogram models.
+# Another solution is to split the variogram plot into subpanels, each with its own linear scale:
+
+# %%
+# **Log scale:**
+xdem.spatialstats.plot_vgm(df, xscale='log')
+
+# %%
+# **Subpanels with linear scale:**
+xdem.spatialstats.plot_vgm(df, xscale_range_split=[100, 1000, 10000])
```
adehecq, line 130: Fantastic tools! Make variograms seem so easy !!

examples/plot_vgm_error.py
```diff
@@ -0,0 +1,256 @@
...
+# **But, in this case, the error is still too small. Why?**
+# The small size of the sampling area against the very large range of the noise means that we might no verify the
+# assumption of second-order stationarity (see :ref:`spatialstats`). We might be missing even longer range correlations
+# in our analysis, due to the limits of the variogram sampling.
+# In other words, a small part of the variance could be fully correlated over a large part of the grid, i.e. have a
+# vertical bias.
+# As a first guess for this, let's examine the difference between mean and median to gain some insight on the central
+# tendency of our sample:
+
+diff_med_mean = np.nanmean(dh.data.data)-np.nanmedian(dh.data.data)
```
adehecq, line 215: Why are you using dh.data.data? This means you're ignoring the mask (so you're including glaciers too).
This made me just realize that the test data set has NaNs ! This should be fixed, and it should be set to nodata instead. Added it as issue #196.

adehecq: Without the NaNs you could do `dh.data.mean() - np.ma.median(dh.data)`.

examples/plot_vgm_error.py
```diff
@@ -0,0 +1,256 @@
...
+plt.xlabel('Averaging area (km²)')
+plt.ylabel('Uncertainty in the mean elevation difference (m)')
+plt.xscale('log')
+plt.yscale('log')
+plt.legend()
+
+# %%
+# Our final estimation is now very close to the empirical error estimate.
+# Take-home points:
+# 1. Long-range correlations are very important to reliably estimate measurement errors integrated in space, even if they have a small partial sill (correlated variance)!
```
adehecq, line 255: There is no breakline in the rendering.

adehecq:
I only looked at the documentation, not the core code, because with the renaming it's difficult to know exactly what is new and what is not. 
I have a few comments on the documentation, but other than that, it looks fantastic !

adehecq (dismissed):

docs/source/code/spatialstats.py
rhugonnet: Fixed in c30dba3

examples/plot_vgm_error.py
rhugonnet: I think it will be more practical if we do a specific PR for this, as specified in #178

examples/plot_vgm_error.py
rhugonnet: =D
The new sampling makes all the difference in terms of speed, and quality of results :+1:

examples/plot_vgm_error.py
rhugonnet: Yes it has some!
```
np.count_nonzero(np.isnan(dh.data))
Out[14]: 4630
```

erikmannerfelt:
99 commits! You can do it!!

erikmannerfelt:
Is it correct that this PR is linked to #152? This would mean that the issue gets closed when this PR is merged.

rhugonnet:
> Is it correct that this PR is linked to #152? This would mean that the issue gets closed when this PR is merged.

It replaces uses with the subsampling function in `spatialstats.py`, but maybe not in other modules. 
@adehecq Where else were you thinking we should update the subsampling when you raised this issue?

rhugonnet:
> 99 commits! You can do it!!

It's done!! :partying_face: :balloon:

adehecq:
> > Is it correct that this PR is linked to #152? This would mean that the issue gets closed when this PR is merged.
> 
> It replaces uses with the subsampling function in `spatialstats.py`, but maybe not in other modules.
> @adehecq Where else were you thinking we should update the subsampling when you raised this issue?

The two lines were this needed replacement are mentioned in the issue. So you can check that both are ok now.

rhugonnet:
> @rhugonnet, with this PR, there is now a minimum version requirement for scikit-gstat. Do you know which? If so, can you specify that in the environment file?

@erikmannerfelt Done!
If it's all good for you, I think we can merge this PR.
There's still some TODOs left in the documentation, but those can be added step by step in the future!

erikmannerfelt (approved):
Great job!

examples/plot_standardization.py
```diff
@@ -0,0 +1,212 @@
...
+slope_arr = slope[~mask_glacier]
+planc_arr = planc[~mask_glacier]
+profc_arr = profc[~mask_glacier]
+maxc_arr = np.maximum(np.abs(planc_arr),np.abs(profc_arr))
+
+# Remove large outliers
+dh_arr[np.abs(dh_arr) > 4 * xdem.spatialstats.nmad(dh_arr)] = np.nan
+
+# Define bins for 2D binning
+custom_bin_slope = np.unique(np.concatenate([np.quantile(slope_arr,np.linspace(0,0.95,20)),
```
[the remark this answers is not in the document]
rhugonnet: Haha, I think the other one is clearer for people who are not crunching Python everyday

xdem/spatialstats.py
```diff
@@ -0,0 +1,1631 @@
...
+    else:
+        # Aggregating several skgstat Variogram after iterative subsampling of specific points in the Raster
+        get_variogram = _aggregate_pdist_empirical_variogram
+
+    return get_variogram(**argdict)
+
+
+def sample_empirical_variogram(values: Union[np.ndarray, RasterType], gsd: float = None, coords: np.ndarray = None,
+                               subsample: int = 10000, subsample_method: str = 'cdist_equidistant',
+                               n_variograms: int = 1, n_jobs: int = 1, verbose=False,
```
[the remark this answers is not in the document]
rhugonnet: Opened #204

examples/plot_nonstationary_error.py
```diff
@@ -0,0 +1,211 @@
...
+
+Digital elevation models have a precision that can vary with terrain and instrument-related variables. However, quantifying
+this precision is complex and non-stationarities, i.e. variability of the measurement error, has rarely been
+accounted for, with only some studies that used arbitrary filtering thresholds on the slope or other variables (see :ref:`intro`).
+
+Quantifying the non-stationarities in elevation measurement errors is essential to use stable terrain as a proxy for
+assessing the precision on other types of terrain (Hugonnet et al., in prep) and allows to standardize the measurement
+errors to reach a stationary variance, an assumption necessary for spatial statistics (see :ref:`spatialstats`).
+
+Here, we show an example in which we identify terrain-related non-stationarities for a DEM difference of Longyearbyen glacier.
```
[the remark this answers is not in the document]
rhugonnet: Corrected in e2b994a

examples/plot_standardization.py
```diff
@@ -0,0 +1,212 @@
...
+# a simple division:
+
+z_dh = dh.data/dh_err
+
+# %%
+# We remove values on glacierized terrain, large outliers, and perform a scale-correction.
+z_dh.data[mask_glacier] = np.nan
+z_dh.data[np.abs(z_dh.data)>4] = np.nan
+fac_std = np.nanstd(z_dh.data)
+z_dh = z_dh/fac_std
```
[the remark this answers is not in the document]
rhugonnet: Clarified in e2b994a

PR #161, opened 2021-06-30
Store ddem example to minimize test/doc computing times

rhugonnet:
### Update examples.py

Added `process_coregistered_example` to `xdem/examples.py` that works similarly as `download_longyearbyen_examples`.
To differentiate the two when loading data from `examples`, I have renamed `FILEPATHS` into `FILEPATHS_DATA` and added a `FILEPATHS_PROCESSED`.

Now one can load the dDEM example through: `examples.FILEPATHS_PROCESSED['longyearbyen_ddem']`. I have refactored previous usages of `FILEPATHS` into `FILEPATHS_DATA`.

@erikmannerfelt I see most of your test cases for volume/ddem and plot scripts in the documentation simply use the difference between the non-aligned Longyearbyen rasters (rather than doing a N&K like I did in spatial_stats). Should we change those to using the dDEM data now that it is available, or keep them as is?

### DIFF
--- a/xdem/examples.py
+++ b/xdem/examples.py
@@ -34,7 +34,7 @@ def download_longyearbyen_examples(overwrite: bool = False):
     :param overwrite: Do not download the files again if they already exist.
     """
     if not overwrite and all(map(os.path.isfile, list(FILEPATHS_DATA.values()))):
-        print("Datasets exist")
+        # print("Datasets exist")
         return
 
     # Static commit hash to be bumped every time it needs to be.
@@ -68,7 +68,7 @@ def download_longyearbyen_examples(overwrite: bool = False):
     # Copy the data to the examples directory.
     copy_tree(dir_name, os.path.join(EXAMPLES_DIRECTORY, "Longyearbyen","data"))
 
-def process_coregistered_example(overwrite=False):
+def process_coregistered_examples(overwrite: bool =False):
     """
        Process the Longyearbyen example files into a dDEM (to avoid repeating this in many test/documentation steps).
 
@@ -90,7 +90,7 @@ def process_coregistered_example(overwrite=False):
                 raise
 
     if not overwrite and all(map(os.path.isfile, list(FILEPATHS_PROCESSED.values()))):
-        print("Processed data exists")
+        # print("Processed data exists")
         return
 
     download_longyearbyen_examples(overwrite=False)
@@ -112,3 +112,19 @@ def process_coregistered_example(overwrite=False):
     # Save it so that future calls won't need to recreate the file
     mkdir_p(os.path.dirname(FILEPATHS_PROCESSED['longyearbyen_ddem']))
     diff.save(FILEPATHS_PROCESSED['longyearbyen_ddem'])
+
+def get_path(name: str) -> str:
+    """
+    Get path of example data
+    :param name: Name of test data (listed in xdem/examples.py)
+    :return:
+    """
+    if name in list(FILEPATHS_DATA.keys()):
+        download_longyearbyen_examples()
+        return FILEPATHS_DATA[name]
+    elif name in list(FILEPATHS_PROCESSED.keys()):
+        process_coregistered_examples()
+        return FILEPATHS_PROCESSED[name]
+    else:
+        raise ValueError('Data name should be one of "'+'" , "'.join(list(FILEPATHS_DATA.keys())+list(FILEPATHS_PROCESSED.keys()))+'".')
+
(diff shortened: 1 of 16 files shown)


erikmannerfelt:
It's great that you're tackling this!

I wonder about the `FILEPATHS_DATA` name, however.. Previously, I've been a little annoyed by the fact that users may need to both run a function that doesn't return anything, and call an uppercase dictionary, just to get an example file:
```python
xdem.examples.download_longyearbyen_examples()
filepath = xdem.examples.FILEPATHS[...]
```
What would have been so much nicer is a function close to what is in geoutils:
```python
filepath = xdem.examples.get_example(...)  # This downloads/creates the example if needed
```
This could also work with the processed data. If `get_example()` requires processing, it will do that under the hood.

Some pseudocode could be:
```python
def get_example(name: str) -> str:
    # Return the filepath if it's already downloaded/exists
    if existing:
          return FILEPATHS[name]  # or FILEPATHS_DATA/FILEPATHS_PROCESSED
    # Process the file if need be (if it doesn't exist)
    if name in FILEPATHS_PROCESSED:
          process_coregistered_example()
          return associated_filepath

    # If no processing is needed but it doesn't exist, download the examples.
    download_longyearbyen_examples()
    return xdem.DEM("path/to/example.tif") 
```

Under the hood, the functions and dicts that you made can be used, but in the documentation, a more simple function call like this would be nice.

Thoughts, @rhugonnet @adehecq ?

rhugonnet:
Fully agree, will modify accordingly!
We also have to keep in mind this is part of `test` and `docs`, so not really meant for people other than developers to use ;)

rhugonnet:
I wrote this:
```python
def get_path(name: str) -> str:
    """
    Get path of example data
    :param name: Name of data
    :return: 
    """
    if name in list(FILEPATHS_DATA.keys()):
        download_longyearbyen_examples()
        return FILEPATHS_DATA[name]
    elif name in list(FILEPATHS_PROCESSED.keys()):
        process_coregistered_examples()
        return FILEPATHS_PROCESSED[name]
    else:
        raise ValueError('Data name should be one of "'+'" , "'.join(list(FILEPATHS_DATA.keys())+list(FILEPATHS_PROCESSED.keys()))+'".')
```

@erikmannerfelt How does this look?

erikmannerfelt:
@rhugonnet, that looks perfect to me!! While you're at it though, could you remove/make optional the print statement in `download_longyearbyen_examples` since this will be run tens of times in the tests?

It's populating the test results everywhere, along with the "No metadata could be read from filename"!

It would be great if all test cases now used this new function instead of the old FILEPATHS system!

rhugonnet:
> @rhugonnet, that looks perfect to me!! While you're at it though, could you remove/make optional the print statement in `download_longyearbyen_examples` since this will be run tens of times in the tests?
> 
> It's populating the test results everywhere, along with the "No metadata could be read from filename"!
> 
> It would be great if all test cases now used this new function instead of the old FILEPATHS system!

On it!

rhugonnet:
> It's populating the test results everywhere, along with the "No metadata could be read from filename"!

I thought we had added a verbose setting to `SatelliteImage`?

rhugonnet:
> > It's populating the test results everywhere, along with the "No metadata could be read from filename"!
> 
> I thought we had added a verbose setting to `SatelliteImage`?

Ah I see it's a `silent` argument but that defaults to `False`! Well that's a one-line PR ;)

erikmannerfelt (approved):
Looks fantastic! Thanks, @rhugonnet!!

xdem/examples.py
```diff
@@ -57,8 +62,69 @@ def download_longyearbyen_examples(overwrite: bool = False):
...
+    copy_tree(dir_name, os.path.join(EXAMPLES_DIRECTORY, "Longyearbyen","data"))
+
+def process_coregistered_examples(overwrite: bool =False):
+    """
+       Process the Longyearbyen example files into a dDEM (to avoid repeating this in many test/documentation steps).
+
+       :param overwrite: Do not download the files again if they already exist.
+       """
+
+    def mkdir_p(out_dir):
```
adehecq, line 78: It's not so useful to define a function that is only used once... Also I'm not sure I understand what the function does, it seems to just pass on error...
Would that not solve your problem? https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

rhugonnet: I feel old sometimes when I write Python code... but not so long ago the `exist_ok` argument of `pathlib.Path.mkdir` did not exist with `parent=True` and it was a nightmare to deal with those Exceptions regularly! 
So I always had my homemade function to create a parent directoy without any error :)

PR #162, opened 2021-06-30
Add example gallery

erikmannerfelt:
The example gallery seems to be working!

Now, an example in the `examples/` directory will show up in the `Example gallery`, and will link to the API if a function from xdem is used. It will also link to the documentation of `geoutils`, `numpy` and `matplotlib` (more can be added).

There are only two weird things:

- Things defined in the `__init__` does not really work (at least now; I'm sure it's fixable). Therefore, `xdem.DEM` refers to `gu.Raster` (since it can't find the definition of `xdem.DEM`). EDIT: This is now fixed!
- Each example has to be prepended with `plot_` in order to actually execute. This is according to https://github.com/sphinx-gallery/sphinx-gallery, and might be possible to circumvent. Still, I think it's a nice way to know which functions will run and which will only render as notebook-like code.

Here's the documentation!
https://xdem-erikm.readthedocs.io/en/example-gallery/index.html

### DIFF
--- a/docs/source/_templates/module.rst
+++ b/docs/source/_templates/module.rst
@@ -1,0 +1,58 @@
+{{ fullname }}
+{{ underline }}
+
+.. automodule:: {{ fullname }}
+
+   .. contents:: Contents 
+      :local:
+
+   {% block functions %}
+   {% if functions %}
+
+   Functions
+   ---------
+
+   {% for item in functions %}
+
+   .. autofunction:: {{ item }}
+
+   .. _sphx_glr_backref_{{fullname}}.{{item}}:
+
+   .. minigallery:: {{fullname}}.{{item}}
+       :add-heading:
+
+   {%- endfor %}
+   {% endif %}
+   {% endblock %}
+
+   {% block classes %}
+   {% if classes %}
+
+   Classes
+   -------
+
+   {% for item in classes %}
+   .. autoclass:: {{ item }}
+      :members:
+
+   .. _sphx_glr_backref_{{fullname}}.{{item}}:
+
+   .. minigallery:: {{fullname}}.{{item}}
+       :add-heading:
+
+   {%- endfor %}
+   {% endif %}
+   {% endblock %}
+
+   {% block exceptions %}
+   {% if exceptions %}
+
+   Exceptions
+   ----------
+
+   .. autosummary::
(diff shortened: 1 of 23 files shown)


docs/source/conf.py
```diff
@@ -43,20 +46,35 @@
...
-autosummary_generate = True
+#autosummary_generate = True
+
+intersphinx_mapping = {
+    "geoutils": ("https://geoutils.readthedocs.io/en/latest", None),
+    "numpy": ("https://numpy.org/doc/stable", None),
+    "matplotlib": ("https://matplotlib.org/stable", None),
+}
+
+sphinx_gallery_conf = {
```
rhugonnet, line 62: reads like that wasn't easy :D

rhugonnet (approved):
Very nice :+1:

docs/source/conf.py
erikmannerfelt: Nope hahaha!

erikmannerfelt:
Thanks for the approval, @rhugonnet! I'll add some more examples today and tweak some settings, so I'll request a review again once that is ready.

erikmannerfelt:
Now, the documentation is ready for review!

Two things that I've included that are only indirectly relevant to the documentation:
* The "longyearbyen_tba_dem_coreg" key now works for `xdem.examples.get_path`
* Some small `xdem.coreg.apply_matrix` tweaks were made to increase its fidelity.

rhugonnet (approved):
I don't have much to say except that it's great!
I'm thinking that maybe in the future we'll have to refine the scope of the examples depending on their relation to other examples, or possibly to try to be more concise to avoid discouraging new users.
But right now I think it's perfect :)

erikmannerfelt:
Thank you, @rhugonnet! There are for example ways of categorizing examples into e.g. "Basic", "Intermediate" and "Advanced" categories.

docs/source/_templates/module.rst
```diff
@@ -0,0 +1,67 @@
+{{ fullname }}
```
adehecq, line 1: It's all black magic to me! ;-)

docs/source/conf.py
adehecq: So that's used to link to the documentation of these packages? Great!!

PR #165, opened 2021-07-01
Add N-dimensional linear interpolation

rhugonnet:
Return a N-D interpolant function for binned explanatory variables.
Very much like np.interp1d but in N-D.

## Example
```python
df = xdem.spstats.nd_binning(values=diff.data.flatten(),list_var=[slope.flatten(),ref.data.flatten()],list_var_names=['slope','elevation'])
fun = robust_interp_nd_binning(df, ['slope','elevation'], statistic = nmad)
```

And you get a function that can return an interpolated value for any point (slope, elevation):
```
fun((5,1000))
Out[100]: array(2.42234615)
```

```
fun((15,1000))
Out[107]: array(4.61452136)
```

Here the NMAD is **~2 m** at 5° slope, 1000 m elevation and **~4 m** 15° slope, 1000 m elevation.
If you run this on the arrays of the explanatory variables, you can get your error map directly:
```python
err_dh = fun((slope.data,elevation.data))
```
Here the elevation is just an example, in practice it doesn't explain any variability in elevation measurement error.

## What's behind it
Both `scipy.interpolate.griddata` i.e. `NearestNDInterpolator` (non-regular grids) and `scipy.interpolate.interpn` i.e. `RegularGridInterpolator` (regular grids).

**The first problem: NaNs**
Regular N-D grid interpolation does not support NaNs (even with nearest neighbour!). With irregular grids you can simply remove the NaNs and provide remaining coordinates.
**The second problem: extrapolation**
Regular N-D grid interpolation supports linear extrapolation, but irregular grids don't (see https://github.com/scipy/scipy/issues/6396)
**My solution:**
- First, I provide a N-D grid based on the N-D binning, cropping the potential NaNs on the edges to be sure a boundary is not completely filled by NaNs (especially relevant after the filtering with `min_count` to be sure each N-D bin has enough sampled to be statistically significant)
- Second, I run `griddata` with N-D **nearest** interpolation on the same grid as the input N-D binning, to fill the potential NaNs with N-D nearest extrapolation (we have to keep in mind that those values won't be used too much if they can't be sampled on the DEM anyway).
- Finally, I run `interpn` (the interpolant function `RegularGridInterpolator` to get a Callable directly), with N-D **linear interpolation** and **extrapolation**. This means that the values in between two bins will be N-D linearly interpolated properly, and values near the N-D edges that were not NaNs before will be linearly extrapolated if the function is called outside the binning grid.

## TODO
Tests
Plot in the documentation
An example in the gallery ;)

### DIFF
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -254,3 +254,96 @@ class TestBinning:
         # dataframe should contain three 1D binning of length 10 and three 2D binning of length 100 and one 2D binning of length 1000
         assert df.shape[0] == (1000 + 3 * 100 + 3 * 10)
 
+
+    def test_interp_nd_binning(self):
+
+        # check the function works with a classic input (see example
+        df = pd.DataFrame({"var1": [1, 1, 1, 2, 2, 2, 3, 3, 3], "var2": [1, 2, 3, 1, 2, 3, 1, 2, 3],
+                                "statistic": [1, 2, 3, 4, 5, 6, 7, 8, 9]})
+        arr = np.array([1, 2, 3, 4, 5, 6, 7, 8, 9]).reshape((3,3))
+        fun = xdem.spstats.interp_nd_binning(df, list_var_names=["var1", "var2"], statistic="statistic", min_count=None)
+
+        # check interpolation falls right on values for points (1, 1), (1, 2) etc...
+        for i in range(3):
+            for j in range(3):
+                x = df['var1'][3 * i + j]
+                y = df['var2'][3 * i + j]
+                stat = df['statistic'][3 * i + j]
+                assert fun((y, x)) == stat
+
+        # check bilinear interpolation inside the grid
+        points_in = [(1.5, 1.5), (1.5, 2.5), (2.5, 1.5), (2.5, 2.5)]
+        for point in points_in:
+            # the values are 1 off from Python indexes
+            x = point[0] - 1
+            y = point[1] - 1
+            # get four closest points on the grid
+            xlow = int(x - 0.5)
+            xupp = int(x + 0.5)
+            ylow = int(y - 0.5)
+            yupp = int(y + 0.5)
+            # check the bilinear interpolation matches the mean value of those 4 points (equivalent as its the middle)
+            assert fun((y + 1, x + 1)) == np.mean([arr[xlow, ylow], arr[xupp, ylow], arr[xupp, yupp], arr[xlow, yupp]])
+
+        # check bilinear extrapolation for points at 1 spacing outside from the input grid
+        points_out = [(0, i) for i in np.arange(1, 4)] + [(i, 0) for i in np.arange(1, 4)] \
+                     + [(4, i) for i in np.arange(1, 4)] + [(i, 4) for i in np.arange(4, 1)]
+        for point in points_out:
+            x = point[0] - 1
+            y = point[1] - 1
+            val_extra = fun((y + 1, x + 1))
+            # the difference between the points extrapolated outside should be linear with the grid edges,
+            # i.e. the same as the difference as the first points inside the grid along the same axis
+            if point[0] == 0:
+                diff_in = arr[x + 2, y] - arr[x + 1, y]
+                diff_out = arr[x + 1, y] - val_extra
+            elif point[0] == 4:
+                diff_in = arr[x - 2, y] - arr[x - 1, y]
+                diff_out = arr[x - 1, y] - val_extra
+            elif point[1] == 0:
+                diff_in = arr[x, y + 2] - arr[x, y + 1]
+                diff_out = arr[x, y + 1] - val_extra
+            # has to be y == 4
+            else:
+                diff_in = arr[x, y - 2] - arr[x, y - 1]
+                diff_out = arr[x, y - 1] - val_extra
+            assert diff_in == diff_out
+
+        # check if it works with nd_binning output
+        ref, diff, mask = load_ref_and_diff()
+        slope, aspect = xdem.coreg.calculate_slope_and_aspect(ref.data.squeeze())
+
+        df = xdem.spstats.nd_binning(values=diff.data.flatten(),list_var=[slope.flatten(),ref.data.flatten(),aspect.flatten()],list_var_names=['slope','elevation','aspect'])
+
+        # in 1d
+        fun = xdem.spstats.interp_nd_binning(df, list_var_names='slope')
+
+        # check a value is returned inside the grid
+        assert np.isfinite(fun([15]))
+        # check the nmad increases with slope
+        assert fun([20]) > fun([0])
+        # check a value is returned outside the grid
+        assert all(np.isfinite(fun([-5,50])))
+
+        # in 2d
+        fun = xdem.spstats.interp_nd_binning(df, list_var_names=['slope','elevation'])
+
+        # check a value is returned inside the grid
+        assert np.isfinite(fun([15, 1000]))
+        # check the nmad increases with slope
+        assert fun([20, 1000]) > fun([0, 1000])
+        # check a value is returned outside the grid
+        assert all(np.isfinite(fun(([-5, 50],[-500,3000]))))
+
+        # in 3d, let's decrease the number of bins to get something with enough samples
+        df = xdem.spstats.nd_binning(values=diff.data.flatten(),list_var=[slope.flatten(),ref.data.flatten(),aspect.flatten()],list_var_names=['slope','elevation','aspect'], list_var_bins=3)
+        fun = xdem.spstats.interp_nd_binning(df, list_var_names=['slope','elevation','aspect'])
+
+        # check a value is returned inside the grid
+        assert np.isfinite(fun([15,1000, np.pi]))
+        # check the nmad increases with slope
+        assert fun([20, 1000, np.pi]) > fun([0, 1000, np.pi])
+        # check a value is returned outside the grid
+        assert all(np.isfinite(fun(([-5, 50],[-500,3000],[-2*np.pi,4*np.pi]))))
+
+
--- a/tests/test_spstats.py
+++ b/tests/test_spstats.py
@@ -254,10 +254,9 @@ class TestBinning:
         # dataframe should contain three 1D binning of length 10 and three 2D binning of length 100 and one 2D binning of length 1000
         assert df.shape[0] == (1000 + 3 * 100 + 3 * 10)
 
-
     def test_interp_nd_binning(self):
 
-        # check the function works with a classic input (see example
+        # check the function works with a classic input (see example)
         df = pd.DataFrame({"var1": [1, 1, 1, 2, 2, 2, 3, 3, 3], "var2": [1, 2, 3, 1, 2, 3, 1, 2, 3],
                                 "statistic": [1, 2, 3, 4, 5, 6, 7, 8, 9]})
         arr = np.array([1, 2, 3, 4, 5, 6, 7, 8, 9]).reshape((3,3))
--- a/xdem/spstats.py
+++ b/xdem/spstats.py
@@ -25,20 +25,51 @@ with warnings.catch_warnings():
     import skgstat as skg
     from skgstat import models
 
-def robust_interp_nd_binning(df: pd.DataFrame, var_names: list[str], statistic : Union[str, Callable, None] = nmad, min_count: Optional[int] = 100) -> Callable:
+def interp_nd_binning(df: pd.DataFrame, list_var_names: Union[str,list[str]], statistic : Union[str, Callable[[np.ndarray],float], None] = nmad,
+                      min_count: Optional[int] = 100) -> Callable[[tuple[np.ndarray, ...]], np.ndarray]:
     """
-    Estimate an interpolated function for an N-dimensional binning.
-    Relies on scipy.ndimage.map_coordinates: see for more information on input arguments.
+    Estimate an interpolant function for an N-dimensional binning. Preferably based on the output of nd_binning.
+    For more details on the input dataframe, and associated list of variable name and statistic, see nd_binning.
 
-    :param df: dataframe of binned values according to explanatory variables
-    :param var_names: list of variable names to use
-    :param statistic: statistic to use
-    :param min_count: filter binned values with a sample count of less than this minimum count
-    :return:
+    If the variable pd.DataSeries corresponds to an interval (as the output of nd_binning), uses the middle of the interval.
+    Otherwise, uses the variable as such.
+
+    Workflow of the function:
+    Fills the no-data present on the regular N-D binning grid with nearest neighbour from scipy.griddata, then provides an
+    interpolant function that linearly interpolates/extrapolates using scipy.RegularGridInterpolator.
+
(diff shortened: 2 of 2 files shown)


xdem/spstats.py
```diff
@@ -9,19 +9,92 @@
...
+from skimage.draw import disk
+from scipy.interpolate import RegularGridInterpolator, LinearNDInterpolator, griddata
+from xdem.spatial_tools import nmad
 
 with warnings.catch_warnings():
     warnings.filterwarnings("ignore", category=DeprecationWarning)
     import skgstat as skg
     from skgstat import models
 
+def robust_interp_nd_binning(df: pd.DataFrame, var_names: list[str], statistic : Union[str, Callable, None] = nmad, min_count: Optional[int] = 100) -> Callable:
```
erikmannerfelt, line 28: The callable type should have type hints. 

For `statistic`: I guess this would be `Callable[[np.ndarray], float]`?

For the return: I guess this may take an arbitrary (read: not easy to predict at definition time) amount of arguments, so the type should be something along `Callable[..., np.ndarray]`, right?

xdem/spstats.py
```diff
@@ -9,19 +9,92 @@
...
+    """
+    Estimate an interpolated function for an N-dimensional binning.
+    Relies on scipy.ndimage.map_coordinates: see for more information on input arguments.
+
+    :param df: dataframe of binned values according to explanatory variables
+    :param var_names: list of variable names to use
+    :param statistic: statistic to use
```
erikmannerfelt, line 35: Could you add something along "The statistic needs to exist in the dataframe"

xdem/spstats.py
```diff
@@ -9,19 +9,92 @@
...
+        raise ValueError('Dataframe is empty.')
+
+    # keep only the binning in the number of dimensions corresponding to the length of the variables
+    df_sub = df[df.nd == len(var_names)]
+    if df_sub.empty:
+        raise ValueError('Dataframe does not contain the number of binned dimensions' + str(len(var_names))+' corresponding to the list of variables.')
+
+    # compute the middle values instead of bin interval
+    for var in var_names:
+        df_sub[var] = pd.IntervalIndex(df_sub[var]).mid.values
```
erikmannerfelt, line 59: Here, you assume that `df` has an `IntervalIndex`, right? What happens if it's not an `IntervalIndex`?

xdem/spstats.py
erikmannerfelt, line 35: Wait, why could `statistic` be a callable if it's never called? The only thing that is used is the `__name__` property, so wouldn't it be easier to just allow `str | None`?

xdem/spstats.py
```diff
@@ -9,19 +9,92 @@
...
     import skgstat as skg
     from skgstat import models
 
+def robust_interp_nd_binning(df: pd.DataFrame, var_names: list[str], statistic : Union[str, Callable, None] = nmad, min_count: Optional[int] = 100) -> Callable:
+    """
+    Estimate an interpolated function for an N-dimensional binning.
+    Relies on scipy.ndimage.map_coordinates: see for more information on input arguments.
+
+    :param df: dataframe of binned values according to explanatory variables
+    :param var_names: list of variable names to use
```
erikmannerfelt, line 34: So `var_names` are the "X" variables, and `statistic` is the "y" variable?

In scikit-learn, the associated terminology would be:
```python
"""
:param X: The training data column(s).
:param y: The target data column.
"""
```
In scipy and numpy, it's `x` and `y`.

I think `var_names` makes sense, but perhaps the `statistic` argument could be `target` (or something similar) instead to be more consistent with other tools?

xdem/spstats.py
erikmannerfelt, line 28: Would it be useful to allow `var_names` to just be a `str` in case of 1D interpolation?

xdem/spstats.py
```diff
@@ -9,19 +9,92 @@
...
+    :param statistic: statistic to use
+    :param min_count: filter binned values with a sample count of less than this minimum count
+    :return:
```
erikmannerfelt, line 37: Here it would be nice with a super-minimal example to show the basic functionality.

```python
"""
:examples:
    >>> df = pd.DataFrame({"var1": [1, 2, 3], "var2": [4, 5, 6], "target": [5, 7, 9]}, min_count=None)
    >>> fun = robust_interp_nd_binning(df, var_names=["var1", "var2"], statistic="target")
    >>> fun((1, 4))
        array(5.)
```
This is just what I assume would work. I haven't tried it.

You can try it with:
```bash
python -m doctest xdem/spstats.py
```

I think such an example (or one that actually works!) would be nice to show the absolute basics of the function!

xdem/spstats.py
```diff
@@ -9,19 +9,92 @@
...
+        shape.append(len(bmid))
+
+    # griddata first to perform nearest interpolation with NaNs (irregular grid)
+    # valid values
+    vals = vals[ind_valid]
+    # coordinates of valid values
+    points_valid = tuple([df_sub[var].values[ind_valid] for var in var_names])
+    # grid coordinates
+    points_grid = tuple([df_sub[var].values for var in var_names])
+    vals_grid = griddata(points_valid, vals, points_grid, method='nearest')
```
erikmannerfelt, line 88: Does one always want nearest neighbour interpolation here or is this worth being an argument?

xdem/spstats.py
erikmannerfelt, line 37: Missing return description.

erikmannerfelt (commented):
If I understand this function correctly, I'm sure we'll have 520359802 uses for it! Great!! As usual, I have some questions regarding how generic the function can be made. Also, (as you mention in the TODO), I'm looking forward to some documentation!

xdem/spstats.py
erikmannerfelt: Oh, and what happens if it's `None`?

xdem/spstats.py
rhugonnet: Thanks for nothign this!
For the return, it's a tuple of any number of arrays, how can we write this? `Callable[tuple[..., np.ndarray]]` ?

erikmannerfelt: If it's always a tuple of ndarrays, it should be:
```python
Callable[[tuple[np.ndarray, ...]], np.ndarray]
```
The `Callable` always takes a list as the first (indexer/argument??), whereby it should be a tuple of arbitrary length containing np.ndarrays: `tuple[np.ndarray, ...]`. The next (indexer/argument) shows the return type: a `np.ndarray`.

xdem/spstats.py
rhugonnet: I assume this is always used after `nd_binning`, which always returns an interval vector but not stored as index (because `MultiIndexes` are a bit of a nightmare to deal with...).

I ran into the problem that MultiIndexes are really hard to manipulate, and if I do:
```python
bins = (2,3,4,5)
df = pd.DataFrame()
df['variable'] = pd.IntervalIndex.from_breaks(bins)
```
then the interval is stored as a vector but you can't access it with the usual methods, for instance:
```python
df.variable.mid
```
does not work.
This is why I have to write:
```python
pd.IntervalIndex(df.slope).mid
```
which works.
If you have any ideas to simplify this!

xdem/spstats.py
rhugonnet: Here again it's for consistency with the input `statistic` of `nd_binning`.
The name might not be obvious to user, more complex to input, for instance: `count` name is `count` but `np.nanmedian` is `nanmedian`.

xdem/spstats.py
erikmannerfelt: Hmm... Not sure I 100% understand the context, but what about just type checking the index? 

If it's an `IntervalIndex`, take the mid, and if not, just take the normal version? Or check if `hasattr(df.variable, "mid")`?

xdem/spstats.py
rhugonnet: It's a bit different here as we are doing binned statistics (not regression), so I try to stay consistent with `scipy.binned_statistic_dd` which uses the terminology `samples`, `values` and `statistic`.

xdem/spstats.py
rhugonnet: Will add!

xdem/spstats.py
rhugonnet: Always nearest, see description of the workflow at the top.

xdem/spstats.py
rhugonnet: Perfect, thanks a lot!

xdem/spstats.py
erikmannerfelt: Ah right.. But still, the "statistic" is the target variable for the interpolation. When I started reading, I thought "statistic" was the error function to use in the error minimization behind the scenes!

erikmannerfelt: I'm sure this confusion can be circumvented either by renaming or by elaborating a bit on the documentation.

xdem/spstats.py
erikmannerfelt: Could this be elaborated upon in the docstring?

xdem/spstats.py
erikmannerfelt: Ah! I didn't read properly!

rhugonnet:
@erikmannerfelt Because N-D binning (and dataframes) can take so many shapes, I simply stated that the structure has to match the output of `nd_binning` already documented in that function.

xdem/spstats.py
rhugonnet: Added with `isinstance(df_sub[var].dtype,pd.IntervalDtype)`

xdem/spstats.py
rhugonnet: Added in 26d3b9e

rhugonnet:
> @erikmannerfelt Because N-D binning (and dataframes) can take so many shapes, I simply stated that the structure has to match the output of `nd_binning` already documented in that function.

In the end I relaxed the structure a bit, allowing more easily for user-created dataframes.
The example should help a lot with this ;)

xdem/spstats.py
```diff
@@ -25,6 +27,109 @@
...
+    :param df: dataframe with statistic of binned values according to explanatory variables (preferably output of nd_binning)
+    :param list_var_names: explanatory variable data series to select from the dataframe (containing interval or float dtype)
+    :param statistic: statistic to interpolate, stored as a data series in the dataframe
+    :param min_count: minimum number of samples to be used as a valid statistic (replaced by nodata)
+    :return: N-dimensional interpolant function
+
+    :examples
+    Using a dataframe created from scratch
+    >>> df = pd.DataFrame({"var1": [1, 1, 1, 2, 2, 2, 3, 3, 3], "var2": [1, 2, 3, 1, 2, 3, 1, 2, 3], "statistic": [1, 2, 3, 4, 5, 6, 7, 8, 9]})
+        In 2 dimensions, the statistic array looks like this
```
erikmannerfelt, line 52: Have you tried running `python -m doctest xdem/spstats.py`? The non-">>>" characters are the expected outputs, so this will probably fail.

xdem/spstats.py
```diff
@@ -25,6 +27,109 @@
...
+        array([
+            [1, 2, 3],
+            [4, 5, 6],
+            [7, 8, 9]
+            ])
+    >>> fun = interp_nd_binning(df, list_var_names=["var1", "var2"], statistic="statistic", min_count=None)
+    Right on point.
+    >>> fun((2, 2))
+        array(5.)
+    Interpolated linearly inside the 2D frame.
```
erikmannerfelt, line 62: These can be expressed as comments after the function call, e.g. `>>> fun((1.5, 1.5))  # Interpolated linearly inside the 2D frame.`

xdem/spstats.py
```diff
@@ -25,6 +27,109 @@
...
+    # compute the middle values instead of bin interval if the variable is a pandas interval type
+    for var in list_var_names:
+        if isinstance(df_sub[var].dtype,pd.IntervalDtype):
+            df_sub[var] = pd.IntervalIndex(df_sub[var]).mid.values
+        # otherwise, leave as is
+
+    # keep only rows where the binning data exists for those variables (meaning they were used)
+    df_sub = df_sub[np.logical_and.reduce([np.isfinite(df_sub[var].values) for var in list_var_names])]
+    if df_sub.empty:
+        raise ValueError('Dataframe does not contain a nd binning with the variables corresponding to the list of variables.')
```
erikmannerfelt, line 96: Could this be more descriptive perhaps? This basically an implicit check for nans, so if this is empty, it means "dataframe has not a single valid row", right? Perhaps the `min_count is not None` logic could be placed above, and then the ValueError could say something along `f"Dataframe does not contain enough valid rows: {df_sub.shape[0]} {'threshold: ' + str(min_count) if min_count is not None else ''}."`

xdem/spstats.py
rhugonnet: Fixed in 84af672
Out of curiosity, are those run at some points by `pytest` or anything in the CI?

xdem/spstats.py
rhugonnet: Done in 84af672

xdem/spstats.py
erikmannerfelt: No but they should! I will add an issue for it.

erikmannerfelt: Added #171 .

rhugonnet:
Tests added, all works as intended :)

For the documentation, I think it'll go better in the documentation PR so that I can add to what I've already modified in the docs, and add the plots. 
So we should merge this PR first!

xdem/spstats.py
rhugonnet: The `min_count` replaces the `statistic` values as NaNs, not the binning values (e.g. slope of 5, 10, 15). This assertion simply verifies that there is binned values to work with.
I can add a check for the `statistic` after filtering with `min_count`

rhugonnet: Added in 8ed31e8

erikmannerfelt (approved):
Coolcool! This looks great!

rhugonnet:
> Coolcool! This looks great!

What, no comments on tests?! I better merge quickly before @adehecq comes around :p

tests/test_spstats.py
```diff
@@ -254,3 +254,96 @@ def percentile_80(a):
         # dataframe should contain three 1D binning of length 10 and three 2D binning of length 100 and one 2D binning of length 1000
         assert df.shape[0] == (1000 + 3 * 100 + 3 * 10)
 
+
```
erikmannerfelt, line 257: THERE ARE TWO LINES HERE WHEN THERE SHOULD BE ONLY ONE. REJECTED

erikmannerfelt (changes requested):

erikmannerfelt:
Careful what you ask for, @rhugonnet !!

erikmannerfelt (approved):
Fantastic! Now I'm happy again! 😈

tests/test_spstats.py
rhugonnet: Haha, so I already made a mistake before even writing the first line

erikmannerfelt: The blank lines are the only bad part of your code!

PR #166, opened 2021-07-01
Fix `load=False` bug with DEM class + silent option `True` as default

### LINKED ISSUE #164: `xdem.DEM(..., load_data=False)` is not working
Example:
```python
>>> import xdem
>>> xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
>>> xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"), load_data=False)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/erik/Projects/ETH/xdem/xdem/dem.py", line 75, in __init__
    if self.nbands > 1:
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

@rhugonnet, any ideas?

rhugonnet:
All in the title
Resolves #164

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -1,0 +1,55 @@
+# This workflow will install Python dependencies, run tests and lint with a single version of Python
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: build
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    name: ${{ matrix.os }}, python ${{ matrix.python-version}}
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: ["ubuntu-latest"]
+        python-version: ["3.7", "3.8", "3.9"]
+
+
+    # Run all shells using bash (including Windows)
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    # Set up the conda-forge environment with mamba
+    - name: Set up Python ${{ matrix.python-version }} and dependencies
+      uses: conda-incubator/setup-miniconda@v2
+      with:
+        auto-update-conda: true
+        python-version: ${{ matrix.python-version }}
+        mamba-version: "*"
+        channels: conda-forge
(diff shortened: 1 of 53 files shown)


rhugonnet:
What, changing three lines are breaking 2 tests :astonished: ? We need to look at this tomorrow

erikmannerfelt:
Ah, it's due to the new geoutils.

Could you please change this to `np.array(dems)` instead of `np.asarray(dems)`
https://github.com/GlacioHack/xdem/blob/46a3edc3436e7b1b60317946441e2f445f886cf6/xdem/demcollection.py#L50

I think this is the only fix that needs to be done.

erikmannerfelt:
I had to tackle the same issue in #147. 

Here's the diff from my fix:
```diff
diff --git a/xdem/demcollection.py b/xdem/demcollection.py
index 522c006..4c8b094 100644
--- a/xdem/demcollection.py
+++ b/xdem/demcollection.py
@@ -47,13 +47,14 @@ class DEMCollection:

         # Find the sort indices from the timestamps
         indices = np.argsort(self.timestamps.astype("int64"))
-        self.dems = np.asarray(dems)[indices]
+        self.dems = np.empty(len(dems), dtype=object)
+        self.dems[:] = [dems[i] for i in indices]
         self.ddems: list[xdem.dDEM] = []
         # The reference index changes place when sorted
         if isinstance(reference_dem, int):
             self.reference_index = np.argwhere(indices == reference_dem)[0][0]
         elif isinstance(reference_dem, gu.georaster.Raster):
-            self.reference_index = np.argwhere(self.dems == reference_dem)[0][0]
+            self.reference_index = [i for i, dem in enumerate(self.dems) if dem is reference_dem][0]

         if outlines is None:
             self.outlines: dict[np.datetime64, gu.geovector.Vector] = {}
```

erikmannerfelt:
Also, could you remove the `xdem.DEM(..., silent=True)` flags that have had to be placed in the documentation everywhere?

It's fine if we just add this as an issue as well.

rhugonnet:
> Also, could you remove the `xdem.DEM(..., silent=True)` flags that have had to be placed in the documentation everywhere?
> 
> It's fine if we just add this as an issue as well.

On it!

erikmannerfelt (approved):
Looks great!!

Issue #171, opened 2021-07-02
Add doctests to test suite


erikmannerfelt:
Doctests are great to show minimal examples:
```python
def add(a, b):
     """
     Simply adds two numbers

     :examples:
         >>> add(1, 2)
         4
     """
     return a + b
```
This will gracefully fail, as the example doesn't return the "expected" number:
```md
**********************************************************************
File "/home/erik/hello.py", line 6, in hello.add
Failed example:
    add(1, 2)
Expected:
    4
Got:
    3
**********************************************************************
1 items had failures:
   1 of   1 in hello.add
***Test Failed*** 1 failures.
```

There are a handful of doctests already in xdem, but these are not run when running pytest.

That should change!!

adehecq:
How do you do that ?!?

erikmannerfelt:
Fixed in #172

PR #172, opened 2021-07-03
Run doctests with pytest

erikmannerfelt:
Solves #171. Doctests now need to pass in the CI.

@rhugonnet, you might want to have a look at one of the doctests in your recent PR. I had to change the return value!

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,6 @@
+[tool.pytest.ini_options]
+addopts = "--doctest-modules"
+testpaths = [
+	"tests",
+	"xdem"
+]
--- a/xdem/spstats.py
+++ b/xdem/spstats.py
@@ -68,7 +68,7 @@ def interp_nd_binning(df: pd.DataFrame, list_var_names: Union[str,list[str]], st
 
     # Extrapolated linearly outside the 2D frame.
     >>> fun((-1, 1))
-    array(-5.)
+    array(-1.)
     """
     # if list of variable input is simply a string
     if isinstance(list_var_names,str):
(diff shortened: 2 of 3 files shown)


xdem/spstats.py
```diff
@@ -68,7 +68,7 @@ def interp_nd_binning(df: pd.DataFrame, list_var_names: Union[str,list[str]], st
 
     # Extrapolated linearly outside the 2D frame.
     >>> fun((-1, 1))
-    array(-5.)
+    array(-1.)
```
rhugonnet, line 71: weird, it worked for me with the other value!

pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[tool.pytest.ini_options]
+addopts = "--doctest-modules"
+testpaths = [
```
rhugonnet, line 3: does this mean we'll potentially test doctests in `xdem/tests` as well?

xdem/spstats.py
erikmannerfelt: Huh!

pyproject.toml
erikmannerfelt: Yes!

erikmannerfelt: The testpaths is there because the `examples/` directory will otherwise be run, and I have no idea why, but then it fails! (there are no doctests there, but it still fails...)

xdem/spstats.py
rhugonnet: oh yes I fixed a bug which affected the dimension order since then, when writing the tests.
so this is now the result along the `x` dimension, but before it was along `y`: if you try `fun((1,-1))` you'll get `array(-5.)`.

So all good! And I just exemplified why we need to add the doctests! :sweat_smile:

erikmannerfelt: Great!

rhugonnet (approved):
Perfect!

pyproject.toml
rhugonnet: Weird one!

Issue #174, opened 2021-07-03
Allow `Raster` (sub-)classes as inputs for `Coreg.fit()` and `Coreg.apply()`


erikmannerfelt:
We want to do two things: make things simple, and make them right. Allowing numpy arrays for DEMs is definitely _right_, as this can be used with `rasterio` and other frameworks. The _simple_ part, however could maybe be improved by allowing `Raster` (sub-)classes as inputs for coregistration.

## Data
```python
>>> import xdem
>>> import geoutils as gu
>>> ref_dem = xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
>>> tba_dem = xdem.DEM(xdem.examples.get_path("longyearbyen_tba_dem")).reproject(ref_dem)  # Reprojection would technically only be required in the "current syntax"
>>> glaciers = gu.Vector(xdem.examples.get_path("glacier_outlines"))
```

## The current syntax
If one wants to coregister a DEM and save it, these are the steps that are involved:
```python
>>> coreg = xdem.coreg.NuthKaab()
>>> coreg.fit(ref_dem.data, tba_dem.data, inlier_mask=~glaciers.create_mask(ref_dem), transform=ref_dem.transform)
>>> aligned_dem_data = coreg.apply(tba_dem.data, transform=tba_dem.transform)
>>> aligned_dem = xdem.DEM.from_array(aligned_dem_data, ref_dem.transform, ref_dem.crs)
>>> aligned_dem.save("path/to/dem_coreg.tif")
```
This is too long!!
#### Some issues I have with it
- The `.fit()` method needs to be called in a separate line (but this will be fixed in #173)
- `Coreg.fit()` requires the `transform` argument, which could just as well have been read from the `ref_dem` raster.
- `Coreg.apply()` returns a numpy array, which needs to be converted to a `DEM`/`Raster` to save.

#### Suggestions
- A `Raster` (sub-)class should be possible to provide `Coreg.fit()`, whereby the `transform` argument should not be needed. A nearest neighbor reprojection could also be made to validate that they are on the same grid.
- If a `Raster` (sub-)class is given to `Coreg.apply()`, it should also return a `Raster` (sub-)class of the same type.

## Potential future syntax
```python
>>> coreg = xdem.coreg.NuthKaab().fit(ref_dem, tba_dem, inlier_mask=~glaciers.create_mask(ref_dem))
>>> aligned_dem = coreg.apply(tba_dem)
[...]

PR #175, opened 2021-07-03
Allow `Raster`s as inputs for `Coreg.fit()` and `Coreg.apply()`

erikmannerfelt:
Solves #174.

See the issue for the new syntax!

### DIFF
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -380,7 +380,7 @@ class TestCoregClass:
         zcorr.fit(self.ref.data, scaled_dem)
 
         # Apply the correction
-        unscaled_dem = zcorr.apply(scaled_dem, None)
+        unscaled_dem = zcorr.apply(scaled_dem, self.ref.transform)
 
         # Make sure the difference is now minimal
         diff = (self.ref.data - unscaled_dem).filled(np.nan)
@@ -409,7 +409,7 @@ class TestCoregClass:
 
         # Try the fit now with the messed up DEM as reference.
         zcorr.fit(dem_with_nans, scaled_dem)
-        unscaled_dem = zcorr.apply(scaled_dem, None)
+        unscaled_dem = zcorr.apply(scaled_dem, self.ref.transform)
         diff = (dem_with_nans - unscaled_dem).filled(np.nan)
         assert np.abs(np.nanmedian(diff)) < 0.05
 
@@ -421,7 +421,7 @@ class TestCoregClass:
         zcorr_nonlinear.fit(dem_with_nans, scaled_dem)
 
         # Make sure the difference is minimal
-        unscaled_dem = zcorr_nonlinear.apply(scaled_dem, None)
+        unscaled_dem = zcorr_nonlinear.apply(scaled_dem, self.ref.transform)
         diff = (dem_with_nans - unscaled_dem).filled(np.nan)
         assert np.abs(np.nanmedian(diff)) < 0.05
 
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -18,6 +18,7 @@ import pytransform3d.transformations
 with warnings.catch_warnings():
     warnings.simplefilter("ignore")
     from xdem import coreg, examples, spatial_tools, misc
+    import xdem
 
 
 def load_examples() -> tuple[gu.georaster.Raster, gu.georaster.Raster, gu.geovector.Vector]:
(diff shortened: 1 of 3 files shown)


tests/test_coreg.py
```diff
@@ -484,6 +485,57 @@ def test_blockwise_coreg(self, pipeline, subdivision):
         # Check that offsets were actually calculated.
         assert np.sum(np.abs(np.linalg.norm(stats[["x_off", "y_off", "z_off"]], axis=0))) > 0
 
+    def test_coreg_raster_and_ndarray_args(_) -> None:
```
rhugonnet, line 488: Is `_` playing a role in the test function arguments? I've only ever seen `self`

xdem/coreg.py
```diff
@@ -468,6 +469,20 @@ def fit(self, reference_dem: Union[np.ndarray, np.ma.masked_array],
         if weights is not None:
             raise NotImplementedError("Weights have not yet been implemented")
 
+        # If both DEMs are Rasters, validate that 'dem_to_be_aligned' is in the right grid. Then extract its data.
+        if isinstance(dem_to_be_aligned, gu.Raster) and isinstance(reference_dem, gu.Raster):
+            dem_to_be_aligned = dem_to_be_aligned.reproject(reference_dem, silent=True).data
```
rhugonnet, line 474: after seeing so many `silent=True`, I'm wondering if we'd not prefer `silent=True` by default on `Raster.reproject` as well!

xdem/coreg.py
```diff
@@ -511,8 +526,18 @@ def fit(self, reference_dem: Union[np.ndarray, np.ma.masked_array],
         # Flag that the fitting function has been called.
         self._fit_called = True
 
-    def apply(self, dem: Union[np.ndarray, np.ma.masked_array],
-              transform: rio.transform.Affine) -> Union[np.ndarray, np.ma.masked_array]:
+    @overload
```
rhugonnet, line 529: very nice

xdem/coreg.py
```diff
@@ -468,6 +469,20 @@ def fit(self, reference_dem: Union[np.ndarray, np.ma.masked_array],
...
+        if isinstance(dem_to_be_aligned, gu.Raster) and isinstance(reference_dem, gu.Raster):
+            dem_to_be_aligned = dem_to_be_aligned.reproject(reference_dem, silent=True).data
+        # If 'reference_dem' is not a Raster, just extract the data (shape validation comes later)
+        elif isinstance(dem_to_be_aligned, gu.Raster):
+            dem_to_be_aligned = dem_to_be_aligned.data
+
+        # If 'reference_dem' is a Raster, use its transform (if not given) and then extract its data.
+        if isinstance(reference_dem, gu.Raster):
+
+            if transform is None:
```
rhugonnet, line 482: I wonder if maybe we should raise a RunTimeWarning if a `RasterType` is provided but another `transform` is given and used? 
That seems quite dangerous to me, and the user should be informed! 
And, in practice, it won't happen so often that the Warning will become annoying. It's really rare to provide a different transform than that of your Raster

tests/test_coreg.py
```diff
@@ -379,7 +380,7 @@ def test_z_scale_corr(self):
         zcorr.fit(self.ref.data, scaled_dem)
 
         # Apply the correction
-        unscaled_dem = zcorr.apply(scaled_dem, None)
+        unscaled_dem = zcorr.apply(scaled_dem, self.ref.transform)
```
rhugonnet, line 383: I'm not sure I get why those needed changing?

rhugonnet (approved):
Tip top!
Birds and heat have a much more productively-positive effect on you than me :laughing:

xdem/coreg.py
erikmannerfelt: A difference is that `Raster.reproject(..., silent=True)` may actually have some unforeseen consequences: it returns a copy, so any action after that will propagate to the original raster too.

While I'm not perfectly happy with the warning that shows up here and there, it would almost have to be replaced by another protection mechanism to be completely safe....

tests/test_coreg.py
erikmannerfelt: `Raster.apply()` now raises an error if an array is provided but no transform. This makes the tools perfectly consistent (there was no previous mechanism to validate that a transform was given, so the error messages were very confusing). In this particular case, it's a bit redundant, but I think that's an okay sacrifice for consistency! If you and @adehecq disagree, I'm sure we can find another way.

xdem/coreg.py
rhugonnet: True, and it's not everyday we reproject on same CRS.
Good as is!

rhugonnet:
Actually, a big question raised by this:
With https://github.com/GlacioHack/GeoUtils/pull/210 and this PR, we now have many use cases where we can use `np.ndarray` and `Raster` objects interchangeably. But there is still several functions for which it does not work, right? Such as terrain attributes.

Should we aim to have xdem working with np.ndarrays and Raster objects interchangeably? If yes, what is left to do? (I think that, @erikmannerfelt, you have the best viewpoint on all this!)
Also, it should be crystal clear when a Raster can be provided simply as a **convenience** and a np.ndarray would suffice, or when georeferenced data is **absolutely necessary**. It will get blurry for the users if we allow to provide both everywhere for convenience, so I'm not sure how to tackle this!

erikmannerfelt:
> Actually, a big question raised by this:
> With [GlacioHack/GeoUtils#210](https://github.com/GlacioHack/GeoUtils/pull/210) and this PR, we now have many use cases where we can use `np.ndarray` and `Raster` objects interchangeably. But there is still several functions for which it does not work, right? Such as terrain attributes.
> 
> Should we aim to have xdem working with np.ndarrays and Raster objects interchangeably? If yes, what is left to do? (I think that, @erikmannerfelt, you have the best viewpoint on all this!)
> Also, it should be crystal clear when a Raster can be provided simply as a **convenience** and a np.ndarray would suffice, or when georeferenced data is **absolutely necessary**. It will get blurry for the users if we allow to provide both everywhere for convenience, so I'm not sure how to tackle this!

Since, `geoutils==0.0.3`, `Raster`s implement the array interface, so they _should_ be possible to use interchangeably with ndarrays. The tests don't cover that "_should_" part, however, and `Raster`s are not consistently returned if provided as an argument. For now, all functions that show a `Raster`/`RasterType` argument/return type hint returns a `RasterType` if given a `RasterType`, and those who don't will return an `ndarray`/`masked_array`.

Long story short, the only way to validate how interchangeable they are is to include more tests!!

xdem/coreg.py
erikmannerfelt: Added in cc0cdb3!

I accidentally included it with a merge... Jaja, I blame the 3 AM birds!

tests/test_coreg.py
erikmannerfelt: An argument needs to be provided as it is a method, but this particular method does not actually need `self`. Potentially, this should be its own function, but currently all `Coreg`-related tests are in the `TestCoregClass`, so it felt wrong to put it outside by itself!

erikmannerfelt:
I think I've fixed all of @rhugonnet 's comments.

Any other thoughts, @rhugonnet  and @adehecq ?

rhugonnet (approved):
All good for me! 
Maybe we should open an issue on the "interchangeable" discussion? 
It would be good if we could say "All xdem functions accept interchangeably `np.ndarray` and `Raster` subclass objects"

erikmannerfelt:
Already done in #178 !

Thanks, @rhugonnet !

tests/test_coreg.py
adehecq: Sounds ok to me

adehecq:
Fantastic !

PR #177, opened 2021-07-03
Use pathlib mkdir

### LINKED ISSUE #176: Use mkdir of pathlib
It's not so useful to define a function that is only used once... Also I'm not sure I understand what the function does, it seems to just pass on error...
Would that not solve your problem? https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

_Originally posted by @adehecq in https://github.com/GlacioHack/xdem/pull/161#discussion_r662980631_

rhugonnet:
To create a parent directory with integrated exception handling
Brought up in #161 
Resolves #176

### DIFF
--- a/xdem/examples.py
+++ b/xdem/examples.py
@@ -5,7 +5,6 @@ import shutil
 import tarfile
 import tempfile
 import urllib.request
-import pathlib
 from distutils.dir_util import copy_tree
 
 import geoutils as gu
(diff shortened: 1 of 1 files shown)


xdem/examples.py
```diff
@@ -113,7 +100,7 @@ def mkdir_p(out_dir):
                                 transform=reference_raster.transform, crs=reference_raster.crs)
 
     # Save it so that future calls won't need to recreate the file
-    mkdir_p(os.path.dirname(FILEPATHS_PROCESSED['longyearbyen_ddem']))
+    pathlib.Path(os.path.dirname(FILEPATHS_PROCESSED['longyearbyen_ddem'])).mkdir(parents=True, exist_ok=True)
```
erikmannerfelt, line 103: Wouldn't just `os.makedirs(..., exist_ok=True)` do the trick? Not that it matters!

erikmannerfelt (dismissed):

xdem/examples.py
rhugonnet: You are right, even better

rhugonnet: I still have those bad habits from coding in old Python

erikmannerfelt (approved):
Lovely!

PR #181, opened 2021-07-05
Fix gap bug in `xdem.volume.norm_regional_hypsometric_interpolation`

erikmannerfelt:
The lowermost half bin of the glacier was not filled because of the midpoint-centered approach; Only the midpoints of the bins were evaluated, so the lower half of the lowermost bin was out of the interpolation range!

Adding `scipy.interpolate.interp1d(..., fill_value="extrapolate")` fixes this issue, so the lower half of the lowermost bin correctly uses the model for the lowermost bin, without introducing NaNs.

I've also added a test for nans to make sure this won't happen again.

### DIFF
--- a/tests/test_volume.py
+++ b/tests/test_volume.py
@@ -283,6 +283,8 @@ class TestNormHypsometric:
         )
         assert not np.array_equal(filled_ddem, idealized_ddem)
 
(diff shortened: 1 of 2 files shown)


rhugonnet (approved):
Nice catch!

PR #183, opened 2021-07-07
Deprecate `subtract_rasters`

### LINKED ISSUE #163: Deprecate `xdem.spatial_tools.subtract_rasters`
This was a very useful function back in the days! With the addition of @adehecq's operators on Rasters, however, it is more or less redundant now.

## How to subtract a raster
Pre-May (when it was needed):
```python
dem2_reproj = dem2.reproject(dem1, resampling="...")

ddem = xdem.DEM.from_array(dem1.data - dem2_reproj.data, transform=dem1.transform, crs=dem1.crs)
```
With the (now redundant function):
```python
ddem = xdem.spatial_tools.subtract_rasters(dem1, dem2, resampling="...")
```
Today:
```python
ddem = dem1 - dem2.reproject(dem1, resampling="...")
```

There are fewer lines when not using my function and instead @adehecq's operators. Therefore, I would consider `xdem.spatial_tools.subtract_rasters` redundant.

Should we remove it, @adehecq @rhugonnet ?

erikmannerfelt:
Closes #163.

All uses of `xdem.spatial_tools.subtract_rasters()` are now gone.

I've added a `xdem.misc.deprecate()` function with associated tests. It triggers a deprecation warning up to the version where it is supposed to be removed, whereby it will raise a ValueError. This means that if the repo is still using the deprecated function by then, it won't pass CI!

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1,0 +1,64 @@
+"""Test the xdem.misc functions."""
+from __future__ import annotations
+
+import warnings
+
+import pytest
+
+import xdem
+import xdem.misc
+
+
+@pytest.mark.parametrize("deprecation_increment", [-1, 0, 1, None])
+@pytest.mark.parametrize("details", [None, "It was completely useless!", "dunnowhy"])
+def test_deprecate(deprecation_increment: int | None, details: str | None) -> None:
+    """
+    Test the deprecation warnings/errors.
+
+    If the version is larger than the current, it should warn.
+    If the version is smaller or equal, it should raise an error.
+
+    :param deprecation_increment: The version number relative to the current version.
+    """
+    warnings.simplefilter("error")
+
+    current_version = xdem.version.version
+
+    # Set the deprecation_version to be the current version plus the increment (e.g. 0.0.5 + 1 -> 0.0.6)
+    deprecation_version = (
+        current_version[:-1] + str(int(current_version.rsplit(".", 1)[1]) + deprecation_increment)
+        if deprecation_increment is not None
+        else None
+    )
+
+    @xdem.misc.deprecate(deprecation_version, details=details)
+    def useless_func() -> int:
+        return 1
+
+    should_warn = deprecation_version is None or deprecation_version > current_version
+
+    text = (
+        "Call to deprecated function 'useless_func'."
+        if should_warn
+        else f"Function 'useless_func' was deprecated in {deprecation_version}."
+    )
+
+    if details is not None:
+        text += " " + details.strip().capitalize()
+
+        if not any(text.endswith(c) for c in ".!?"):
+            text += "."
+
+    
+    if should_warn and deprecation_version is not None:
+        text += f" This functionality will be removed in version {deprecation_version}."
+    elif not should_warn:
+        text += f" Current version: {current_version}."
+
+
+    if should_warn:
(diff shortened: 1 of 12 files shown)


erikmannerfelt:
FYI, this is how the deprecation warning looks like. It has the shape "Call to deprecated XX", "description", "this will be removed in XX"
```pythn
>>> import xdem
>>> dem = xdem.DEM(xdem.examples.get_path('longyearbyen_ref_dem'))
>>> _ = xdem.spatial_tools.subtract_rasters(dem, dem)
DeprecationWarning: Call to deprecated function 'subtract_rasters'. This function is redundant after the '-' operator for rasters was introduced. Use 'dem1 - dem2.reproject(dem1, resampling_method='cubic_spline')' instead. This functionality will be removed in version 0.0.6.

```

tests/test_misc.py
```diff
@@ -0,0 +1,67 @@
...
+
+import warnings
+
+import pytest
+
+import xdem
+import xdem.misc
+
+
+@pytest.mark.parametrize("deprecation_increment", [-1, 0, 1, None])
```
rhugonnet, line 12: so this is the equivalent of 2 independent `for` loops?

xdem/misc.py
```diff
@@ -23,12 +30,76 @@ def generate_random_field(shape: tuple[int, int], corr_size: int) -> np.ndarray:
...
+            ksize=(2 * corr_size + 1, 2 * corr_size + 1),
+            sigmaX=corr_size,
+        )
+        / 255,
+        dsize=(shape[1], shape[0]),
     )
     return field
+
+
+def deprecate(removal_version: str | None = None, details: str | None = None):
```
rhugonnet, line 51: very nice addition

tests/test_misc.py
erikmannerfelt: Yes, so this one evaluates into 12 independent tests!

xdem/misc.py
```diff
@@ -23,12 +30,76 @@ def generate_random_field(shape: tuple[int, int], corr_size: int) -> np.ndarray:
...
+
+            # Add the details explanation if it was given, and make sure the sentence is ended.
+            if details is not None:
+                details_frm = details.strip()
+                if details_frm[0].islower():
+                    details_frm = details_frm[0].upper() + details_frm[1:]
+
+                text += " " + details_frm
+
+                if not any(text.endswith(c) for c in ".!?"):
```
rhugonnet, line 88: Apparently you really want those sentence to be rightly punctuated :p

rhugonnet (approved):
Amazing functionality for the deprecate function :+1: 
Without it, I could already see us completely forgetting to remove a deprecated function when changing versions, now CI is going remind us of this!! Very practical

xdem/misc.py
adehecq: Fancy!

adehecq: And including a test for deprecating functions is even fancier !!

xdem/misc.py
adehecq: ^^

adehecq (approved):
Reads all good!

erikmannerfelt:
Thank you both!

xdem/misc.py
erikmannerfelt: I don't trust myself, and even less others!!

PR #186, opened 2021-07-09
Exclude empty blocks in BlockwiseCoreg and handle warnings.

### LINKED ISSUE #185: Blockwise coregistration fails with large data gaps
If certain blocks (or chunks) created by the blockwise coregistration include only NaNs, the function fails with error:
> ValueError: Fitting failed for XX chunks:

Instead, the empty chunks should be ignored.

This is useful in many cases, as most DEMs footprint is rotated compared to the geographic grid, creating large gaps around the DEM (see illustration below).

A way to reproduce is to modify the script in the example at this line: https://github.com/GlacioHack/xdem/blob/2fb86d85ad9cf5a7476013f55a38aa9a38c7b422/examples/plot_blockwise_coreg.py#L33
Add:
```
reference_dem = reference_dem.reproject(dst_crs='EPSG:3413', dst_res=reference_dem.res, resampling='bilinear')
dem_to_be_aligned = dem_to_be_aligned.reproject(dst_ref=reference_dem, resampling='bilinear')
```
This will create a nicely rotated DEM with lots of gaps :smile:

Figure_1

erikmannerfelt:
Closes #185.

Completely empty blocks are now skipped, and are not considered for the success percentage.

Also, a few warnings may occur before coreg approaches fail, so these are now caught and are by default not shown. For debug purposes, it can be shown by enabling the `warn_failures` attribute to `True` (can also be set as an argument to the `__init__()` method), whereby each failure and warning will be shown as warnings.

### DIFF
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -485,6 +485,21 @@ class TestCoregClass:
         # Check that offsets were actually calculated.
         assert np.sum(np.abs(np.linalg.norm(stats[["x_off", "y_off", "z_off"]], axis=0))) > 0
 
+
+    def test_blockwise_coreg_large_gaps(self):
+        """Test BlockwiseCoreg when large gaps are encountered, e.g. around the frame of a rotated DEM."""
+        warnings.simplefilter("error")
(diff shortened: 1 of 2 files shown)


adehecq (approved):
Seems great! A simple fix for a much more intuitive use!

Issue #178, opened 2021-07-03
Add tests+functionality to assess the interchangeability of `Raster`s and arrays


erikmannerfelt:
Brought up by @rhugonnet in #175.

Here is the conversation:

> Actually, a big question raised by this:
> With [GlacioHack/GeoUtils#210](https://github.com/GlacioHack/GeoUtils/pull/210) and this PR, we now have many use cases where we can use `np.ndarray` and `Raster` objects interchangeably. But there is still several functions for which it does not work, right? Such as terrain attributes.
> 
> Should we aim to have xdem working with np.ndarrays and Raster objects interchangeably? If yes, what is left to do? (I think that, @erikmannerfelt, you have the best viewpoint on all this!)
> Also, it should be crystal clear when a Raster can be provided simply as a **convenience** and a np.ndarray would suffice, or when georeferenced data is **absolutely necessary**. It will get blurry for the users if we allow to provide both everywhere for convenience, so I'm not sure how to tackle this!

Since, `geoutils==0.0.3`, `Raster`s implement the array interface, so they _should_ be possible to use interchangeably with ndarrays. The tests don't cover that "_should_" part, however, and `Raster`s are not consistently returned if provided as an argument. For now, all functions that show a `Raster`/`RasterType` argument/return type hint returns a `RasterType` if given a `RasterType`, and those who don't will return an `ndarray`/`masked_array`.

Long story short, the only way to validate how interchangeable they are is to include more tests!!

PR #188, opened 2021-07-09
Allow `Raster`s as arguments -> return types for terrain attributes

erikmannerfelt:
Tackles #178 

Now, this works:
```python
import xdem

dem = xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))

slope = xdem.terrain.slope(dem)  # This is a Raster since a RasterType was given as an argument.

slope.save(...)

```

### DIFF
--- a/xdem/terrain.py
+++ b/xdem/terrain.py
@@ -8,8 +8,8 @@ import numba
 import numpy as np
 
 import xdem.spatial_tools
-from xdem.dem import DEMLike
 import geoutils as gu
+from geoutils.georaster import RasterType
 
 
 @numba.njit(parallel=True)
@@ -225,7 +225,7 @@ def get_terrain_attribute(
 
 @overload
 def get_terrain_attribute(
-    dem: DEMLike,
+    dem: RasterType,
     attribute: str,
     resolution: tuple[float, float] | float | None,
     degrees: bool,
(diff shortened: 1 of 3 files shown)


xdem/terrain.py
```diff
@@ -463,14 +530,31 @@ def aspect(dem: np.ndarray | np.ma.masked_array, degrees: bool = True) -> np.nda
...
+) -> np.ndarray: ...
+
+def hillshade(
+    dem: np.ndarray | np.ma.masked_array,
+    resolution: float | tuple[float, float] | None = None,
     azimuth: float = 315.0,
     altitude: float = 45.0,
     z_factor: float = 1.0,
-) -> np.ndarray:
+) -> np.ndarray | Raster:
```
rhugonnet, line 557: how come for hillshade it can be a `Raster` wth a `np.ndarray` as input?

rhugonnet: my bad, this is not the `@overload` !

xdem/terrain.py
```diff
@@ -414,12 +454,28 @@ def get_terrain_attribute(
 
     output_attributes = [terrain_attributes[key].reshape(dem.shape) for key in attribute]
 
+    if isinstance(dem, gu.Raster):
```
rhugonnet, line 457: we're going to write this line quite often I think! but it's not too much, it's fine :smile: !

rhugonnet (approved):
Perfect :+1:

xdem/terrain.py
```diff
@@ -209,26 +213,56 @@ def get_terrain_attribute(
...
+    degrees: bool,
+    hillshade_altitude: float,
+    hillshade_azimuth: float,
+    hillshade_z_factor: float,
+    fill_method: str,
+    edge_method: str
 ) -> list[np.ndarray]:
     ...
 
+@overload
```
adehecq, line 226: Those overloads are becoming increasingly complex!

adehecq:
Nice addition! 
But all those overloads don't make the code look simpler :sweat_smile:

PR #192, opened 2021-07-19
Fix BlockwiseCoreg.stats() KeyError and improved tests.

### LINKED ISSUE #189: Blockwise coregistration example fails with large gaps
At line https://github.com/GlacioHack/xdem/blob/71aa08516ea19f9b0b1967a8100b54467c787ee0/examples/plot_blockwise_coreg.py#L87
the script will fail if in the example the DEM is rotated after reprojection e.g. with
```
reference_dem = reference_dem.reproject(dst_crs='EPSG:3413', dst_res=reference_dem.res, resampling='bilinear')
dem_to_be_aligned = dem_to_be_aligned.reproject(dst_ref=reference_dem, resampling='bilinear')
```
The error is 
> <ipython-input-5-d8980d4be069> in <module>
> ----> 1 for _, row in blockwise.stats().iterrows():
>       2     plt.annotate(round(row["z_off"], 1), (row["center_x"], row["center_y"]), ha
>       3 ="center")
> 
> ~/development/GlacioHack/xdem/xdem/coreg.py in stats(self)
>    1783                     "y_off": points[i, 1, 1] - points[i, 1, 0],
>    1784                     "z_off": points[i, 2, 1] - points[i, 2, 0],
> -> 1785                     "inlier_count": chunk_meta[i]["inlier_count"],
>    1786                     "nmad": chunk_meta[i]["nmad"],
>    1787                     "median": chunk_meta[i]["median"]
> 
> KeyError: 0
> 
This is due to the fact that statistics do not exist for empty chunks. Hence the method self.stats should be updated accordingly.

Somewhat related, if commenting out the problematic line, the estimated vertical bias for that case does not look very convincing... :thinking: 
Figure_1

erikmannerfelt:
Closes #189.

The blocks that were empty are now correctly skipped.

Also, I found that there were `inf`s in the stats! That is because the  `Coreg.residuals()` for some reason calculated an incorrect mask sometimes. I fixed that as well, so all of the statistics are finite.

These two issues are now fixed and correctly tested, so they should not reoccur.

### DIFF
--- a/xdem/coreg.py
+++ b/xdem/coreg.py
@@ -705,6 +705,10 @@ class Coreg:
         # Calculate the DEM difference
         diff = ref_arr - aligned_dem
 
+        # Sometimes, the float minimum (for float32 = -3.4028235e+38) is returned. This and inf should be excluded.
+        if "float" in str(diff.dtype):
+            full_mask[(diff == np.finfo(diff.dtype).min) | np.isinf(diff)] = False
+
         # Return the difference values within the full inlier mask
         return diff[full_mask]
 
@@ -756,7 +760,6 @@ class Coreg:
                     f"Invalid 'error_type'{'s' if len(error_type) > 1 else ''}: "
                     f"'{error_type}'. Choices: {list(error_functions.keys())}"
                     ) from exception
-
 
         return errors if len(errors) > 1 else errors[0]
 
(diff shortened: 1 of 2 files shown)


xdem/coreg.py
```diff
@@ -705,6 +705,10 @@ def residuals(self, reference_dem: Union[np.ndarray, np.ma.masked_array],
         # Calculate the DEM difference
         diff = ref_arr - aligned_dem
 
+        # Sometimes, the float minimum (for float32 = -3.4028235e+38) is returned. This and inf should be excluded.
```
rhugonnet, line 708: how come this extreme float value is returned in the first place?
should we not fix it there instead?

rhugonnet (approved):
:+1:

xdem/coreg.py
erikmannerfelt: Added #193!

PR #197, opened 2021-07-30
Fix erroneous masks for Coreg classes

### LINKED ISSUE #195: Blockwise coregistration apply method does not take nodata properly into account
When the DEM is resampled during the blockwise coregistration, if nodata value exist, they will "leak" around the gaps.
A way to visualize it is to run the plot_blockwise_coreg.py example script and at [L 76](https://github.com/GlacioHack/xdem/blob/a5036f30b4ae01ac85791d423d13f57a546bdd9c/examples/plot_blockwise_coreg.py#L76)* to create an artificial gap in the DEM, e.g. with:
```
dem_to_be_aligned.data[0, 450:500, 450:500] = -9999
dem_to_be_aligned.set_ndv(-9999)
```
The nodata "leak" shows up on the upper edge of the gap. See attached figure.
The issue is that the nodata values are used in the resampling, so the "leaked" values do not match the nodata value and can be extremely difficult to filter out.
Figure_2

I just checked and apparently, the issue shows up when running the same example with the Nuth & Kaab coregistration instead, but to a lesser extent.
Figure_1

* **Note:** the script must be interrupted at that line. For some reason, if the entire script is run and then L 75-77 are re-run, without any change, they fail with the following error. Maybe one of the variables is being modified later in the script?
> LinAlgError: Matrix is singular.

erikmannerfelt:
Fixes #195 .

Changes:
- `warp_dem()` now has mask dilation by default.
- `get_array_and_mask()` takes and properly respects Rasters now (it was used with rasters before, but the mask was not respected).
- -9999 values in `warp_dem()` were replaced by np.nan
- Added a warning catcher in `apply_matrix()` (due to skimage + numpy deprecation warning)
- Improved test to validate that the above changes fixes the problem.

Something that is not tested automatically but was tested manually was the example by @adehecq in #195 in `examples/plot_blockwise_coreg.py`. As seen in the screenshot in #195, these fixes solve that problem.

### DIFF
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -518,7 +518,13 @@ class TestCoregClass:
 
         # Validate that the zshift is not something crazy high and that no negative values exist in the data.
         assert np.nanmax(np.abs(zshift)) < 50
-        assert np.count_nonzero(aligned.data.filled(0) < -50) == 0
+        assert np.count_nonzero(aligned.data.compressed() < -50) == 0
+
+        # Check that coregistration improved the alignment
+        ddem_post = (aligned - self.ref).data.compressed()
+        ddem_pre = (tba - self.ref).data.compressed()
+        assert abs(np.nanmedian(ddem_pre)) > abs(np.nanmedian(ddem_post))
+        assert np.nanstd(ddem_pre) > np.nanstd(ddem_post)
 
 
     def test_coreg_raster_and_ndarray_args(_) -> None:
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -505,7 +505,20 @@ class TestCoregClass:
         # Statistics are only calculated on finite values, so all of these should be finite as well.
         assert np.all(np.isfinite(stats))
 
+        # Copy the TBA DEM and set a square portion to nodata
+        tba = self.tba.copy()
+        tba.data[0, 450:500, 450:500] = -9999
+        tba.set_ndv(-9999)
 
+        blockwise = xdem.coreg.BlockwiseCoreg(xdem.coreg.NuthKaab(), 8, warn_failures=False)
+
+        # Align the DEM and apply the blockwise to a zero-array (to get the zshift)
+        aligned = blockwise.fit(self.ref, tba).apply(tba)
+        zshift = blockwise.apply(np.zeros_like(tba.data), transform=tba.transform)
+
+        # Validate that the zshift is not something crazy high and that no negative values exist in the data.
+        assert np.nanmax(np.abs(zshift)) < 50
+        assert np.count_nonzero(aligned.data.filled(0) < -50) == 0
 
 
     def test_coreg_raster_and_ndarray_args(_) -> None:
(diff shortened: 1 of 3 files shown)


erikmannerfelt:
@adehecq, I've honestly forgot why these lines need to exist, but they do right now:

https://github.com/GlacioHack/xdem/blob/dd91748ac9ee64122521348c3752789730aa52e8/xdem/coreg.py#L1442

If I remove them, the test_nuth_kaab() fails since an edge effect becomes predominant.

tests/test_coreg.py
```diff
@@ -505,7 +505,20 @@ def test_blockwise_coreg_large_gaps(self):
...
 
+        blockwise = xdem.coreg.BlockwiseCoreg(xdem.coreg.NuthKaab(), 8, warn_failures=False)
+
+        # Align the DEM and apply the blockwise to a zero-array (to get the zshift)
+        aligned = blockwise.fit(self.ref, tba).apply(tba)
+        zshift = blockwise.apply(np.zeros_like(tba.data), transform=tba.transform)
+
+        # Validate that the zshift is not something crazy high and that no negative values exist in the data.
+        assert np.nanmax(np.abs(zshift)) < 50
+        assert np.count_nonzero(aligned.data.filled(0) < -50) == 0
```
adehecq, line 521: I've seen you use 'filled' quite often. I believe your intention is to exclude nodata from the statistics. If so, I recently found out you can use data.compressed(), which returns only the valid values as a 1D array (I know, the name is not very intuitive...).

adehecq: A better test would be to check that the elevation range before/after coregistration is more or less identical.

xdem/coreg.py
```diff
@@ -1453,24 +1452,27 @@ def apply_matrix(dem: np.ndarray, transform: rio.transform.Affine, matrix: np.nd
...
+        )
+        # Warp the NaN mask, setting true to all values outside the new frame.
+        tr_nan_mask = skimage.transform.warp(
+            nan_mask.astype("uint8"),
+            inds,
+            order=resampling_order,
+            mode="constant",
+            cval=1,
+            preserve_range=True
+        ) > 0.1  # Due to different interpolation approaches, everything above 0.25 is assumed to be 1 (True)
```
adehecq, line 1475: You've changed the threshold value but not the comment.

adehecq (dismissed):
I don't understand why those changes are needed, but if that fixes the issue and does not break any previous test, I guess it should be better!

adehecq:
Thanks for fixing the bugs!

tests/test_coreg.py
erikmannerfelt: Changed to use compressed() in 2d13111!

I also added tests to check that coregistration improved. I see now that I misread your suggestion haha, but this should be more than apt, no?

xdem/coreg.py
erikmannerfelt: Fixed in 468029f

tests/test_coreg.py
adehecq: Well, the overall statistics might not be too sensitive to just a few outliers. But the range will be, so I would still recommend checking the range.
But it's a good idea to check that the dispersion is reduced too!

PR #207, opened 2021-09-06
Update skgstat version to latest fixes

### LINKED ISSUE #203: Update skgstat version in environment file to `0.6.8`
To do once it is available through the conda-forge channel to include this fix: https://github.com/mmaelicke/scikit-gstat/pull/115

rhugonnet:
Was not working for CI before (skgstat release too recent?), should work now.
Resolves #203


rhugonnet:
Got a multiprocessing error building the docs that doesn't seem linked to our code, and that I don't get locally :thinking: 

Rerunning...

rhugonnet:
@erikmannerfelt @adehecq As you can see, during CI, sphinx fails in test_docs/test_build with a multiprocessing error. And this even without changing anything to the current `main`. So it was passing a couple weeks ago and now it fails... Any ideas?

rhugonnet:
I checked all package versions of sphinx and sphinx submodules fetched by conda/mamba between the CI that use to pass in the PRs of a couple weeks ago, and this CI, and they are exactly the same.
Must come from somewhere else...

rhugonnet:
I temporarily changed the number of cores for sphinx to use during CI to 1, and opened an issue so that we can fix it.

rhugonnet:
> I temporarily changed the number of cores for sphinx to use during CI to 1, and opened an issue so that we can fix it.

Literaly one issue fixed, one issue opened :sweat:

rhugonnet:
FYI: force merging those 3 lines of change so that the CI works for other PRs

erikmannerfelt:
Wow, that was weird...  Looking at the test execution time, however, I don't really see a difference, so that's good at least

adehecq:
Weird error indeed... Sorry I can't really help on this.

PR #214, opened 2021-09-10
Re-enable blockwise test

### LINKED ISSUE #211: `test_blockwise_coreg_large_gaps` fails with all NaN assertion in `coreg.apply_matrix`
The CI passed yesterday, and this new error was triggered today. The error can be reproduced locally.
Contrary to the other fixes triggered by rasterio version 1.2.7 release (see #210), it seems this behaviour change does not depend on rasterio (I still had the error locally with version 1.2.6).
So it must have been triggered by something else?
Another mystery to solve!

erikmannerfelt:
Closes #211.

I don't know what you're smoking, @rhugonnet, but this works fine on my end! Was it CI or locally that it failed for you? For me, both pass.

### DIFF
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -485,7 +485,6 @@ class TestCoregClass:
         # Check that offsets were actually calculated.
         assert np.sum(np.abs(np.linalg.norm(stats[["x_off", "y_off", "z_off"]], axis=0))) > 0
 
-    pytest.mark.skip('This test started triggering "AssertionError: Transformed DEM has all nans." on Sept. 7 2021. Opening issue.')
     def test_blockwise_coreg_large_gaps(self):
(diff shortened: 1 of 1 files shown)


rhugonnet:
> Closes #211.
> 
> I don't know what you're smoking, @rhugonnet, but this works fine on my end! Was it CI or locally that it failed for you? For me, both pass.

I wish I was smoking good stuff! :smoking: :love_you_gesture:  
But it's probably a CI issue from another package that's been solved since yesterday... I'm certain it wasn't passing on CI (but I'm not sure they keep the logs except for the last run of a PR).

Well all good! This way we don't have to go chasing all the bugs introduced by other packages... just being patient is enough :smiley: 

On this same subject: we could also try resetting multiprocessing for the sphinx build at some point, they might have corrected the issue if it came from there (not directly them at least as I checked the version hadn't changed... but maybe a dependency).

rhugonnet (approved):

PR #218, opened 2021-09-29
Add more options to spatialstats functions

rhugonnet:
Some more flexibility for some spatialstats functions, without changing the default behaviour.
- `fit_sum_model_variogram` can now take an inital guess of variogram parameters that are later passed to `scipy.curve_fit`: `p0` and bounds `bounds`
- `plot_vgm` can be passed `xlabel`, `ylabel`, `xlim` and `ylim` to more easily customize the plot (which can be hard with `plt.gca()`as the Figure is parted into subpanels within a grid)
- fixed a bug in `patches_method` to allow for a "only 100% valid data" when checking patches no-data with the input threshold

For plots: it would be good to have a consistent approach at plotting and passing axes/grid objects to allow easy user edits without duplicating input parameters. But, realistically, this would have to be done in a later PR focusing on a `plotting` library. (`skgstat` plotting functionalitiers allow to pass ax, grid, and fig object back and forth, maybe we can check that out? or maybe there's better solutions out there, it's quite a common functionality people want to have)


rhugonnet:
@adehecq @erikmannerfelt 
Ready to be merged, just a tiny bug fix and some more flexiblity in the fit/plot vgm functions.

erikmannerfelt (approved):

PR #219, opened 2021-09-29
Minor fix in interp_nd_binning

rhugonnet:
Minor fix of `interp_nd_binning` that was failing when containing a `pd.Interval` vector with NaNs in a certain order.
- Added a better type checking of the binning variable vector
- Added a test that failed without the above fix


rhugonnet:
@adehecq @erikmannerfelt Merging this PR without review as it's only a couple line fix

PR #224, opened 2021-10-14
Fix NaNs in "longyearbyen_ddem" example

### LINKED ISSUE #196: Example "longyearbyen_ddem" contains NaNs
The data calculated for the "longyearbyen_ddem" example contains NaNs. This should be properly set to nodata so that it is properly taken into account by numpy.

rhugonnet:
- Resolves #196
- Also fixes the use of `Raster` as an input of `sample_empirical_variogram` with a tiny bug on the ground sampling distance definition fixed, and tests updated
- Opened issue #225 to add the right behaviour directly in `Coreg` classes, so that there is no operation to perform for the user downstream


rhugonnet:
It seems that https://github.com/GlacioHack/GeoUtils/pull/257 fixed the issues that appeared in the tests.
Some tests are not passing in `test_spatial_tools` which is now deleted, so I simply added a skipping marker.

erikmannerfelt:
I'm afraid I don't really understand the fix here! You simplify the syntax and clean up the variogram test, which is great. How does this relate to the nans in the test?

rhugonnet:
@erikmannerfelt The example bug fix is in the example. The rest is a side fix for using the Raster class directly as input, I added a line in the main description of the PR:
- Also fixes the use of `Raster` as an input of `sample_empirical_variogram` with a tiny bug on the ground sampling distance definition fixed, and tests updated

erikmannerfelt (approved):
Ah! Weirdly the example code diff didn't load on my GitHub. Now it does, and it looks good!

xdem/examples.py
```diff
@@ -94,9 +96,10 @@ def process_coregistered_examples(overwrite: bool =False):
     nuth_kaab.fit(reference_raster.data, to_be_aligned_raster.data,
                   inlier_mask=inlier_mask, transform=reference_raster.transform)
     aligned_raster = nuth_kaab.apply(to_be_aligned_raster.data, transform=reference_raster.transform)
```
adehecq, line 98: This line and the few lines below could be simplified now that Coreg methods take Rasters as arguments and with the arithmetic overloading functions !

PR #229, opened 2021-11-05
Fix nan skipping in patches method

rhugonnet:
Fixes a bug in the patches method where a patch was skipped if the central pixel was NaN, a behavior that discarded lots of valid patches within the `perc_valid` range.
I think this line was originally introduce to improve processing time for raster with a lot of no data, but I didn't see any difference when generating the docs, so seems all good. I used it on a Raster with quite a bit of NaNs, but maybe not as much as those where there is almost no stable terrain available...
I think we discussed this at one point @erikmannerfelt? Do you remember the ouput?


erikmannerfelt (approved):
Seems good! I don't remember this issue, but this line definitely seems extraneous.

PR #230, opened 2021-11-05
Fixes bugs in `spatialstats`

rhugonnet:
- Fixes legend plotting zorder in `plot_vgm`
- Fixes reshaping of final array in `interp_nd_binning` (needed to specify the `indexing='ij'` argument for `np.meshgrid` to work as intended)
- Added tests in `test_spatialstats.py` to check dimension order and of varying length for `interp_nd_binning`

Ready to be merged :smile:


rhugonnet:
@erikmannerfelt Don't forget there's this one as well, as you've reviewed the other two :wink:

rhugonnet:
@erikmannerfelt Thanks! All part of the same heading comment so I left as is.
FYI: this means that the ND-binning interpolation was inverting dimensions, if you were using it. The example in docs will change consequently.

Issue #232, opened 2021-11-15
Nuth & Kaab coregistration spreads NaNs


adehecq:
The Nuth & Kaab coregistration (and I assume other types of coregistration as well) spread NaNs because of the resampling.
An easy way to see it is to run the [Nuth & Kaab example](https://github.com/GlacioHack/xdem/blob/main/examples/plot_nuth_kaab.py) and to add some random gaps in the DEM. See attached script (rrnamed to txt so that GitHub does not complain) and figure below.

![Figure_1](https://user-images.githubusercontent.com/3285905/141816075-bfb7f1ff-9821-4378-beeb-dd554cd6067d.png)

[plot_nuth_kaab.txt](https://github.com/GlacioHack/xdem/files/7540043/plot_nuth_kaab.txt)

adehecq:
A simple solution is to replace the whole _apply_matrix method by something like this:
```
dem_to_be_aligned_copy = dem_to_be_aligned.copy()
dem_to_be_aligned_copy.shift(-nuth_kaab._meta['offset_east_px']*nuth_kaab._meta['resolution'], -nuth_kaab._meta['offset_north_px']*nuth_kaab._meta['resolution'])
aligned_dem = dem_to_be_aligned_copy.reproject(reference_dem)
```

adehecq:
Actually, cubic interpolation spreads the NaNs even more, so bilinear is preferable, see below.

With bilinear interpolation
![image](https://user-images.githubusercontent.com/3285905/141820869-76be37f7-fda6-49ab-a2df-39cfff6dce4a.png)

With cubic interpolation
![image](https://user-images.githubusercontent.com/3285905/141820718-fef6190d-27e4-4b8c-8bdc-2ffea378b3ac.png)

erikmannerfelt:
This is most likely due to the `dilate_mask` argument, which we deemed was a preferable behavior over potentially introducing weird artefacts. I'm not saying this isn't an issue, just that the effect is an intentional "better of two evils"!

https://github.com/GlacioHack/xdem/pull/87#discussion_r618416956

adehecq:
The NaN spread is even worth with two coregistrations in a row, like `coreg = xdem.coreg.ICP() + xdem.coreg.NuthKaab()` (see attached script and figure).
Turning off the 'dilate_mask' option reduces the gaps, but it is not perfect yet, because currently the `apply` method is run twice with two items in the CoregPipeline.

[...]

Issue #225, opened 2021-10-14
Preserve nodata in `masked_array` or `Raster` subclass during `Coreg` operations


rhugonnet:
EDIT with reproducible example (of NaNs created by `Coreg` but not masked), while both `reference_raster` and `to_be_aligned_raster` are fine:
```python
import xdem
import geoutils as gu
import numpy as np

xdem.examples.download_longyearbyen_examples()

reference_raster = gu.georaster.Raster(xdem.examples.get_path("longyearbyen_ref_dem"))
to_be_aligned_raster = gu.georaster.Raster(xdem.examples.get_path("longyearbyen_tba_dem"))
glacier_mask = gu.geovector.Vector(xdem.examples.get_path("longyearbyen_glacier_outlines"))
inlier_mask = ~glacier_mask.create_mask(reference_raster)

nuth_kaab = xdem.coreg.NuthKaab()
nuth_kaab.fit(reference_raster.data, to_be_aligned_raster.data,
              inlier_mask=inlier_mask, transform=reference_raster.transform)
aligned_raster = nuth_kaab.apply(to_be_aligned_raster)
diff = reference_raster - aligned_raster

print(np.min(diff.data))
print(np.ma.min(diff.data))
print(np.ma.min(aligned_raster.data))
print(np.ma.min(reference_raster.data))
print(np.ma.min(to_be_aligned_raster.data))
```
> nan
> nan
> nan
> 8.052481
> 8.376112

For opening issue: we don't want to repeat this operation after every `Coreg` operation to preserve nodata along calculations:
```python
aligned_raster.data[~np.isfinite(aligned_raster)] = reference_raster.nodata
diff = reference_raster - \
           gu.Raster.from_array(aligned_raster.data, transform=reference_raster.transform, crs=reference_raster.crs, nodata=reference_raster.nodata)
```

_Originally posted by @rhugonnet in https://github.com/GlacioHack/xdem/issues/224#issuecomment-943226975_

rhugonnet:
@adehecq I think this is closed by https://github.com/GlacioHack/GeoUtils/pull/257, right?

rhugonnet:
Actually, it's not as it happens within the `Coreg` classes. Though similar issue as #178 and #145

adehecq:
So I'm not sure I understand. Can it be simplified now or not?
I believe you don't need the call to `gu.Raster.from_array` anymore if you provide a Raster directly to a Coreg method following #175. 
[...]

Issue #140, opened 2021-06-03
Improve tests for Coreg, in particular with NaNs


adehecq:
As mentioned in #135, many issues arose when using DEMs with NaNs. Current tests do not include nans, so it should be improved.
The lower level function such as `get_horizontal_shift` or `calculate_slope_aspect` should also be tested, to more easily identify the source of the issue.

rhugonnet:
And also for all Coreg subclasses, see #202

PR #236, opened 2021-11-16
Improve coreg methods with NaNs

adehecq:
Tackles several issues in Coreg methods with large gaps:
- issues #225 and #239, that are related: NaNs or nodata value that spread during resampling are not properly masked in certain cases
- partially solves issue #232: Coreg apply method tend to spread NaNs/gaps a lot when calling `apply_matrix`, especially with iterative pipelines. This PR makes the argument `dilate_mask` optional to turn this off, if needed.

### DIFF
--- a/xdem/coreg.py
+++ b/xdem/coreg.py
@@ -545,22 +545,23 @@ class Coreg:
         return self
 
     @overload
-    def apply(self, dem: RasterType, transform: rio.transform.Affine | None) -> RasterType: ...
+    def apply(self, dem: RasterType, transform: rio.transform.Affine | None, **kwargs) -> RasterType: ...
 
     @overload
-    def apply(self, dem: np.ndarray, transform: rio.transform.Affine | None) -> np.ndarray: ...
+    def apply(self, dem: np.ndarray, transform: rio.transform.Affine | None, **kwargs) -> np.ndarray: ...
 
     @overload
-    def apply(self, dem: np.ma.masked_array, transform: rio.transform.Affine | None) -> np.ma.masked_array: ...
+    def apply(self, dem: np.ma.masked_array, transform: rio.transform.Affine | None, **kwargs) -> np.ma.masked_array: ...
 
 
     def apply(self, dem: np.ndarray | np.ma.masked_array | RasterType,
-              transform: rio.transform.Affine | None = None) -> RasterType | np.ndarray | np.ma.masked_array:
+              transform: rio.transform.Affine | None = None, **kwargs) -> RasterType | np.ndarray | np.ma.masked_array:
         """
         Apply the estimated transform to a DEM.
 
         :param dem: A DEM array or Raster to apply the transform on.
         :param transform: The transform object of the DEM. Required if 'dem' is an array and not a Raster.
+        :param kwargs: Any optional arguments to be passed to either self._apply_func or apply_matrix.
 
         :returns: The transformed DEM.
         """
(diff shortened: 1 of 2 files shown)


adehecq:
Trying to improve the CoregPipeline apply method to use a single matrix when possible. Based on the [N&K example](https://github.com/GlacioHack/xdem/blob/main/examples/plot_nuth_kaab.py), I just ran the following commands afterwards:
```
# Current method for 2 coregistrations
coreg = xdem.coreg.ICP() + xdem.coreg.NuthKaab()
coreg.fit(reference_dem, dem_to_be_aligned, inlier_mask, verbose=True)
aligned_dem = coreg.apply(dem_to_be_aligned)

# Attempt to do the apply in one go
test = xdem.coreg.apply_matrix(dem_to_be_aligned.data.squeeze(), dem_to_be_aligned.transform, coreg.to_matrix())

# Plot difference
diff = aligned_dem.data - test
plt.imshow(diff.squeeze(), interpolation='none', cmap='RdYlBu', vmin=-1, vmax=1)
plt.show()
```
Figure_2

The results show that the two approches are not exactly the same. I believe this is because of the `centroid` arguments, which is not taken into account in the `self.to_matrix()` method.

erikmannerfelt:
It is probably indeed the centroid missing. You can get it with `Coreg._meta.get("centroid")` (defaults to `None` if it doesn't exist, which makes `apply_matrix()` fall back to a corner (I forget which one)

adehecq:
> It is probably indeed the centroid missing. You can get it with `Coreg._meta.get("centroid")` (defaults to `None` if it doesn't exist, which makes `apply_matrix()` fall back to a corner (I forget which one)

The default is the DEM center coordinate. But the problem is when running two transformations with a different centroid. In that case I don't know if it can be simplified to a single transformation...

adehecq:
Commit 9381cd5 should solve #225 and #239.
However, as a consequence the coregistered DEM is a masked array (instead of an array with NaNs) and since many tests rely on this assumption, they fail.

xdem/coreg.py
```diff
@@ -585,27 +586,40 @@ def apply(self, dem: np.ndarray | np.ma.masked_array | RasterType,
...
                     centroid=self._meta.get("centroid"),
-                    dilate_mask=True
+                    dilate_mask=dilate_mask,
+                    **kwargs
                 )
             else:
                 raise ValueError("Coreg method is non-rigid but has no implemented _apply_func")
 
+        # Calculate final mask
+        final_mask = dem_mask + np.isnan(applied_dem)
```
rhugonnet, line 613: This line is essentially the fix?

xdem/coreg.py
```diff
@@ -585,27 +586,40 @@ def apply(self, dem: np.ndarray | np.ma.masked_array | RasterType,
         # See if a _apply_func exists
         try:
             # Run the associated apply function
-            applied_dem = self._apply_func(dem_array, transform)  # pylint: disable=assignment-from-no-return
+            applied_dem = self._apply_func(dem_array, transform, **kwargs)  # pylint: disable=assignment-from-no-return
         # If it doesn't exist, use apply_matrix()
         except NotImplementedError:
             if self.is_affine:  # This only works on it's affine, however.
+                # If dilate_mask is not specified, set it to True by default
```
rhugonnet, line 593: Nice to have this option accessible through kwargs!

rhugonnet (approved):
Pretty much all good!
Only comments: 
- It might be good to add couple tests to ensure the intended behaviour will be maintained in the future regarding input/ouput + presence of NaNs (especially if we update geoutils soon!). Though it might have to be done on a large scale for all our functions, so this PR could also simply be linked to the already opened issue on this #178 .
- You should probably edit the PR top message to summarize the changes if we ever get back to it (the addition of the `kwargs` handling of `dilate_mask` is invisible without looking at the code right now!)

xdem/coreg.py
adehecq: This was actually the initial intention of this PR !

xdem/coreg.py
adehecq: Regarding issues #225 and #239, yes!

adehecq:
>     * It might be good to add couple tests to ensure the intended behaviour will be maintained in the future regarding input/ouput + presence of NaNs (especially if we update geoutils soon!). Though it might have to be done on a large scale for all our functions, so this PR could also simply be linked to the already opened issue on this #178 .
> 
I think this is actually more related to #140. Since it's already an open issue, which will take a long time to address, I will ignore it for now.

>     * You should probably edit the PR top message to summarize the changes if we ever get back to it (the addition of the `kwargs` handling of `dilate_mask` is invisible without looking at the code right now!)
Done.

PR #240, opened 2021-11-17
Fix `DEM.copy` with `new_array` argument

### LINKED ISSUE #235: Argument `new_array` of `DEM.copy()` not working
The `copy` function of the `DEM` class is not working when passing a `new_array` (array not updated).
Reproducible example:
```python
import xdem
r = xdem.dem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
r2 = r + 5
np.nanmedian(r)
>>> 360.64624
np.nanmedian(r2)
>>> 365.64624
r3 = r.copy(new_array=r2.data)
np.nanmedian(r3)
>>> 360.64624
r4 = r.copy(r2.data)
np.nanmedian(r4)
>>> 360.64624
```
This problem is probably common with the `SatImg` class. I'll check and open an issue if it's indeed the case.

rhugonnet:
Resolves #235
Easy fix, the argument was simply not passed to `super()` for `SatImg`


xdem/dem.py
```diff
@@ -84,9 +85,9 @@ def __init__(self, filename_or_dataset, vref_name=None, vref_grid=None, silent=T
         # trying to get vref from product name (priority to user input)
         self.__parse_vref_from_fn(silent=silent)
 
-    def copy(self, new_array=None):
+    def copy(self, new_array: np.ndarray | None = None) -> DEM:
```
adehecq, line 88: Simple fix ! But wouldn't it be more generic to use kwargs? This way if one adds new functionalities to the parent function, it is also passed.

rhugonnet: Ah true, I don't know if that maps the argument for the subclass function tnough.
That's applicable to so many other things, we would have to rethink a lot of Subclasses functions! So not something we want to do in the frame of this PR.

adehecq: Well, you can at least try in that case to replace your argument 'new_array' with **kwargs and see if that solves your issue.
The only problem might be with the docstrings. But one can always refer to the parent's documentation.

rhugonnet: Just checked, and it works as expected.
Though, to be honest, it seems a lot less practical because the keywords argument are not mapped by the function:
Screenshot from 2021-11-17 17-13-33
Screenshot from 2021-11-17 17-13-43
My opinion is that we should implement this kind of syntax for overriden class function syntax either everywhere or nowhere. Right now it's nowhere in xdem, we repeat arguments everywhere. Copying all arguments is also what the editor recommends when one starts the writing of such an overriden parent class function.
So it is best to raise an issue and tackle this consistently and all at once in another PR.

adehecq (approved):
Added our discussion as a new issue, so that one is good to merge.

Issue #237, opened 2021-11-16
`NuthKaab` shifted grids in `fit_func` contain interpolated nodata


rhugonnet:
Sorry for opening so many issues on the Nuth and Kaab in the same day, @erikmannerfelt! It means we are testing (sometimes struggling) thoroughly :wink: 
The Nuth and Kaab shift function based on `scipy.interpolate.RectBivariateSpline` has a wrong NaN propagation that keeps interpolated NaNs.
Reproducible example:
```python
import numpy as np
dem = np.arange(12.).reshape((4, 3))
dem[1, 1]= np.nan
print(dem.astype(int))
```
> [[ 0.  1.  2.]
 [ 3. nan  5.]
 [ 6.  7.  8.]
 [ 9. 10. 11.]]
```python

import scipy.interpolate

# Make index grids for the east and north dimensions
east_grid = np.arange(dem.shape[1])
north_grid = np.arange(dem.shape[0])

# Make a function to estimate the aligned DEM (used to construct an offset DEM)
elevation_function = scipy.interpolate.RectBivariateSpline(
    x=north_grid, y=east_grid, z=np.where(np.isnan(dem), -9999, dem), kx=1, ky=1
)

# Make a function to estimate nodata gaps in the aligned DEM (used to fix the estimated offset DEM)
# Use spline degree 1, as higher degrees will create instabilities around 1 and mess up the nodata mask
nodata_function = scipy.interpolate.RectBivariateSpline(
    x=north_grid, y=east_grid, z=np.isnan(dem), kx=1, ky=1
)

offset_east = 0.5
offset_north = 0.5

origin_elevation = elevation_function(y=east_grid, x=north_grid)
print(origin_elevation.astype(int))
```
> [[    0     1     2]
 [    3 -9999     5]
 [    6     7     8]
 [    9    10    11]]
```python
new_elevation = elevation_function(y=east_grid + offset_east, x=north_grid - offset_north)

# Set NaNs where NaNs were in the original data
new_nans = nodata_function(y=east_grid + offset_east, x=north_grid - offset_north)
new_elevation[new_nans >= 1] = np.nan

print(new_nans)
```
> [[0.   0.   0.  ]
 [0.25 0.25 0.  ]
 [0.25 0.25 0.  ]
 [0.   0.   0.  ]]
```python
print(new_elevation.astype(int))
```
> [[    0     1     2]
 [-2498 -2497     3]
 [-2495 -2494     6]
 [    8     9     9]]

[...]

PR #242, opened 2021-11-19
Fix issue with NaNs masks in Coreg methods

adehecq:
This is a fix to #237.
During coreg.apply, a resampling is needed. When NaNs exist in the input DEMs, the NaNs also propagate and must be handled carefully. Currently, such resampling is done at 3 different places, and NaNs are not always caught properly.
This PR tries to fix this issue.

I also fixed a small issue in BlockwiseCoreg. Errors during fitting were caught, but not during apply.

### DIFF
--- a/tests/test_spatialstats.py
+++ b/tests/test_spatialstats.py
@@ -280,7 +280,7 @@ class TestPatchesMethod:
 
         # Check the sampling is always fixed for a random state
         assert df['tile'].values[0] == '31_184'
-        assert df['nanmedian'].values[0] == pytest.approx(2.28, abs=0.01)
+        assert df['nanmedian'].values[0] == pytest.approx(2.3, abs=0.01)
 
         # Check that all counts respect the default minimum percentage of 80% valid pixels
         assert all(df['count'].values > 0.8*np.max(df['count'].values))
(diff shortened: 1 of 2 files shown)


adehecq:
Note that BlockwiseCoreg is not very reliable with large gaps, but the current fix improved it a bit. See figure of elevation difference after Blockwise coregistration (See attached script inspired from [test_blockwise_coreg_large_gaps](https://github.com/GlacioHack/xdem/blob/edde09b1e53701f2825b7cacf51d280646171fef/tests/test_coreg.py#L489)): [test_blockwise.txt](https://github.com/GlacioHack/xdem/files/7569255/test_blockwise.txt)

**Before the PR**
Figure_1

**After the PR**
Figure_2



The underlying NuthKaab method should probably be improved to be more robust with few points.

adehecq:
Finally, this is another indication that Coreg methods should be thoroughly tested for datasets with large gaps (#140).

rhugonnet:
Thanks for tackling this! It's good to have all these fixes incorporated right away.

I agree, we'll have to work on a testing scheme for outlier robustness! This can probably be done in the future by simulating data gaps/NaNs, which will also be super useful if we want to write a paper comparing the performance of `Coreg` methods and their robustness :smile: !

rhugonnet (approved):
Looks good!

PR #245, opened 2021-11-21
Fix `get_terrain_attribute` bug with `edge_method='none'`

### LINKED ISSUE #244: Argument `edge_method='none'` of `get_terrain_attribute` returns void/constant grid
Reproducible example:
```python
import xdem
import numpy as np

fn_img = xdem.examples.get_path("longyearbyen_ref_dem")

dem = xdem.DEM(fn_img)

slope = xdem.terrain.get_terrain_attribute(dem, attribute='slope')
print(np.nanmedian(slope))
slope = xdem.terrain.get_terrain_attribute(dem, attribute='slope', fill_method='none')
print(np.nanmedian(slope))
slope = xdem.terrain.get_terrain_attribute(dem, attribute='slope', edge_method='none')
print(np.nanmedian(slope))
```
>15.067599254391457
>15.067599254391457
>0.0

For aspect:
>191.47179257973147
>191.47179257973147
>90.0

Same using `dem.data` and `resolution=dem.res[0]`.
I am also surprised that fill_method='none' has virtually no effect on the median, not sure if it is working correctly either. Maybe @erikmannerfelt has an idea?

rhugonnet:
Resolves #244

- Fixes an indexation issueof `get_terrain_attribute` with `edge_method='none'`,
- Removes wrong referencing of slope equation in comments,
- Changes the default edge behaviour to match that of GDAL (i.e. `edge_method='none'` and `fill_method=`none`),
- Adds doctests for `slope` convenience function, and adapt doctests of other attributes with new default methods.

We should have more specific, in-depth tests for the edge and filling methods. Not the aim of this PR, so opening an issue for it.


xdem/terrain.py
```diff
@@ -67,6 +67,9 @@ def _get_quadric_coefficients(
                 elif edge_method_n == 1:
                     row_indexer = (row + k) % dem.shape[0]
                     col_indexer = (col + j) % dem.shape[1]
+                else:
+                    row_indexer = row + k
```
erikmannerfelt, line 71: Are you sure this won't overflow at any point?

erikmannerfelt (commented):
Thanks for improving on these methods! 

It's uncontroversial to default to the GDAL functionality, but I think GDAL is wrong in propagating NaNs like that. If you think we should default to GDAL-behavior, then sure.

xdem/terrain.py
rhugonnet: Not sure I got this. You mean you'd prefer `elif edge_method_n == 2`?

rhugonnet:
> Thanks for improving on these methods!
> 
> It's uncontroversial to default to the GDAL functionality, but I think GDAL is wrong in propagating NaNs like that. If you think we should default to GDAL-behavior, then sure.

When you are missing only one point in the corner, I agree the results is still OK! But when you are missing half of the cell, the terrain attribute value will be aberrant. 
Maybe an argument could control when the filling method is applied, depending on the amount of gaps to fill in the 3x3 window? Then I would certainly be fine having a default method that fills when there's only one missing pixel for example :wink:
With the current approach, we should definitely default to GDAL to avoid all those pixels that end up with erroneous estimates (usually we have enough pixels in DEM data anyway!)

xdem/terrain.py
erikmannerfelt: It may be just me not remembering my own code haha, but what happens if `row == dem.shape[0]` and `k` is nonzero? Then that would result in an `IndexError`

rhugonnet: It is the non-remembering haha, you wrote a loop further up to avoid that issue :wink::
```python
  if edge_method_n == 2:
            if (row < 1) or (row > (dem.shape[0] - 2)) or (col < 1) or (col > (dem.shape[1] - 2)):
                continue
```
And tests would have picked that up for sure. 
So all ready to be merged :+1:

erikmannerfelt: Past me is great and so are you!

erikmannerfelt (approved):
Great!

PR #248, opened 2021-12-14
Fix issue in volume.calculate_hypsometry_area when ddem_bins contain NaNs

adehecq:
`volume.calculate_hypsometry_area` raises an `AssertionError` when `ddem_bins` contains gaps/NaNs. This should not be the case when `timeframe` is set to "reference", as the ddem_bins values are not used.
Fixed this.


adehecq:
Forcing merge despite test failures, which are not related to my edits...

Issue #264, opened 2022-04-22
Subsections of sphinx-gallery + Content headers for documentation


rhugonnet:
We already have **12** gallery examples! Many more will come, and it's starting to be a bit messy!! :worried: 

Two points to improve this:

1. Organize the gallery in subsections. Thankfully, this is done super easily by the default behaviour of sphinx-gallery that splits galleries in sub-galleries if the `examples/` folder has subfolders. Unfortunately, we already have the "Easton" and "Longyearbyen" subfolders in the `example` folder. Should we (A) move those somewhere else (`tests/data` or similar?)? Or (B) create a subfolder of `examples/gallery/` that we make the parent of all sphinx-gallery? 

2. Add **content headers** to the documentation. Right now we just concatenate sections without permanent headers. Essentially, we would have several `toctree` instead of a single one "Content" in our `index.py`, to split section, for example like the blue delineators of the Sphinx-Gallery documentation (https://sphinx-gallery.github.io/stable/index.html).
For this, we should reflect on a tree, for instance:

> **Getting started**
> 
> - What is xdem?
> - Installing xdem
> - First steps
> 
> **Background**
> 
> - Aligning digital elevation models
> - Analysis of accuracy and precision
> 
> **Features**
> 
> - Vertical reference
> - Terrain attributes
> - Coregistration
> - Bias correction
> - Filtering
> - Uncertainty analysis
> 
> **Example gallery**
> 
> - Beginner
> - Advanced
> - Expert
> 
> **Reference**
> 
> - API
> - Development guidelines

Any input @erikmannerfelt @adehecq ? I'd be for option 1.A and 2. with something similar to this structure.
This is quite straightforward to do, but would be easier to implement once the `terrain.py` PR is merged.

rhugonnet:
Forgot this was already opened partially in #258 by @adehecq! More issues to close :grin:

Issue #206, opened 2021-09-03
Split and streamline information in README 


rhugonnet:
> Just a minor thought, but should we not have a dedicated INSTALL file or so with all these considerations. I feel like this is not the best use of the README, which should be rather clear and concise. What do you think?

Fully agreed, there is quite a bit of work on the README to make it a more agreeable read for users. I think we should focus more on the DEM thematic and a couple examples with clear links to documentation rather than all the coding stuff (how to contribute, how to install, etc... should live in separate files simply referred in the README).

_Originally posted by @rhugonnet in https://github.com/GlacioHack/xdem/pull/200#r702048886_

rhugonnet:
Solved by #303

PR #265, opened 2022-04-23
Restructure documentation and add or streamline beginner-friendly examples

### LINKED ISSUE #258: Split doc into "Getting started" and "Advanced guide"
We need to make the first pages of the documentation more easy for newcomers, by highlighting high-level functions.

rhugonnet:
## Final PR description

**This PR includes:**

- A restructuration of documentation for clearer sections "Getting started", "Background", "Features", "Gallery of examples",
- An update of the landing page,
- The addition of beginner pages on "About xDEM", "How to install" and "Quick start",
- A streamlining of existing examples to be more beginner-friendly,
- A change in `dev-environment.yml` to pass tests dependant on GeoUtils without requiring a new release. See new issue #268 to switch back once our packages are mature enough.

New documentation can be viewed here: https://xdem-rhugonnet.readthedocs.io/en/doc_struc/

Resolves #258

We merge this PR early to match @adehecq's EGU talk.
Similar objectives described below will be addressed in later PRs.

### Points of issue in original PR

1) Structure of documentation (see #264)
2) Location of test data VS Location of gallery examples (see #264 and #258)
3) Structure of information in README (see #206)


docs/source/quick_start.rst
```diff
@@ -0,0 +1,53 @@
...
+===========
+
+Sample data
+-----------
+
+*xdem* comes with some sample data that is used throughout this documentation to demonstrate the features. If not done already, the sample data can be downloaded with the command
+
+.. code-block:: python
+
+        xdem.examples.download_longyearbyen_examples(overwrite=False)
```
erikmannerfelt, line 13: Overwrite is implicitly `False`. Honestly it took me a while to even figure out what that was, so I think it's best to remove here!

docs/source/vertical_ref.rst
```diff
@@ -0,0 +1,6 @@
+.. _vertical_ref:
+
+Vertical referencing
+====================
+
+In construction
```
erikmannerfelt, line 6: Same comment as for the DEM. I suggest adding something easy to `grep -r` when the time comes!

docs/source/index.rst
```diff
@@ -3,41 +3,52 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-Welcome to xdem's documentation!
+xDEM: Analysis of Digital Elevation Models for geosciences
 ================================
-``xdem`` aims to make Digital Elevation Model (DEM) analysis easy.
-Coregistration, subtraction (and volume measurements), and error statistics should be available to anyone with the correct input data.
+*xdem* aims at simplifying the analysis of Digital Elevation Models (DEM). This module provides typical post-processing steps such as coregistration, elevation/volume change measurements and error statistics.
```
rhugonnet, line 8: DEM with an "s" on this line! :D 
The capitalization is also not so common I think (e.g., https://en.wikipedia.org/wiki/Digital_elevation_model, and I remember it was decapitalized by co-authors several times when I drafted my papers). We should probably decide on one and stay consistent throughout the package. I'd be in favour of lowercase :grin:

docs/source/install_xdem.rst
```diff
@@ -0,0 +1,60 @@
...
+.. code-block:: bash 
+
+  pip install xdem
+
+**NOTE**: Setting up GDAL and PROJ may need some extra steps, depending on your operating system and configuration.
+
+
+Installing for contributors
+---------------------------
+Recommended: Use conda for depencency solving.
```
rhugonnet, line 48: typo: dependency

docs/source/quick_start.rst
rhugonnet: +1

docs/source/quick_start.rst
```diff
@@ -0,0 +1,53 @@
+.. _quick_start:
+
+Quick start
+===========
+
+Sample data
+-----------
+
+*xdem* comes with some sample data that is used throughout this documentation to demonstrate the features. If not done already, the sample data can be downloaded with the command
```
rhugonnet, line 9: When we talk about the package name we write "xDEM", and for the module sometimes ``xdem`` or *xdem*. Which one should we pick and stick to?
In the NumPy docs, looks like they stick to the package name "NumPy" without any kind of highlighting, and that the lower case, package-style naming ``numpy`` is only used with a submodule e.g. ``numpy.array``

docs/source/vertical_ref.rst
rhugonnet: +1, we can add TODO everywhere (I didn't have that in mind and just wrote the "In construction")

docs/source/what_is_xdem.rst
```diff
@@ -0,0 +1,16 @@
+.. _what_is_xdem:
+
+What is xDEM?
+=============
+
+*xdem* aims at simplifying the analysis of Digital Elevation Models (DEM). This module provides typical post-processing steps such as coregistration, elevation/volume change measurements and error statistics.
```
rhugonnet, line 6: Right now this line is a copy/paste of the `index.rst` first line. If a reader reads the introduction page, then goes to "What is xDEM" which is the first link below, he re-reads this same line twice in a row.
We could leave this shortened version only on the landing page, and elaborate in this one?

docs/source/what_is_xdem.rst
```diff
@@ -0,0 +1,16 @@
...
+As of today, we are a group of research scientists mostly with interests in glaciology, but the tools we're developing are designed for all sort of geoscience applications.
+
+The core concepts behind *xdem* are:
+
+**Ease of use**: Most operations require only a few lines of code. The module was developed with a focus on remote sensing and geoscience applications. 
+
+**Flexibility**: We offer a set of options, rather than favoring a single method. Everyone should be able to contribute to xdem and add to its functionalities.
+
+**Reproducible**: Version-controlled, releases saved with DOI and test-based development ensure our code always performs as expected.
```
rhugonnet, line 15: Perfect :smile: !

rhugonnet (commented):
I think this is perfect to have a more intuitive landing page + getting started examples for EGU :+1: 

We could merge this PR just for EGU, and advance further on the documentation (e.g. splitting of the examples) in a later one. Is that what you have in mind @adehecq ?

docs/source/intro_dems.rst
```diff
@@ -0,0 +1,6 @@
+.. _intro_dems:
+
+Digital elevation models
+========================
+
+In construction
```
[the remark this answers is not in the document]
adehecq: Replaced with "TODO"

docs/source/quick_start.rst
adehecq: Done

docs/source/index.rst
adehecq: Used lowercase here. I did not check if this occurred in other areas though.

docs/source/quick_start.rst
adehecq: I agree. I have no strict opinion on what we should use. In case of NumPy, the name is somewhat highlighted with the camel syntax. It feels a bit weird to write xdem in the text as if it was a regular word, so I like having a way to highlight it.
We agreed that lowercase is better than uppercase so that's why I went with the italic (*xdem*) and I think the code highlight (`xdem`) should be reserved for actual code so not in the text.
What do you think?

rhugonnet: Just looked at what is done in the docs of Matplotlib, NumPy, SciPy, PyTorch, GPyTorch, I'd do the same: use the "normal text" package syntax that is camel-case/acronym-coded.
For us, it would correspond to using "xDEM" in normal text everywhere (in a way, its capitalization would serve to highlight the name).

rhugonnet: **Still time to change this later:** to avoid the discrepancy between the ``xdem`` and *xdem* in the docs of this PR, I changed into xDEM everywhere to mirror the syntax used in big packages as described above. It will be easier to replace if we want to, because with the capitalized letters it doesn't mix with the code ocurrences :)

adehecq (approved):
Thanks a lot for making the final push !!

PR #271, opened 2022-07-07
Add gaussian and exponential models to `fit_sum_model_variogram`

rhugonnet:
This PR adds the exponential and spherical models for fitting a sum of model to an empirical variogram.


rhugonnet:
One issue brought up by this PR and the work in https://github.com/rhugonnet/dem_error_study is that we inconsistently use skgstat.models or home-defined variogram models in our functions. 
This should be dealt with in the future by relying on single source, preferably from an existing package.

rhugonnet:
Admin-merging this PR as it is only a couple lines

PR #276, opened 2022-08-01
Homogenize and simplify `spatialstats.py`

### LINKED ISSUE #204: Warnings for `sample_empirical_variogram` with multiple jobs
I'm getting lots of warnings when running with multiple jobs:
```
Numba: Attempted to fork from a non-main thread, the TBB library may be in an invalid state in the child process.
```

_Originally posted by @erikmannerfelt in https://github.com/GlacioHack/xdem/pull/159#discussion_r686603907_

### LINKED ISSUE #215: `double_sum_covar()` clarify and improve function arguments
Thanks for making all these great features, @rhugonnet! There were some minor things that felt could make the user-friendliness even higher.

- Fix the `list_tuple_errs` type hint; it's currently `list[float]` when I guess it should be `list[tuple[float, ...]]`?
- Add an option to use generic x/y (easting/northing) coordinates instead of lat/lon. Could be a boolean flag to enable conversion or not.
- Potentially add a minimal example to show some mock inputs and an output?
- Add a verbosity setting to enable/disable the print statements. Would also be nice with tqdm progress bars!

### LINKED ISSUE #238: Correct `nd_binning` example in the documentation
Since #230 the behaviour of `interp_nd_binning` is corrected. 
However, now some aspects of the gallery example might be erroneous as a result. Need to adjust!

### LINKED ISSUE #250: Vectorize covariance sums in `spatialstats`
The speed can probably be optimized quite a bit by computing through basic linear algebra, instead of for loops.

### LINKED ISSUE #252: Bugs and suggestions of improvements in spatialstats
While I was testing the different examples related to the spatialstats.py module on the documentation, I found a few issues and maybe some ways from improvement:

- [x] Functions `plot_1d_binning` and `plot_2d_binning` display the NMAD (in many examples the default). But this may not be calculated, and the functions plot NaNs instead of raising an error. The issue is at [this line](https://github.com/GlacioHack/xdem/blob/674c94785df07097da1aeedbdb60ae87392a6636/xdem/spatialstats.py#L1462), where a blank "nmad" (or whatever `statistic_name` is) column is created.
One way to test is to run the `plot_standardization.py` example and trying plotting the histograms, e.g.:
```
xdem.spatialstats.plot_1d_binning(df, "slope", "nmad", "Slope (degrees)", "NMAD of dh (m)")
plt.show()

xdem.spatialstats.plot_2d_binning(df, "slope", "maxc", "nmad")
plt.show()
```
It works fine if "nanstd" is used instead of "nmad". So this is more about catching a misuse (encouraged by most examples) than an actual bug.

- [x] related to the previous point, to avoid confusion with the other example (e.g. `plot_nonstationary_error.py`, one could calculate the nmad instead/on top of the nanstd.

- [x] the doctstring of `fit_sum_model_variogram` could be a bit more detailed, e.g.:
   - ":param list_model: list of K variogram models to sum for the fit: from short-range to long-ranges" -> what are "variogram models", it's unclear what should be passed (from the doc examples, "sph" is an option. This should be clearly stated what the options are here.
   - ":param empirical_variogram: empirical variogram" -> specify "as returned by sample_empirical_variogram"
   - ":return: modelled variogram function, coefficients" -> it should say which coefficients and in which order, so that one can figure out what are the sills and ranges
   
- [x] in the docstring of `exact_neff_sphsum_circular`, also detail what units are expected for the inputs (they should at least all be the same, but most likely should be in m or m2).

- [x] related to the previous point, in the example `plot_standardization.py`, Neff is calculated using the following area attribute `svendsen_shp.ds['Shape_Area'].values`. This example can be a bit more difficult to translate to another dataset if that particular attribute is not available. Maybe in that example, we could calculate the area using geopandas functionalities, e.g. `glacier_outlines.ds.area`. In my case, I was tricked because I used the RGI "Area" attribute, which is in km2 so not the right unit... 

I may add more bullet points in the future as I play with the examples.
Apart from these minor issues, the examples are great !

### LINKED ISSUE #253: Elaborate on the `bins` column in empirically sampled variograms
I had a journey trying to figure out if the `bins` column represented the upper bound or the centrepoint of the bin!

Because of this line, I made the assumption that this is in fact decided by skgstat (since no default argument is given):
https://github.com/GlacioHack/xdem/blob/674c94785df07097da1aeedbdb60ae87392a6636/xdem/spatialstats.py#L467

And skgstat in turn (currently) has the default that it's the upper bound:
https://github.com/mmaelicke/scikit-gstat/blob/69074feefb3759ba75c4ee91a5a9fced8d872801/skgstat/Variogram.py#L1790

This should be mentioned in the docstring of `xdem.spatial_stats.sample_empirical_variogram()` and anywhere else that it might be relevant!

### LINKED ISSUE #261: Add "count" on top of user-defined `statistics` argument for `nd_binning`
Otherwise the plotting function, and other functionality will fail. Those should check that the DataFrame contains this attribute, and raise warning in any case, as well.

### LINKED ISSUE #272: Use a single, consistent set of variogram models functions (either skgstat, or xDEM) across the package
One issue brought up by this PR and the work in https://github.com/rhugonnet/dem_error_study is that we inconsistently use skgstat.models or home-defined variogram models in our functions. 
This should be dealt with in the future by relying on single source, preferably from an existing package.

_Originally posted by @rhugonnet in https://github.com/GlacioHack/xdem/issues/271#issuecomment-1177231534_

rhugonnet:
This PR aims to address several issues linked to `spatialstats`, in particular the ease-of-use, clarity of parameters and tests!

# Summary of changes

- The arguments to pass to `sample_empirical_variogram` in relation to `skgstat.MetricSpace` have been simplified, and further described. In particular the `subsample` argument is now automatically calculated to match the number pairwise samples found in a single ensemble (N**2/2 if N is the subsample), even if the method uses more complex distinct ensembles (as is the case for the default `cdist_equidistant` method),
- Some minors fixes have been implemented to avoid user-induced mistakes in `nd_binning`, `plot_1d_binning` and `plot_2d_binning`.
- The format of variogram model functions has been homogenized to use only `skgstat.models`, which results in the deletion of the `vgm`, `cov` functions that existed in `xdem.spatialstats`, and the replacement of occurences in `fit_sum_model_variogram` and `neff` functions (for fitting to empirical variograms, and spatial integration),
- The format for returning/passing modelled variograms in xDEM has been homogenized to a `pd.Dataframe` object with columns "models", "range", "psills" and "smooth", described in related function output/input. The format is tested before each function run to provide clear `ValueError` messages in case of wrong user input,
- Three basic functions `get_func_sum_vgm_models`, `covariance_from_vgm` and `correlation_from_vgm` have been added to convert variogram model parameters of a `pd.Dataframe` into a function of sum of variogram with spatial lags, a spatial covariance function with spatial lags, and also a spatial correlation function with spatial lags,
- The `neff` functions to estimate a number of effective samples based on spatial correlation have been improved, clarified, and further tested. In particular, the `neff_circular_approx_theoretical` now contains exact circular integration formulas for any number of summed `spherical`, `gaussian`, `exponential` and `cubic` models. It is used to thoroughly test the `neff_circular_approx_numerical` function that can integrate numerically any number of summed variogram of any form. Those two fonctions are to be used when the user solely provides an area value. When the shape of the area is provided, the `neff_exact` or `neff_approx_hugonnet`, based on double sum of covariance with exact coordinates, can be used to derive the number of effective samples. The double covariance sum has been vectorized for computational speed, and the random sampling of the `neff_approx_hugonnet` is now tested with `random_state` arguments.
- For now, the documentation examples are updated with the new function names and parameters, but those need further simplification ("beginner example", "advanced example").

Additional changes after answers to questions below:

- Renamed all occurences of `vgm` into parameters or functions names into `variogram`,
- Corrected typos, unclear descriptions and parameters,
- Added `_` in function names `std_err`, `std_err_finite`, `distance_latlon`, `create_circular_mask` and `create_circular_ring` to remove from API.
- Added additional tests on output of `sample_empirical_variogram` and `_choose_cdist_equidistant_sampling_parameters`,
- Corrected sphinx syntax error,
- Addressed issues related to testing and added two pages to the wiki: https://github.com/GlacioHack/xdem/wiki/Stuck-on-a-mysterious-error-during-CI%3F and https://github.com/GlacioHack/xdem/wiki/Memo-on-Python-and-numpy-%60dtype%60,
- **Questions 3-4 will resolved in a separate PR**.

**(Old) Questions for @adehecq and @erikmannerfelt to move forward with this PR:**

1. Are all the function and parameters names fine with you? There's probably a bit improvement to do there, and once we have some good names we stick to them! But it's a bit hard to find those alone sometimes :sweat_smile:.
2. Should I do a core `neff` function, which defaults to one of the other functions, and can be used to choose which calculation? Should I change the others to hidden functions `_neff_...`? This would remove them from API.
3. My idea for the updated documentation is to have: (i) beginner examples that fit in just 10 lines of code to exemplify each parameter and the basic behaviour of a function, (ii) advanced examples where things are more detailed/discussed.
4. For the purpose of 3., I would add an "error_pipeline" that automatically combines steps that go together: `nd_binning` and `interp_nd_binning`, a new "`standardization`" function, and a new "`error_map`" function; then another one that combines `sample_empirical_variogram`, `fit_sum_model_variogram` and `neff` to get the error in an area? Something along these lines.

# To-do-list for related issues

- [x] Resolves #272,
- [x] Resolves #261,
- [x] Resolves #253, 
- [x] Resolves #252, 
- [x] Resolves #250, 
- [x] Resolves #238, 
- [x] Resolves #215, 
- [x] Resolves #204. 

As well as:

- [x] Add more tests for neff function, including exact integration for spherical, exponential, gaussian and cubic models,
- [x] Add tests for plotting functions of `spatialstats` to ensure clear error message with wrong user input.


xdem/spatialstats.py
```diff
@@ -47,6 +49,119 @@ def nmad(data: np.ndarray, nfact: float = 1.4826) -> float:
...
+    each bin. By default, the sample count, the median and the normalized absolute median deviation (NMAD). The count
+    is always computed, no matter user input.
+    Values input is a (N,) array and variable input is a list of flattened arrays of similar dimensions (N,).
+    For more details on the format of input variables, see documentation of scipy.stats.binned_statistic_dd.
+
+    :param values: Values array (N,)
+    :param list_var: List (L) of explanatory variables array (N,)
+    :param list_var_names: List (L) of names of the explanatory variables
+    :param list_var_bins: Count, or list (L) of counts or custom bin edges for the explanatory variables; defaults to 10 bins
+    :param statistics: List (X) of statistics to be computed; defaults to count, median and nmad
```
adehecq, line 67: What's the meaning of "List (L)" and "List (X)"?

xdem/spatialstats.py
```diff
@@ -473,20 +484,82 @@ def _get_pdist_empirical_variogram(values: np.ndarray, coords: np.ndarray, **kwa
 
     return df
 
+def _choose_cdist_equidistant_sampling_parameters(**kwargs):
+    """Add a little calculation to partition the "subsample" argument automatically into the "run" and "samples"
```
adehecq, line 488: It's better to have the """ and then a breakline. Same for the bottom ones.

xdem/spatialstats.py
```diff
@@ -473,20 +484,82 @@ def _get_pdist_empirical_variogram(values: np.ndarray, coords: np.ndarray, **kwa
...
+    arguments of RasterEquidistantMetricSpace, to have a similar number of points than with a classic pdist method.
+
+    The number of pairwise samples for a classic pdist is N0(N0-1)/2 with N0 the number of samples of the ensemble. For
+    the cdist equidistant calculation it is M*N*R where N are the subsamples in the center disk, M is the number of
+    samples in the rings which amounts to X*N where X is the number of rings in the grid extent, as each ring draws N
+    samples. And R is the number of runs with a different random center point.
+    X is fixed by the extent and ratio_subsample parameters, and so N0**2/(2X) = N**2*R, and we want at least 30 runs
+    with 10 subsamples."""
+
+    # First, the extent
```
adehecq, line 498: Not very informative comment 😉

xdem/spatialstats.py
```diff
@@ -735,47 +819,171 @@ def sample_empirical_variogram(values: Union[np.ndarray, RasterType], gsd: float
...
+
+    return model_name
+
+def get_func_sum_vgm_models(params_vgm: pd.DataFrame) -> Callable[[np.ndarray], np.ndarray]:
+    """
+    Construct the sum of spatial variogram function from a dataframe of variogram parameters.
+
+    :param params_vgm: Dataframe of variogram models to sum with three to four columns, "model" for the model types
+        (e.g., ["spherical", "matern"]), "range" for the correlation ranges (e.g., [2, 100]), "psill" for the partial
+        sills (e.g., [0.8, 0.2]) and "smooth" for the smoothness paramter if exists for this model (e.g., [None, 0.2]).
```
adehecq, line 869: typo

xdem/spatialstats.py
```diff
@@ -820,213 +1028,384 @@ def vgm_sum(h, *args):
...
+    if not all([c in params_vgm for c in expected_columns]):
+        raise ValueError('The dataframe with variogram parameters must contain the columns "model", "range" and "psill".')
+
+    # Check that the format of variogram models are correct
+    for m in params_vgm['model'].values:
+        _get_scikitgstat_vgm_model_name(m)
+
+    # Check that the format of ranges, sills are correct
+    for r in params_vgm['range'].values:
+        if not isinstance(r, (float, np.integer)):
```
adehecq, line 1077: Would that condition fail is the value was a np.float? Maybe use `np.number` rather than `np.integerer` to allow any numpy number? 
I just checked how we handle any number in geoutils here: https://github.com/GlacioHack/GeoUtils/blob/c101c3ed1e7424e0492b453b08f34aaa52b0569a/geoutils/_typing.py#L14
Same for the other conditions below.

xdem/spatialstats.py
```diff
@@ -820,213 +1028,384 @@ def vgm_sum(h, *args):
...
-    over the area.
-    If SE is the standard error, SD the standard deviation and N_eff the number of effective samples, we have:
+    Number of effective samples approximated from exact disk integration of a sum of any number of variogram models
+    of spherical, gaussian, exponential or cubic form over a disk of a certain area. This approximation performs best
+    for areas with a shape close to that of a disk.
+    Inspired by Rolstad et al. (2009): http://dx.doi.org/10.3189/002214309789470950.
+    The input variogram parameters match the format of the dataframe returned by `fit_sum_variogram_models`, also
+    detailed in the parameter description to be passed manually.
+
+    This function is contains the exact integrated formulas and is mostly used for testing the numerical integration
```
adehecq, line 1112: typo: "is contains"

xdem/spatialstats.py
```diff
@@ -820,213 +1028,384 @@ def vgm_sum(h, *args):
...
+    :return: Integral between lower and upper bound
+    """
+    return integrate.quad(fun, low_b, upp_b)[0]
+
+def neff_circular_approx_numerical(area: float, params_vgm: pd.DataFrame) -> float:
     """
-    Number of effective samples derived from numerical integration for any sum of variogram models a circular area
-    (generalization of Rolstad et al. (2009): http://dx.doi.org/10.3189/002214309789470950)
-    The number of effective samples N_eff serves to convert between standard deviation/partial sills and standard error
+    Number of effective samples derived from numerical integration for any sum of variogram models a circular area.
```
adehecq, line 1210: "a circular area" -> "over a circular area"?

xdem/spatialstats.py
```diff
@@ -820,213 +1028,384 @@ def vgm_sum(h, *args):
...
-        psill_tot += vario[2]
 
+    # Check input dataframe
+    _check_validity_params_vgm(params_vgm)
+
+    # Get the total sill from the sum of partial sills
+    total_sill = np.nansum(params_vgm.psill)
+
+    # Define the covariance sum function times the spatial lag, for later integration
     def hcov_sum(h):
```
adehecq, line 1235: it seems like you have a similar function at L 972. Would it not be possible to use the same for both?

xdem/spatialstats.py
```diff
@@ -820,213 +1028,384 @@ def vgm_sum(h, *args):
...
-    :param model1: model of first variogram
-    :param crange2: correlation range of second variogram
-    :param psill2: partial sill of second variogram
-    :param model2: model of second variogram
-
-    :returns: number of effective samples
+     Exact number of effective samples derived from a double sum of covariance with euclidean coordinates based on
+     the provided variogram parameters. This method works for any shape of area.
+
+    :param coords: Coordinates at the scale of the spatial support (typically, pixel)
```
adehecq, line 1276: I don't fully understand what these two arrays are. Do you expect 1D, 2D, or any dimension? Is `errors` the error calculated for the standardization, or what?

xdem/spatialstats.py
```diff
@@ -165,110 +280,6 @@ def interp_nd_binning(df: pd.DataFrame, list_var_names: Union[str,list[str]], st
...
-
-    # concatenate everything
-    list_all_dfs = list_df_1d + list_df_2d + [df_stats_nd]
-    df_concat = pd.concat(list_all_dfs)
-    # commenting for now: pd.MultiIndex can be hard to use
-    # df_concat = df_concat.set_index(list_var_names)
-
-    return df_concat
-
 def create_circular_mask(shape: Union[int, Sequence[int]], center: Optional[list[float]] = None,
```
adehecq, line 283: Should this function be hidden? I don't think it is meant to be really used by the user.

adehecq: Same for create_ring_mask.

xdem/spatialstats.py
```diff
@@ -735,47 +819,171 @@ def sample_empirical_variogram(values: Union[np.ndarray, RasterType], gsd: float
...
+            raise ValueError('Variogram model name {} not recognized. Supported models are: '.format(model)+
+                             ', '.join(list_supported_models)+'.')
+
+    else:
+        raise ValueError('Variogram models can be passed as strings or skgstat.models function. '
+                         'Supported models are: '+', '.join(list_supported_models)+'.')
+
+    return model_name
+
+def get_func_sum_vgm_models(params_vgm: pd.DataFrame) -> Callable[[np.ndarray], np.ndarray]:
```
adehecq, line 863: Rename `get_vgm_models_func`?

xdem/spatialstats.py
```diff
@@ -735,47 +819,171 @@ def sample_empirical_variogram(values: Union[np.ndarray, RasterType], gsd: float
...
+    # Get covariance from sum of variogram
+    cov = covariance_from_vgm(params_vgm)
+
+    def rho(h):
+        return cov(h)/total_sill
+
+    return rho
+
+
+def fit_sum_model_variogram(list_models: list[str] | list[Callable], empirical_variogram: pd.DataFrame,
```
adehecq, line 953: Some functions use the term `vgm`, some `variogram`. We should try to be consistent. 
Maybe rename `fit_sum_vgm_models`?

xdem/spatialstats.py
```diff
@@ -820,213 +1028,384 @@ def vgm_sum(h, *args):
...
+        var = np.sum(errors.reshape((-1, 1)) @ errors_sub.reshape((1, -1)) * rho(pds_matrix_sub.flatten()).reshape(pds_matrix_sub.shape))
+
+    # The number of effective sample is the fraction of total sill by squared standard error
+    squared_se_dsc = var / (n * subsample)
+    neff = np.mean(errors)**2 / squared_se_dsc
+
+    return neff
 
 
 def std_err_finite(std: float, neff_tot: float, neff: float) -> float:
```
adehecq, line 1406: This will probably call for some explanations in the documentation.

adehecq:
Thanks for this huge effort ! 👏 I made some (mostly minor) comments directly in the code. We will see with practice how it goes.
Below are an answer to your questions.

> * Are all the function and parameters names fine with you? There's probably a bit improvement to do there, and once we have some good names we stick to them! But it's a bit hard to find those alone sometimes 😅.
> 
I went through each function's name and commented when I thought some changes could be useful. We should particularly try to be consistent between the use of `vgm` and `variogram`.
Of course, the different neff functions are a bit puzzling, but with the proper documentation it should be ok.
I could not add a comment at that line because it was not edited, but apparently `distance_latlon` is not used anymore and could be removed.

> * Should I do a core `neff` function, which defaults to one of the other functions, and can be used to choose which calculation? Should I change the others to hidden functions `_neff_...`? This would remove them from API.
> 
Mmmh. Maybe a good idea to have just one function with a parameter to use one or the other.

> * My idea for the updated documentation is to have: (i) beginner examples that fit in just 10 lines of code to exemplify each parameter and the basic behaviour of a function, (ii) advanced examples where things are more detailed/discussed.
> 
Sounds perfect!

> * For the purpose of 3., I would add an "error_pipeline" that automatically combines steps that go together: `nd_binning` and `interp_nd_binning`, a new "`standardization`" function, and a new "`error_map`" function; then another one that combines `sample_empirical_variogram`, `fit_sum_model_variogram` and `neff` to get the error in an area? Something along these lines.

Yes, I think it is important to have a simple pipeline that runs all the steps.

adehecq (dismissed):

xdem/spatialstats.py
rhugonnet: Done!

xdem/spatialstats.py
rhugonnet: Improved :sweat_smile:

xdem/spatialstats.py
rhugonnet: Solved with this: https://github.com/GlacioHack/xdem/wiki/Memo-on-Python-and-numpy-%60dtype%60

xdem/spatialstats.py
rhugonnet: Renamed everything with `variogram`, 6 letters doesn't make much of a difference so I went for clarity! :smile:

rhugonnet:
Merging now that comments are accounted for. 
The error pipeline + new documentation will be done in a separate PR.

PR #279, opened 2022-08-09
Move to python `3.10` in CI and remove python `3.7`

### LINKED ISSUE #249: Test python 3.10 in CI
The exact same motivation as https://github.com/GlacioHack/GeoUtils/issues/266

### LINKED ISSUE #260: Possible update in NumPy triggering warning for NaN addition
I've seen tests of `volume.py` fail several times in the past days due to a RuntimeWarning triggered during NaN addition that we use to initiate our empty arrays (in #227).
For example in line 552 of `volume.py`:

```python
    values = np.empty((n_bins, unique_indices.shape[0]), dtype=float) + np.nan
```

The failure was a bit random, so I guess it depends on the version of `numpy` set up in the environment, and this warning might be a recent addition. If this is the case (I couldn't find confirmation with a quick look), we might have to update a lot of initializations!

rhugonnet:
Main change in the title.

Also ran into the issue of RuntimeWarning with NaNs of #260 again. I fixed it by using `np.full(shape, np.nan)` to initialize a NaN array instead of multiplying or adding `np.zeros()`, `np.ones()` or `np.empty()` with `np.nan` like we have the (_now bad?_) habit of doing.
The `dtype` of output attributes defined as that of the DEM was removed. This was triggered by failing doctests because the `dtype` argument works differently if initiated in `np.full` or in `np.empy` later multiplied by NaN. In short, it now works as intended (before it was ignored because multiplied by a NaN). This is an issue, however. Because, if a DEM is an integer, all terrain attributes would now be derived as integers (and some attributes are between 0 and 1, for instance). 

Added a point on this in the wiki (https://github.com/GlacioHack/xdem/wiki/Stuck-on-a-mysterious-error-during-CI%3F):

> 4. The unexpected: Initialization of np.ndarray full of NaNs could historically be done in many ways: multiplying or summing a np.zeros, np.ones or np.empty with a np.nan. However, recently, RuntimeWarning have started to pop up a bit randomly with these methods, resulting in test failres. We could not track down the exact version change that triggered this (might be unexpected behaviour, so not in the logs), but it does seems to depend on the version of numpy. Solution: use np.full(shape, np.nan, dtype) to initialize those arrays!
> ```python
> values = np.empty((n_bins, unique_indices.shape[0]), dtype=float) * np.nan
> E       RuntimeWarning: invalid value encountered in multiply
> ```

Resolves #249 
Resolves #260


xdem/terrain.py
```diff
@@ -59,7 +59,7 @@ def _get_quadric_coefficients(
     L = resolution
 
     # Allocate the output.
-    output = np.empty((12,) + dem.shape, dtype=dem.dtype) + np.nan
```
adehecq, line 62: Is there any reason why the dtype argument has been removed?

adehecq (approved):
Thanks !

xdem/terrain.py
rhugonnet: Edited the answer in the PR description.

PR #280, opened 2022-08-09
Fix previously skipped tests that used to fail randomly

### LINKED ISSUE #209: Basinhopping randomly fails in CI
> I'm still desperately trying to understand just the base: why

After a few hours, still can't trace it back. :cry: 
I'm giving up for now, let's just open an issue that I'll reference to this comment.
Description here:

`test_robust_sumsin_fit` randomly fails, in CI only, for the following assertion:
```python
        for i in range(2):
            assert coefs[3*i] == pytest.approx(true_coefs[3*i], abs=0.02)
```
The function call is based on `scipy.optimize.basinhopping`. I cannot reproduce the issue locally, even when setting up `xdem-dev` environment the same way as in CI.
This is very strange because the test uses a random state `random_state=42` that is used both by `xdem.spatial_tools.subsample_raster` and by `basinhopping` (seed argument). And locally the output is indeed not subject to any random variation (constant).
The error seems to occur 75% of the time in CI. When it does, the `scipy` version (1.7.1) and `numpy` version (1.20.3) are exactly the same as the ones I have locally without error after running 15+ times. So the issue doesn't seem to be from there either...
The mystery remains!

_Originally posted by @rhugonnet in https://github.com/GlacioHack/xdem/issues/151#issuecomment-914391165_

### LINKED ISSUE #259: Warp_dem tests failing after environment update between 10.21 and 04.22
When starting this PR again, there was a couple of failures in tests that had nothing to do with `terrain.py` despite the tests passing 6 months ago.
I managed to solve most of them (e.g. deprecation warnings, etc), but there is still 2 tests that are failing which are directly related to the `warp_dem` function and I couldn't fully grasp the source. In the meantime I added a `skip` decorator, and I will open a related issue so that @erikmannerfelt can check this later.

_Originally posted by @rhugonnet in https://github.com/GlacioHack/xdem/issues/227#issuecomment-1104266663_

rhugonnet:
This PR fixes 1 random failing test in `test_fit.py` by removing the `pytest.skip` (wrong behaviour fixed directly by scipy) and 2 failing tests in `test_coreg.py` due to changes in `scipy.transform.warp` NaN propagation.

**Details below:**

For the failure in `test_fit.py`, my best trick for fixing yet: waiting :smile:! The issue with the random seed of `scipy.optimize.basin_hopping` seemed to have been fixed directly in `scipy` since the issue arose. I ran the test several times and it always passes now.

For the two tests that were skipped in `test_coreg.py`, they relate to issues with `skimage.transform.warp`. The tests let me think the behaviour has changed with NaNs, but I couldn't find anything in the release notes that would indicate it (https://scikit-image.org/docs/0.19.x/release_notes.html). We should keep in mind it's quite a young package.
Basically, `warp` now propagates NaNs with the order of interpolation, exactly the same as the "homemade" mask + dilation calculation that we use to do by default. So, using default parameters, results are unchanged. However, we cannot "ignore" the NaN propagation anymore, if that was part of the intended behaviour (I let @erikmannerfelt check that). There's probably more work needed to decide if we consistently use these functions within the package, as they are currently a bit apart.

Resolves #259
Resolves #209


tests/test_coreg.py
```diff
@@ -822,7 +821,7 @@ def test_warp_dem():
     # The warped DEM should have the value 'elev_shift' in the upper left corner.
     assert warped_dem[0, 0] == elev_shift
     # The corner should be zero, so the corner pixel (represents the corner minus resolution / 2) should be close.
-    assert warped_dem[-1, -1] < 1.0
+    assert warped_dem[-2, -1] < 1
```
adehecq, line 824: Why -2?

rhugonnet: Because the last row is NaNs propagated by the warp function. I added a comment to specify it in the test.

adehecq (approved):
Great that the issues with tests are fixed !

adehecq:
Which version of scipy is needed for the test_fit.py tests to run successfully?

PR #288, opened 2022-08-24
Force geoutils to 0.0.7

### LINKED ISSUE #268: Switch back from git-based to release-based GeoUtils in dev-environment
To avoid having to put a release out every time we change something in GeoUtils in order for the tests to pass, I switched the `dev-environment.yml` source for GeoUtils from conda:
```
- geoutils
```
to the latest git version:
```
- pip:
  - git+https://github.com/GlacioHack/GeoUtils.git
```
This was done in #265.

We should switch back to release-based GeoUtils once both packages have reached a minimal level of maturity and documentation (e.g., version 0.1 ?).

rhugonnet:
To avoid all hell breaking loose from https://github.com/GlacioHack/GeoUtils/pull/265, but also for long-term stability with releases.

Resolves #268

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -33,5 +33,6 @@ dependencies:
   - sphinx-gallery
 
   - pip:
-    - git+https://github.com/GlacioHack/GeoUtils.git
+    - geoutils==0.0.7
+#    - git+https://github.com/GlacioHack/GeoUtils.git
 
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -33,6 +33,6 @@ dependencies:
   - sphinx-gallery
 
   - pip:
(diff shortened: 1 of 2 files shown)


dev-environment.yml
```diff
@@ -33,5 +33,6 @@ dependencies:
   - sphinx-gallery
 
   - pip:
-    - git+https://github.com/GlacioHack/GeoUtils.git
+    - geoutils==0.0.7
```
adehecq, line 36: Should it not be moved out of the pip commands, directly in the conda install?

adehecq (approved):

dev-environment.yml
rhugonnet: Conda takes a day to update, pip is faster

PR #290, opened 2022-08-25
Add `HOW_TO_RELEASE` file

rhugonnet:
All in the title.
Also adds the incrementation of version number in "the hard way" following https://github.com/GlacioHack/geoutils/issues/269.

Ready to merge.


rhugonnet:
Exact same as in geoutils, should be no issue so merging now.

PR #291, opened 2022-08-25
Fix bugs following geoutils update

adehecq:
Because of the [MemoryFile PR](https://github.com/GlacioHack/GeoUtils/pull/265) in geoutils, several attributes of Raster objects do not exist. Fix bugs that arise with that change. Things important to keep in mind:
- the rasterio DatasetReader initially accessed with `self.ds` was removed
- the method `self._get_rio_attrs` was removed
- the method `self._update` was removed

Fixed several issues, but I still have a few that fail locally:

- [x] tests/test_docs.py::TestDocs::test_build -> don't understand why
- [x] tests/test_fit.py::TestRobustFitting::test_robust_sumsin_fit -> fail with AssertionError @rhugonnet do you want to check?
- [x] tests/test_spatialstats.py::TestVariogram::test_sample_multirange_variogram_default -> @rhugonnet?
- [x] xdem/terrain.py::xdem.terrain.slope -> there is some rounding error in the docstring. Not sure how we can set the threshold for doc checks.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -53,7 +53,7 @@ jobs:
       # pytest will run asynchronously, so test data need to be downloaded first.
     - name: Download example data
       run: |
-        python -c "from xdem.examples import *; download_longyearbyen_examples(overwrite=False)"
+        python -c "from xdem.examples import *; download_longyearbyen_examples(overwrite=True)"
     - name: Test with pytest
       run: |
         pip install pytest-cov coveralls
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -12,7 +12,7 @@ dependencies:
   - matplotlib
   - opencv
   - pyproj
-  - rasterio
+  - rasterio<=1.2.10
   - scipy
   - tqdm
   - scikit-image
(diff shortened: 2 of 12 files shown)


rhugonnet:
**Good news:** most of these tests you mention don't fail on the CI. 
Two reasons:
- Need to overwrite your example files locally,
- Need to update your environment (this will surely fix the `tests/test_fit.py`::TestRobustFitting::test_robust_sumsin_fit, which used to fail randomly but not anymore, see #209)

**Bad news:** the different value for variograms is also an issue in the CI, and I don't understand that one yet.

rhugonnet:
OK, so I added `test_example.py` with two functions that tests the example data to see if it stays the same. In particular to test if the data we process and store as examples (to avoid processing multiple times) stays the same. The data we fetch from `xdem-data` should never change, but now we'll be sure of it too :smile:.

I ran the test changing from `geoutils==0.0.8` to after the MemoryFile PR. The values in the arrays are still the same, but the nodata in the array are not. They change from **2316 points with nodata** (before MemoryFile PR) to 0 after.
@adehecq Any idea why?

rhugonnet:
Tests finally passing, in the end it was a `rasterio` issue with >1.3. Opened https://github.com/GlacioHack/geoutils/issues/293 and we'll have to create a reproducible example to open an issue there as well.

PR #301, opened 2022-09-10
Fix tests with GeoUtils `data.setter` and `__array_interface__` update

rhugonnet:
Updating a couple tests that used syntax only supported by the previous `__array_interface__` or raised `UserWarning` due to the new `data.setter` of `Raster`.
Will make CI pass once https://github.com/GlacioHack/geoutils/pull/302 is merged.

### DIFF
--- a/xdem/demcollection.py
+++ b/xdem/demcollection.py
@@ -47,8 +47,7 @@ class DEMCollection:
 
         # Find the sort indices from the timestamps
         indices = np.argsort(self.timestamps.astype("int64"))
-        self.dems = np.empty(len(dems), dtype=object)
-        self.dems[:] = [dems[i] for i in indices]
+        self.dems = [dems[i] for i in indices]
         self.ddems: list[xdem.dDEM] = []
         # The reference index changes place when sorted
         if isinstance(reference_dem, (int, np.integer)):
--- a/xdem/terrain.py
+++ b/xdem/terrain.py
@@ -1012,7 +1012,7 @@ def get_terrain_attribute(
     output_attributes = [terrain_attributes[key].reshape(dem.shape) for key in attribute]
 
     if isinstance(dem, gu.Raster):
-        output_attributes = [gu.Raster.from_array(attr, transform=dem.transform, crs=dem.crs, nodata=None) for attr in output_attributes]
+        output_attributes = [gu.Raster.from_array(attr, transform=dem.transform, crs=dem.crs, nodata=-99999) for attr in output_attributes]
 
     return output_attributes if len(output_attributes) > 1 else output_attributes[0]
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -57,7 +57,7 @@ jobs:
     - name: Test with pytest
       run: |
         pip install pytest-cov coveralls
-        pytest -ra -n=auto --cov=xdem/
+        pytest -ra --cov=xdem/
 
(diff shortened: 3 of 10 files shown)


xdem/demcollection.py
```diff
@@ -47,8 +47,7 @@ def __init__(self, dems: Union[list[gu.georaster.Raster], list[xdem.DEM]],
 
         # Find the sort indices from the timestamps
         indices = np.argsort(self.timestamps.astype("int64"))
-        self.dems = np.empty(len(dems), dtype=object)
-        self.dems[:] = [dems[i] for i in indices]
+        self.dems = [dems[i] for i in indices]
```
adehecq, line 50: Are you sure both are equivalent? Before `self.dems` was a `np.ndarray`, but now it's not.

xdem/terrain.py
```diff
@@ -1012,7 +1012,7 @@ def get_terrain_attribute(
     output_attributes = [terrain_attributes[key].reshape(dem.shape) for key in attribute]
 
     if isinstance(dem, gu.Raster):
-        output_attributes = [gu.Raster.from_array(attr, transform=dem.transform, crs=dem.crs, nodata=None) for attr in output_attributes]
+        output_attributes = [gu.Raster.from_array(attr, transform=dem.transform, crs=dem.crs, nodata=-99999) for attr in output_attributes]
```
adehecq, line 1015: It would probably be better to use the default for the float32 or float64 dtype (which I think is -99999, but it would be more transparent!)

xdem/demcollection.py
rhugonnet: Yes I think so (all tests passing so the `demcollection.py` module seems to be functioning correctly). 
Before, `self.dems` was an `np.array` just used for iterations through the `xdem.DEM` as `object` types. By using `[:]` on previous line 51, it broadcasted them differently with the `__array_interface__` within the list. Now it's consistent (and understandable). I guess we'll fix it later if we find an issue (and make the tests more complete).

adehecq: 👍

xdem/terrain.py
rhugonnet: It complains when I do that because `_default_ndv` is not supposed to be used outside of GeoUtils "access to a protected member of a module".
And, actually, we don't define the types of outputs properly in `terrain.py` yet (some like hillshade are `uint8`). We could improve this once we specify the output type.

rhugonnet: Opened #302

rhugonnet:
I finally found the source of the errors (it's been 2 weeks)! Parallelized tests run automatically with `pytest-xargs`, and this started impacting `Raster` reading/writing (maybe because of some issue in a low-level dependency).
The issue for tracking this, is that the effect was that sometimes (randomly) the grid was read as full of nodata.

Locally, we rarely run with `pytest -n=auto`, which is why we could not reproduce it. Here's the fix: [0a59b42](https://github.com/GlacioHack/xdem/pull/301/commits/0a59b426cd4eef979e6721aa706d5fe1129dfd21)

Closed https://github.com/GlacioHack/geoutils/issues/293. In the end, this is not related to floating nodata at all...

rhugonnet:
Merging this, as only 2 lines have changed: the `pytest-xargs` fix + the commit hash for the new `xdem-data` with nodata changed to -9999. We can decide later if we keep this new -9999, or revert to the floating nodata.

PR #304, opened 2022-09-18
Add Mac and Windows to CI

### LINKED ISSUE #126: Add Windows CI with GitHub
Many people use Windows unfortunately!

Luckily, GitHub's CI has support for testing on Windows machines, so this should be added as a check.

rhugonnet:
## Summary

This PR adds Windows and Mac to the full test suite run during CI. We now have 9 jobs running in CI (Python `3.8`, `3.9` and `3.10` for Ubuntu, Windows and Mac).
All core functionalities are working, however a couple little issues remain, opened for later work: making outputs of SciPy fully consistent between OS due to floating point considerations, making Sphinx work on all OS, and making PROJ work on Windows to use `proj-data` grids.

## In details

This PR:

- Unsets the `PROJ_LIB` variable during CI to allow `rasterio` to import correctly, otherwise failing on Windows. This fix is recommended by https://rasterio.readthedocs.io/en/latest/faq.html#why-can-t-rasterio-find-proj-db-rasterio-from-pypi-versions-1-2-0. Curiously, this is not needed in GeoUtils' CI, I couldn't figure why it fails in xDEM exactly (similar environments);
- Performs rounding of the results of `scipy.optimize.least_squares` after setting a lower `xtol` to ensure the optimized parameters found are the same between OSs. This works to get the exact same output of `NuthKaab()` co-registration on Windows and Linux, but somehow still differs on Mac... Added a comment on this in issue #310.
- Tries to do the same rounding operation for `scipy.optimize.basin_hopping` but the `xtol` parameter does not exist as such, we would need to derive the tolerance of the parameter from the scipy method and `ftol` or `gtol` parameters, opened issue #310;
- Skips tests related to grids stored by `proj-data` in `test_dem.py` on Windows, that currently can't be found, opened issue #311;
- Skips the building of the documentation with Sphinx that fails on both Mac and Windows with the same `TypeError: 'module' object is not callable` when `xdem.examples.get_path()` is called, notably in `examples/*/plot_*.py` scripts for `sphinx-gallery`, opened #316.

Unrelated but needed to fix bugs related to package updates, this PR: 

- Forces the version of `openh264` in `environment.yml` at `2.3.0` (on which `opencv` depends), as it started failing with `2.3.1` on 24.09.22 (probably an erroneous release from them...), opened #317;
- Upgrades `pre-commit` hooks to solve a `mypy` issue with the latest Python 3.10 version: https://github.com/python/mypy/issues/13627;
- Updates the regex pattern to catch `RuntimeWarning` with `true_divide`, as those can be simply called `divide` now.

Resolves #126


adehecq:
Haha, I see some failure on Windows 😆

rhugonnet:
One issue closed, four open! But at least now we know that all the important stuff works as intended on all OSs! :partying_face: 

Ready to merge, then I'll add the cache in xDEM to speed up CI, that can take up to 30min right now...

.github/workflows/python-package.yml
```diff
@@ -48,6 +47,7 @@ jobs:
     # development-specific dependencies by differencing the env and dev-env yml files
     - name: Check normal environment import
       run: |
+        unset PROJ_LIB
```
adehecq, line 50: maybe add a comment here to state why this is needed?

dev-environment.yml
```diff
@@ -4,14 +4,16 @@ channels:
...
   - python>=3.8
   - proj>=7.2
+  - geopandas>=0.10.0
   - fiona
   - shapely
   - numba
   - numpy
   - matplotlib
   - opencv
+  - openh264=2.3.0
```
adehecq, line 14: to be removed later when the issue is fixed, no?

dev-environment.yml
```diff
@@ -34,4 +36,4 @@ dependencies:
   - richdem
 
   - pip:
-    - git+https://github.com/GlacioHack/geoutils.git
+    - geoutils==0.0.9
```
adehecq, line 39: Is the plan then to always link a version of xdem to a version of geoutils?

adehecq (dismissed):
Looking good !
CI will take even longer now, but hopefully the cache will speed things up.

dev-environment.yml
adehecq: Added an issue so we remove it in the future.

.github/workflows/python-package.yml
rhugonnet: Done

dev-environment.yml
rhugonnet: :+1:

dev-environment.yml
rhugonnet: Yes, for stable versions at least.

rhugonnet:
Merging as there's only two comments added since last approval

Issue #318, opened 2022-10-14
opencv no longer an optional dependency


adehecq:
Because of the filters.py module, that import opencv at [this line](https://github.com/GlacioHack/xdem/blob/50fa7351cd257b110c6cdcbed75f2e37f7171b02/xdem/filters.py#L6), opencv is no longer an optional dependency, so if not installed, it will fail.
I'm surprised this does not show up in the CI, but I had the issue when running a binder with xdem as a dependency.
A quick fix is to update the import as in coreg.py for example.

PR #319, opened 2022-10-14
Make opencv import optional

adehecq:
Solves #318.


rhugonnet:
Nice! I noticed the CI error was because you needed to keep `import cv2 as cv` instead of just `import cv2` and fixed it :smile:. Should be ready to merge after checks.

rhugonnet (dismissed):

rhugonnet (approved):

Issue #310, opened 2022-09-27
Fix results of `scipy.optimize` between OS


rhugonnet:
The output of `scipy.optimize.basinhopping` varies between operating systems, added TODOs in the tests in #304. It does not look like it's due to the random `seed` that seems correctly passed, but the output tolerance.
For `scipy.optimize.least_squares`, I managed to have the same results on different OS by rounding the result 100 times above the variable tolerance `xtol` provided to the optimize function (not sure if it's 100% robust, need to think more on it).

For `basinhopping` and the method we use in it, one can only pass `ftol` (cost function) or `gtol` (gradient), which does not translate directly into `xtol` (variables of interest). So the tolerances might be too low compared to the rounding, hence different results on different OS. The equations of the tolerances are provided in the `scipy` doc, so we should be able to work back the influence on the variable tolerance to ensure all results are fixed.

rhugonnet:
**Update:** in the PR #304, the output of `scipy.optimize.least_squares` after rounding with a lower tolerance than `xtol` is the same for Linux and Windows, but not for Mac.
Need to find a way to solve this more robustly...

rhugonnet:
Revisiting this about this 3 years later: As long as the precision is sufficient small, it's normal that the results would differ on different OS.

Given that we now test that precision in details in the new `test_affine` (with the `__example` test that check a fixed value output), typically at 10e-5), I think this is solved.

rhugonnet:
PR that introduced the new tests: https://github.com/GlacioHack/xdem/pull/585

Issue #257, opened 2022-04-06
Create a one-liner for DEM coregistration


adehecq:
Currently, the easiest way to run DEM coregistration is to follow the example [here](https://github.com/GlacioHack/xdem/blob/main/examples/plot_nuth_kaab.py). This is several lines long, including reading files, reprojecting, creating mask, running coreg.fit and apply, saving to disk...
Ideally this could be done in a single function with some default parameters, as done for [RAGMAC](https://github.com/adehecq/ragmac_xdem/blob/master/ragmac_xdem/dem_postprocessing.py#L325 ).

erikmannerfelt:
@adehecq, is this solved now?

rhugonnet:
Essentially, but there was still a list of TODOs in #267. Maybe it can be moved to a new issue "Refine one-liner for DEM coregistration"?

rhugonnet:
Actually an issue for refining that PR is opened: #324, closing!

PR #322, opened 2022-10-17
Get consistent `scipy.optimize` outputs out of `NuthKaab`

rhugonnet:
This PR fixes the outputs of `NuthKaab` for all operating systems. This is done by rounding the outputs of `scipy.optimize.least_squares` that have varying floating precision to 1e-2.
A test is added to `test_coreg.py` to specifically check the `Coreg` metadata out of the example, to know where changes are ocurring.

Opened #327. This issue mentions in particular that `scipy.optimize` functions should be called from `fit.py`, in order to have consistent tests on the tolerance and outputs of fitting methods, to continue addressing #310.

Partially addresses #310


rhugonnet:
Will merge this now so that xDEM's build passes again after #257

adehecq (approved):
Ok, so basically all you did was to round the output of the parameter fit, and then update the expected output?

rhugonnet:
Yes !

PR #328, opened 2022-10-25
Fix loading of processed examples in `examples.py`

rhugonnet:
This PR fixes a bug in the function `process_coregistered_examples` of `examples.py` that made the `ddem` processing run every time the file was called with `xdem.example.get_path("longyearbyen_ddem")`. This should save us some time in testing and building documentation! :smile:
 
Additionally, it makes the variables `FILEPATHS_DATA` and `FILEPATHS_PROCESSED` not accessible to users (adding `_` in front), and create the `xdem.examples.available` variable that list keys of both dictionaries that can be called with `xdem.examples.get_path()`, for consistency with GeoUtils https://github.com/GlacioHack/geoutils/pull/284.

### DIFF
--- a/docs/source/quick_start.rst
+++ b/docs/source/quick_start.rst
@@ -16,7 +16,7 @@ The dataset keys and paths can be found from
 
 .. code-block:: python
 
-        xdem.examples.FILEPATHS_DATA
+        xdem.examples.available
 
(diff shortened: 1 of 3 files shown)


adehecq (approved):
Great ! And nice that it is now consistent with geoutils.

Issue #327, opened 2022-10-25
Re-organization of `coreg.py`


rhugonnet:
Several ideas here in relation to the one-liner coregistration: https://github.com/GlacioHack/xdem/pull/267#pullrequestreview-1145346346

More generically:

- The core functionalities of `Coreg` methods should live outside of `Coreg` classes for readability and flexibility (e.g., functions to estimate the horizontal shift, to deramp, etc). In other words all `Coreg` methods should only be wrappers of already existing functions;
- Optimization functions should call the methods of `fit.py` instead of directly calling `scipy.optimize`, so that their tolerance and outputs are consistently tested, and that any type of optimization method can be used;
- While all class method should retain flexibility to work with `np.ndarray`, they should still be optimized to function with `Raster` objects (e.g., modifying the geotransform instead of resampling in case of a horizontal shift). To have a similar default behaviour for `np.ndarray` objects, we could potentially introduce a return of a `transform` object as output of `apply()`.

Maybe @adehecq has more specific things to add!

adehecq:
Yes, I agree with those points. Below I outlined the main steps to be done, hopefully in logical order:
1. Step 2 requires that the data is convertible to a DEM instance -> Change coreg apply and fit functions so that they accept either an xdem.DEM instance, or a numpy array, transform and CRS.
2. Replace all Coreg core functionalities with either standalone function or xdem/geoutils functions 
    * Replace calculate_slope_and_aspect with xdem.terrain attributes
    * Replace subsampling with geoutils’ subsampling
    * Remove duplicates of deramping
3. Add function to calculate standard stats before/after coregistration (mean, median, NMAD, std, others?). Call this function in apply.
4. Add plotting step in all `apply` methods, and in some `fit` methods, like N&K.
5. Update Coreg so that coregistration params are more easily output than currently with _meta
6. Add function to create inlier mask (load vectors, remove slopes, outliers…)
7. Save input/output DEM during fit and apply functions to avoid redoing calc?
8. Remove unused/redundant functions (`filter_by_range`, `filtered_slope`,`apply_xy_shift`, `apply_z_shift`,  `mask_as_array`, `dilate_mask` option, `Coreg.residuals` and `error`)

rhugonnet:
Adding some points after closing #158:
9. Move `apply_matrix` from `base.py` to `affine.py` by overridding `apply()` in `AffineCoreg` (as done in `BiasCorr`), and related functions,
[...]

Issue #150, opened 2021-06-21
`xdem.coreg.calculate_slope_and_aspect` should (soon) be deprecated.


erikmannerfelt:
Brought up in #149.

This function technically does the job but returns invalid and inconsistent units.

- Slope is returned in m/pixel (not m/m or degrees; the resolution is not taken into account)
- Aspect is returned in radians (technically correct, but unusual)

The point of the function is to use it in the NuthKaab approach. Problems may arise, however, if a user finds it and tries to use it for other purposes.

The best approach would be to deprecate it and modify NuthKaab to not use it. With #147, these functionalities will exist in a much more standard way. At least, some kind of warning or renaming should be done to the function, to avoid the case of people using the function for the purpose (it sounds like) it is intended for.

adehecq:
Just make sure we don't break the Nuth&Kaab doing so! ;-)
The "slope" that is returned by that function is the one that we want in the algorithm, i.e tan(slope).
Having a slope function that calculates the arctan, then applying tan to it is not super efficient, and it may lead to some numerical rounding maybe?

erikmannerfelt:
We don't have to remove it necessarily! We can just mark it as a private ("_blabla") function and rename it to something that says "get slope/aspect for nuth_kaab"

adehecq:
Just came across this issue @rhugonnet. This is related to our discussions about the slope calculation.

Issue #315, opened 2022-10-06
Remove dilate_mask option of coreg.apply_matrix


adehecq:
Following PR #267 and issue #314, the dilate_mask option was commented out.
We need to remove that option completely, which requires updating all occurrences of that option in coreg.py and in the tests.

PR #329, opened 2022-10-25
Re-organize coreg.py

adehecq:
## Summary
This PR is a (big) step towards making the use of coreg simpler and more consistent, as stated in #327.
**NB:** Some previous commits were accidentally added to this PR, to see the actual changes, use [this link](52b72260c9d6ebecad50c98e632a2ff832f44cb2).

Here is a summary of the main changes:
- All coreg `apply` and `fit` functions accept either an xdem.DEM instance, or a numpy array, transform and CRS, i.e. the minimum information required to convert the output to a Raster/DEM (only array and transform were needed before).
- Replace all Coreg core functionalities with either standalone functions or xdem/geoutils functions:
    - ~~Replace `calculate_slope_and_aspect` with xdem.terrain attributes~~ -> actually reverted to old implementation, that is much faster. I left the xdem option commented, and could be reverted once xdem terrain attributes are sped-up, but in any case, it seems sufficient and more computationally efficient, to use a simple `np.gradient` rather than more complex gradient. Renamed the function to `_calculate_slope_and_aspect_nuthkaab` to avoid misuse as discussed in #150.
    - Replace subsampling block with geoutils’ subsampling
    - Add a standalone function to create inlier_mask (load vectors, remove low/high slopes, outliers…)
    - Add function to calculate standard stats before/after coregistration. Defaults are "count", "mean", "median", "nmad", "std" but can be set by the user. 
- Add an option `resample` to the apply methods to avoid resampling the data (and hence spreading NaNs). This is currently implemented for `BiasCorr`, `Deramp`, `ZScaleCorr` and `NuthKaab`. Resampling is necessary for all non-rigid transformations. In the case of `NuthKaab`, the output DEM has a different transform (and hence must be reprojected before calculating a ddem). This is useful if other processing follow the coregistration such as a second coreg method or regridding.
- Update `dem_coregistration` function to use apply's option `resample=False` and take as input xdem.DEM instances on top of filenames.
- remove the hmode/vmode in `dem_coregistration` since it's not used at the moment. The idea was possibly to allow two different inlier_masks for the horizontal vs vertical alignment.
- Remove unused/redundant functions (`filter_by_range`, `filtered_slope`, `apply_xy_shift`, `apply_z_shift`, `dilate_mask` option, duplicates of deramping)
- Add associated tests (in particular for `resample` option, `create_inlier_mask` and `dem_coregistration`).

## To be addressed later
- [ ] Call `calculate_ddem_stats` in `coreg.apply` to get standard statistics -> This is more complex than expected, because reference DEM and inlier_mask are unknown in `apply`. 
- [ ] Add plotting step in all apply methods, and in some fit methods, like N&K. -> same issue as previous
- [ ] Update Coreg so that coregistration params are more easily output than currently with _meta
- [ ] Modify `fit` to return coregistered DEM, so that `apply` is optional (to avoid re-doing calculation) -> turns out it only matters for CoregPipeline and NuthKaab (in the latter case, the shift is calculated in one go with `apply` instead of sequentially, so it may be beneficial).
- [ ] Update `Coreg.error` to not require re-running `apply` another time. -> will be addressed later
- [ ] Move function `mask_as_array`?
- [ ] Update documentation
- [ ] Make the resampling method (bilinear, cubic) an argument along the new `resample` argument
- [ ] Make the statistic for vertical shift (currently hardcoded `np.nanmedian` in N&K a parameter


tests/test_coreg.py
```diff
@@ -911,3 +1027,241 @@ def test_warp_dem() -> None:
...
+        ref,
+        [
+            outlines,
+        ],
+        inout=[
+            -1,
+        ],
+        filtering=False,
+    )
+    assert np.all(inlier_mask_comp == inlier_mask)
```
rhugonnet, line 1075: Here you could just compare with `~inlier_mask_comp` without re-generating it above after the start of the "Masking outside" case, I think

rhugonnet (dismissed):
Amazing!! :tada:
Thanks for all the work @adehecq, this is really a big step towards a `Coreg` module that is modular, robust and practical :)
I noted two things that I added at the bottom of the description on the next things to-do: 1/ Have the resampling algorithm as a parameter, to go with `resample`; 2/ Have the vertical shift statistic as a parameter in the Nuth & Kääb, hardcoded as `np.median` right now.

tests/test_coreg.py
rhugonnet: Done in [88391b1](https://github.com/GlacioHack/xdem/pull/329/commits/88391b18ee18c0ff67e4a2c2f1ffa4e226a2a81c)

xdem/coreg.py
```diff
@@ -203,69 +167,97 @@ def residuals(parameters: tuple[float, float, float], y_values: NDArrayf, x_valu
...
-        valid_diffs = valid_diffs[random_indices]
-        valid_x_coords = valid_x_coords[random_indices]
-        valid_y_coords = valid_y_coords[random_indices]
-
-    # Create a function whose residuals will be attempted to minimise
-    def estimate_values(
-        x_coordinates: NDArrayf, y_coordinates: NDArrayf, coefficients: NDArrayf, degree: int
-    ) -> NDArrayf:
+    # Extract only valid pixels
+    valid_mask = ~spatial_tools.get_mask(ddem)
```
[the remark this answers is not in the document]
rhugonnet: Done in [88391b1](https://github.com/GlacioHack/xdem/pull/329/commits/88391b18ee18c0ff67e4a2c2f1ffa4e226a2a81c)

xdem/coreg.py
```diff
@@ -203,69 +167,97 @@ def residuals(parameters: tuple[float, float, float], y_values: NDArrayf, x_valu
...
+    if len(stats_list) != len(stats_labels):
+        raise ValueError("Number of items in `stats_list` and `stats_labels` should be identical.")
+    for stat, label in zip(stats_list, stats_labels):
+        if not callable(stat):
+            raise ValueError(f"Item {stat} in `stats_list` should be a callable/function.")
+        if not isinstance(label, str):
+            raise ValueError(f"Item {label} in `stats_labels` should be a string.")
+
+    # Get the mask of valid and inliers pixels
+    nan_mask = spatial_tools.get_mask(ddem)
```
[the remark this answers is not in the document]
rhugonnet: Done in [88391b1](https://github.com/GlacioHack/xdem/pull/329/commits/88391b18ee18c0ff67e4a2c2f1ffa4e226a2a81c)

xdem/coreg.py
```diff
@@ -203,69 +167,97 @@ def residuals(parameters: tuple[float, float, float], y_values: NDArrayf, x_valu
...
+    ddem: NDArrayf | MArrayf,
+    inlier_mask: NDArrayf | None = None,
+    stats_list: tuple[Callable[[NDArrayf], AnyNumber], ...] | None = None,
+    stats_labels: tuple[str, ...] | None = None,
+) -> dict[str, float]:
+    """
+    Calculate standard statistics of ddem, e.g., to be used to compare before/after coregistration.
+    Default statistics are: count, mean, median, NMAD and std.
+
+    :param ddem: The DEM difference to be analyzed
```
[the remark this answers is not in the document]
rhugonnet: Done in [88391b1](https://github.com/GlacioHack/xdem/pull/329/commits/88391b18ee18c0ff67e4a2c2f1ffa4e226a2a81c)

xdem/coreg.py
```diff
@@ -203,69 +167,97 @@ def residuals(parameters: tuple[float, float, float], y_values: NDArrayf, x_valu
     return east_offset, north_offset, c_parameter
 
 
-def calculate_slope_and_aspect(dem: NDArrayf) -> tuple[NDArrayf, NDArrayf]:
+def apply_xy_shift(transform: rio.transform.Affine, dx: float, dy: float) -> NDArrayf:
```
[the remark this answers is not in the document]
rhugonnet: Done in [88391b1](https://github.com/GlacioHack/xdem/pull/329/commits/88391b18ee18c0ff67e4a2c2f1ffa4e226a2a81c)

xdem/coreg.py
```diff
@@ -603,33 +609,58 @@ def apply(
...
+                applied_dem,
+                destination=applied_dem,
+                src_transform=out_transform,
+                dst_transform=transform,
+                src_crs=crs,
+                dst_crs=crs,
+                resampling=rio.warp.Resampling.bilinear,  # Could make this an argument
+                dst_nodata=dst_nodata,
+            )
+            applied_dem[applied_dem == dst_nodata] = np.nan
```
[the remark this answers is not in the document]
rhugonnet: Done in [ca28258](https://github.com/GlacioHack/xdem/pull/329/commits/ca282587b04e0d17eb1d68cc7fd8ec0c3a59fa28)

adehecq (commented):
Thanks for the small fixes @rhugonnet !
Now there's a weird thing, you need to approve again, as I can't approve the PR myself 😆 
We also need to fix the test error.

xdem/coreg.py
```diff
@@ -1506,7 +1534,6 @@ def apply_matrix(
     :param invert: Invert the transformation matrix.
     :param centroid: The X/Y/Z transformation centroid. Irrelevant for pure translations. Defaults to the midpoint (Z=0)
     :param resampling: The resampling method to use. Can be `nearest`, `bilinear`, `cubic` or an integer from 0-5.
-    :param dilate_mask: DEPRECATED - This option does not do anything anymore. Will be removed in the future.
```
erikmannerfelt, line 1509: Wait sorry I missed something here. Why is this not needed any more?

xdem/coreg.py
```diff
@@ -1979,7 +1997,14 @@ def subdivide_array(self, shape: tuple[int, ...]) -> NDArrayf:
             shape = (shape[1], shape[2])
         return spatial_tools.subdivide_array(shape, count=self.subdivision)
 
-    def _apply_func(self, dem: NDArrayf, transform: rio.transform.Affine, **kwargs: Any) -> NDArrayf:
+    def _apply_func(
+        self, dem: NDArrayf, transform: rio.transform.Affine, crs: rio.crs.CRS, **kwargs: Any
+    ) -> tuple[NDArrayf, rio.transform.Affine]:
+
+        # The option resample=False is not implemented for this case
+        if "resample" in list(kwargs.keys()):
```
erikmannerfelt, line 2005: Is this not as simple as `"resample" in kwargs and kwargs["resample"] is not True`?

Note the more explicit boolean comparison. `not None` or `not ""` is also True.

`is False` is of course easier, but then a weirdly typed resample would be okay, like `resample = None`

xdem/coreg.py
```diff
@@ -2155,134 +2180,233 @@ def warp_dem(
...
+    - pixels with low/high slope (with filtering=True and set slope_lim values)
+
+    :param src_dem: the source DEM to be coregistered, as a Raster or DEM instance.
+    :param ref_dem: the reference DEM, must have same grid as src_dem. To be used for filtering only.
+    :param shp_list: a list of one or several paths to shapefiles to use for masking. Default is none.
+    :param inout: a list of same size as shp_list. For each shapefile, set to 1 (resp. -1) to specify whether \
+to mask inside (resp. outside) of the polygons. Defaults to masking inside polygons for all shapefiles.
+    :param filtering: if set to True, pixels will be removed based on dh values or slope (see next arguments).
+    :param dh_max: remove pixels for which abs(dh) is more than this value. Default is None.
+    :param nmad_factor: pixels where abs(dh) differ by nmad_factro * NMAD from the median
```
erikmannerfelt, line 2211: Typo at `nmad_factro`

xdem/coreg.py
```diff
@@ -2155,134 +2180,233 @@ def warp_dem(
...
+    nmad_factor: AnyNumber = 5,
+    slope_lim: list[AnyNumber] | tuple[AnyNumber, AnyNumber] = (0.1, 40),
+) -> NDArrayf:
+    """
+    Create a mask of inliers pixels to be used for coregistration. The following pixels can be excluded:
+    - pixels within polygons of file in shp_list (with corresponding inout element set to 1) - useful for masking \
+unstable terrain like glaciers.
+    - pixels outside polygons of file in shp_list (with corresponding inout element set to -1) - useful to \
+delineate a known stable area.
+    - pixels with absolute dh (=src-ref) are larger than a given threshold
```
erikmannerfelt, line 2199: I first thought `src-ref` was a typo. Mind doing `(src_dem-ref_dem).abs()`  or `|src - ref|`, or `(src - ref).abs()`, whatever you prefer?

xdem/coreg.py
```diff
@@ -2155,134 +2180,233 @@ def warp_dem(
...
+    filtering: bool = True,
+    dh_max: AnyNumber = None,
+    nmad_factor: AnyNumber = 5,
+    slope_lim: list[AnyNumber] | tuple[AnyNumber, AnyNumber] = (0.1, 40),
+) -> NDArrayf:
+    """
+    Create a mask of inliers pixels to be used for coregistration. The following pixels can be excluded:
+    - pixels within polygons of file in shp_list (with corresponding inout element set to 1) - useful for masking \
+unstable terrain like glaciers.
+    - pixels outside polygons of file in shp_list (with corresponding inout element set to -1) - useful to \
```
erikmannerfelt, line 2197: Change to `file(s)`

xdem/coreg.py
```diff
@@ -2155,134 +2180,233 @@ def warp_dem(
...
+    :param coreg_method: The xdem coregistration method, or pipeline.
+    :param grid: The grid to be used during coregistration, set either to "ref" or "src".
+    :param resample: If set to True, will reproject output Raster on the same grid as input. Otherwise, only \
+the array/transform will be updated (if possible) and no resampling is done. Useful to avoid spreading data gaps.
+    :param shp_list: A list of one or several paths to shapefiles to use for masking.
+    :param inout: A list of same size as shp_list. For each shapefile, set to 1 (resp. -1) to specify whether \
+to mask inside (resp. outside) of the polygons. Defaults to masking inside polygons for all shapefiles.
+    :param filtering: If set to True, filtering will be applied prior to coregistration.
+    :param dh_max: Remove pixels where abs(src - ref) is more than this value.
+    :param nmad_factor: Remove pixels where abs(src - ref) differ by nmad_factor * NMAD from the median.
```
erikmannerfelt, line 2327: This is much better than in the mask generation docstring! (See my comment above)

xdem/coreg.py
```diff
@@ -2155,134 +2180,233 @@ def warp_dem(
...
+    src_dem: RasterType,
+    ref_dem: RasterType,
+    shp_list: list[str | gu.Vector | None] | tuple[str | gu.Vector] | tuple[()] = (),
+    inout: list[int] | tuple[int] | tuple[()] = (),
+    filtering: bool = True,
+    dh_max: AnyNumber = None,
+    nmad_factor: AnyNumber = 5,
+    slope_lim: list[AnyNumber] | tuple[AnyNumber, AnyNumber] = (0.1, 40),
+) -> NDArrayf:
+    """
```
erikmannerfelt, line 2193: Generally, the docstring of `dem_coregistration()` reads much better than this one at the same keys. Mind copying the `dem_coregistration()` versions here?

In the future, perhaps we should have a system like pandas where documentation can be shared among multiple functions. That makes it easier to be consistent!

xdem/coreg.py
```diff
@@ -2155,134 +2180,233 @@ def warp_dem(
...
     else:
-        raise ValueError(f"`grid` must be either 'ref' or 'src' - currently set to {grid}")
+        ref_dem = ref_dem_path
+        src_dem = src_dem_path
+        if grid == "ref":
+            src_dem = src_dem.reproject(ref_dem, silent=True)
+        elif grid == "src":
+            ref_dem = ref_dem.reproject(src_dem, silent=True)
 
     # Convert to DEM instance with Float32 dtype
```
erikmannerfelt, line 2377: I know this is already merged, but why do they have to be `float32` specifically? I get that integers are bad for coregistration, but I think this should at least be an argument. What if I'm working with float16 or float64?

erikmannerfelt (commented):
A lot of great work, Amaury! I have some comments but generally I agree with all decisions made.

As you may notice, I've been out of the game for a while so some comments/questions may be self-explanatory to you and Romain at this point. I'm trying my best to get back!

xdem/coreg.py
```diff
@@ -278,72 +273,50 @@ def estimate_values(
         if len(coefficients) != coefficient_size:
             raise ValueError()
 
-        # Do Amaury's black magic to estimate the values.
```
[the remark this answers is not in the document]
adehecq: 😆

xdem/coreg.py
```diff
@@ -473,7 +447,8 @@ def fit(
         :param reference_dem: 2D array of elevation values acting reference.
         :param dem_to_be_aligned: 2D array of elevation values to be aligned.
         :param inlier_mask: Optional. 2D boolean array of areas to include in the analysis (inliers=True).
-        :param transform: Optional. Transform of the reference_dem. Mandatory in some cases.
+        :param transform: Optional. Transform of the reference_dem. Mandatory if DEM provided as array.
+        :param crs: Optional. CRS of the reference_dem. Mandatory if DEM provided as array.
```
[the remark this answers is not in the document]
adehecq: Yes, the CRS is mandatory only since this PR (or previous one). It's needed as you say just to check that some calculations make sense, in particular for the slope calculation for NuthKaab.
I'll see if I can easily make it optional, but only mandatory whenever it matters.

xdem/coreg.py
adehecq: Because the code that was using it was deleted in a previous PR (and it was on TODO list to take the time to fully remove it).
See the issue about it https://github.com/GlacioHack/xdem/issues/315.

xdem/coreg.py
```diff
@@ -203,69 +167,100 @@ def residuals(parameters: tuple[float, float, float], y_values: NDArrayf, x_valu
...
+
+    return stats
 
 
 def deramping(
-    elevation_difference: NDArrayf,
-    x_coordinates: NDArrayf,
-    y_coordinates: NDArrayf,
+    ddem: NDArrayf | MArrayf,
+    x_coords: NDArrayf | MArrayf,
```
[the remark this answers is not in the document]
adehecq: Removed `MArrayf` in 73fbd47

xdem/coreg.py
```diff
@@ -278,72 +273,50 @@ def estimate_values(
...
+
+    # Optimize polynomial parameters
+    coefs = scipy.optimize.leastsq(
+        func=residuals,
+        x0=np.zeros(shape=((degree + 1) * (degree + 2) // 2)),
+        args=(x_coords, y_coords, ddem),
+    )
 
-    def ramp(x_coordinates: NDArrayf, y_coordinates: NDArrayf) -> NDArrayf:
+    def fit_func(x: NDArrayf, y: NDArrayf) -> NDArrayf:
```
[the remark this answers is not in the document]
adehecq: Not so sure why it matters because the function is already hidden within another function... 🤔 But I renamed it anyway in ada836d.

xdem/coreg.py
```diff
@@ -278,72 +273,50 @@ def estimate_values(
...
 
         :param x_coordinates: x-coordinates of interest.
         :param y_coordinates: y-coordinates of interest.
 
-        :returns: The estimated ramp offsets.
+        :returns: The estimated elevation difference bias.
         """
-        return estimate_values(x_coordinates, y_coordinates, coefficients, degree)
-
-    if metadata is not None:
```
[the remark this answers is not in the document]
adehecq: This functionality was not used anywhere in xdem. And because I tried to unify the way errors are calculated with `calculate_ddem_stats`.

xdem/coreg.py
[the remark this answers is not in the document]
adehecq: I had a quick look and I think it would be too much effort and make the code more complex than needed to make CRS optional. 
Because most functionalities are meant to be performed on Raster instances, it makes sense that we request the minimum information needed for a Raster i.e. a transform and CRS. I can't see a reasonable case for which a CRS would not be defined, apart from tests. So I'll leave it as it is. I don't think we need to change the code for hypothetical use cases for the moment. 😇

xdem/coreg.py
adehecq: Done

xdem/coreg.py
adehecq: Done

xdem/coreg.py
adehecq: Done in 94c6a07

xdem/coreg.py
adehecq: Done in 94c6a07

Yes, if you're motivated, I'm happy to see such an example !

xdem/coreg.py
adehecq: Yes, seems to be working too, thanks!
Changed in bde8967

xdem/coreg.py
```diff
@@ -2155,134 +2180,233 @@ def warp_dem(
...
+
+        # Exclude steep slopes for coreg
+        slope = xdem.terrain.slope(ref_dem)
+        inlier_mask[slope.data < slope_lim[0]] = False
+        inlier_mask[slope.data > slope_lim[1]] = False
+
+    return inlier_mask
 
 
 def dem_coregistration(
```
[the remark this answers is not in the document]
adehecq: Done in 7fed61d

xdem/coreg.py
adehecq: That's a good point! My intention was to avoid having int dtypes, but indeed, it doesn't have to be float32. On the other hand, it yields probably very similar results whether it is float16 or float64 and I feel that float32 is a good compromise.
I'll add a comment anyway to remind later us !

xdem/coreg.py
```diff
@@ -662,10 +662,10 @@ def apply(
                 resampling=rio.warp.Resampling.bilinear,  # Could make this an argument
                 dst_nodata=dst_nodata,
             )
-            applied_dem[applied_dem == dst_nodata] = np.nan
+
 
         # Calculate final mask
-        final_mask = ~np.isfinite(applied_dem)
+        final_mask = np.logical_or(~np.isfinite(applied_dem), applied_dem == dst_nodata)
```
adehecq, line 674: There's ab ug here, as dst_nodata is not defined for `resample=False`.

adehecq: Fixed in e7b847b

adehecq:
Some tests fail. I've checked all of them:
- test_coreg fails due to a bug in an older version of geoutils, but fix in the latest here: https://github.com/GlacioHack/geoutils/pull/326
- test_docs fails for reasons that do not seem related to this PR. It fails on `code/spatialstats_heterosc_slope.py`, but the code itself runs without problem locally. But the exec statement in test_docs raises a weird error: "*** RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject" When trying to run the commands manually, it even fails in an earlier script...
- I haven't looked in detail at the errors in spatialstats.py, but again, I doubt they are related to this PR, but maybe to recent changes in geoutils?
Your input would be useful on this @rhugonnet !

erikmannerfelt:
> Some tests fail. I've checked all of them:
> 
>     * test_coreg fails due to a bug in an older version of geoutils, but fix in the latest here: [Fix small bug in spatial_tools geoutils#326](https://github.com/GlacioHack/geoutils/pull/326)
> 
>     * test_docs fails for reasons that do not seem related to this PR. It fails on `code/spatialstats_heterosc_slope.py`, but the code itself runs without problem locally. But the exec statement in test_docs raises a weird error: "*** RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject" When trying to run the commands manually, it even fails in an earlier script...
> 
>     * I haven't looked in detail at the errors in spatialstats.py, but again, I doubt they are related to this PR, but maybe to recent changes in geoutils?
>       Your input would be useful on this @rhugonnet !

@adehecq, do you want to merge anyway and add an issue ASAP, since this seems unrelated to your PR? The only consequence is an ugly "Faling" badge on the front page.

rhugonnet:
For the errors in `test_spatialstats`, they might actually be related to the change in nodata of the `ddem`?

We also had to change this test in this PR for it to pass (`2316` changed in `0`)
```python
    @pytest.mark.parametrize("rst_and_truenodata", [(ref_dem, 0), (tba_dem, 0), (ddem, 2316)])  # type: ignore
    # Note: Following PR #329, no gaps on DEM edges after coregistration
```
One way to be sure: you could run your tests locally with/without the PR (main and this branch) to check if it fails?

And I agree for leaving aside the weird error in `test_docs`  and opening an issue! :smile:

rhugonnet:
Just checked this: with the same environment, all tests passing on the `main` branch. The change for values in `test_spatialstats` are definitely linked to the new data that is replacing NaNs in the `ddem`.
Updating tests...

rhugonnet:
Actually, all the tests that were failing except the one linked to https://github.com/GlacioHack/geoutils/pull/326 were due to the changes in this PR. I fixed them all.
We can merge!

rhugonnet:
Now I'll open a PR from a new branch pulling the latest GeoUtils, to see if releasing a new GeoUtils version will fix the issue and if it will create others.

PR #332, opened 2022-11-17
Fix `flake8` hook address in `pre-commit`

rhugonnet:
All in the title, same done in GeoUtils and KH-9 repo.


adehecq (dismissed):
Yes, I noticed the issue in my repository and fixed it already!

PR #336, opened 2023-01-31
Fix xDEM with a new GeoUtils release

rhugonnet:
An update from GeoUtils is required after #329, this PR prepares that by pulling the latest GeoUtils directly
And fixing some remaining bugs that didn't make it through in the other PR (docs, Windows).

One tricky problem introduced in the tests of #329, the `NamedTemporaryFile` has permissions issues on Windows (see https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file). Details:
> Whether the name can be used to open the file a second time, 
> while the named temporary file is still open, varies across platforms 
> **(it can be so used on Unix; it cannot on Windows NT or later)**.
After some tweaking, I fixed it using the `mode` and `delete` arguments of, trick added to the Wiki: https://github.com/GlacioHack/xdem/wiki/Stuck-on-a-mysterious-error-during-CI%3F

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -75,7 +75,7 @@ repos:
 
         # Sort imports using isort
         - repo: https://github.com/PyCQA/isort
-          rev: 5.10.1
+          rev: 5.12.0
           hooks:
                   - id: isort
                     args: ["--profile", "black"]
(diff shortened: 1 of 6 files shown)


rhugonnet:
Force-merging as it's just a few lines of bug fixes

adehecq:
Thanks ! 
Since you modified the requirements from the latest geoutils to conda geoutils, we need to update geoutils regularly no?
The issue is that most of us develop xdem while using the latest geoutils... So tests that work locally may fail when pushed.

PR #340, opened 2023-03-02
Edit bands into indexes of DEM class

### LINKED ISSUE #338: Small text corrections
Hello,
Here is 3 small text mistakes I spotted:
- in the python doc of the function xdem.dem.DEM.reproject, the line ":param crs: Specifiy [...]" does not correspond to the argument name, which is "dst_crs".
- in the same doc, the "dst_bounds" argument descriptions says coordinates must be in the “source CRS" but it seems more it's the destination one, as the dst_bounds name also suggests it.
- A typing problem: for coregistration, the nuth_kaab.fit() verbose feature prints at one point "shit" instead of "shift".

### LINKED ISSUE #339: 🐛 need change 'bands' into 'indexes' in DEM class
### Bug

The DEM class in xdem/dem.py has following lines:

```
[86] # self.nbands can be None when data is not loaded through the Raster class
[87] if self.nbands is not None and self.nbands > 1:
[88]     raise ValueError("DEM rasters should be composed of one band only")
```

### Error

it raises 
`AttributeError: 'DEM' object has no attribute 'nbands'
`

### Reason

The 'bands' were changed into 'indexes' since geoutils 0.0.10.

liuh886:
Edit a few lines of the DEM class in xdem/dem.py, from

```
[86] # self.nbands can be None when data is not loaded through the Raster class
[87] if self.nbands is not None and self.nbands > 1:
[88]     raise ValueError("DEM rasters should be composed of one band only")
```

to 

```
[86] # self.indexes can be None when data is not loaded through the Raster class
[87] if self.indexes is not None and len(self.indexes) > 1:
[88]     raise ValueError("DEM rasters should be composed of one band only")
```

so that it fixed #339

EDIT from @rhugonnet:
Resolves #338 as well

### DIFF
--- a/tests/test_dem.py
+++ b/tests/test_dem.py
@@ -4,6 +4,7 @@ import warnings
 
 import geoutils.georaster as gr
 import geoutils.satimg as si
+import rasterio as rio
 import numpy as np
 import pyproj
 import pytest
@@ -64,6 +65,10 @@ class TestDEM:
                 np.all(dem3.data.mask == dem4.data.mask),
             )
         )
+
+        # Check that an error is raised when more than one band is provided
+        with pytest.raises(ValueError, match="DEM rasters should be composed of one band only"):
+            xdem.DEM.from_array(data=np.zeros(shape=(2, 5, 5)), transform= rio.transform.from_bounds(0, 0, 1, 1, 5, 5), crs=None, nodata=None)
 
     def test_copy(self) -> None:
         """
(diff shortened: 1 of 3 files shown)


rhugonnet:
Thanks! Added a new test to ensure this doesn't happen again. 
And for the record: actually good to test on `indexes` and not `count` (which was the equivalent of `nbands`), in case a single band is implicitly loaded by GeoUtils on opening (new big PR coming soon).

rhugonnet (approved):

adehecq:
Thanks for the edits !

Issue #347, opened 2023-03-30
Typo in import error message


erikmannerfelt:
I had some, ehrm, problems installing xdem recently and a "geoutils" error popped up. Turns out it was xdem, but geoutils is written in the string!

https://github.com/GlacioHack/xdem/blob/f40e897761704bffe5ad54ff6b43e3db0730a6f1/xdem/__init__.py#L19

sjdex:
Fixed the typo with #348

PR #348, opened 2023-03-30
changed geoutils to xdem in import error message

sjdex:


sjdex:
This closed #347 .

rhugonnet:
Sorry @sjdex, but we don't do single-line PR changes for something like this here. This will be resolved in a dedicated PR by a maintainer.

PR #350, opened 2023-04-04
Rework `DEM.vcrs` to support any 3D transformation (and fix with PROJ update)

### LINKED ISSUE #223: Write documentation and gallery example for `DEM` class and `vref` manipulation

### LINKED ISSUE #262: DEM.to_vref fails if CRS does not have an EPSG code
When running the command `my_dem.to_vref('WGS84')`, it fails with the following error:

> ~/development/GlacioHack/xdem/xdem/dem.py in ccrs(self)
>     146             with warnings.catch_warnings():
>     147                 warnings.filterwarnings("ignore", module="pyproj")
> --> 148                 self._ccrs = pyproj.Proj(init="EPSG:" + str(int(crs.to_epsg())), geoidgrids=self.vref_grid).crs
>     149         else:
>     150             self._ccrs = None
> 
> TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

This is because in that case, `my_dem.crs.to_epsg()` returns None. The CRS of that DEM was set with a proj4 string and never with an EPSG code, so gdal probably does not find it.

You can reproduce the error by reprojecting the xdem sample DEM in a different projection e.g.,:
```
import geoutils as gu
import matplotlib.pyplot as plt
import xdem

reference_dem = xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
test_dem = reference_dem.reproject(dst_crs='+proj=aea +lat_0=30 +lon_0=95 +lat_1=15 +lat_2=65 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs=True')
test_dem.set_vref('EGM08')
test_dem.to_vref('WGS84')
# -> Fails
reference_dem.set_vref('EGM08')
reference_dem.to_vref('WGS84')
# -> works
```

### LINKED ISSUE #311: `PROJ` grids not found on Windows
Windows has an issue with its PROJ environment variable (`PROJ_DATA` or `PROJ_LIB`) in the CI. To fix this, #304 unsets the variables which is then reset by rasterio (see https://rasterio.readthedocs.io/en/latest/faq.html#why-can-t-rasterio-find-proj-db-rasterio-from-pypi-versions-1-2-0).

This works for all OS for importing `rasterio` and using `pyproj`. However, it fails when trying to locate the grids in `test_dem.py`. More needs to be investigated here...

### LINKED ISSUE #331: the datum shift bug of xDEM
Hi, I noticed a bug in the function dem.to_vref().

and have [a quick validation.](https://github.com/liuh886/etc/blob/main/xdem_to_crs.ipynb)

It shows the xDEM does not give the correct datum shift by using pyproj.

Please notice this if anyone uses .to_vref().

### LINKED ISSUE #344: DEM changing of geoid not working
The function _xdem.DEM.set_vref()_ does not seems to perfectly set the ccrs with my data. Or, _.to_vref()_ should not use the ccrs created by _.set_vref()_?

**To reproduce it:**
[GLO90_sub.tif.zip](https://github.com/GlacioHack/xdem/files/11027505/GLO90_sub.tif.zip)
#First, unzip the DEM .tif file in the attached archive, 
`import xdem`
`mnt = xdem.DEM(PathToDEM) # path of the .tif`
`mnt.set_vref('EGM08')`
`mnt.ccrs # prints +geoidgrids is set with EGM08, but ellipsoid is set to WGS84`
`print(mnt.data)`
`mnt.to_vref('WGS84')`
`print(mnt.data) # no elevation change noticed`

In the example, we set the geoid to EGM08 and ask to transform it to WGS84 ellipsoid.
The _mnt.ccrs_ prints ellipsoid WGS84 (not the geoid EGM08), and _to_ref()_ uses the _.ccrs_ attribute and not the _vref_ value (which indeed is EGM08), so no transformation is done.

OS: Ubuntu 22
Environment : xdem (0.0.7) from the repository

### LINKED ISSUE #345: Remove proj-data dependency
`proj-data` is rather big to download and 99% of the files are not used. It looks like there is the possibility to download the grids afterwards and even to download the geoid grids by groups: https://pyproj4.github.io/pyproj/stable/transformation_grids.html#transformation-grids
We could remove proj-data as a dependency and either download the data when needed, or recommend to install proj-datumgrid.

rhugonnet:
This PR reworks the vertical CRS attribute of `DEM` from `.vref` into `.vcrs` and uses more rigorously `pyproj` to define `.vcrs` as a vertical CRS (1D vertical axis) and `ccrs` as a 3D CRS (compound 2D + 1D, or promoted 2D to 3D CRS) required for vertical transformation.

The previous implementation stopped working because, since PROJ6, having 3D CRS as both input and output of a transformation is necessary to transform `z` (otherwise runs but without transforming). See: https://pyproj4.github.io/pyproj/stable/advanced_examples.html#promote-crs-to-3d.

Now, two methods are implemented in `DEM`: `to_vcrs()` and `set_vcrs` (very much like `to_crs()` and `set_crs()` in GeoPandas). The functionalities resides in a separate module `vcrs.py`.
I ran into some difficulties implementing the functionality because the function `pyproj.CRS.to_2d()` did not exist while `pyproj.CRS.to_3d()` does. This is now solved as it existed in PROJ and just needed to be wrapped, see related discussion: https://github.com/pyproj4/pyproj/discussions/1265 and related PR: https://github.com/pyproj4/pyproj/pull/1267.

Also removes one time test that still failed randomly (forgotten in #355).

## Details

This PR:
- Moves all vertical CRS operations in a `vcrs.py` module with his own tests `test_vcrs.py`,
- Adds a `_vcrs_from_user_input()` function to parse any vertical CRS from either common names ("EGM96", "EGM08", "Ellipsoid"), or a PROJ grid path, or an EPSG code or a CRS object,
- Adds a `_build_vcrs_from_grid()` function to consistently build a vertical CRS from a grid. If the grid does not exist locally, it attempts to download it from the PROJ fixed repo: https://cdn.proj.org/. 
- Adds a `_build_ccrs_from_crs_and_vcrs()` function to consistently build a 3D CRS from a 2D/3D CRS stored in `.crs` by `Raster` (dropping the 3D if needed), and a vertical CRS in `.vcrs`. Additionally, if the target is the ellipsoid, the function casts to 3D.
- Adds a `_transform_zz` function that creates a `TransformerGroup` to list all possible transformers and, if the `best_available` is not there, downloads the grids automatically,
- Re-works `DEM.set_vcrs` and `DEM.to_vcrs` using the previous functions of `vcrs.py`,
- Adds an overridden `DEM.from_array()` to set `vcrs` during DEM creation from array,
- Adds a detection of 3D CRS in `DEM.__init__` to set a `vcrs` if the `.crs` is 3D,
- Removes `proj-data` as a dependency,
- Forces `pyproj>=3.4` to ensure that `Transformer.transform()` returns a `np.ma.MaskedArray`,
- Writes the documentation page on "Vertical referencing".

## Issues

Resolves #344 
Resolves #345
Resolves #331
Resolves #262
Resolves #311 
Resolves #223


xdem/dem.py
```diff
@@ -105,151 +128,103 @@ def copy(self, new_array: NDArrayf | None = None) -> DEM:
...
+        Convert the DEM to another vertical coordinate reference system.
 
-        :param vref_name: Vertical reference name
-        :param vref_grid: Vertical reference grid (any grid file in https://github.com/OSGeo/PROJ-data)
+        :param dst_vcrs: Destination vertical CRS. Either as a name ("WGS84", "EGM08", "EGM96"),
+            a EPSG code or pyproj.crs.VerticalCRS, or a path to a PROJ grid file (https://github.com/OSGeo/PROJ-data)
+        :param src_vcrs: Force a source vertical CRS (uses metadata by default). Same formats as for `dst_vcrs`.
 
         :return:
         """
```
erikmannerfelt, line 199: Might be good here to test if `(src_vcrs or self.vcrs) == dst_vcrs` to avoid running this unnecessarily. Imagine if you want to compare two DEMs and want to make sure that they have a vcrs beforehand. Then this might be called even though it may already be set properly.

erikmannerfelt: I guess technically this check would be better further down when `src_vcrs` is a `VerticalCRS` and nothing else.

xdem/vcrs.py
```diff
@@ -0,0 +1,256 @@
...
+    :param yy: Y coordinates.
+    :param zz: Z coordinates.
+
+    :return: Transformed Z coordinates.
+    """
+
+    # Transform the grid
+    transformer = Transformer.from_crs(crs_from=crs_from, crs_to=crs_to, always_xy=True)
+
+    # Will preserve the mask of the masked-array since pyproj 3.4
```
erikmannerfelt, line 253: This is quite essential. Pyproj 3.4 is now a minimum version. Fine by me, but it should be specified in the requirements and environment files.

erikmannerfelt: I tried to find a snarky comment saying wHaT abOUT mASKS! But instead I found this hehe

xdem/dem.py
```diff
@@ -1,76 +1,92 @@
...
+        :param vcrs: Vertical coordinate reference system either as a name ("WGS84", "EGM08", "EGM96"),
+            a EPSG code or pyproj.crs.VerticalCRS, or a path to a PROJ grid file (https://github.com/OSGeo/PROJ-data).
+        :param silent: Whether to display vertical reference parsing.
         """
 
         self.data: NDArrayf
+        self._vcrs: VerticalCRS | None = None
+        self._vcrs_name: str | None = None
+        self._vcrs_grid: str | None = None
+        self._ccrs: CompoundCRS | None = None
```
erikmannerfelt, line 89: I can't find where `self._ccrs` is ever used. Doesn't the property always create a new one?

xdem/vcrs.py
```diff
@@ -0,0 +1,256 @@
...
+    # This requires transforming the horizontal CRS to 2D in case it was 3D
+    # Using CRS() because rasterio.CRS does not allow to call .name otherwise...
+    if isinstance(vcrs, CRS):
+        ccrs = CompoundCRS(
+            name="Horizontal: " + CRS(crs).name + "; Vertical: " + vcrs.name,
+            components=[CRS(crs).to_2d(), vcrs],
+        )
+    # Else if "Ellipsoid" was passed, there is no vertical reference
+    # We still have to return the CRS in 3D
+    else:
```
erikmannerfelt, line 78: This will run if `vcrs` is a sausage too. I'd prefer:
```python
elif vcrs == "Ellipsoid":
   ccrs = CRS(crs).to_3d()
else:
  raise ValueError(f"Invalid vcrs given. Must be a VerticalCRS or the literal string 'Ellipsoid'")

```

erikmannerfelt: I saw the tests now. I guess `str(vcrs).lower() == "ellipsoid"` is required to make the tests pass.

erikmannerfelt (commented):
Wow, what a big change with so few things for me to say! This looks like a great modification of the current behaviour and I only have nitpicks.

xdem/dem.py
rhugonnet: Good point, removed!

xdem/dem.py
rhugonnet: Done, it now also raises a `UserWarning` to tell the user that the operation did nothing. And I added tests for it! :smile:

rhugonnet: In [ad871ba](https://github.com/GlacioHack/xdem/pull/350/commits/ad871baef73bac8e714772625c2f013d5aa3d877)

xdem/vcrs.py
rhugonnet: Done!

xdem/vcrs.py
rhugonnet: Done! :D

rhugonnet:
Thanks a lot for the review @erikmannerfelt! :smile:

I realized there was still a bit more to do to have everything practical and consistent!
What's new:
- Added the automatic downloading of PROJ grids and removed the `proj-data` dependency (fixes #311 and #345 in one go),
- Now, before transformation, `pyproj.transformer.TransformerGroup` is called to get all existing transformations for the area-of-interest sorted by best accuracy. If the best one is not available, it is downloaded automatically. Then it is used,
- I added an overridden `from_array` that accepts `vcrs`,
- I added a loop during `DEM.__init__` that tries to estimate the vertical CRS from `DEM.crs` (in case it is 3D, which won't be the case for old archives, but is now getting more and more common! :smile:).

And, most importantly, wrote the documentation page! Here is it: https://xdem-rhugonnet.readthedocs.io/en/fix_vref/vertical_ref.html

rhugonnet:
Finally fixed the last little bit! All yours to review @erikmannerfelt!

**One big question remains for me:** Should we integrate the `vcrs` into the `crs` (overridding `Raster.crs` in `DEM.crs`)? This way the DEM would be written to file/read with the correct 3D CRS! :smile: 
(only small downside I see, is that people unaware of 3D CRS might not understand why their DEM CRS has changed, but that should become obvious from the doc)

doc/source/vertical_ref.md
```diff
@@ -1,5 +1,233 @@
...
+
+## Methods to manipulate vertical CRSs
+
+The parsing, setting and transformation of vertical CRSs revolves around **three methods**, all described in details further below:
+- The **instantiation** of {class}`~xdem.DEM` that implicitly tries to set the vertical CRS from the metadata (or explicitly through the `vcrs` argument),
+- The **setting** method {func}`~xdem.DEM.set_vcrs` to explicitly set the vertical CRS of a {class}`~xdem.DEM`,
+- The **transformation** method {func}`~xdem.DEM.to_vcrs` to explicitly transform the vertical CRS of a {class}`~xdem.DEM`.
+
+```{caution}
+As yet, **[Rasterio](https://rasterio.readthedocs.io/en/stable/) does not support vertical transformations during CRS reprojection** (even if the CRS provided contains a vertical axis).
```
erikmannerfelt, line 48: Is "As yet" properly spelled? "As of now" would probably read better.

doc/source/vertical_ref.md
```diff
@@ -1,5 +1,233 @@
...
+```{code-cell} ipython3
+:tags: [remove-cell]
+
+# Replace this with a new DEM in xdem-data
+import rasterio as rio
+dem = xdem.DEM.from_array(data=np.ones((2,2)),
+                     transform=rio.transform.from_bounds(0, 0, 1, 1, 2, 2),
+                     crs=pyproj.CRS("EPSG:4326"),
+                     nodata=None)
+dem.save("SETSM_WV03_20151101_104001001327F500_104001001312DE00_seg2_2m_v3.0_dem.tif")
```
erikmannerfelt, line 107: Could you add a comment describing the choice of the string here? People might not be aware of that this is the filename pattern of an ArcticDEM. It's written below, but here it's a bit out of the blue.

doc/source/vertical_ref.md
```diff
@@ -1,5 +1,233 @@
...
+```
+
+```{code-cell} ipython3
+# Open an ArcticDEM strip
+dem = xdem.DEM("SETSM_WV03_20151101_104001001327F500_104001001312DE00_seg2_2m_v3.0_dem.tif")
+# The vertical CRS is set from the filename (no information in the CRS that is 2D)
```
erikmannerfelt, line 113: Technically, it's set since the ArcticDEM vcrs is hardcoded, right? So the filename "only" helps to show that it's an ArcticDEM. In practice it's the same, but rephrasing might make it easier to understand the exact chain of events.

tests/test_spatialstats.py
```diff
@@ -434,9 +433,9 @@ def test_sample_empirical_variogram_speed(self) -> None:
         subsample = 10
 
         # First, run the xDEM wrapper function
-        t0 = time.time()
```
erikmannerfelt, line 437: Weird that this shows up, since it's already fixed in the merged #355. Or am I missing something?

erikmannerfelt: if it's new, perhaps refer to the issue to make it easier to backtrack what changed what in the future

erikmannerfelt (commented):
I only have documentation comments now!

Regarding your suggestion to override crs with vcrs, do you mean to override crs with ccrs, or am I missing something? I'm still a bit new to vcrs so it might be a dumb question!

doc/source/vertical_ref.md
rhugonnet: Done!

doc/source/vertical_ref.md
rhugonnet: Good idea, added!

doc/source/vertical_ref.md
rhugonnet: You're right, rephrased!

tests/test_spatialstats.py
rhugonnet: One that was forgotten, refering back.

doc/source/vertical_ref.md
```diff
@@ -1,5 +1,233 @@
...
+```{code-cell} ipython3
+# Set a vertical CRS extended from the ellipsoid of the DEM's CRS
+dem.set_vcrs("Ellipsoid")
+dem.vcrs
+```
+
+- **Any PROJ grid name available at [https://cdn.proj.org/](https://cdn.proj.org/)**,
+
+```{tip}
+**No need to download the grid!** This is done automatically during the setting operation, if the grid does not already exist locally.
```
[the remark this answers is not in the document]
rhugonnet: I'm not sure where you would place the addition?

rhugonnet:
Thanks for the comments!

> Regarding your suggestion to override crs with vcrs, do you mean to override crs with ccrs, or am I missing something? I'm still a bit new to vcrs so it might be a dumb question!

Yes, exactly, overriding it with `ccrs` so that it's always 3D! (then I guess we could remove `ccrs` and have it just be `crs`).

@adehecq: Curious about your feedback on this PR as well! 
FYI: It's a problem also of interest to the ICESat-2 SlideRule team that wants to integrate vertical transforms in their cloud processing, so we're discussing it a bit there too in parallel.

rhugonnet:
Merging as @erikmannerfelt is on fieldwork so I assume the review is finished! :smile: 
Opening an issue for overridding `crs`.

Issue #353, opened 2023-04-18
Test suite will soon stop working (but there's an easy fix)


erikmannerfelt:
As of 31 May 2023, the `echo "::set-output name={name}::{value}"` syntax of GitHub Actions will cease functioning. Below is a link to the announcement. I saw it in a warning and decided to see if there's a fix, which there is.
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

 I've seen the highly impressive work by @rhugonnet to figure out mamba caching (great work! I tried and failed many moons ago), and I have since shamelessly stolen most of that code for [variete](https://github.com/erikmannerfelt/variete) 😈. Unfortunately, the caching function uses that syntax, so it will stop working soon.

# But fear not

There's an easy fix for it using new GitHub syntax. I'm just a good lad making an issue for it before submitting a PR to fix it!

PR #354, opened 2023-04-18
Fix GitHub Actions mamba caching

erikmannerfelt:
See #353 for the rationale.

For once, it seems the fix was rather simple! But maybe the test suite fails for other reasons now.. Oh well, you can see that the exact syntax works in [variete](https://github.com/erikmannerfelt/variete/blob/main/.github/workflows/python-test.yml)

Closing #353


rhugonnet (approved):
Amazing that you picked that up in time! :star_struck:

You can open the exact same PR in GeoUtils ;)

PR #355, opened 2023-04-19
Remove `spatialstats` test based on `time` that randomly fails

rhugonnet:
Opening an issue, there's probably a better way of testing this.


erikmannerfelt (approved):
I guess the GitHub CI can get congested with other stuff and randomly introduce delays? I'd say it's good enough with just the approx equals test!

Issue #371, opened 2023-05-03
Nodata value not preserved for terrain attributes when using richdem


adehecq:
The terrain attributes calculated using richdem in xdem.terrain do not properly take into account no data.
Here is a MWE (requires to have richdem installed):
```
import xdem
dem = xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))

# Add nodata
mask = np.zeros_like(dem.data.data)
mask[:100, :100] = 1
dem.set_mask(mask)

# Calculate slope with xdem implemetation
slope = xdem.terrain.slope(dem)
print(slope.data[0, 0])
# -> masked value

# Calculate slope with richdem implementation
slope = xdem.terrain.slope(dem, use_richdem=True)
print(slope.data[0, 0])
# -> -572900.44

# Same issue with max curvature for example, but a different value replaces nodata
maxc = xdem.terrain.maximum_curvature(dem, use_richdem=True)
print(maxc.data[0, 0])
# -> -9999.0 
```

adehecq:
The fix is probably as simple as adding `terrattr[terrattr == terrattr.no_data] = np.nan` after [this line](https://github.com/GlacioHack/xdem/blob/47c55e4231005dfbd6acaa45dca11df4135ccdf8/xdem/terrain.py#L44).
But this requires updating the tests to make sure it works for all cases.

adehecq:
Solved by #372

PR #372, opened 2023-05-03
Fix nodata not preserved when calculating terrain attributes with richdem

adehecq:
A one-line fix to solve #371.
A test has been updated to flag this issue in the future.


rhugonnet (approved):
Perfect!

PR #389, opened 2023-08-01
Extrapolate using nearest neighbour in `interp_nd_binning`

### LINKED ISSUE #380: Negative values predicted by `interp_nd_binning` problematic for error analysis
Opening after a discussion started by Enrico Mattea:

The linear extrapolation of `interp_nd_binning` can result in negative values (and, more generally, yield curious values) for points outside of the binning maximum bounds, especially as the "edge bins" tend to have less points to sample from, and thus more variability:
https://github.com/GlacioHack/xdem/blob/d91bf1cc9b3f36d77f3729649bc8e9edc6b42f9f/xdem/spatialstats.py#L342

It does happen that slope/curvature have very high values because of outliers, and those can be ignored during error prediction (`min_vals` argument of `interp_nd_binning`). 
If those outliers remain unfiltered in the dDEM, some of the predicted errors could then be negative? (untested)

So: should we rather default to nearest neighbour outside of the bin bounds, and leave the linear extrapolation option to the user?

rhugonnet:
Resolves #380 

The extrapolation is always done using "nearest", and interpolation is "linear" by default.
The workflow is a bit complex in order to deal with nodata in the grid, see comments in the code.

Added tests to ensure all works as intended.

### DIFF
--- a/tests/test_biascorr.py
+++ b/tests/test_biascorr.py
@@ -216,7 +216,7 @@ class TestBiasCorr:
         elev_fit_params.update({"bias_vars": bias_vars_dict})
 
         # Run with input parameter, and using only 100 subsamples for speed
-        bcorr.fit(**elev_fit_params, subsample=1000, random_state=42)
+        bcorr.fit(**elev_fit_params, subsample=10000, random_state=42)
 
         # Apply the correction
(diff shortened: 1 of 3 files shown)


adehecq:
Great, thanks ! 🎉

PR #395, opened 2023-08-10
Silence numba warning for `@jit`

### LINKED ISSUE #393: Numba deprecation warning
With an update of numba (0.57.1 in my case), I now get a deprecation warning:

> NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit"

I guess the best is to set "nopython=True" and hope it works?

rhugonnet:
Resolves #393 

Should already be fixed in `scikit-gstat`: https://github.com/mmaelicke/scikit-gstat/issues/146
(`@njit` is equivalent to `@jit(nopython=True)`)

Opened #396

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -17,7 +17,6 @@ dependencies:
   - scipy
   - tqdm
   - scikit-image
-  - scikit-gstat>=1.0.11
   - pytransform3d
   - geoutils==0.0.11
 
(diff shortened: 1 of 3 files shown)


adehecq:
Cool, thanks !

adehecq:
Actually, on conda-forge, only version 1.0.10 exists... So the warnings still show up with the latest conda install.

rhugonnet:
Yes, there's no way around it unfortunately because pip doesn't work in conda-forge recipes. I specified this there: https://github.com/GlacioHack/xdem/pull/398

PR #398, opened 2023-08-12
Only use `conda-forge` channel in environment

### LINKED ISSUE #396: Revert `scikit-gstat` to `conda-forge` channel once `1.0.11` is released
Only available on pip right now, to solve #395

rhugonnet:
Forgot `pip` cannot be called from `conda build`, so versions would be inconsistent between PyPI and conda.

Reverting back `scikit-gstat` to `conda-forge`, the warnings (#395) will go away by themselves when its `1.0.11` release is made on `conda-forge`.

Additionally, still trying to figure out the circular import issue in `xdem-feedstock`...

Resolves #396

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -17,6 +17,7 @@ dependencies:
   - scipy
   - tqdm
   - scikit-image
+  - scikit-gstat>=1.0
   - pytransform3d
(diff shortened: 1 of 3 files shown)


adehecq:
Good to know about xdem.coreg !

Issue #406, opened 2023-08-16
Make `CoregPipeline` work with point data


erikmannerfelt:
There is currently no  `_fit_pts_func` in `CoregPipeline`, meaning it fails on point data. This should be an easy addition, however, since it just calls the underlying `Coreg` objects and requires no functionality in itself. 20 minute job, in and out.

Issue #403, opened 2023-08-15
Argument descriptions are missing in `Coreg.fit_pts` docstring.


erikmannerfelt:
The `samples` and `mask_high_curv` arguments are not described in the docstring. I'm honestly struggling a bit to understand the difference between `samples` and `subsample`. @rhugonnet, do you know?

https://github.com/GlacioHack/xdem/blob/02902c095ffb2bfbf422e90c13d4fce227da6583/xdem/coreg/base.py#L831

Issue #402, opened 2023-08-15
Revamp and simplify the raster-point/P-R/P-P functionality


erikmannerfelt:
ICP is inherently a point-to-point registration method. Therefore, by logical extension, it should have a `fit_pts_func`! Currently it does not, meaning point-to-raster co-registration does not work.

EDIT: The original title and topic of this issue was "ICP doesn't have a fit_pts_func", but it evolved into a more interesting conversation.

rhugonnet:
I'm wondering: is there any coregistration method that relies on the 2D array shape?

If not, maybe the `Coreg.fit_pts()` function should NOT be subclassed by a `_fit_pts_func`, but simply pass two vectors to the `_fit_pts` method that every method has? This would simplify things a lot. Essentially, `fit_pts()` would only be a convenience wrapper that handles the different input shape of the data.

erikmannerfelt:
Hmm, could you please elaborate on that, @rhugonnet? I'm all for simplification, but I'm not sure I understand what you mean.

- ICP works by converting rasters to point clouds and then running the main functionality.
- NuthKaab currently has TWO almost identical functions for `fit` and `fit_pts`. Both work on the same principle but look slightly different.
- GradientDescending does not have a Raster-Raster fit function at all at the moment.

So to answer your first question (I'll call a 2D array a raster); some methods require at least one raster to solve the otherwise existing topology problem. Any 2D (or 2.5D if we want to call it that) approach requires that at least one surface is given, or else there is no easy way to compare points. If we want to make all of them fully 3D (for point-point to work), we need to implement a fast and accurate topology approximation. It could be as simple as just doing delunay triangulation of the points, but extreme outliers may be a nuisance. The best is some kind of robust averaging of sorts (but then perhaps as a preprocessing filter). My point is that for all to be fully 3D, complexity can increase. 

- ICP does not require any raster or topology information; it works entirely on point data
- NuthKaab requires one raster in its current implementation as it's reliant on the topology
- GradientDescending has identical requirements to NuthKaab, but would be easier to convert to point-point as it only works on dh (not aspect+slope+dh).

One simple (and safe) workaround is to add some default fallback behaviour unless overridden, as suggested below. Honestly, it feels like we should have four different functions, not two which we have at the moment, since there are four potential cases of co-registration:
### Raster-raster: 
1. Use a raster-raster function if available
2. Use a raster-point function if available (by converting one raster to points)
3. Convert to points and use point-point

### Raster-point
[...]

Issue #401, opened 2023-08-15
`Coreg.fit_pts` does not respect the `z_name` kwarg.


erikmannerfelt:
The `z_name` kwarg exists for specifying what the z column is called in a reference "DEM" dataframe. However, at L898 (see below), it's hardcoded to "z", so it fails if "z" is not available. 

https://github.com/GlacioHack/xdem/blob/02902c095ffb2bfbf422e90c13d4fce227da6583/xdem/coreg/base.py#L898

A simple workaround for now is of course just to rename the column to "z", but this could have unintended consequences in the future if not fixed as the kwarg doesn't work like advertised.

The fix is most likely just to replace the "z" with `z_name`. Also, a test should exist for this case.

PR #407, opened 2023-08-16
Multiple coreg `fit_pts` fixes.

erikmannerfelt:
My first PR in quite some time!

Two things have been done in this PR; re-implementing my PR (https://github.com/liuh886/xdem_snow/pull/7) to the fork that was #346, and solving some small issues that I've found along the way.

## The most important change: smashing the half-pixel shift inconsistency
The first part is arguably the most important because the fit_pts functions of NuthKaab and GradientDescending have been at risk of being half a pixel off since #346 was merged! The tests introduced in the PR did not actually check for the fit_pts result on points, but only on rasters converted to points in a predefined manner. My PR (https://github.com/liuh886/xdem_snow/pull/7) that fixed this was unfortunately lost along the way in #346...  Anyway, here it is again, which fixes and tests the fit_pts half-pixel problem. The reason for the issue was that sampling was not done consistently in the centers of the pixels. The solution is quite hacky I feel, because I'm shifting coordinates and then sampling corners, but it works, as proven by the tests. Paging @liuh886 and @mmazzolini just to let you know that this problem exists in the current release of xdem. EDIT: It might be that the functionality worked before, but it wasn't tested properly, and could be half a pixel off with the wrong GDAL `AREA_OR_POINT` settings. This PR makes it so this won't happen (at least according to the tests).

## Improvements
- The fit_pts functionality can now use a geometry column if available instead of E/N columns ([d81ceb5](https://github.com/GlacioHack/xdem/pull/407/commits/d81ceb536c92a0c984806c5a5327a6914426c81c))
- Added ICP fit_pts functionality (#402) (fixed in [9f28eeb](https://github.com/GlacioHack/xdem/pull/407/commits/9f28eeb676697623e18374994c99b485e75b6ca2)).

## Other fixes

- Solves #406. Pipelines now work with point data (I added a small rudimentary test for it) (fixed in [fd4fde3](https://github.com/GlacioHack/xdem/pull/407/commits/fd4fde3d9b3dc7908e9d8aaa04a3808aa8196786))
- Solves #401 (fixed in [d81ceb5](https://github.com/GlacioHack/xdem/pull/407/commits/d81ceb536c92a0c984806c5a5327a6914426c81c))
- Solves #403 (fixed in [4305eae](https://github.com/GlacioHack/xdem/pull/407/commits/4305eaed4533143143f0f344e2134d6a726d2dfc))

### DIFF
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -368,6 +368,25 @@ class TestCoregClass:
         # Assert that the combined vertical shift is 2
         assert pipeline2.to_matrix()[2, 3] == 2.0
 
+    def test_pipeline_pts(self) -> None:
+        warnings.simplefilter("ignore")
+
+        pipeline = coreg.NuthKaab() + coreg.GradientDescending()
+        ref_points = self.ref.to_points(as_array=False, subset=5000, pixel_offset="center").ds
+        ref_points["E"] = ref_points.geometry.x
+        ref_points["N"] = ref_points.geometry.y
+        ref_points.rename(columns={"b1": "z"}, inplace=True)
+
+        # Check that this runs without error
+        pipeline.fit_pts(reference_dem=ref_points, dem_to_be_aligned=self.tba)
+
+        for part in pipeline.pipeline:
+            assert np.abs(part._meta["offset_east_px"]) > 0
+
+        assert pipeline.pipeline[0]._meta["offset_east_px"] != pipeline.pipeline[1]._meta["offset_east_px"]
+
+
+
     def test_coreg_add(self) -> None:
         warnings.simplefilter("error")
         # Test with a vertical shift of 4
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -272,6 +272,9 @@ class TestCoregClass:
 
         if coreg_class.__name__ == "ICP":
             matrix = coreg_obj.to_matrix()
+            # The ICP fit only creates a matrix and doesn't normally show the alignment in pixels
+            # Since the test is formed to validate pixel shifts, these calls extract the approximate pixel shift
+            # from the matrix (it's not perfect since rotation/scale can change it).
             coreg_obj._meta["offset_east_px"] = -matrix[0][3] / res
             coreg_obj._meta["offset_north_px"] = -matrix[1][3] / res
 
--- a/tests/test_coreg.py
+++ b/tests/test_coreg.py
@@ -230,7 +230,10 @@ class TestCoregClass:
 
         # Run co-registration
         gds = xdem.coreg.GradientDescending(downsampling=downsampling)
-        gds.fit_pts(self.ref, self.tba, inlier_mask=inlier_mask, verbose=verbose, samples=samples)
+        ref_pts = self.ref.to_points().ds.rename(columns={"b1": "z"})
+        ref_pts["E"] = ref_pts.geometry.x
+        ref_pts["N"] = ref_pts.geometry.y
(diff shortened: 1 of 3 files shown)


liuh886:
brilliant! Thanks, Eric,  I am on the way back to Oslo, and I am going to have a test on the half-pixel shift bug next week!

erikmannerfelt:
Oh and @rhugonnet, I acknowledge that the ICP improvement efforts could be seen as misplaced to do now if we're to revamp everything either way. But this way, even the current functionality of xdem has more features. And more importantly for me, I can do my work like I intended now that ICP and CoregPipeline works with ICESat-2!

tests/test_coreg.py
```diff
@@ -218,64 +218,75 @@ def test_coreg_example(self, verbose: bool = False) -> None:
...
+        best_east_diff = 1e5
+        best_north_diff = 1e5
+        if points_or_raster == "raster":
+            coreg_obj.fit(shifted_ref, self.ref, verbose=verbose)
+        elif points_or_raster == "points":
+            coreg_obj.fit_pts(shifted_ref_points, self.ref, verbose=verbose)
+
+        if coreg_class.__name__ == "ICP":
+            matrix = coreg_obj.to_matrix()
+            coreg_obj._meta["offset_east_px"] = -matrix[0][3] / res
```
rhugonnet, line 275: I'm not sure I understand why you overwrite the `_meta` here for ICP? A comment to explain the logic would be helpful

xdem/coreg/affine.py
```diff
@@ -31,6 +31,7 @@
     _transform_to_bounds_and_res,
     deramping,
 )
+from xdem.dem import DEM
```
rhugonnet, line 34: I was thinking that, eventually, we could wrap the `coreg` class in a `DEM.coregister()` method. Using it here would trigger a circular import issue.

It looks like we can replace `DEM` with `Raster` below. But, eventually, it would also be nice to be able to verify that 3D CRS used during a coregistration are the same, and for this we'd need to use `DEM`... A problem for later, I guess! :sweat_smile:

rhugonnet (dismissed):
Looking good, thanks for all the fixes and the tests!

I agree, we can leave the big refactoring to a separate PR, in particular:
- Use a single `fit()` method,
- Define "raster-raster", "point-raster" and "point-point" logic,
- Consistent point preprocessing/postprocessing.

Very weird to suddenly have a segmentation fault on `dem_coregistration`, but nowhere else (especially when it is just a wrapper of other functions)... :thinking:

xdem/coreg/affine.py
erikmannerfelt: It was lazy to make them DEMs anyway because they don't contain elevation data. Good catch! Fixed in [d471412](https://github.com/GlacioHack/xdem/pull/407/commits/d471412ded75c80548519a840f97ed80eb94e7ca)

tests/test_coreg.py
erikmannerfelt: Added comments in [df66946](https://github.com/GlacioHack/xdem/pull/407/commits/df66946a6f614d75e725c7a8bb7430bede0a6142). 

The metadata is not overridden as `offset_east_px` and  `offset_north_px`  don't usually exist.

erikmannerfelt:
Thanks for the inputs, @rhugonnet !

adehecq:
> @adehecq, I had to disable your test (`tests/test_coreg.py::test_dem_coregistration`) because it segfaults both on my machine and in CI! Do you have any idea of why?

Weird... Did you add an issue for this? I can have a look at it later, once we work on the big coreg changes.

PR #412, opened 2023-08-21
Update README.md to point to xdem instead of geoutils

erikmannerfelt:
In the end of the README, it states to push a feature branch to `geoutils:main` which is obviously wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Citing the related study:
 ## Start contributing
 
 1. Fork the repository, make a feature branch and push changes.
-2. When ready, submit a pull request from the feature branch of your fork to `GlacioHack/geoutils:main`.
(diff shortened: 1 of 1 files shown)


rhugonnet:
Strange test failure, I restarted all CI runs!

For the "Contributing" file + README summary, I shortened it earlier this year to avoid repetition with other files (License, Install, etc) and keep it simple. But I now find it a bit "formal" (not too inviting)... What do you think?

I was thinking we could take some ideas from `pyproj`, which has a very inviting page (a bit too long to read to my taste, maybe we could mostly work on our intro, then keep our "Short" summary for experienced developers and add a "Long" one for new developers?): https://github.com/pyproj4/pyproj/blob/main/CONTRIBUTING.md
I also really like the effort to recognize contributors on the front page: https://github.com/pyproj4/pyproj#contributors-

rhugonnet (approved):

PR #432, opened 2023-09-06
Update contributing guidelines to be more accessible

### LINKED ISSUE #414: Make "Contributing" section more inviting
Add a "long" description to `CONTRIBUTING`, and list contributors in `README`? See original comment below.

For the "Contributing" file + README summary, I shortened it earlier this year to avoid repetition with other files (License, Install, etc) and keep it simple. But I now find it a bit "formal" (not too inviting)... What do you think?

I was thinking we could take some ideas from `pyproj`, which has a very inviting page (a bit too long to read to my taste, maybe we could mostly work on our intro, then keep our "Short" summary for experienced developers and add a "Long" one for new developers?): https://github.com/pyproj4/pyproj/blob/main/CONTRIBUTING.md
I also really like the effort to recognize contributors on the front page: https://github.com/pyproj4/pyproj#contributors-

_Originally posted by @rhugonnet in https://github.com/GlacioHack/xdem/issues/412#issuecomment-1686967693_

rhugonnet:
Resolves #414


CONTRIBUTING.md
```diff
@@ -30,24 +41,37 @@ mamba activate xdem-dev  # Or any other name specified above
...
 pytest
 ```
 
+Running `pytest` will trigger a script that automatically downloads test data from [https://github.com/GlacioHack/xdem-data](https://github.com/GlacioHack/xdem-data) before running all tests.
+
+### Documentation
+
+If your changes need to be reflected in the documentation, update the related pages located in `doc/source/`. The documentation is written in MyST markdown syntax, similar to GitHub's default Markdown (see [MyST-NB](https://myst-nb.readthedocs.io/en/latest/authoring/text-notebooks.html) for details).
+
+To ensure that the documentation is building properly after your changes, you can run `pytest tests/test_doc.py` on Linux. On Windows and Mac, the documentation is not maintained, so you can wait to open the PR for it to be checked by the CI.
```
adehecq, line 57: I believe there is a command to build the documentation locally, no? It would be great to add it here. Everytime I make changes to the documentation, I forget about it...

CONTRIBUTING.md
```diff
@@ -30,24 +41,37 @@ mamba activate xdem-dev  # Or any other name specified above
...
-You can then run pre-commit manually:
 ```bash
-pre-commit run --all-files
+pre-commit run --all
 ```
 
+You can then commit and push those changes.
 Optionally, `pre-commit` can be installed as a git hook to ensure checks have to pass before committing.
 
+### Final steps
```
adehecq, line 71: Maybe add a word that the tests have to succeed and the PR be approved for it to be merged.

adehecq (dismissed):
Thanks for updating this outdated file.
I anticipate that some of it will be outdated in some time, for example the Python version used etc. Is there a good way to remind us to update this file when we make such changes?
Maybe we should have a list of things to do when marking a new (major) release? Including updating this file.

CONTRIBUTING.md
rhugonnet: Added!

rhugonnet: (but FYI: `tests/test_doc` also builds the documentation, so does exactly the same thing more easily)

CONTRIBUTING.md
rhugonnet: Done!

rhugonnet:
Good point, I think it's more related to specific dev-changes than necessarily version release, so I added a check-box to the PR template for now (easy to delete the line when not relevant). We can see how that works! :wink:

Issue #435, opened 2023-09-06
Way forward for co-registration module


rhugonnet:
We currently have **25 issues** open on co-registration :scream_cat:, and a lot of notes and plans for the way forward dispersed everywhere. I initially tried to organize this in a project: https://github.com/orgs/GlacioHack/projects/3/views/2, but I think that the level of detail of the issues is too inconsistent for this, so I'm writing this summary instead!

**Architecture:**
We already did a lot in https://github.com/GlacioHack/xdem/pull/158 and https://github.com/GlacioHack/xdem/pull/329 to re-organize the coding structure to be more consistent and robust, but some points are left:
- A detailed list in #327 especially about cleaning + having consistent subfunctions of class methods + side-features (see next point),
- Additional cleaning mentioned in #266,
- The discussion of https://github.com/GlacioHack/xdem/issues/402 about merging point methods into a single `fit` and `apply`, and call methods depending on Point-Point, Point-Raster or Raster-Raster support, which would also completely solve #134.

**Features:**
We've been needing for a while to have consistent:
- Plotting: #139 (also in #327),
- Statistics (in #327),
- Weight support: #59 ,
- Subsampling: #243 and mainly the long discussion in #428.

Most of them are not too far since we introduced a consistent structure for optimizers or binning in https://github.com/GlacioHack/xdem/pull/158. I makes weight support almost directly supported (but we'll need to raise proper warnings for methods that currently ignore, such as ICP), and will make plotting more easy by consistently treating the type of methods (binning, fit, or both) and the dimensionality of the variables (1D, 2D, ND), which can be re-used for `Tilt` or `NuthKaab` fits in the 
affine functions.

**Tests:**
Some tests are still a bit old or slow, several related issues could be solved all at once:
- Consistent data fixtures in test modules: #427,
- More varied test data #64, #140 and https://github.com/GlacioHack/geoutils/issues/310, with synthetic data examples from GeoUtils here: https://github.com/GlacioHack/xdem/pull/358#issuecomment-1522200021.

**Performance:**
:warning: **We really need to think ahead for a structure that will allow memory-efficient computations using Dask:** 
- For reading/writing with georeferencing, this will depend directly on `rioxarray`, and https://github.com/GlacioHack/geoutils/issues/383 and #392, so not too much to think about! 
[...]

Issue #293, opened 2022-08-26
Write specific functions to collapse/rebuild input/outputs as `Raster`, `np.ma.masked_array` or `np.ndarray` consistently


rhugonnet:
We have tons of "wrapper" functions that can take either `Raster`, `np.ma.masked_array` or `np.ndarray` as input, but then only work on the `np.ndarray` (and possibly some metadata), and for output rebuild a `Raster`, `np.ma.masked_array` or `np.ndarray` to match the input.
Right now this is done manually everywhere, we need to have consistent input/output functions for those conversions. Might be triggering issues in #291 due to https://github.com/GlacioHack/GeoUtils/pull/265.

erikmannerfelt:
That's a good idea. A small problem right now (or an unintended feature) is that all metadata are lost after raster recreation. Only the CRS and transform are identical to the input.

rhugonnet:
Yes, the 10 lines in `Coreg.apply()` for "reconstructing" for example: https://github.com/GlacioHack/xdem/blob/main/xdem/coreg/base.py#L1160

rhugonnet:
We could do the same to construct masks from a `Vector`, a `Mask`, an array, etc... It is used everywhere!
Something like this: https://github.com/GlacioHack/xdem/blob/main/xdem/spatialstats.py#L593

PR #436, opened 2023-09-06
Improve `subsample` across `Coreg` subclasses and pipelines

### LINKED ISSUE #243: Add `random_state` argument to `Coreg` class for `subsample` cases
To be able to control the random effects of subsampling in tests or examples of `Coreg` class, which will become increasingly important with higher resolution data!

### LINKED ISSUE #428: Argument `subsample` does not work for a `CoregPipeline`
Could also solve #137 at the same time...

rhugonnet:
This PR makes the `subsample` argument consistent across all `Coreg` subclasses, and allows them to function in pipelines independently for each step. It also reworks `Coreg.fit()` to be more durable for future chunking routines.

All details in the full discussion here: https://github.com/GlacioHack/xdem/issues/428
**Only difference with the discussion:** Now passing directly `inlier_mask` to each subclass' `_fit_func`, so that the subsampling mask can be computed on the final mask of valid pixels within the subclass (valid data changes in the subclass because of NaNs in auxiliary data not yet available to `Coreg.fit()`: slope, curvature, etc...). This way we will always have the exact number of valid points asked when subsampling :smile:! 
To perform the subsampling consistently on the final valid mask, each `Coreg` subclass calls the method `get_subsample_on_valid_mask`.

Tests significantly improved.

As now each class has a default `subsample` argument during instantiation, methods that currently rely on optimization without binning (`NuthKaab`, `Deramping`)* have `5e5` to avoid very long computing time, otherwise `1` for all others.

Also a small fix to `geoutils.raster.subsampling` was necessary: https://github.com/GlacioHack/geoutils/pull/402.

Resolves #428 
Resolves #243


dev-environment.yml
```diff
@@ -52,4 +52,4 @@ dependencies:
     - noisyopt
 
     # To run CI against latest GeoUtils
-#     - git+https://github.com/GlacioHack/GeoUtils.git
+    - git+https://github.com/GlacioHack/geoutils.git
```
erikmannerfelt, line 55: Wasn't there a reason we left pulling from github? I think if we make a breaking change on geoutils, xdem will fail on new PRs without it being the "PR's fault". Any way around this? Otherwise, please make an issue for it so we can revert this asap!

tests/test_coreg/test_affine.py
```diff
@@ -139,15 +139,15 @@ def test_coreg_example(self, verbose: bool = False) -> None:
...
         # Run co-registration
         nuth_kaab = xdem.coreg.NuthKaab()
-        nuth_kaab.fit(self.ref, self.tba, inlier_mask=self.inlier_mask, verbose=verbose)
+        nuth_kaab.fit(self.ref, self.tba, inlier_mask=self.inlier_mask, verbose=verbose, random_state=42)
 
         # Check the output metadata is always the same
-        assert nuth_kaab._meta["offset_east_px"] == pytest.approx(-0.46255704521968716)
-        assert nuth_kaab._meta["offset_north_px"] == pytest.approx(-0.13618536563846081)
-        assert nuth_kaab._meta["vshift"] == pytest.approx(-1.9815309753424906)
+        assert nuth_kaab._meta["offset_east_px"] == pytest.approx(-0.45061858808956284)
```
erikmannerfelt, line 145: Did this change because it now works on different samples? If so, then this comment can be disregarded. If not, I'd like to know why it just changed!

tests/test_coreg/test_affine.py
```diff
@@ -139,15 +139,15 @@ def test_coreg_example(self, verbose: bool = False) -> None:
...
+        assert nuth_kaab._meta["offset_north_px"] == pytest.approx(-0.14225262689582596)
+        assert nuth_kaab._meta["vshift"] == pytest.approx(-1.987523471566405)
 
     def test_gradientdescending(
-        self, downsampling: int = 10000, samples: int = 5000, inlier_mask: bool = True, verbose: bool = False
+        self, subsample: int = 10000, samples: int = 5000, inlier_mask: bool = True, verbose: bool = False
```
erikmannerfelt, line 150: Could we remove the `samples` kwarg now or should we do it later? I still find it confusing!

tests/test_coreg/test_affine.py
```diff
@@ -286,7 +286,7 @@ def test_deramping(self) -> None:
         assert np.abs(np.mean(periglacial_offset)) < np.abs(np.mean(pre_offset))
 
         # Check that the mean periglacial offset is low
-        assert np.abs(np.mean(periglacial_offset)) < 1
+        assert np.abs(np.mean(periglacial_offset)) < 2
```
erikmannerfelt, line 289: Why did this change?

tests/test_coreg/test_base.py
```diff
@@ -106,66 +107,143 @@ def test_ij_xy(self, i: int = 10, j: int = 20) -> None:
...
+            # Calculate the difference in the full vs. subsampled matrices
+            matrix_diff = np.abs(coreg_full.to_matrix() - coreg_sub.to_matrix())
+            # Check that the x/y/z differences do not exceed 30cm
+            assert np.count_nonzero(matrix_diff > 0.5) == 0
+
+        elif coreg_name == "Tilt":
+            # Check that the estimated biases are similar
+            assert coreg_sub._meta["coefficients"] == pytest.approx(coreg_full._meta["coefficients"], rel=1e-1)
+
+    def test_subsample__pipeline(self) -> None:
```
erikmannerfelt, line 189: The double underscore is a typo, right?

tests/test_coreg/test_base.py
```diff
@@ -106,66 +107,143 @@ def test_ij_xy(self, i: int = 10, j: int = 20) -> None:
...
+        assert pipe.pipeline[0]._meta["subsample"] == 200
+        assert pipe.pipeline[1]._meta["subsample"] == 5000
+
+        # Check definition during fit
+        pipe = coreg.VerticalShift() + coreg.Deramp()
+        pipe.fit(**self.fit_params, subsample=1000)
+        assert pipe.pipeline[0]._meta["subsample"] == 1000
+        assert pipe.pipeline[1]._meta["subsample"] == 1000
+
+    def test_subsample__errors(self) -> None:
```
erikmannerfelt, line 205: Maybe unintentional double underscore

tests/test_coreg/test_base.py
```diff
@@ -106,66 +107,143 @@ def test_ij_xy(self, i: int = 10, j: int = 20) -> None:
...
+        """Check proper errors are raised when using the subsample argument"""
+
+        # A warning should be raised when overriding with fit if non-default parameter was passed during instantiation
+        vshift = coreg.VerticalShift(subsample=100)
+
+        with pytest.warns(
+            UserWarning,
+            match=re.escape(
```
erikmannerfelt, line 213: What does escape do? Nothing wrong, I'm just interested!

tests/test_coreg/test_biascorr.py
```diff
@@ -144,7 +144,7 @@ def test_biascorr__errors(self) -> None:
             scipy.optimize.curve_fit,
         ],
     )  # type: ignore
-    def test_biascorr__fit_1d(self, fit_func, fit_optimizer) -> None:
+    def test_biascorr__fit_1d(self, fit_func, fit_optimizer, capsys) -> None:
```
erikmannerfelt, line 147: Double underscore

requirements.txt
```diff
@@ -13,7 +13,7 @@ scipy
 tqdm
 scikit-image
 scikit-gstat>=1.0
-geoutils==0.0.13
 pip
+geoutils
```
erikmannerfelt, line 17: Why is the version removed? This means installs might get weird since there's a minimum version requirement of 0.0.13 (?).

tests/test_coreg/test_base.py
```diff
@@ -106,66 +107,143 @@ def test_ij_xy(self, i: int = 10, j: int = 20) -> None:
...
+        else:
+            # Otherwise the number of pixels
+            subsample_val = subsample
+        assert np.count_nonzero(subsample_mask) == min(subsample_val, np.count_nonzero(valid_mask))
 
-        # Measure the start and stop time to get the duration
-        # start_time = time.time()
-        nuthkaab_full.fit(**self.fit_params)
-        # icp_full_duration = time.time() - start_time
+    # TODO: Activate NuthKaab once subsampling there is made consistent
```
erikmannerfelt, line 137: Can you add an issue for this TODO when this PR is merged?

xdem/coreg/base.py
```diff
@@ -725,16 +722,56 @@ def is_affine(self) -> bool:
...
+            )
+
+        return subsample_mask
+
     def fit(
         self: CoregType,
         reference_dem: NDArrayf | MArrayf | RasterType,
         dem_to_be_aligned: NDArrayf | MArrayf | RasterType,
-        inlier_mask: NDArrayf | Mask | None = None,
+        inlier_mask: NDArrayb | Mask | None = None,
```
erikmannerfelt, line 769: Does this have a check if the user passes a uint8 mask? It could just be converted or error out.

erikmannerfelt (commented):
I love this and I love you

Just minor comments!

dev-environment.yml
rhugonnet: Yes, this should only be used to test that CI passes temporarily on the main branch for devs, we don't want to make a release with it! Should we add a reminder in `HOW_TO_RELEASE`?

Now that GeoUtils 0.0.15 is published, I can revert it back!

tests/test_coreg/test_affine.py
rhugonnet: Yes, different samples!

tests/test_coreg/test_affine.py
rhugonnet: Ah yes, done! I didn't touch `fit_pts` too much in the end, because we will need to change when we merge it in `fit`, and that shouldl smooth out the current differences (for subsampling but also other things!)

tests/test_coreg/test_affine.py
rhugonnet: The subsample changed and the result is ~1.2 now. Let me check before...

rhugonnet: Before the test would have passed even with `< 0.01` :scream_cat:  ! OK, something is wrong with this now, I'll try to figure out why.

tests/test_coreg/test_base.py
rhugonnet: No haha it's for subdividing tests more clearly! Many packages do it, and I liked it a lot. I thought it could improve the clarity/organization of our tests that were getting a bit messy. 
I think we talked about this in the biascorr PR a bit already! Maybe we should establish guidelines for our tests?

tests/test_coreg/test_base.py
rhugonnet: Same as above

tests/test_coreg/test_base.py
rhugonnet: It's for recognizing matching regex expressions that have special characters, for example a parenthesis, an interrogation point, etc. Otherwise the matching fails (and `pytest` actually recommends you to use `re.escape()`).
I think tt's the equivalent of applying an escape "`\`" to all characters in the string.

tests/test_coreg/test_biascorr.py
rhugonnet: Same as above

requirements.txt
rhugonnet: This will revert back with `environment.yml` when removing the current `git+geoutils` for CI tests! 

Right now I just added an exception in the `pre-commit` for `conda_to_pip_deps` so that CI does run when pulling from Git, and that just adds this versionless `geoutils`.

tests/test_coreg/test_affine.py
rhugonnet: Found it! The `inlier_mask` was never used in `Tilt`. This should get more consistent once we do the restructuration in Coreg :)

tests/test_coreg/test_base.py
rhugonnet: Yes! Added a new comment in #327: https://github.com/GlacioHack/xdem/issues/327#issuecomment-1734564317

xdem/coreg/base.py
rhugonnet: I think the check is still the same one you wrote originally, looks like we only accept `bool`:
```python
            assert inlier_mask.dtype == bool, f"Invalid mask dtype: '{inlier_mask.dtype}'. Expected 'bool'"
```

rhugonnet: This joins #293!
Added a comment: https://github.com/GlacioHack/xdem/issues/293#issuecomment-1734573084

rhugonnet:
> I love this and I love you
> 
> Just minor comments!

Happy you like it!! Thanks a lot for the quick review :blush: 
Next will come the re-structuration to have a single `fit()` function for all inputs + be able to pass any optimizer/binning also for Affine classes + some of the rest mentioned in #435.

rhugonnet:
All accounted for, merging and focusing on the next steps! :smile:

PR #439, opened 2023-09-28
Fix sampling in `NuthKaab` and set better default values for other classes

rhugonnet:
This PR fixes:
- The sampling in `NuthKaab` that was done a bit too early, where NaNs introduced by slope/aspect could become problematic (somehow the tests were less sensitive to this).
- Classes such as `Deramp` that didn't have a default `subsample` value set for more reasonable computing times, now all changed to `5e5`.
- The randomness of certain tests with a subsample, such as for `Tilt`.

Additionally, the previous way of installing the base environment for CI (concatenating YML dependencies without the Python version from a script `get_yml_env_nopy`) was failing on Windows when dependencies were pinned with a `<`, for example when I tried `scipy<1.11.1` to ensure the randomness problem didn't come from there.
Now the script is renamed `generate_yml_env_fixed_py`, and generates a temporary YML file during CI to feed to `conda`, which parses all types of pinning without issues on Windows!

I'm leaving `scipy<1.11.1`, which also makes variogram modelling fail, will reactivate once they release `1.11.4`: #430


adehecq (approved):
Seems good! 
The changes in the tests are related to the fact that the automatically generated ddem has changed because of the change in the coregistration subsamplimg?

adehecq:
Some other tests seem to fail though...

rhugonnet:
Thanks! Yes it's the new subsampling that yields different values.

For the test not passing:  it looks like the randomness of results is not fixed between different OS again... Joins https://github.com/GlacioHack/xdem/issues/310. Will investigate...

adehecq:
Haha, I like the many different attempts to make it work !! 🤣

PR #450, opened 2023-11-28
Fix bug in `CoregPipeline.fit()`

### LINKED ISSUE #447: Coregistration: Tilt and VerticalShift pipelines in 0.0.16
I've encountered a regression with coregistration between 0.0.13 and 0.0.16, affecting (at least) Tilt and VerticalShift when used in combination with NuthKaab.

Attached the relevant dh plots: coregistration result is wrong in 0.0.16 with any combination of NuthKaab + one of the two methods mentioned above; when used individually, NuthKaab and VerticalShift behave as expected also in 0.0.16. Also attached the input files to reproduce.

demdiff_orig_plot

[xdem_coreg_issue_data.zip](https://github.com/GlacioHack/xdem/files/13333797/xdem_coreg_issue_data.zip)

rhugonnet:
This PR fixes a bug in `CoregPipeline` that made all fits erroneous (using the original `tba_dem` instead of the updated one).

This was only 4 characters, but not easy to find.

 - [x] Resolves #447,
 - [x] Tests added.

### DIFF
--- a/xdem/coreg/base.py
+++ b/xdem/coreg/base.py
@@ -1544,6 +1544,26 @@ class CoregPipeline(Coreg):
 
         return self
 
+    def _fit_pts_func(
+        self: CoregType,
+        ref_dem: NDArrayf | MArrayf | RasterType | pd.DataFrame,
+        tba_dem: RasterType,
+        verbose: bool = False,
+        **kwargs: Any,
+    ) -> CoregType:
+
+        tba_dem_mod = tba_dem.copy()
+
+        for i, coreg in enumerate(self.pipeline):
+            if verbose:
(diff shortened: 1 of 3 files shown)


rhugonnet:
Will fix the documentation error with #451, merging.

xdem/coreg/base.py
```diff
@@ -1516,7 +1516,7 @@ def fit(
 
             main_args_fit = {
                 "reference_dem": ref_dem,
-                "dem_to_be_aligned": tba_dem,
+                "dem_to_be_aligned": tba_dem_mod,
```
adehecq, line 1519: Good catch !

adehecq (commented):
Thanks for fixing the bug !

Issue #378, opened 2023-06-20
Re-structure `spatialstats.py`


rhugonnet:
The module `spatialstats.py` is starting to be a bit messy!

My thoughts right now:
- Move the spatial variogram `scikit-gstat` wrappers into a `stats/spatial.py` in GeoUtils,
- Move the N-D binning from `scipy` wrappers into a `stats/binning.py` in GeoUtils,
- Rename `xdem/spatialstats` into `xdem/uncertainty.py`? Move the fitting of a sum of variograms directly to `scikit-gstat`? 

Any thoughts @adehecq @erikmannerfelt?

rhugonnet:
From @erikmannerfelt's comment in #158: 
Should also move `_pandas_str_to_interval` to `misc.py`

adehecq:
Good point ! Yes, it can indeed be a bit difficult to find its way through all the functions... Some ideas/comments:
- I agree that the scikit-gstat wrappers could be moved to a separate (hidden?) submodule
- I'm also fine having a module named `uncertainty.py` containing the higher level functions highlighted in the examples. I don't know if I would move the binning into a separate module though. Wouldn't it be a very small submodule?
- after our discussions in PR #158, and my experience working with these tools yesterday: wouldn't it make sense to also have a binning and parametric option to calculate the heteroscedasticity? I was working on limited point observations, and in that case, the nearest neighbor extrapolation for values outside the domain didn't make so much sense, while a parametric fit would be super useful !
- I also noticed while going through the documentation that in the API, one cannot get access to the full docstring of functions in spatialstats.py (and other modules), only a summary. This forced me to check in the code directly, which is not so nice for beginners. Why is that?

rhugonnet:
OK, perfect!
The binning stuff takes quite a bit of space, happy to add parametric fits (I also added a piecewise application of the original binning in `spatialstats.py` in #158).
Yes, the API of xDEM is not yet ready, we'll need to organize stuff a bit like in GeoUtils.

Issue #275, opened 2022-07-22
Describe GeoUtils-xDEM link in documentation


rhugonnet:
As discussed in the meeting of July 6th with @adehecq
Further brought up by #274 and also linked to #178

rhugonnet:
Point made by a user related to this issue: we could copy the `Raster` API of GeoUtils in xDEM to make it easier to find class functions of `DEM`.

PR #478, opened 2024-02-24
Wrap all functionalities as `DEM` methods

### LINKED ISSUE #474: Add wrappers for all features (terrain, coregistration) as `DEM` class methods
For simplicity and consistency of the API, we should be able to do:
```python
dem = xdem.DEM(filename)

dem.slope()
dem.coregister_3d(other_elev, stable_terrain)
dem.error(ref_elev, stable_terrain)
```

It's simply writing wrappers.
The existing function modules could stay for modularity (coreg) or performance (terrain?), but I think this is a top priority for new users!

rhugonnet:
This PR:
- Adds all of xDEM's methods as wrappers in the `DEM` class, for instance `DEM.slope()`, `DEM.coregister_3d()`, `DEM.estimate_uncertainty()`,
- Adds tests to check these run correctly (simple comparison to outputs of module functions),
- Adds a `copy_doc` decorator in `misc.py` that allows to copy the docstring of `terrain` function descriptions in the `DEM` class without having to re-write it all,
- Adds to the documentation a new "Elevation data objects" section containing a "The digital elevation model" page that describes the `DEM` object (and soon an "Elevation point cloud", and "Xarray DEM accessor"),
- Adds a "Quick use" subsection on the pages "Coregistration" and "Terrain attributes" to mention the options of calling the functions from a `DEM`, or to use the module directly,
- Updates the "Spatial statistics" page into an "Uncertainty analysis" page, in preparation for #378,
- Updates the API of xDEM to mirror part of GeoUtils, or point towards it.

Resolves #474
Partially addresses #275 

I will further improve the uncertainty analysis page in #378, and we can improve the one of the coregistration after our hackathon! Then we'll be close to a fully populated and homogeneous documentation for a `v0.1`! :smile:


rhugonnet:
Ready to merge, other documentation change will take place in future PRs. 
Maybe the most efficient is to review the entire documentation at once it's over? By default it won't be visible for users until we release a new version (will be accessible in `latest`, but `stable` is the default one).

rhugonnet:
Functions already tested, only wrappers. And for the documentation, let's review all at once before the first minor release, it'll be more practical!
Merging.

Issue #463, opened 2024-01-31
Add an `ElevationPointCloud` class to facilitate operations with point data?


rhugonnet:
Could be a subclass of a `PointCloud` class in GeoUtils (consistent Point interface is a feature before `v1.0`: https://github.com/GlacioHack/geoutils/wiki/Roadmap-to-v1.0, that could benefit from:
- Vertical reference transformations,
- Coregistration.

See `PointCloud` issue in GeoUtils: https://github.com/GlacioHack/geoutils/issues/463

PR #480, opened 2024-03-01
Fuse `Coreg` point functions and add consistent Raster-Point logic

### LINKED ISSUE #134: Allow point clouds for `Coreg`
A suggestion for the memory problem.

A DEM is often quite large, so how does one do coregistration on just a subsample?

* A geoutils function to "read N pixels as a X/Y/Z point cloud"
* Allow X/Y/Z point clouds to be used for the `Coreg` class. They could be of shape (rows, 3), and all with shape[1] == 3 and len(shape) == 2 could be parsed as point clouds.

### LINKED ISSUE #376: RichDEM does not build with latest GDAL versions
Environment build fails in `mamba` when `richdem` builds with latest GDAL versions.

Related tests are skipped and dependency in `dev-environment.yml` removed temporarily, see #375.

### LINKED ISSUE #377: New warning make tests fails in `test_terrain`
The test failing is `test_attribute_functions_against_gdaldem`.
Removed the filter that made warnings raise an error in `pytest` temporarily, see #375.

### LINKED ISSUE #402: Revamp and simplify the raster-point/P-R/P-P functionality

### LINKED ISSUE #404: `Coreg.fit_pts(..., mask_high_curv=True)` does not seem to work.
I'm getting an error when enabling the `mask_high_curv=True` flag (after adhering to the warning: "Warning: There is no curvature in dataframe. Set mask_high_curv=True for more robust results"). 

```python
File /nix/store/kyrwb500hfcyyabb66msvq2whpvhmh0d-python3-3.10.10-env/lib/python3.10/site-packages/xdem/coreg/base.py:907, in Coreg.fit_pts(self, reference_dem, dem_to_be_aligned, inlier_mask, transform, samples, subsample, verbose, mask_high_curv, order, z_name, weights)
    903 ref_dem = reference_dem[ref_valid]
    905 if mask_high_curv:
    906     maxc = np.maximum(
--> 907         np.abs(get_terrain_attribute(tba_dem, attribute=["planform_curvature", "profile_curvature"])), axis=0
    908     )
    909     # Mask very high curvatures to avoid resolution biases
    910     mask_hc = maxc.data > 5.0

TypeError: bad operand type for abs(): 'Raster'
```

The code in question is here:

https://github.com/GlacioHack/xdem/blob/02902c095ffb2bfbf422e90c13d4fce227da6583/xdem/coreg/base.py#L905-L910

I don't know if this piece of code was ever tested, or if something new has broken it! Either way, it doesn't seem to work right now, as `np.abs(Raster)` doesn't work it seems. @adehecq  or @rhugonnet, do you know if this has ever worked? I see three potential solutions:

- Make `np.abs(Raster)` work in GeoUtils.
- Change the problematic line to: `np.max(np.abs(get_terrain_attribute(tba_dem, attribute=[...]).data), axis=0)` (note the `.data` addition to make it a masked array instead of a raster).
- Remove this functionality here altogether and make it a `Filter`

Also, I don't think it's great that a maxc limit of 5.0 is hardcoded. Should that be made a keyword argument?

What do you think @rhugonnet?

### LINKED ISSUE #472: Make `invert_matrix` accessible from `xdem.coreg`
Currently only possible calling `xdem.coreg.base`

### LINKED ISSUE #475: Fix existing `fit_pts`/`apply_pts` functionalities
There are:
- Many hard-coded filters that run in the background, such as `min_count` in `_get_horizontal_shift`,
- The input `GeoDataFrame` in `fit_pts` expects the x/y values copied in "E" and "N" columns while they are already available,
- The `apply_pts` functions expects a different format of a 3 x N `np.ndarray`...

A lot of homogenization to do!

### LINKED ISSUE #479: Store final number of points used by `subsample` in `Coreg` class
Right now if `subsample=500000` and there is only 1000 points to use during Coreg, `Coreg._meta["subsample"]` is still 500000.
We need to add another attribute `subsample_final` that counts the number of points actually used.

### LINKED ISSUE #489: Capture or solve all warnings raised in tests
Mentioned in #477. Quite important otherwise users also get warnings for no reason at many occurences.

Instead of listing `warnings.simplefilter("error")` in target test functions, every `UserWarning` should be dealt with at the source! We can mirror https://github.com/GlacioHack/geoutils/pull/458 for this.

Will do this directly in #480 to avoid too many conflicts.

rhugonnet:
## Summary

This PR adds point-raster and point-point support in the base `Coreg` class, with modifications to properly inherit it through `CoregPipeline` and `BlockwiseCoreg`, making the treatment of point-raster coregistration fully consistent for the entire coregistration module. It adds point-raster support for all `BiasCorr` classes, and maintains current point-raster support for `AffineCoreg` classes.

## In practice

Any `gpd.GeoDataFrame` can now be passed with a `z_name` argument defining the column name representing the elevation point cloud data, and it can be passed either as `reference_elev`, or `to_be_aligned_elev`, or both. In the future, it will be easy to pass an `xdem.ElevationPointCloud()` that knows the `z_name` to avoid multiplying arguments, see #463.

A `gpd.GeoDataFrame` is always compared to a `gu.Raster` using `gu.Raster.interp_points`, fully consistent and tested since https://github.com/GlacioHack/geoutils/pull/484. It mirrors `Xarray.interp()` behaviour, but should be faster on equal grids thanks to using directly `map_coordinates`.
A `gu.Raster` is converted into points using `to_points()` (for tests or for fallback mechanism in the `Coreg` class discussed in #402). This still requires some fixes in GeoUtils to avoid passing too many arguments: https://github.com/GlacioHack/geoutils/issues/499.

Everything is done through `fit()` and `apply()`.

## Changes

This PR:
- Merges the functions `apply_pts` and `apply`, and `fit_pts` and `fit`,
- Moves the pre/post-processing of `fit` or `apply` into non-public `_preproc_...` and `_postproc_...` functions for better organisation and clarity depending on input type (point or raster),
- Renames `reference_dem` and `dem_to_be_aligned` into `reference_elev` and `to_be_aligned_elev` as inputs can be DEMs or elevation point clouds,
- Renames `_fit_func()` into `_fit_rst_rst()` (only valid for "Raster-Raster") for all subclasses, `_fit_pts_func()` into `_fit_rst_pts()` (for "Raster-Point"), and adds a `_fit_pts_pts()` (for "Point-Point"); also renames `_apply_func()` into `_apply_rst()` (apply to a raster) and `_apply_pts_func()` into `_apply_pts()` (to point),
- Adds the fallback established by @erikmannerfelt discussed in #402 in a single parent `Coreg._fit_func()` which distributes to all other `fit_rst_rst()`, `fit_rst_pts()` and `fit_pts_pts()` of subclasses. Same for `Coreg._apply_func()` that distributes to `_apply_rst()` and `_apply_pts()` depending on input type.
- Changes accepted input point type to only `gpd.GeoDataFrame` with a column name `z_name` (in the future will easily accept `gu.PointCloud` that can be created easily from an array and read from a point cloud file e.g. LAZ, but need to wait for GeoUtils PRs for this),
- Replace `CoregPipeline._apply_rst()` function by an overridden `apply()` function to work properly for any point/raster input and with `BlockwiseCoreg`, this mirrors the overridden `fit()` function added in #436,
- Makes relevant changes to raster-point functions `_fit_rst_pts()` that had inconsistent input/output,
- Renames the `apply_matrix()` function into `_apply_matrix_rst()`, and moves the code used to apply matrix transforms to point data into a new `_apply_matrix_pts()` function, and wraps them into a new `apply_matrix()` function that works on any raster or point! Also makes `invert_matrix()` public through `coreg`,
- Renames `BiasCorr.fit_rst_rst()` into `BiasCorr._fit_biascorr()` to be able to call it through `BiasCorr.fit_rst_pts()` without being overridden by subclass functions, and now calls directly this parent class function in most subclasses, also renamed in `BiasCorr1D`, `2D`, etc, to maintain the chain of checks,
- Adds a `subsample_final` key in `Coreg._meta` to save the number of final pixels used during Coreg, see #479,
- Creates a `NotImplementCoregFit` and `NotImplementedCoregApply` subclasses of `NotImplementedError` to differentiate these specific type of errors made to track the implementation of Coreg subclass methods to other potential `NotImplementedError`.
- Adds tests in `test_biascorr` to check any combination of raster-raster, raster-point or point-raster input for all functionalities, with checks verifying that all corrections indeed remove 99% of the variance introduced artificially,
- Makes minor adjusts in other tests of `test_coreg/`, almost everything was passing the same way with the new structure!
- Fixes the `GradientDescending` test values by half a pixel due to previous enforcement of "area" or "point" coordinates,
- Improving tests of `test_affine` will have to come in another PR, see below.

Additionally, this PR makes all `UserWarning` fail tests at the package level, and removes the individual `warnings.simplefilter("error")`. All other warnings that depend on our routines are either fixed, filtered or caught if expected during the tests.

Resolves #472 
Resolves #479 
Resolves #402
Resolves #475
Resolves #134 
Resolves #404 
Resolves #489 
Resolves #376 
Resolves #377 

TO-DO:

- [x] Add point-raster support in `biascorr`,
- [x] Add point-raster tests in `test_biascorr`,
- [x] Fix point-raster behaviour for `CoregPipeline`.

## For later

Reworking tests in `test_affine` and adapting some of the functions: more efficient if done in another PR after working on the `AffineCoreg` restructuration for being more modular. Opened #485.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -13,9 +13,9 @@ dependencies:
   - tqdm
   - scikit-image=0.*
   - scikit-gstat>=1.0
-  - geoutils=0.1.*
+#  - geoutils=0.1.*
   - pip
 
-  # To run CI against latest GeoUtils
-  # - pip:
-  #   - git+https://github.com/GlacioHack/geoutils.git
+   # To run CI against latest GeoUtils
+  - pip:
+    - https://github.com/rhugonnet/geoutils.git@fix_to_points
--- a/environment.yml
+++ b/environment.yml
@@ -18,4 +18,4 @@ dependencies:
 
    # To run CI against latest GeoUtils
   - pip:
-    - https://github.com/rhugonnet/geoutils.git@fix_to_points
+    - git+https://github.com/rhugonnet/geoutils@fix_to_points
--- a/environment.yml
+++ b/environment.yml
@@ -18,4 +18,4 @@ dependencies:
 
    # To run CI against latest GeoUtils
   - pip:
-    - git+https://github.com/rhugonnet/geoutils@fix_to_points
+    - git+https://github.com/rhugonnet/geoutils.git@fix_to_points
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - python>=3.9,<3.12
-  - geopandas>=0.12.0,<0.14
+  - geopandas>=0.12.0
   - numba=0.*
   - numpy=1.*
   - matplotlib=3.*
@@ -13,7 +13,7 @@ dependencies:
   - tqdm
   - scikit-image=0.*
   - scikit-gstat>=1.0
-  - geoutils=0.0.15
+  - geoutils=0.1.*
   - pip
 
 #  - pip:
--- a/environment.yml
+++ b/environment.yml
@@ -16,5 +16,6 @@ dependencies:
   - geoutils=0.1.*
   - pip
 
-#  - pip:
-#     - git+https://github.com/GlacioHack/geoutils.git
+  # To run CI against latest GeoUtils
+  # - pip:
+  #   - git+https://github.com/GlacioHack/geoutils.git
--- a/environment.yml
+++ b/environment.yml
@@ -2,18 +2,16 @@ name: xdem
 channels:
   - conda-forge
 dependencies:
-  - python>=3.9
-  - geopandas>=0.10.0
-  - fiona
-  - shapely
-  - numba
-  - numpy
-  - matplotlib
-  - pyproj>=3.4
-  - rasterio>=1.3
-  - scipy
+  - python>=3.9,<3.12
+  - geopandas>=0.12.0,<0.14
+  - numba=0.*
+  - numpy=1.*
+  - matplotlib=3.*
+  - pyproj>=3.4,<4
+  - rasterio>=1.3,<2
+  - scipy=1.*
   - tqdm
-  - scikit-image
+  - scikit-image=0.*
   - scikit-gstat>=1.0
   - geoutils=0.0.15
   - pip
--- a/environment.yml
+++ b/environment.yml
@@ -13,9 +13,9 @@ dependencies:
   - tqdm
   - scikit-image=0.*
   - scikit-gstat>=1.0
-#  - geoutils=0.1.*
+  - geoutils>=0.1.2
   - pip
 
    # To run CI against latest GeoUtils
-  - pip:
-    - git+https://github.com/rhugonnet/geoutils.git@fix_to_points
+#  - pip:
+#    - git+https://github.com/rhugonnet/geoutils.git@fix_to_points
--- a/xdem/coreg/affine.py
+++ b/xdem/coreg/affine.py
@@ -583,7 +583,6 @@ class ICP(AffineCoreg):
             )
 
         # If raster was reference, invert the matrix
-        # TODO: Move matrix/invert_matrix to affine module?
         if ref == "raster":
             matrix = xdem.coreg.base.invert_matrix(matrix)
 
--- a/xdem/coreg/affine.py
+++ b/xdem/coreg/affine.py
@@ -1213,7 +1213,7 @@ class GradientDescending(AffineCoreg):
 
         ref_elev = (
             Raster.from_array(ref_elev, transform=transform, crs=crs, nodata=-9999.0)
-            .to_points(as_array=False, pixel_offset="center")
+            .to_pointcloud()
             .ds
         )
         ref_elev["E"] = ref_elev.geometry.x
--- a/xdem/coreg/affine.py
+++ b/xdem/coreg/affine.py
@@ -11,8 +11,9 @@ try:
     import cv2
 
     _has_cv2 = True
-except ImportError:
+except ImportError as e:
     _has_cv2 = False
+    raise e
 import geopandas as gpd
 import numpy as np
 import rasterio as rio
--- a/xdem/coreg/affine.py
+++ b/xdem/coreg/affine.py
@@ -878,7 +878,7 @@ projected CRS. First, reproject your DEMs in a local projected CRS, e.g. UTM, an
         if verbose:
             print("Running Nuth and Kääb (2011) coregistration. Shift pts instead of shifting dem")
 
-        rst_elev = Raster.from_array(rst_elev, transform=transform, crs=crs)
+        rst_elev = Raster.from_array(rst_elev, transform=transform, crs=crs, nodata=-9999)
         tba_arr, _ = get_array_and_mask(rst_elev)
 
         bounds, resolution = _transform_to_bounds_and_res(ref_elev.shape, transform)
@@ -1120,7 +1120,7 @@ class GradientDescending(AffineCoreg):
             rst_elev = ref_elev
             ref = "raster"
 
-        rst_elev = Raster.from_array(rst_elev, transform=transform, crs=crs)
+        rst_elev = Raster.from_array(rst_elev, transform=transform, crs=crs, nodata=-9999)
 
         # Perform downsampling if subsample != None
         if self._meta["subsample"] and len(point_elev) > self._meta["subsample"]:
@@ -1213,7 +1213,7 @@ class GradientDescending(AffineCoreg):
 
         ref_elev = (
             Raster.from_array(ref_elev, transform=transform, crs=crs, nodata=-9999.0)
-            .to_pointcloud()
+            .to_pointcloud(force_pixel_offset="center")
             .ds
         )
         ref_elev["E"] = ref_elev.geometry.x
--- a/xdem/coreg/affine.py
+++ b/xdem/coreg/affine.py
@@ -5,20 +5,22 @@ from __future__ import annotations
 import warnings
 from typing import Any, Callable, TypeVar
 
+import xdem.coreg.base
+
 try:
     import cv2
 
     _has_cv2 = True
 except ImportError:
     _has_cv2 = False
+import geopandas as gpd
 import numpy as np
-import pandas as pd
 import rasterio as rio
 import scipy
 import scipy.interpolate
 import scipy.ndimage
 import scipy.optimize
-from geoutils.raster import Raster, RasterType, get_array_and_mask
+from geoutils.raster import Raster, get_array_and_mask
 from tqdm import trange
 
 from xdem._typing import NDArrayb, NDArrayf
@@ -298,36 +300,6 @@ class AffineCoreg(Coreg):
 
         raise NotImplementedError("This should be implemented by subclassing")
 
-    def _fit_func(
-        self,
-        ref_dem: NDArrayf,
-        tba_dem: NDArrayf,
-        inlier_mask: NDArrayb,
-        transform: rio.transform.Affine,
-        crs: rio.crs.CRS,
-        weights: NDArrayf | None,
-        bias_vars: dict[str, NDArrayf] | None = None,
-        verbose: bool = False,
-        **kwargs: Any,
-    ) -> None:
-        # FOR DEVELOPERS: This function needs to be implemented.
-        raise NotImplementedError("This step has to be implemented by subclassing.")
-
-    def _apply_func(
-        self,
-        dem: NDArrayf,
-        transform: rio.transform.Affine,
-        crs: rio.crs.CRS,
-        bias_vars: dict[str, NDArrayf] | None = None,
-        **kwargs: Any,
-    ) -> tuple[NDArrayf, rio.transform.Affine]:
-        # FOR DEVELOPERS: This function is only needed for non-rigid transforms.
-        raise NotImplementedError("This should have been implemented by subclassing")
-
-    def _apply_pts_func(self, coords: NDArrayf) -> NDArrayf:
-        # FOR DEVELOPERS: This function is only needed for non-rigid transforms.
-        raise NotImplementedError("This should have been implemented by subclassing")
-
 
 class VerticalShift(AffineCoreg):
     """
@@ -350,14 +322,15 @@ class VerticalShift(AffineCoreg):
 
         super().__init__(meta={"vshift_func": vshift_func}, subsample=subsample)
 
-    def _fit_func(
+    def _fit_rst_rst(
         self,
-        ref_dem: NDArrayf,
-        tba_dem: NDArrayf,
+        ref_elev: NDArrayf,
+        tba_elev: NDArrayf,
         inlier_mask: NDArrayb,
         transform: rio.transform.Affine,
         crs: rio.crs.CRS,
-        weights: NDArrayf | None,
+        z_name: str,
+        weights: NDArrayf | None = None,
         bias_vars: dict[str, NDArrayf] | None = None,
         verbose: bool = False,
         **kwargs: Any,
@@ -366,7 +339,7 @@ class VerticalShift(AffineCoreg):
 
         if verbose:
             print("Estimating the vertical shift...")
-        diff = ref_dem - tba_dem
+        diff = ref_elev - tba_elev
 
         valid_mask = np.logical_and.reduce((inlier_mask, np.isfinite(diff)))
         subsample_mask = self._get_subsample_on_valid_mask(valid_mask=valid_mask)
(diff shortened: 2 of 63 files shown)


rhugonnet:
Tests all passing locally.
To fix it in CI: probably need to release a `0.1.1` for GeoUtils that integrates the recent `interp_points` fixes!

rhugonnet:
@erikmannerfelt Long time since we had that revamping discussion, it's finally done!! :grin:

rhugonnet:
Tests pass locally but require [#](https://github.com/GlacioHack/geoutils/pull/501)

environment.yml
```diff
@@ -13,9 +13,9 @@ dependencies:
...
-  - geoutils=0.1.*
+  - geoutils>=0.1.2
   - pip
 
-  # To run CI against latest GeoUtils
-  # - pip:
-  #   - git+https://github.com/GlacioHack/geoutils.git
+   # To run CI against latest GeoUtils
+#  - pip:
+#    - git+https://github.com/rhugonnet/geoutils.git@fix_to_points
```
adehecq, line 21: Should we not remove these lines?

xdem/coreg/base.py
```diff
@@ -379,7 +379,252 @@ def _preprocess_coreg_raster_input(
...
+
+    return rst_elev, point_elev, inlier_mask, transform, crs
+
+
+def _preprocess_coreg_fit_point_point(
+    reference_elev: gpd.GeoDataFrame, to_be_aligned_elev: gpd.GeoDataFrame
+) -> tuple[gpd.GeoDataFrame, gpd.GeoDataFrame]:
+    """Pre-processing and checks of fit for point-point input."""
+
+    ref_dem = reference_elev
```
adehecq, line 432: Is it really worth renaming into "ref_dem". This is a bit comfusing... Same for "tba_dem" also this one is modified at least. Maybe use ref_elev and tba_elev.

xdem/coreg/base.py
```diff
@@ -379,7 +379,252 @@ def _preprocess_coreg_raster_input(
...
+        if isinstance(reference_elev, (np.ndarray, gu.Raster)):
+            raster_elev = reference_elev
+            point_elev = to_be_aligned_elev
+            ref = "raster"
+        else:
+            raster_elev = to_be_aligned_elev
+            point_elev = reference_elev
+            ref = "point"
+
+        rst_elev, point_elev, inlier_mask, transform, crs = _preprocess_coreg_fit_raster_point(
```
adehecq, line 475: here you're overwriting all variables, expect "raster_elev" that is replaced by "rst_elev". Maybe use the same name?

xdem/coreg/base.py
```diff
@@ -379,7 +379,252 @@ def _preprocess_coreg_raster_input(
...
+
+    return elev_out, new_transform, new_crs
+
+
+def _postprocess_coreg_apply_pts(
+    applied_elev: gpd.GeoDataFrame,
+) -> gpd.GeoDataFrame:
+    """Post-processing and checks of apply for point input."""
+
+    # TODO: Convert CRS back if the CRS did not match the one of the fit?
```
adehecq, line 545: Hum... probably.
Do we need to open an issue for that?

xdem/coreg/base.py
```diff
@@ -379,7 +379,252 @@ def _preprocess_coreg_raster_input(
...
+    return applied_elev
+
+
+def _postprocess_coreg_apply_rst(
+    elev: NDArrayf | gu.Raster,
```
adehecq, line 550: What is "elev" here? 
It's a bit easier to follow the logic if you give that info in the docstring. I know these functions are not visible for the user and it's a bit annoying to fill the docstring, but for debugging, it will help for sure! 😄

adehecq: From what I understand, it's the original elevation data, used only to get the output grid?

xdem/coreg/base.py
```diff
@@ -482,12 +727,55 @@ def invert_matrix(matrix: NDArrayf) -> NDArrayf:
...
+    if isinstance(elev, gpd.GeoDataFrame):
+        return _apply_matrix_pts(epc=elev, matrix=matrix, invert=invert, centroid=centroid, z_name=z_name)
+    else:
+        if isinstance(elev, gu.Raster):
+            transform = elev.transform
+            dem = elev.data
+        else:
+            dem = elev
+
+        # TODO: Add exception for translation to update only geotransform, maybe directly in apply_matrix?
```
adehecq, line 769: Would be great indeed!

xdem/coreg/base.py
```diff
@@ -826,21 +1181,25 @@ def fit(
         if self._meta["subsample"] != 1:
             self._meta["random_state"] = random_state
 
-        # Pre-process the inputs, by reprojecting and subsampling
-        ref_dem, tba_dem, inlier_mask, transform, crs = _preprocess_coreg_raster_input(
-            reference_dem=reference_dem,
-            dem_to_be_aligned=dem_to_be_aligned,
+        # TODO: Add preproc for points too
```
adehecq, line 1184: do you plan to address those in this PR?

xdem/coreg/affine.py
```diff
@@ -473,93 +454,113 @@ def _fit_func(
...
+        point_elev["N"] = point_elev.geometry.y.values
+
+        if any(col not in point_elev for col in ["nx", "ny", "nz"]):
             for key, raster in [("nx", normal_east), ("ny", normal_north), ("nz", normal_up)]:
                 raster.tags["AREA_OR_POINT"] = "Area"
-                ref_dem[key] = raster.interp_points(
-                    ref_dem[["E", "N"]].values, shift_area_or_point=True, mode="nearest"
+                point_elev[key] = raster.interp_points(
+                    point_elev[["E", "N"]].values,
+                    shift_area_or_point=True,
```
adehecq, line 557: Why did argument "mode" disappear? Btw should we fix it (to bilinear I assume here) or leave it as an optional argument?

xdem/coreg/affine.py
```diff
@@ -580,6 +582,11 @@ def _fit_pts_func(
                 f"'dem_to_be_aligned' had {points['tba'].size} valid points."
             )
 
+        # If raster was reference, invert the matrix
+        # TODO: Move matrix/invert_matrix to affine module?
```
adehecq, line 586: to be addressed in this PR?

xdem/coreg/biascorr.py
```diff
@@ -317,9 +322,114 @@ def _fit_func(  # type: ignore
...
+        valid_pts = np.floor(inlier_rst.interp_points(pts)).astype(bool)  # Interpolates boolean mask as integers
+
+        # If there is a subsample, it needs to be done now on the point dataset to reduce later calculations
+        subsample_mask = self._get_subsample_on_valid_mask(valid_mask=valid_pts, verbose=verbose)
+        pts = pts[subsample_mask]
+
+        # Now all points should be valid, we can pass an inlier mask completely true
+        inlier_pts_alltrue = np.ones(len(pts), dtype=bool)
+
+        # Below, we derive 1D arrays for the rst_rst function to take over after interpolating to the point coordinates
```
adehecq, line 394: do you mean "pts_pts"?

xdem/coreg/biascorr.py
```diff
@@ -317,9 +322,114 @@ def _fit_func(  # type: ignore
...
+        if bias_vars is not None:
+            bias_vars_pts = {}
+            for var in bias_vars.keys():
+                bias_vars_pts[var] = gu.Raster.from_array(
+                    bias_vars[var], transform=transform, crs=crs, nodata=-9999
+                ).interp_points(pts)
+        else:
+            bias_vars_pts = None
+
+        # Send to raster-raster fit
```
adehecq, line 416: same question as above. Your data are points now, right?

xdem/coreg/biascorr.py
```diff
@@ -839,48 +1058,87 @@ def __init__(
...
+        crs: rio.crs.CRS,
+        z_name: str,
+        bias_vars: dict[str, NDArrayf] | None = None,
+        weights: None | NDArrayf = None,
+        verbose: bool = False,
+        **kwargs,
+    ) -> None:
+
+        # The number of parameters in the first guess defines the polynomial order when calling np.polyval2d
+        p0 = np.ones(shape=((self._meta["poly_order"] + 1) * (self._meta["poly_order"] + 1)))
```
adehecq, line 1076: why not use square?

adehecq (dismissed):
Wow, impressive work !! Thanks ! 🙏 
It's neat that the coreg methods now work similarly for raster or point data! It seems all good to me, I didn't see any big problem, just some minor comments. 
I haven't checked the tests and will do later, but you don't need to wait for that.

environment.yml
rhugonnet: It's good to keep them commented for CI testing of the same kind (not having to release a version in GeoUtils to check it passes CI in xDEM). They served me well here! :smile:

xdem/coreg/base.py
rhugonnet: Done

xdem/coreg/base.py
rhugonnet: Done

xdem/coreg/base.py
rhugonnet: Done: https://github.com/GlacioHack/xdem/issues/481

xdem/coreg/base.py
rhugonnet: Exactly, I added an additional description in the function to explain this for our sake later on. 
I don't think we need a full parameter descriptions because except for this one, all other parameters are simply repeated from the main function that calls these non-public ones.

xdem/coreg/base.py
rhugonnet: Opened #499 for it

xdem/coreg/base.py
rhugonnet: This is done, I forgot to remove the TODOs, good catch!

xdem/coreg/base.py
```diff
@@ -1713,13 +2057,15 @@ def fit(
...
-            dem_to_be_aligned=dem_to_be_aligned,
+        ref_dem, tba_dem, inlier_mask, transform, crs = _preprocess_coreg_fit(
+            reference_elev=reference_elev,
+            to_be_aligned_elev=to_be_aligned_elev,
             inlier_mask=inlier_mask,
             transform=transform,
             crs=crs,
         )
+
+        # TODO: Blockwise can only work if one of the two is a Raster... or by defining a grid somehow?
```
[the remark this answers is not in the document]
rhugonnet: Done and opened issue #500

xdem/coreg/biascorr.py
rhugonnet: Actually "rst_rst" but it's because BiasCorr methods can work on 1D arrays that match exactly in coordinates (which wouldn't be the case for "pts_pts").
I added a comment to clarify this behaviour

xdem/coreg/biascorr.py
rhugonnet: Clarified with new comments

xdem/coreg/biascorr.py
rhugonnet: I don't know :see_no_evil:, haha. I'll try with a square I guess it should still work the same.

xdem/coreg/affine.py
rhugonnet: In the end I think we can keep it in the base module for now. Moving to affine.py would require overridding several `Coreg` functions and moving a lot of things to avoid circular imports...

xdem/coreg/affine.py
rhugonnet: The "mode" argument does not exist anymore, it's "method" now to be consistent with SciPy interpn. And actually it did not do anything before, it went into the void! (defaulting to bilinear due to the "order" argument that defaulted to one)
So now it's still bilinear everywhere... For easy use, we might want to add a `geoutils.config` argument for users to change it module-wide!

rhugonnet: Opened https://github.com/GlacioHack/geoutils/issues/530

adehecq: Haha ok that's good this is fixed now ! 🤦 
And good idea to use the config file in the future !

rhugonnet:
Alelujah OpenCV finally installs on CI again... I've opened a Wiki page to remember how to solve this: https://github.com/GlacioHack/xdem/wiki/Installing-OpenCV-in-CI.

adehecq:
Just thinking: should we add a point cloud to our sample data to document these new features in the documentation?

rhugonnet:
Yes, maybe once we have #463!

PR #486, opened 2024-03-12
Speed-up terrain attribute calculation using convolution

rhugonnet:
Ongoing

For the (frequent) case of desiring no edge-filling or nodata-filling method during terrain attribute derivation, we can use a 3D convolution (already implemented with a function hidden in `spatialstats.py`) for fast computing of several terrain attributes. This should work for almost all but fractal rugosity.
Convolution can be done through SciPy or Numba (which one is fastest depends on matrix size and convolution kernel size).
Additionally, our current functions are deriving all coefficients of Zevenberg and Thorne, while less than half are typically use during the calculation!

# To-do

- [ ] Add implementation for windowed attributes,
- [ ] Add tests,
- [ ] Homogenize current implementation to support the new functions.

# Bonus?

- [ ] Check if we can't derive 2D planar fit coefficients analytically for a certain grid size and turn them into kernels (we should be able to!)... To provide the means of deriving slope/aspect/curvature over any grid size! :smile:

### DIFF
--- a/xdem/terrain.py
+++ b/xdem/terrain.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 import warnings
-from typing import Sized, overload
+from typing import Sized, overload, Literal
 
 import geoutils as gu
 import numba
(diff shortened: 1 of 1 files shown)


rhugonnet:
@adehecq @erikmannerfelt @adebardo @belletva This PR is ready for your review, see description above for details.

xdem/terrain.py
```diff
@@ -776,9 +1253,9 @@ def get_terrain_attribute(
     :param hillshade_z_factor: Vertical exaggeration factor.
     :param slope_method: Method to calculate the slope, aspect and hillshade: "Horn" or "ZevenbergThorne".
     :param tri_method: Method to calculate the Terrain Ruggedness Index: "Riley" (topography) or "Wilson" (bathymetry).
-    :param fill_method: See the 'get_quadric_coefficients()' docstring for information.
-    :param edge_method: See the 'get_quadric_coefficients()' docstring for information.
     :param window_size: Window size for windowed ruggedness and roughness indexes.
+    :param engine: Engine to use for computing the attributes with convolution or other windowed calculations, currently
```
adehecq, line 1257: Maybe the choice of engine should be slightly more documented, here or in the doc? Basically, what could be reasons for chosing scipy over the default numba?
From your benchmark tests, I see that scipy is faster for small arrays, but does it matter if the array is small anyway and computation time is small? Are there other reasons that could push a user to use scipy?

adehecq (approved):
Looking good to me! Thanks!! 🙌 🙌 
It will be nice to see the benchmark repo to continue tracking performance in the future.

xdem/terrain.py
rhugonnet: I've added an explanation in the doc.
What you see for small arrays is actually the initial compile time of Numba, which is unavoidable. I've explained that.

rhugonnet: I basically didn't do a first independent call to cache before timing the performance, so the first array size is wrong in the plot.
See: https://numba.readthedocs.io/en/stable/user/5minguide.html#how-to-measure-the-performance-of-numba

rhugonnet:
> Do you plan to add the "edge_method" problem in a new issue?

You're right, opened #747!

PR #502, opened 2024-04-08
Update documentation and warnings before `0.1` release

### LINKED ISSUE #275: Describe GeoUtils-xDEM link in documentation

### LINKED ISSUE #285: Improve DEM background documentation page
I know it is not part of this PR, but reading the doc makes me think that we could quicly move the first paragraph of this page to the "background" section about DEMs. I would take the first sentence as it is (maybe later we can develop more about the methods used to generate DEMs), then say that a DEM is defined by:
-  a geotransform, that sets the georeferencing information: position of the top-left corner, pixel size and possible rotation. It can be obtained with `self.transform` (similar to geoutils.raster.transform).
- a projection to set the coordinates into World space. Obtained with `self.crs`.
- a nodata value to handle missing observations (`self.nodata`)
- specifically for DEMs, a vertical reference datum must be set (`self.vref`).

_Originally posted by @adehecq in https://github.com/GlacioHack/xdem/pull/283#discussion_r952287311_

### LINKED ISSUE #431: Decide on author attribution across GeoUtils and xDEM
Right now we have:
- "The GlacioHack team" on PyPI: https://pypi.org/project/xdem/,
- "GlacioHack" in the LICENSE (which ships with Conda): https://github.com/GlacioHack/xdem/blob/main/LICENSE,
- "Erik Mannerfelt, Romain Hugonnet, Amaury Dehecq and others" in the docs: https://xdem.readthedocs.io/,
- "xDEM contributors" on Zenodo: https://zenodo.org/record/4809698.

It'd probably be better to be more consistent... What are your thoughts @adehecq @erikmannerfelt?

### LINKED ISSUE #434: Improve `Coreg` docstrings
Maybe to-do after things have changed in the next weeks.

              Maybe a minor comment. Should we expand the docstrings at the beginning of each coreg submodule to better explain what it contains?
E.g in base.py: """Base coregistration classes to define generic methods and pre/post-processing of input data.""" -> add a short description of the base Coreg class and attributes, the Pipeline and the Blockwise? Or is it too redundant with the doc and or the specific docstrings?

_Originally posted by @adehecq in https://github.com/GlacioHack/xdem/issues/424#issuecomment-1707929766_

### LINKED ISSUE #464: Add a "Correct-your-DEM cheatsheet" page to the documentation?
Most interactions I have with users are about knowing the right tool to deal with their DEM issues, typically "here's this bias I have, what's the cause and how to correct it?".
The most common scenarios:
- Positive/Negative errors that look like the landforms = Horizontal coregistration,
- Big vertical error that is only positive or negative error, but not completely consistent everywhere = Vertical reference (geoid/ellipsoid),
- Patterns of errors after co-registration: What bias correction or other affine registration to use?
- Positive/negative errors around point samples: What method to use to compare point and raster data? (rasterizing points within pixels creates location biases, need to interpolate the raster at point locations).

We could have a "Cheatsheet" for users to look up if their scenario falls among the typical ones (with illustrated example), and point towards the right xDEM tools. :slightly_smiling_face:

### LINKED ISSUE #477: To-do before `v0.1`
The things to do before releasing a `0.1` for xDEM are:

- [x]  #474 then update related documentation example to make them even more accessible,
- [x]  #275 without forgetting to copy the `Raster` API to help users,
- [x] Deal with all warnings in tests making `UserWarning` fail tests everywhere,
- [x] Write more useful error message that propose solutions.

Coregistration improvements can be left for a next minor version, as well as elevation point clouds, etc.

### LINKED ISSUE #505: Add Deramp and Blockwise to the documentation (coregistration section)
The current documentation ([here](https://xdem.readthedocs.io/en/stable/coregistration.html)) only describes in detail the Nuth & Kaab, Tilt, ICP and Pipeline algorithms. Some algorithms (I'm thinking about Deramp and Blockwise, maybe others?) are missing from that part and so some users miss these tools.
One more issue to add to #435 !

### LINKED ISSUE #528: Collect references of projects using xdem
Out of interest, for referencing in papers, and for some light boasting, it would be nice to have a list of papers using xdem in the documentation! A short non-comprehensive list is:

- https://doi.org/10.1016/j.rse.2022.113379
- https://doi.org/10.33799/jokull2023.73.035
- https://doi.org/10.5194/tc-17-3251-2023
- https://doi.org/10.5194/isprs-archives-XLVIII-4-W1-2022-5-2022
- https://doi.org/10.1109/JSTARS.2022.3188922
- https://doi.org/10.5194/tc-16-3249-2022
- https://doi.org/10.1016/j.qsa.2021.100041
- https://doi.org/10.1016/j.geomorph.2023.108805
- https://doi.org/10.1038/s41467-024-48886-x

MSc theses:
- https://doi.org/10.3929/ethz-b-000637956
- https://hdl.handle.net/11250/3057200
- http://hdl.handle.net/10852/103730

### LINKED ISSUE #532: Update coreg examples to use `fit_and_apply`
Following #527, the documentation should encourage the use of the function `fit_and_apply` rather than the separate `fit` and `apply` functions. In particular, the examples should be updated to reflect this.

### LINKED ISSUE #562: Fix date in surface roughness reference
Should be "Dartnell et al. (2004)" and not "Dartnell (2000)".
To address in #502

### LINKED ISSUE #583: Automatically extract translations/rotations from affine matrix
For more easy manipulation from non-expert users. We already do this in tests.

Could be class method of `AffineCoreg` named `to_translations` and `to_rotations`?

### LINKED ISSUE #614: Resolution of figures generated by Sphinx-Gallery is blurry
Mirrors https://github.com/GlacioHack/geoutils/issues/612

rhugonnet:
This PR improves the documentation in preparation for a first minor release `0.1`! :partying_face: 

The documentation now separates scientific background content and code examples, adds better links and plots to all feature pages, and adds some auxiliary pages for diverse uses based on user feedback (how to cite, method overview, cheatsheet for helping selecting a method, configuration, ecosystem, etc).

Link to rendered doc: https://xdem-rhugonnet.readthedocs.io/en/towards_0.1/index.html 

## Detailed documentation changes

This PR:
- Moves scientific background content that was mixed with the code examples, and adds new content, into a series of **guides pages** "Guides for elevated analysis" in the new section "Resources", illustrated with plots generated just for the documentation and a list for references for "More reading" at the bottom of the pages,
- Adds a "Cheatsheet: How to correct... ?" page in the new section "Resources" to help users read patterns of errors in elevation data to select a correction method,
- Adds an "Ecosystem" page in the new section "Resources" that points towards other relevant packages to pair with xDEM,
- Adds a "Citing and method overview" page in the section "Getting started", that points to xDEM's Zenodo and lists the references to use for all methods implemented in the packages,
- Adds a "Use in publication" page in the section "Reference", that lists scientific articles/theses that are using xDEM (to our knowledge),
- Adds a "Configuration" page in the section "Reference", that explains how to configure verbosity level, and georeferenced operations in xDEM via GeoUtils (resampling, etc),
- Streamlines and adds plots to the "Coregistration", "Bias correction" and "Uncertainty analysis" **features pages** for better understanding of these methods,
- Add "Quick use" subsections at the start of all feature pages, to save time for user who don't want to get into the details,
- Adds tables listing all methods at the start of "Terrain attributes", "Coregistration" and "Uncertainty analysis",
- Adds up-to-date descriptions of modular arguments and metadata for "Coregistration" based on recent changes,
- Adds equations used for "Terrain attributes",
- Adds more links to GeoUtils with explanation of the inheritance between packages,
- Adds links between **feature pages** and **guide pages**,
- Streamlines the API reference into a manually ordered list of submodule functions meant to be publicly referenced, and repeating functions that are from GeoUtils to facilitate user searches.

## Other small changes

This PR also has some small changes in the core modules related to the documentation/visualization:
- Adds `AffineCoreg.to_translations()` and `AffineCoreg.to_rotations()` and `AffineCoreg.from_rotations()` (as `from_translations()` already existed) to quickly extract from/build the affine matrix, which might not be easy otherwise for a user (added in a note on the Coregistration page), which directly use `pytransform3d` (so don't require testing),
- Adds a `resample` option to `apply_matrix` for practicality (using a the subfunction `_reproject_horizontal_shift_samecrs` that was already tested, and is moved to the `base` module instead of `affine`),
- Adds more formatting options to `Coreg.info()` for types that were not yet supported (after some testing in the documentation), and a `CoregPipeline.info()` that simply wraps the `Coreg.info()` of each pipeline step,
- Updates the recent function `DEM.estimate_uncertainty()` to have clearer arguments, choice of method mirroring the new documentation.

## TO-DO:

- [x] Add profile curvature and plan curvature formulas on "Terrain attributes" page,
- [x] Add `BlockwiseCoreg` to coregistration page,
- [x] Add "Cheatsheet" page,
- [x] Add link to PyTransform3d on Affine registration page, 
- [x] Add "Ecosystem" page,
- [x] Add "Georeferencing intricacies" page,
- [x] Add documentation for "DhMinimize" method,
- [x] Streamline "Coregistration" and "Bias correction" pages, 
- [x] Detail `.meta` keys for users (and make the attribute public in a separate PR),
- [x] Add `.info()` method for `CoregPipeline`,
- [x] Remove temporarily `BlockwiseCoreg`,
- [x] Update reference for TPI: #562,
- [x] Re-structure/streamline "Spatial statistics for error analysis" guide page,
- [x] Finalize "Uncertainty analysis" page, add warning that interface might change with Scikit-GStat now supporting directly sum of variogram models,
- [x] Streamline "Differencing and volume change" into a "Gap-filling" page and add warning that it might be moved to a different package,
- [x] Proof-read/streamline all gallery examples,
- [x] Prepare banner to communicate breaking changes with `0.1`,
- [x] Check if figures in "Quick start" are fixed (normal colorbar, vector plotting on top),
- [x] Check if PROJ error still appears or not in the rendered documentation, fix it if yes,
- [x] Add "read more" section at the end of each guide page?
- [x] Move down `dDEM` and `DEMCollection` in API, remove `dDEM` from README, and add warning of potential removal in future,
- [x] Fix Nuth&Kääb gallery example colorbar title, 
- [x] Shorten detailed intros from gallery example, 
- [x] Fix all hyperlinks and warnings,
- [x] Fix warnings of sphinx build,
- [x] Release new GeoUtils with default tight layout,
- [ ] Fix cheatsheet vertical CRS failing to download on RTD, 
- [x] Fix bias-correction sinusoid example sometimes failing to correct randomly. Couldn't! Opened #637 

Resolves #477 (last step)
Resolves #505
Resolves #464
Resolves #434
Resolves #285
Resolves #275
Resolves #431
Resolves #532 
Resolves #528
Resolves #562 
Resolves #583
Resolves #614

Opens #637
Opens #638


doc/source/accuracy_precision.md
```diff
@@ -0,0 +1,115 @@
...
+- a large number of [outliers](https://en.wikipedia.org/wiki/Outlier) **that remain difficult to filter** as they can originate from various sources (e.g., blunders, clouds).
+
+All of these factors lead to difficulties in assessing the reliability of elevation data, required to
+perform additional quantitative analysis, which calls for defining the concepts relating to accuracy and precision for elevation data.
+
+## Accuracy and precision of elevation data
+
+### What are accuracy and precision?
+
+[Accuracy and precision](https://en.wikipedia.org/wiki/Accuracy_and_precision) describe random and systematic errors, respectively.
```
erikmannerfelt, line 25: Flip "random and systematic", as accuracy is systematic and precision is random.

doc/source/accuracy_precision.md
```diff
@@ -0,0 +1,115 @@
...
+
+:::{figure} imgs/precision_accuracy.png
+:width: 80%
+
+Source: [antarcticglaciers.org](http://www.antarcticglaciers.org/glacial-geology/dating-glacial-sediments2/precision-and-accuracy-glacial-geology/), accessed 29.06.21.
+:::
+
+### Translating these concepts for elevation data
+
+However, elevation data rarely consists of a single independent measurement but of a **series of measurement** (image grid,
```
erikmannerfelt, line 41: "series of measurement**s**"

doc/source/accuracy_precision.md
```diff
@@ -0,0 +1,115 @@
...
+
+Source: [Hugonnet et al. (2022)](https://doi.org/10.1109/jstars.2022.3188922).
+:::
+
+### Absolute or relative elevation accuracy
+
+Accuracy is generally considered from two focus points:
+
+- **Absolute elevation accuracy** describes systematic errors to the true positioning, usually important when analysis focuses on the exact location of topographic features at a specific epoch.
+- **Relative elevation accuracy** describes systematic errors with reference to other elevation data that does not necessarily matches the true positioning, important for analyses interested in topographic change over time.
```
erikmannerfelt, line 72: "does not necessarily **match** the true positioning"

doc/source/conf.py
```diff
@@ -25,8 +25,8 @@
 # -- Project information -----------------------------------------------------
 
 project = "xDEM"
-copyright = "2021, Erik Mannerfelt, Romain Hugonnet, Amaury Dehecq and others"
-author = "Erik Mannerfelt, Romain Hugonnet, Amaury Dehecq and others"
+copyright = "2020, GlacioHack"
```
erikmannerfelt, line 28: I see some packages leaving the original year while others keep it updated. Any ideas on implications of this?

doc/source/ecosystem.md
```diff
@@ -0,0 +1,27 @@
+(ecosystem)=
+
+# Ecosystem
+
```
erikmannerfelt, line 4: Maybe not here, but somewhere; should we refer to demcompare as that's still used quite much? Sorry if it already says somewhere.

doc/source/ecosystem.md
```diff
@@ -0,0 +1,27 @@
...
+xDEM is but a single tool among a large landscape of open tools for geospatial elevation analysis! Below is a list of 
+other **tools that you might find useful to combine with xDEM**, in particular for retrieving elevation data or to perform complementary analysis.
```
erikmannerfelt, line 6: Add geoutils, as it might not be evident how closely related these two projects are.

doc/source/spatial_stats.md
```diff
@@ -0,0 +1,186 @@
...
+not well-suited to grid data that contains many millions of points and would be impossible to comupute. Thus, in order
+to estimate a variogram for large grid data, subsampling is necessary.
+Random subsampling of the grid samples used is a solution, but often unsatisfactory as it creates a clustering
+of pairwise samples that unevenly represents lag classes (most pairwise differences are found at mid distances, but too
+few at short distances and long distances).
+
+For more details on variography, **we refer to [the documentation of SciKit-GStat](https://scikit-gstat.readthedocs.io/en/latest/userguide/userguide.html).**
+
+
+## Error propagation
```
erikmannerfelt, line 156: Add a TODO here or remove it!

erikmannerfelt: Or an issue

doc/source/static_surfaces.md
```diff
@@ -0,0 +1,81 @@
+(static-surfaces)=
+
+# Static surfaces as error proxy
+
+## The great benefactor of quantitative elevation analysis
+
+Elevation data benefits from an uncommon asset, which is that **large proportions of planetary surface elevations
+usually remain virtually unchanged through time**. Those static surfaces, sometimes also referred to as "stable terrain",
```
erikmannerfelt, line 8: The geologist in me wants you to add "within human time-scales". Everything moves significantly if you wait long enough!

doc/source/static_surfaces.md
```diff
@@ -0,0 +1,81 @@
...
+generally refer to bare-rock, grasslands, and are often isolated by excluding dynamic surfaces such as glaciers,
```
erikmannerfelt, line 9: ..., rivers, human settlements, lakes, ....

xdem/coreg/biascorr.py
```diff
@@ -460,7 +460,7 @@ def __init__(
         | Literal["norder_polynomial"]
         | Literal["nfreq_sumsin"] = "norder_polynomial",
         fit_optimizer: Callable[..., tuple[NDArrayf, Any]] = scipy.optimize.curve_fit,
-        bin_sizes: int | dict[str, int | Iterable[float]] = 100,
+        bin_sizes: int | dict[str, int | Iterable[float]] = 1000,
```
erikmannerfelt, line 463: Does this link to an issue? I'm sure it's a good update but it's a bit off topic!

xdem/coreg/base.py
```diff
@@ -2627,7 +2701,7 @@ def copy(self: CoregType) -> CoregType:
         """Return an identical copy of the class."""
         new_coreg = self.__new__(type(self))
 
-        new_coreg.__dict__ = {key: copy.deepcopy(value) for key, value in self.__dict__.items() if key != "pipeline"}
+        new_coreg.__dict__ = {key: copy.copy(value) for key, value in self.__dict__.items() if key != "pipeline"}
```
erikmannerfelt, line 2704: What does this change do?

xdem/coreg/affine.py
```diff
@@ -848,6 +788,34 @@ def to_matrix(self) -> NDArrayf:
         """Convert the transform to a 4x4 transformation matrix."""
         return self._to_matrix_func()
 
+    def to_translations(self) -> tuple[float, float, float]:
+        """
+        Convert the affine transformation matrix to only its X/Y/Z translations.
```
erikmannerfelt, line 793: I think something funky will happen here if scaling is involved (I don't remember which order the components are applied in). I feel like a warning should be added to this and the `to_rotations()` saying something like "Warning: Information may be lost".

And the conversion is less of a conversion and more of an extraction I would say! Since there's no validation that scaling won't mess with the result.

tests/test_coreg/test_base.py
```diff
@@ -960,8 +960,7 @@ def test_blockwise_coreg_large_gaps(self) -> None:
         ddem_post = (aligned - self.ref).data.compressed()
         ddem_pre = (tba - self.ref).data.compressed()
         assert abs(np.nanmedian(ddem_pre)) > abs(np.nanmedian(ddem_post))
-        # TODO: Figure out why STD here is larger since PR #530
-        # assert np.nanstd(ddem_pre) > np.nanstd(ddem_post)
```
erikmannerfelt, line 964: Why is an identical assertion commented out here?

erikmannerfelt (commented):
Wow, such an improvement to the documentation!! Super great job @rhugonnet!

I have some questions, generally pertaining to the non-documentation part of the PR (perhaps in the future changes like this could be a separate PR? :) )

doc/source/accuracy_precision.md
rhugonnet: Good catch! Fixed

doc/source/accuracy_precision.md
rhugonnet: Fixed

doc/source/conf.py
rhugonnet: No, we should ask advice from CNES/CS on this. Maybe we can adjust when we'll change the license.

doc/source/ecosystem.md
rhugonnet: Good point! Added for now, CNES/CS can orient us on how they'd like demcompare to show up in the doc.

doc/source/ecosystem.md
rhugonnet: Good idea, done!

doc/source/static_surfaces.md
rhugonnet: Good point, added!

tests/test_coreg/test_base.py
rhugonnet: This is linked to #584

xdem/coreg/biascorr.py
rhugonnet: Those default parameters were a bit arbitrary (one of my PRs 2 years ago) and didn't render well, noticed that in the doc so adjusted it!

xdem/coreg/affine.py
rhugonnet: I agree that it's not a conversion, the idea was just to have an extraction! We could change this in the description? 
Leaving this conversation open for @adehecq to comment.

rhugonnet: And I'll look into it for when scaling is involved, great point :wink:

doc/source/spatial_stats.md
rhugonnet: Was unfinished, now done!

xdem/coreg/affine.py
```diff
@@ -1521,7 +1520,7 @@ def _fit_rst_pts(
 
         # Write output to class
         # (Mypy does not pass with normal dict, requires "OutAffineDict" here for some reason...)
-        output_affine = OutAffineDict(shift_x=-easting_offset, shift_y=-northing_offset, shift_z=vertical_offset)
+        output_affine = OutAffineDict(shift_x=easting_offset, shift_y=northing_offset, shift_z=vertical_offset)
```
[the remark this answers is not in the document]
rhugonnet: This line removes a temporary fix that was done in parallel of #595 to run this PR.
So there is actually no change at this line when comparing to `main`! (but you had it pop up when clicking on "See changes since last review")

rhugonnet:
@vschaffn @adebardo FYI: I used the opportunity of merging the new `main` in this big documentation PR to integrate the documentation on logging of #610 (because the page  was not listed in the index so not rendered, and same for the Sphinx-Gallery example as it had the wrong name pattern!).

If you want to check the rendering of the docs in advance in next PRs, you can add your fork (xdem-yourfork) in Readthedocs to render certain branches (Import Project, then Versions/Activate a version/Choose hidden, then the doc builds automatically in Builds at each commit on that branch!): https://readthedocs.org/dashboard/ :wink: 

In short, I have made the following small changes:
- Made both page and example render,
- Integrated the "logging" page into a section of a more generic "Configuration" page,
- Added the name "Verbosity level" in addition to "Logging" in titles and description as, after a bit of asking around, few people knew what logging referred to!
- Re-structured the Sphinx-Gallery example to only show logging within the xDEM function calls for clarity (otherwise it was a bit much at once, after seeing the rendered output!),
- Added a one-liner example for those who just want to know how to show "logging.INFO" level quickly,
- Added some Sphinx configuration-file specifics to make it work (resetting the logging between examples).

xdem/coreg/base.py
rhugonnet: Forgot to add this in the `CoregPipeline` to mirror an issue fixed in #530 for `Coreg`

rhugonnet:
@adehecq @erikmannerfelt This PR is finalized, will be merging soon. Only two small issues I can't seem to fix (with `DirectionalBias` and vertical CRS download through readthedocs...), will likely open issues.
I'll then open a separate PR to add documentation notes on the fusion with demcompare, and get feedback from CNES/CS.

doc/source/background.md
```diff
@@ -0,0 +1,66 @@
...
+xDEM was created during the **[GlacioHack](https://github.com/GlacioHack) hackathon**, that was initiated by
+Amaury Dehecq<sup>2</sup> and took place online on November 8, 2020.
+
+```{margin}
+<sup>3</sup>Check-out [glaciology.ch](https://glaciology.ch) on our founding group of VAW glaciology!
+```
+
+The initial core development of xDEM was performed by members of the Glaciology group of the Laboratory of Hydraulics, Hydrology and
+Glaciology (VAW) at ETH Zürich<sup>3</sup>, with contributions by members of the University of Oslo, the University of Washington, and University
+Grenobles Alpes.
```
adehecq, line 61: Grenoble without s :wink:

doc/source/quick_start.md
```diff
@@ -46,6 +56,11 @@ fn_glacier_outlines = xdem.examples.get_path("longyearbyen_glacier_outlines")
 print(f"DEM 1: {fn_dem_ref}, \nDEM 2: {fn_dem_tba}, \nOutlines: {fn_glacier_outlines}")
 ```
 
+```{tip}
+:class: margin
+Set up your {ref}`verbosity` to manage outputs to the console (or a file) during execution!
+```
+
```
adehecq, line 63: Just a question on line 78 below, which was not edited. 
I assume this is just for the example, as both DEMs are originally on the same vertical datum? It must impact coregistration a bit... I think it's great to showcase this functionality, but it woul dbe good to have a sample data set with actual vertical datum difference.

adehecq: + L 119: "An basic" -> "a basic"

doc/source/citation.md
```diff
@@ -0,0 +1,114 @@
+(citation)=
+
+# Citing and method overview
+
+When using a method implemented in xDEM, one should cite both the package and the original study behind the method (if there is any)!
+
+## Citing xDEM
+
+To cite the package, use the Zenodo DOI: [![Zenodo](https://zenodo.org/badge/doi/10.5281/zenodo.4809697.svg)](https://zenodo.org/doi/10.5281/zenodo.4809697).
```
adehecq, line 9: I think we need to be more detailed than this. The best way is to have a ref like "(xDEM contributors, 202X)" and in the reference list: "Xdem Contributors: xdem, Zenodo [code], https://doi.org/10.5281/zenodo.YYYY, 202X." with year and zenodo link to updated depending on the version and year used.

adehecq: It will be easier when we have a proper article for it...

doc/source/citation.md
```diff
@@ -0,0 +1,114 @@
...
+
+   * - Method
+     - Reference
+   * - Deramp
+     - N/A
+   * - Directional bias (sum of sinuoids)
+     - [Girod et al. (2017)](https://doi.org/10.3390/rs9070704)
+   * - Directional bias (other)
+     - N/A
+   * - Terrain bias (maximum curvature)
```
adehecq, line 75: Hopefully we can update this with the work of Arnaud!

doc/source/citation.md
```diff
@@ -0,0 +1,114 @@
...
+   :widths: 1 1
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Method
+     - Reference
+   * - Bilinear
+     - N/A
+   * - Local and regional hypsometric
+     - [McNabb et al. (2019)](https://tc.copernicus.org/articles/13/895/2019/)
```
adehecq, line 97: It is definitely a good resource to understand the concept of the hypsometric approach, but this is not the paper that actually proposes this method. It has been used for years before Bob's paper. I will check if I can find an original reference for this.

adehecq: McNabb 2019 refers to Arendt (2002) and Kääb (2008). In the first one, there is no occurrence of "hypso". The second one uses the term "hypsography" ! So not sure what is the best way to acknowledge...

doc/source/dem_class.md
```diff
@@ -81,9 +89,12 @@ import geoutils as gu
 fn_glacier_outlines = xdem.examples.get_path("longyearbyen_glacier_outlines")
 vect_gla = gu.Vector(fn_glacier_outlines)
 
+# Crop outlines to those intersecting the DEM
+vect_gla = vect_gla.crop(dem)
+
 # Plot the DEM and the vector file
 dem.plot(cmap="terrain", cbar_title="Elevation (m)")
-vect_gla.plot(dem)  # We pass the DEM as reference for the plot CRS/extent
+vect_gla.plot(dem, ec="k", fc="none")  # We pass the DEM as reference for the plot CRS/extent
```
adehecq, line 97: Actually, on the figure, the plot extent is still that of the outlines, not the DEM...

doc/source/dem_class.md
```diff
@@ -140,7 +151,7 @@ dem_tba = xdem.DEM(filename_tba)
 dem_tba_coreg = dem_tba.coregister_3d(dem)
```
adehecq, line 151: I was not aware of this method ! 😅

doc/source/dem_class.md
```diff
@@ -157,10 +168,11 @@ stable terrain as a proxy.
 
 ```{code-cell} ipython3
 # Estimate elevation uncertainty assuming both DEMs have similar precision
-sig_dem, rho_sig = dem.estimate_uncertainty(dem_tba_coreg, precision_of_other="same")
+sig_dem, rho_sig = dem.estimate_uncertainty(dem_tba_coreg, precision_of_other="same", random_state=42)
```
adehecq, line 171: I was also not aware of this method!!
Maybe explain in the comment why random_state is used and that it should not be used in the general case.

doc/source/vertical_ref.md
```diff
@@ -12,15 +12,64 @@ kernelspec:
...
 # Vertical referencing
 
-xDEM supports the use of **vertical coordinate reference systems (vertical CRSs)** and vertical transformations for DEMs
+xDEM supports the use of **vertical coordinate reference systems (vertical CRSs) and vertical transformations for elevation data**
 by conveniently wrapping PROJ pipelines through [Pyproj](https://pyproj4.github.io/pyproj/stable/) in the {class}`~xdem.DEM` class.
 
-```{important}
+```{note}
 **A {class}`~xdem.DEM` already possesses a {class}`~xdem.DEM.crs` attribute that defines its 2- or 3D CRS**, inherited from
 {class}`~geoutils.Raster`. Unfortunately, most DEM products do not yet come with a 3D CRS in their raster metadata, and
```
adehecq, line 31: Maybe it can be good also to explain that the 2D CRS may contain a term "datum" that is misleading. Even for 2D projections, one needs a datum, which is very often WGS84, for projecting from degrees to meters, but this is not the same as the vertical datum. There was some confusion about it at the winter school.

adehecq: For example, projection UTM zone 44N (EPSG:32644), if printed as WKT string, starts with:
PROJCS["WGS 84 / UTM zone 44N",
    GEOGCS["WGS 84",
        **DATUM["WGS_1984",**
            SPHEROID["WGS 84",6378137,298.257223563,
                AUTHORITY["EPSG","7030"]],
            AUTHORITY["EPSG","6326"]],

doc/source/terrain.md
```diff
@@ -10,8 +24,87 @@ and tested for consistency against [gdaldem](https://gdal.org/programs/gdaldem.h
...
+```
+
+```{code-cell} ipython3
+# Slope from DEM method
+slope = dem.slope()
+# Or from terrain module using an array input
+slope = xdem.terrain.slope(dem.data, resolution=dem.res)
+```
+
+If computational performance is key, xDEM can rely on [RichDEM](https://richdem.readthedocs.io/) by specifying
```
adehecq, line 59: Maybe add a note that we hope to speed-up calculation (using convoutions)?

adehecq: Maybe you can also mention that if several attributes are needed, for optimization is is also better to call `get_terrain_attribute`? I know it is described later as well, but maybe have all "performance" related points at one location is good.

doc/source/terrain.md
```diff
@@ -10,8 +24,87 @@ and tested for consistency against [gdaldem](https://gdal.org/programs/gdaldem.h
...
+`use_richdem=True` for speed-up of specific supported attributes (slope, aspect, curvature).
+
+## Summary of supported methods
+
+```{list-table}
+   :widths: 1 1 1
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Attribute
```
adehecq, line 69: Very useful table! Especially the unit column.
I am also thinking, should we raise a warning when trying to call many of these functions from geographic coordinates? Because it will return unrealistic values (e.g. slope).
Also for the hillshade, we could detect this and add the correction factor.

adehecq: Added issue #634.

doc/source/terrain.md
```diff
@@ -20,20 +113,63 @@ The slope of a DEM describes the tilt, or gradient, of each pixel in relation to
...
+```
 
-The differences between methods are illustrated in the {ref}`sphx_glr_basic_examples_plot_terrain_attributes.py`
-example.
 
-```{image} basic_examples/images/sphx_glr_plot_terrain_attributes_001.png
-:width: 600
-```
+Finally, $\Delta x$
+and $\Delta y$ correspond to the pixel resolution west-east and south-north, respectively.
```
adehecq, line 162: Add a note that we do not support non square pixels in xDEM at the moment? It will raise an error.

doc/source/terrain.md
```diff
@@ -42,169 +178,212 @@ The aspect describes the orientation of strongest slope.
...
+$$
+\theta = arctan(\frac{p}{q}),
+$$
 
-```{image} basic_examples/images/sphx_glr_plot_terrain_attributes_002.png
-:width: 600
+with $p$ and $q$ defined in the slope section.
+
+```{warning}
+A north aspect represents the upper direction of the Y axis in the coordinate reference system of the
```
adehecq, line 192: good point !!

doc/source/terrain.md
```diff
@@ -42,169 +178,212 @@ The aspect describes the orientation of strongest slope.
...
 ```
 
+(curv)=
 ## Curvature
 
 {func}`xdem.DEM.curvature`
 
-The curvature map is the second derivative of elevation, which highlights the convexity or concavity of the terrain.
+The curvature is the second derivative of elevation, which highlights the convexity or concavity of the terrain.
 If a surface is convex (like a mountain peak), it will have positive curvature.
```
adehecq, line 233: I thought we had discussed it before, but it's the opposite. convex = valley, concave = summit.

doc/source/terrain.md
```diff
@@ -42,169 +178,212 @@ The aspect describes the orientation of strongest slope.
...
+curvature.plot(cmap="RdGy_r", cbar_title="Curvature (100 / m)", vmin=-1, vmax=1, interpolation="antialiased")
 ```
 
+(plancurv)=
 ## Planform curvature
 
 {func}`xdem.DEM.planform_curvature`
 
-The planform curvature is the curvature perpendicular to the direction of slope, reported in 100 m{sup}`-1`.
+The planform curvature is the curvature perpendicular to the direction of slope, reported in 100 m{sup}`-1`, also based
```
adehecq, line 255: I had to check because I find the names counter-intuitive. To me, profile would mean perpendicular to the slope. I guess I am biased by the fact we often make cross profiles in glaciology 😉

doc/source/coregistration.md
```diff
@@ -16,250 +16,472 @@ kernelspec:
...
+aligned_dem = my_coreg_pipeline.fit_and_apply(ref_dem, tba_dem)
+```
 
-Each coregistration approach in xDEM inherits their interface from the {class}`~xdem.Coreg` class<sup>1</sup>.
+Information about the coregistration inputs and outputs is summarized in {func}`~xdem.coreg.Coreg.info`.
 
-```{margin}
-<sup>1</sup>In a style resembling [scikit-learn's pipelines](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html#sklearn-linear-model-linearregression).
+```{tip}
+Often, an `inlier_mask` has to be passed to {func}`~xdem.coreg.Coreg.fit` to isolate static surfaces to utilize during coregistration (for instance removing vegetation, snow, glaciers). This mask can be easily derived using {func}`~geoutils.Vector.create_mask`.
```
adehecq, line 93: I would recommend using the function `create_inlier_mask` from the workflow submodule.

doc/source/coregistration.md
```diff
@@ -16,250 +16,472 @@ kernelspec:
...
-It estimates translation iteratively by solving a cosine equation to model the direction at which the DEM is most likely offset.
-First, the DEMs are compared to get a dDEM, and slope/aspect maps are created from the reference DEM.
-Together, these three products contain the information about in which direction the offset is.
-A cosine function is solved using these products to find the most probable offset direction, and an appropriate horizontal shift is applied to fix it.
-This is an iterative process, and cosine functions with suggested shifts are applied in a loop, continuously refining the total offset.
-The loop stops either when the maximum iteration limit is reached, or when the NMAD between the two products stops improving significantly.
+The [Nuth and Kääb (2011)](https://doi.org/10.5194/tc-5-271-2011) coregistration approach estimates a horizontal
+translation iteratively by solving a cosine equation between the terrain slope, aspect and the elevation differences.
+The iteration stops if it reaches the maximum number of iteration limit or if the tolerance does not improve.
```
adehecq, line 170: it's not the tolerance that improves but the shift falls below a certain threshold.

doc/source/coregistration.md
```diff
@@ -16,250 +16,472 @@ kernelspec:
...
-The opencv implementation of ICP includes outlier removal, since extreme outliers will heavily interfere with the nearest neighbour distances.
-This may improve results on noisy data significantly, but care should still be taken, as the risk of landing in [local minima](https://en.wikipedia.org/wiki/Maxima_and_minima) increases.
+- **Performs:** Rigid transform transformation (3D translation + 3D rotation).
+- **Does not support weights.**
+- **Pros:** Efficient at estimating rotation and shifts simultaneously.
+- **Cons:** Poor sub-pixel accuracy for horizontal shifts, sensitive to outliers, and runs slowly with large samples.
 
-### Limitations
+Iterative Closest Point (ICP) coregistration is an iterative point cloud registration method from [Besl and McKay (1992)](https://doi.org/10.1117/12.57955). It aims at iteratively minimizing the distance between closest neighbours by applying sequential rigid transformations. If DEMs are used as inputs, they are converted to point clouds.
+As for Nuth and Kääb (2011), the iteration stops if it reaches the maximum number of iteration limit or if the tolerance does not improve.
```
adehecq, line 311: same, tolerance does not improve

doc/source/biascorr.md
```diff
@@ -15,146 +17,420 @@ kernelspec:
...
+:tags: [hide-input]
+:mystnb:
+:  code_prompt_show: "Show plotting code"
+:  code_prompt_hide: "Hide plotting code"
+
+# Plot before and after
+f, ax = plt.subplots(1, 2)
+ax[0].set_title("Before directional\nde-biasing")
+(tbc_dem_sumsin - ref_dem).plot(cmap='RdYlBu', vmin=-30, vmax=30, ax=ax[0])
+ax[1].set_title("After directional\nde-biasing")
```
adehecq, line 257: The example plotted in the doc is not working... The bias is still very visible.

adehecq: Actually, the bias seem to be doubled.

doc/source/biascorr.md
```diff
@@ -15,146 +17,420 @@ kernelspec:
...
+### Terrain biases
 
 {class}`xdem.coreg.TerrainBias`
 
-- **Performs:** Correct biases along a terrain attribute of the DEM.
-- **Supports weights** Yes.
-- **Recommended for:** Different native resolution between DEMs.
+- **Performs:** Correct biases along a terrain attribute.
+- **Supports weights:** Yes.
+- **Pros:** Useful to correct for instance curvature bias due to different native resolution between elevation data.
```
adehecq, line 314: in Arnaud's project we ended up calling it the "resolution bias" as curvature is not the best proxy

doc/source/biascorr.md
```diff
@@ -15,146 +17,420 @@ kernelspec:
...
+- **Cons:** For curvature biases, only works for elevation data with relatively close natire resolution.
```
adehecq, line 315: natire -> native

doc/source/biascorr.md
adehecq, line 1: It looks like Blockwise coregistration is not documented in this section. Only in teh API and in the example. Would it make sense to have a dedicated section on "highly non linear bias correction"?

doc/source/gapfill.md
```diff
@@ -0,0 +1,272 @@
...
+dem_1990 = dem_1990.crop(bounds)
+```
+
+We create a difference of DEMs object {class}`xdem.ddem.dDEM` to experiment on:
+
+```{code-cell} ipython3
+ddem = xdem.dDEM(raster=dem_2009 - dem_1990, start_time=dem_1990.datetime, end_time=dem_2009.datetime)
+
+# The example DEMs are void-free, so let's make some random voids.
+# Introduce 50000 nans randomly throughout the dDEM.
```
adehecq, line 82: actually it is not 50000 nans, but 1/5 of all pixels.

doc/source/gapfill.md
```diff
@@ -0,0 +1,272 @@
...
+mask = np.zeros_like(ddem.data, dtype=bool)
+mask.ravel()[(np.random.choice(ddem.data.size, int(ddem.data.size/5), replace=False))] = True
+ddem.set_mask(mask)
+```
+
+## Linear spatial interpolation
+
+Linear spatial interpolation (also often called bilinear interpolation) of dDEMs is arguably the simplest approach: voids are filled by an average of the surrounding pixels values, weighted by their distance to the void pixel.
```
adehecq, line 90: Actually it is not a bilinear interpolation, but an Inverse Distance Weighting interpolation. Bilinear would be to interpolate between 4 pixels. At least according to wikipedia: "Bilinear interpolation considers the closest 2 × 2 neighborhood of known pixel values surrounding the unknown pixel's computed location."

doc/source/gapfill.md
```diff
@@ -0,0 +1,272 @@
...
+ddem.plot(cmap='RdYlBu', vmin=-10, vmax=10, ax=ax[0])
+ax[1].set_title("After linear\ngap-filling")
+ddem_linear.plot(cmap='RdYlBu', vmin=-10, vmax=10, ax=ax[1], cbar_title="Elevation differences (m)")
+_ = ax[1].set_yticklabels([])
+```
+
+## Local hypsometric interpolation
+
+This approach assumes that there is a relationship between the elevation and the elevation change in the dDEM, which is often the case for glaciers.
+Elevation change gradients in late 1900s and 2000s on glaciers often have the signature of large melt in the lower parts, while the upper parts might be less negative, or even positive.
```
adehecq, line 116: I would replace "melt" by thinning, as we cannot know if the thinning is due to melt or dynamic.

doc/source/gapfill.md
```diff
@@ -0,0 +1,272 @@
...
+:mystnb:
+:  code_prompt_show: "Show plotting code"
+:  code_prompt_hide: "Hide plotting code"
+
+ddem_linear = ddem.copy(new_array=ddem_linear)
+
+# Plot before and after
+f, ax = plt.subplots(1, 2)
+ax[0].set_title("Before linear\ngap-filling")
+ddem.plot(cmap='RdYlBu', vmin=-10, vmax=10, ax=ax[0])
```
adehecq, line 107: the color scale is very saturated. I would increase the vmin/vmax values.

doc/source/gapfill.md
```diff
@@ -0,0 +1,272 @@
...
+
+This approach assumes that there is a relationship between the elevation and the elevation change in the dDEM, which is often the case for glaciers.
+Elevation change gradients in late 1900s and 2000s on glaciers often have the signature of large melt in the lower parts, while the upper parts might be less negative, or even positive.
+This relationship is strongly correlated for a specific glacier, and weakly correlated on regional scales.
+With the local (glacier specific) hypsometric approach, elevation change gradients are estimated for each glacier separately.
+This is simply a linear or polynomial model estimated with the dDEM and a reference DEM.
+Then, voids are interpolated by replacing them with what "should be there" at that elevation, according to the model.
+
+```{code-cell} ipython3
+ddem_localhyps = ddem.interpolate(method="local_hypsometric", reference_elevation=dem_2009, mask=glaciers_1990)
```
adehecq, line 123: The API for the function `interpolate` is not available, so it is difficult to get detailed info on the arguments. I was actually wondering about the type of argument `mask` as it looks like you are passing a Vector and not a mask.
I know this module needs more work, but good to raise an issue if this is not already mentioned.
I was also thinking that in the same way you generalized the bias correction to take any predictor variable, we could generalize the hypsometric method to work with any 1, 2 or Nd variables.

doc/source/gapfill.md
```diff
@@ -0,0 +1,272 @@
...
+:mystnb:
+:  code_prompt_show: "Show plotting code"
+:  code_prompt_hide: "Hide plotting code"
+
+ddem_reghyps = ddem.copy(new_array=ddem_reghyps)
+
+# Plot before and after
+f, ax = plt.subplots(1, 2)
+ax[0].set_title("Before regional\nhypsometric\ngap-filling")
+ddem.plot(cmap='RdYlBu', vmin=-10, vmax=10, ax=ax[0])
```
adehecq, line 222: I could not find where the issue is, but on the figure "before gap-filling" there is no gap...

doc/source/uncertainty.md
```diff
@@ -12,369 +14,385 @@ kernelspec:
...
+and thus generally encompass methods described in other studies on the topic (e.g., [Anderson et al. (2019)](http://dx.doi.org/10.1002/esp.4551)).
+
+The tables below summarize the characteristics of these methods.
+
+### Estimating and modelling the structure of error
+
+Frequently, in spatial statistics, a single correlation range is considered ("basic" method below).
+However, elevation data often contains errors with correlation ranges spanning different orders of magnitude.
+For this, [Rolstad et al. (2009)](http://dx.doi.org/10.3189/002214309789470950) and
+[Hugonnet et al. (2022)](http://dx.doi.org/10.1109/JSTARS.2022.3188922) considers
```
adehecq, line 112: considers -> consider

doc/source/uncertainty.md
```diff
@@ -12,369 +14,385 @@ kernelspec:
...
 # Calculate the area-averaged uncertainty with these models
-neff = xdem.spatialstats.number_effective_samples(area=1000, params_variogram_model=params_variogram_model)
+neff = xdem.spatialstats.number_effective_samples(area=outline_brom, params_variogram_model=params_variogram_model)
 ```
 
-TODO: Add this section based on Rolstad et al. (2009), Hugonnet et al. (in prep)
+**Step 3: Derive final error**
 
-### Propagation of correlated errors
+And can now compute our final random error for the mean elevation change in this area of interest:
```
adehecq, line 376: "and we can" -> "We can now"

doc/source/code/intricacies_datatypes.py
```diff
@@ -0,0 +1,58 @@
...
+coords = ref_dem.coords(grid=False)
+cont = ax[1, 0].contour(
+    np.flip(coords[0]), coords[1], np.flip(ref_dem.get_nanarray()), levels=15, cmap="terrain", vmin=280, vmax=420
+)
+plt.colorbar(mappable=cont, ax=ax[1, 0], label="Elevation (m)", pad=0.02)
+ax[1, 0].set_xticklabels([])
+ax[1, 0].set_yticklabels([])
+ax[1, 0].set_aspect("equal")
+
+plt.title("Types of elevation data")
```
adehecq, line 55: it looks like this line is overwriting the "point cloud" title. Maybe you need to use "suptitle"?

doc/source/elevation_intricacies.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+```{eval-rst}
+.. plot:: code/intricacies_datatypes.py
+    :width: 90%
+```
+
+
+Additionally, there are a critical differences for elevation point clouds depending on point density:
+- **Sparse elevation point clouds** (e.g., altimetry) are generally be stored as small vector-type datasets (e.g., SHP). Due to their sparsity, for subsequent analysis, they are rarely gridded into a DEM, and instead compared with DEMs at the point cloud coordinates by interpolation of the DEM,
```
adehecq, line 31: "be stored" -> "stored"

doc/source/robust_estimators.md
```diff
@@ -1,17 +1,21 @@
...
 
-Digital elevation models often contain outliers that can be traced back to instrument acquisition or processing artefacts, and which hamper further analysis.
+Elevation data often contain outliers that can be traced back to instrument acquisition or processing artefacts, and which hamper further analysis.
 
-In order to mitigate their effect, xDEM integrates [robust statistics](https://en.wikipedia.org/wiki/Robust_statistics) at different levels:
-- Robust optimizers for the fitting of parametric models during {ref}`coregistration` and {ref}`biascorr`,
-- Robust measures for the central tendency (e.g., mean) and dispersion (e.g., standard deviation), to evaluate DEM quality and converge during {ref}`coregistration`,
-- Robust measures for estimating spatial autocorrelation for uncertainty analysis in {ref}`spatialstats`.
+In order to mitigate their effect, the analysis of elevation data can integrate [robust statistics](https://en.wikipedia.org/wiki/Robust_statistics) at different levels:
+- **Robust estimators for the central tendency (e.g., mean) and dispersion (e.g., standard deviation)**, to evaluate DEM quality and converge during {ref}`coregistration`,
```
adehecq, line 8: do you mean "median" and "NMAD" instead of mean and std?

adehecq: "converge" -> "convergence"? I am not sure if you mean "evaluate convergence" or something else

doc/source/robust_estimators.md
adehecq, line 1: pretty cool !

doc/source/terrain.md
erikmannerfelt: Uuuh what do you mean here @adehecq? A convex surface is summit-like while a concave surface is valley-like.

doc/source/quick_start.md
rhugonnet: Good catch. And opened https://github.com/GlacioHack/xdem/issues/635 for the vertical datum of the DEM.

doc/source/citation.md
rhugonnet: The link I put is the generic Zenodo link that works for all versions.
I wanted to add a comment about the option to point to a specific version, but we haven't done this so well until now... (We only have 0.0.2 and 0.0.18, which is a bit weird).

@erikmannerfelt Can you activate user sharing on the Zenodo for @adehecq and me (https://help.zenodo.org/docs/share/user-sharing/), so that any one of us can update the versioning when others are too busy. Or maybe we can even set-up auto-updates on version + authors? I've seen several packages do it.

doc/source/citation.md
rhugonnet: Yes! :slightly_smiling_face:

rhugonnet: I thought exactly the same when redacting it.

doc/source/citation.md
rhugonnet: Good point. Indeed, Anthony is binning per elevation band in his study. I trust @iamdonovan will have done the work on this :stuck_out_tongue:  
We can also check quickly with Anthony: @aaarendt Do you remember if there was another study before your 2002 paper doing "elevation change binning per elevation band" to compute glacier volume change (what we now call hypsometric gap-filling)?

doc/source/dem_class.md
rhugonnet: Corrected, the reference is only for the CRS

doc/source/dem_class.md
rhugonnet: :wink:

doc/source/dem_class.md
rhugonnet: Done!

doc/source/terrain.md
rhugonnet: Thank god you commented here, I forgot to remove the RichDEM support since #558.
Rest done!

doc/source/terrain.md
rhugonnet: Perfect. I got exactly this issue the other day.

doc/source/terrain.md
rhugonnet: Added

doc/source/terrain.md
rhugonnet: Thanks @erikmannerfelt, I also find several sources backing the current statements!
https://www.britannica.com/science/valley/Valley-cross-profiles
https://avalanche.ca/glossary/terms/convex-slope
Leaving as is :wink:

doc/source/terrain.md
rhugonnet: Double-checked now too!

doc/source/coregistration.md
rhugonnet: This doesn't seem to have been limiting to users (and `workflows` is not in the API yet). Maybe we should discuss on the organization of this separately once you're back, also in relation to the CLI for coregistration.

doc/source/coregistration.md
rhugonnet: Nice catch! :sweat_smile:

doc/source/biascorr.md
rhugonnet: Yep, this is the ongoing issue described in the PR comment above that I can't seem to fix (it either corrects perfectly, or not at all... with the same `random_state`). :cry:

doc/source/biascorr.md
rhugonnet: Good point. Hard to change the name here as I add a bias based on curvature in the example below, we can adjust once we have the TPI implementation :wink:

doc/source/biascorr.md
rhugonnet: Nice catch

doc/source/biascorr.md
rhugonnet: The section is already written but not added for now, it's copied in #584 in the meantime

rhugonnet: Because of the errors still not fixed

doc/source/code/intricacies_datatypes.py
rhugonnet: Nice catch, yes!

doc/source/robust_estimators.md
rhugonnet: It's true it was a bit convoluted, reworded!

doc/source/gapfill.md
rhugonnet: That's an important detail! Changing everywhere.

doc/source/gapfill.md
rhugonnet: As it's glacier-focused, I think we need to discuss removing it altogether in its current state, including potentially the `dDEM` class. I've already added warning everywhere (API, pages and gallery examples) that these might be removed/re-factored.
Opening #636

doc/source/gapfill.md
```diff
@@ -0,0 +1,272 @@
...
+plt.show()
+```
+
+*Caption: Hypsometric elevation change of Scott Turnerbreen on Svalbard from 1990--2009. The width of the bars indicate the standard deviation of the bin. The light blue background bars show the area distribution with elevation.*
+
+## Regional hypsometric interpolation
+
+Similarly to local hypsometric interpolation, the elevation change is assumed to be largely elevation-dependent.
+With the regional approach (often also called "global"), elevation change gradients are estimated for all glaciers in an entire region, instead of estimating one by one.
+This is advantageous in respect to areas where voids are frequent, as not even a single dDEM value has to exist on a glacier in order to reconstruct it.
```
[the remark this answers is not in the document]
rhugonnet: Removed

doc/source/gapfill.md
rhugonnet: Fixed

doc/source/citation.md
rhugonnet: My bad! I forgot you added us, and we can edit when logged in Zenodo with our ORCID. All good!

doc/source/vertical_ref.md
rhugonnet: A future improvement that would fit better on the guide pages, the feature pages should not get too much into details.

rhugonnet: :wink:

xdem/coreg/affine.py
rhugonnet: For the record: translation is OK with scaling, but the rotation is affected. Opening an issue to add this in the future.

rhugonnet:
Accounted for all comments, opened issues otherwise. All tests passing.
Merging!

vschaffn:
@rhugonnet It's probably not deliberate, but the logging.md doc has been removed from your updates, and the logging_configuration.py example has been renamed plot_logging_configuration.py

rhugonnet:
In short: None of the two pages were rendering after #610. For instance, renaming into "plot_xxx" is required for it to be rendered by Sphinx-Gallery, otherwise it is shown without execution of the code.
In the comment above I explain how you can check the documentation rendering in the future with your own RTD fork setup :slightly_smiling_face:

vschaffn:
@rhugonnet Sorry, I had forgotten about that comment, thanks for the reminder 😃

PR #511, opened 2024-04-27
Remove legacy NumPy random number generator

### LINKED ISSUE #510: Remove legacy NumPy random number generator
Mirrors https://github.com/GlacioHack/geoutils/issues/536

rhugonnet:
This PR converts all legacy calls `np.random.seed(seed=)` into `np.random.default_rng(seed=)`.
It also removes the `random_state_definition` function that was essentially useless (`np.random.default_rng()` also accepts None, int or another `np.random.Generator` and returns the right result).

We stop the support for `np.random.RandomState` as input, which only provided access to specific legacy generator seeds, so was essentially useless for our applications here (we're already satisfied passing an integer as seed, or a specific random generator from `np.random.Generator`, we don't need any kind of repeatability with old-schools generators).
And renames `rnd` into `rng` everywhere for consistency.

Resolves #510


rhugonnet:
Requires a new release of GeoUtils to have a consistent subsampling output. On the way...

PR #517, opened 2024-05-06
Remove `Bias1D/2D/ND` classes in favor of an `nd` metadata attribute

### LINKED ISSUE #514: Remove classes `Bias1D`, `Bias2D` in favor of a `nd` argument defined on instantiation?
These intermediate classes seem to have no real purpose except checking the input data dimension, which could be done by simply defining an `nd` attribute in the `._meta` of each method, during `__init__()`. 

It would simply the inheritance structure to present in the documentation as well.

rhugonnet:
This PR removes the `BiasCorr1D/2D/ND` classes that were effectively useless except for checking the number of input variables, now moved to the main `BiasCorr` class, with tests moved as well.

Resolves #514

Needed for #502, will merge pre-review now as it's a minimal change.


adehecq:
Nice simplification !

PR #525, opened 2024-05-21
Dask for coregistration

ameliefroessl:
Use dask to improve the memory usage of the `BiasCorr` methods.  

Remaining todo item: 
- [x] Map remaining logic from `_postprocess_coreg_apply_rst()` to `_postprocess_coreg_apply_xarray`
- [x] `bias_vars` mask that you mentioned https://github.com/GlacioHack/xdem/pull/525#discussion_r1608774867
- [ ] Adapt other methods 
    - [ ]  Method ... 
- [x] Generic method for  map_blocks in the fit function
- [x] Create unit tests for the introduced functions
- [x] Fix unit tests
    - [x] Update the examples files to use the new COG files
    - [x] Create a mask file on the fly vs adding a new file to the example data?
    - [x] Once #531 merged to upstream, rebase 
- [x] Save output to file that works out of memory
- [x] Update geoutils dependency to 0.1.6
- [x] Rebase to xDEM main
- [ ] Add documentation on how to load the inputs such that the data is processed with dask. (`rioxarray` + `chunks`)

### DIFF
--- a/xdem/coreg/biascorr.py
+++ b/xdem/coreg/biascorr.py
@@ -241,6 +241,19 @@ class BiasCorr(Coreg):
             else:
                 bin_sizes = self._meta["bin_sizes"]
 
+        if isinstance(diff, np.ndarray):
+            ydata = diff[subsample_mask]
+            xdata = [var[subsample_mask] for var in bias_vars.values()]
+            sigma = weights[subsample_mask] if weights is not None else None
+
+        elif isinstance(diff, da.Array):
+            ydata = diff.vindex[subsample_mask].compute()  # type:ignore [assignment]
+            xdata = [var.vindex[subsample_mask].compute() for var in bias_vars.values()]
+            # TODO - where do the weights come from? Are they also dask arrays?
+            sigma = weights.vindex[subsample_mask].compute() if weights is not None else None
+        else:
+            raise TypeError(f"Incompatible input type for arrays {type(diff)}.")
+
         # Option 1: Run fit and save optimized function parameters
         if self._fit_or_bin == "fit":
 
@@ -251,24 +264,12 @@ class BiasCorr(Coreg):
                     "with function {}.".format(", ".join(list(bias_vars.keys())), self._meta["fit_func"].__name__)
                 )
 
-            if isinstance(diff, np.ndarray):
-                ydata = diff[subsample_mask].flatten()
-                xdata = np.array([var[subsample_mask].flatten() for var in bias_vars.values()]).squeeze()
-                sigma = weights[subsample_mask].flatten() if weights is not None else None
-            elif isinstance(diff, da.Array):
-                ydata = diff.vindex[subsample_mask].flatten().compute()  # type:ignore [assignment]
-                xdata = np.array([var.vindex[subsample_mask].flatten().compute() for var in bias_vars.values()])
-                # TODO - where do the weights come from? Are they also dask arrays?
-                sigma = weights.vindex[subsample_mask].flatten() if weights is not None else None
-            else:
-                raise TypeError(f"Incompatible input type for arrays {type(diff)}.")
-
             # we dont need to call the fit_chunked here because the data going in is not a chunked dask array.
             results = self._meta["fit_optimizer"](
                 f=self._meta["fit_func"],
-                xdata=xdata,
-                ydata=ydata,
-                sigma=sigma,
+                ydata=ydata.flatten(),
+                xdata=np.array([data.flatten() for data in xdata]),
+                sigma=sigma.flatten() if sigma is not None else None,
                 absolute_sigma=True,
                 **kwargs,
             )
@@ -283,8 +284,8 @@ class BiasCorr(Coreg):
                 )
 
             df = xdem.spatialstats.nd_binning(
-                values=diff[subsample_mask],
-                list_var=[var[subsample_mask] for var in bias_vars.values()],
+                values=ydata,
+                list_var=xdata,
                 list_var_names=list(bias_vars.keys()),
                 list_var_bins=bin_sizes,
                 statistics=(self._meta["bin_statistic"], "count"),
@@ -305,8 +306,8 @@ class BiasCorr(Coreg):
                 )
 
             df = xdem.spatialstats.nd_binning(
-                values=diff[subsample_mask],
-                list_var=[var[subsample_mask] for var in bias_vars.values()],
+                values=ydata,
+                list_var=xdata,
                 list_var_names=list(bias_vars.keys()),
                 list_var_bins=bin_sizes,
                 statistics=(self._meta["bin_statistic"], "count"),
--- a/xdem/coreg/base.py
+++ b/xdem/coreg/base.py
@@ -64,7 +64,6 @@ except ImportError:
 
 
 import dask.array as da
-
 from xarray.core.dataarray import DataArray
 
 ###########################################
@@ -437,7 +436,6 @@ def _preprocess_coreg_fit_xarray_xarray(
     )
 
     # TODO handle mask has no inliers -> np.all(~mask)
-
     # outputs are dask arrays
     return reference_dem.data, dem_to_be_aligned_reprojected, inlier_mask_all, transform, crs
 


xdem/coreg/biascorr.py
```diff
@@ -219,11 +234,26 @@ def _fit_biascorr(  # type: ignore
...
 
+            if isinstance(diff, np.ndarray):
+                ydata = diff[subsample_mask].flatten()
+                xdata = np.array([var[subsample_mask].flatten() for var in bias_vars.values()]).squeeze()
+                # TODO - there is a bug here still
+                # sigma = (weights[subsample_mask].flatten() if weights is not None else None,)
+                sigma = None
+            elif isinstance(diff, da.Array):
+                ydata = diff.vindex[subsample_mask].flatten().compute()  # type:ignore [assignment]
+                xdata = [var.vindex[subsample_mask].flatten() for var in bias_vars.values()]  # type:ignore [assignment]
```
rhugonnet, line 245: Don't you need a `.compute()` here as well?

rhugonnet: Yes, even if the `bias_vars` are not delayed (no operation since their creation/passing in the `_fit_rst_rst`) function, the compute seems needed, otherwise you get a dask array output that won't be understood by the following steps:
```python
da.ones((10000, 10000)).vindex[[0, 1], [0, 1]]
Out[12]: dask.array<vindex-merge, shape=(2,), dtype=float64, chunksize=(2,), chunktype=numpy.ndarray>
da.ones((10000, 10000)).vindex[[0, 1], [0, 1]].compute()
Out[13]: array([1., 1.])
```

xdem/coreg/biascorr.py
```diff
@@ -1077,7 +1107,10 @@ def _fit_rst_rst(  # type: ignore
         p0 = np.ones(shape=((self._meta["poly_order"] + 1) ** 2))
 
         # Coordinates (we don't need the actual ones, just array coordinates)
-        xx, yy = np.meshgrid(np.arange(0, ref_elev.shape[1]), np.arange(0, ref_elev.shape[0]))
+        if type(ref_elev) == da.Array:
+            xx, yy = da.meshgrid(da.arange(0, ref_elev.shape[1]), da.arange(0, ref_elev.shape[0]))
```
rhugonnet, line 1111: We should define the `chunks` here to ensure the auto ones are not completely irrelevant and result in a lot of memory usage. Maybe simply copying the chunks of the reference DEM?

xdem/coreg/biascorr.py
```diff
@@ -176,17 +179,29 @@ def _fit_biascorr(  # type: ignore
         # Compute difference and mask of valid data
         # TODO: Move the check up to Coreg.fit()?
 
-        diff = ref_elev - tba_elev
-        valid_mask = np.logical_and.reduce(
-            (inlier_mask, np.isfinite(diff), *(np.isfinite(var) for var in bias_vars.values()))
-        )
+        if all(isinstance(dem, (da.Array, Delayed)) for dem in (ref_elev, tba_elev)):
+            diff = da.subtract(ref_elev, tba_elev)
+
```
rhugonnet, line 184: We also need to compute the `valid_mask` here for the delayed inputs to ensure we don't introduce outliers from the `ref_elev` or `tba_elev`, same as below:
```python
            valid_mask = np.logical_and.reduce(
                (inlier_mask, np.isfinite(diff), *(np.isfinite(var) for var in bias_vars.values()))
            )
```
Not sure if this is supported directly, or if we need to use `map_blocks` for it...

rhugonnet: I guess `map_blocks` the same way you used it further up would work perfectly :)

xdem/coreg/biascorr.py
ameliefroessl: thank you! nice catch :) - changed it!

xdem/coreg/biascorr.py
ameliefroessl: So the reason I have not calculated this is because I calculate the inlier_mask from all of the inputs [here](https://github.com/GlacioHack/xdem/pull/525/files#diff-1430c7a606ab459441322127ec6b1c671169f487050d53481cc23723099bc834R344) . Taking also the non-valid values into account from the `ref_elev` and `tba_elev`. I think this would cover that? Or did I miss something?

xdem/coreg/biascorr.py
ameliefroessl: So it seems like chunks is not a supported argument to `da.meshgrid` unfortunately. i tried rechunking it but it doesn't seem to make that big of a difference computation-wise :thinking:

xdem/coreg/biascorr.py
rhugonnet: It's for the `bias_vars` (that can have different non-finite values that will propagate during the fit), not covered by ` _get_valid_data_mask`! 
And actually this will also be relevant in affine functions (for instance for slope/aspect in `NuthKaab` that propagate NaNs of the DEM)

xdem/coreg/biascorr.py
rhugonnet: Strange indeed, it seems important... We can check if that matters later with a memory usage-specific test!

ameliefroessl: yes i agree! I'll add a note so we remember to have a closer look at it.

xdem/coreg/biascorr.py
ameliefroessl: ahhh ok ok. I missed that thank you! :+1: ill make sure to add it :slightly_smiling_face:

xdem/coreg/biascorr.py
rhugonnet: Amazing!

I still found it weird that they didn't ask a chunk argument directly in Dask, and found this interesting discussion:
https://github.com/dask/dask/pull/2943#issuecomment-350478499.

If I understood correctly, doing `da.meshgrid(da.arange(, chunks=), da.arange(, chunks=)` should have the same behaviour as chunking! (or even more optimized)

ameliefroessl: yes! Adding the chunks to the input `da.arange()` makes it possible to chunk the output array. However, I didn't find a way to combine it with delayed. Meaning even if it's chunked I think it loads the entire array into memory. Instead of "creating"? / "reading" ? that part only when needed in the processing. But I'd love to be proven wrong :smiley:

xdem/coreg/biascorr.py
ameliefroessl: this shouldn't have an effect on the `Deramp` algorithm though since the `bias_vars` is created from the input coordinates of the `reference_elevation` input correct? I'm just asking because I'm running into some unexpected outputs when I have nans in the original input data.

rhugonnet: Yes it shouldn't have any effect in this case, the meshgrid should be always a full array :thinking:

ameliefroessl: Ok yes that makes sense. I did end up figuring out what the issue was with the nan data. I was calculating the mask on the output from the `delayed_reproject` ([here](https://github.com/GlacioHack/xdem/pull/525/commits/32eb94b31acafa38855f6dcf8fefe558ecb4e6a9) is the change). I'm not 100% sure, but my assumption is that there is some smoothing/ resampling happening and i guess the nodata values accidentally get smeared into neighboring pixels?... Meaning it was potentially sampling some very high values and the estimated function was pretty off. This was my assumption after looking at the input y values going into the fit function.

xdem/coreg/biascorr.py
ameliefroessl: I don't know why exactly but when I use this version of meshgrid the entire application just hangs without doing anything... maybe i'll investigate further but currently I can't seem to figure out what's the issue :thinking:

xdem/coreg/biascorr.py
```diff
@@ -230,11 +251,24 @@ def _fit_biascorr(  # type: ignore
                     "with function {}.".format(", ".join(list(bias_vars.keys())), self._meta["fit_func"].__name__)
                 )
 
+            if isinstance(diff, np.ndarray):
```
rhugonnet, line 254: Here is the data subsampling that can be moved out of the if loop "Option 1: Run fit...", then `xdata`/`ydata`/etc can be passed to every option similarly :slightly_smiling_face:

xdem/coreg/biascorr.py
```diff
@@ -460,7 +494,12 @@ def _apply_rst(  # type: ignore
 
         # Apply function to get correction (including if binning was done before)
         if self._fit_or_bin in ["fit", "bin_and_fit"]:
-            corr = self._meta["fit_func"](tuple(bias_vars.values()), *self._meta["fit_params"])
+            if isinstance(list(bias_vars.values())[0], da.Array):
+                corr = fit_chunked(
```
rhugonnet, line 498: You should be able to do almost exactly the same right below, replacing `self._meta["fit_func"]` for the line `corr = bin_interpolator(*args)`, and the line `corr = get_perbin_nd_binning(*args)`!

xdem/coreg/biascorr.py
ameliefroessl: I guess [this](https://github.com/GlacioHack/xdem/pull/525/commits/1c57a389b43bb252eedbf7f92f4164d6b9e443cc) is what you meant? :slightly_smiling_face:

ameliefroessl: which of the unit test would you expect to pass if passed an xarray instead of the normal input?

rhugonnet: Yes exactly!
I guess for tests with only `fit()`, everything should pass exactly the same with just the Xarray input.
For tests that also have `apply()`, you'd have to write a check on the input type and call `.compute()` to be able to run the final checks on the output arrays.

rhugonnet: So I think you should be able to pass everything by just adding an Xarray input in the current list: https://github.com/ameliefroessl/xdem/blob/1c57a389b43bb252eedbf7f92f4164d6b9e443cc/tests/test_coreg/test_biascorr.py#L97
And adding a `compute()` only for tests using the output of `apply()` (the ones just calling it don't need to), like here: https://github.com/ameliefroessl/xdem/blob/1c57a389b43bb252eedbf7f92f4164d6b9e443cc/tests/test_coreg/test_biascorr.py#L520, and other synthetic tests further below.

PR #526, opened 2024-05-23
Make coregistration metadata consistent and public

rhugonnet:
This PR makes the `.meta` attribute a public property of `Coreg`, and makes the attribute names consistent.

In particular:
- `"offset_east_px"` is now `"shift_x"`, and is in georeferenced units (instead of pixels) which allows to remove the `"resolution"` key in the metadata,
- `"offset_north_px"` is now `"shift_y"`,
- `"vshift"` is now `"shift_z"`,
Others are renamed to be consistent with `BiasCorr`.

Will be documented in #502!

### DIFF
--- a/xdem/coreg/affine.py
+++ b/xdem/coreg/affine.py
@@ -412,7 +412,8 @@ class ICP(AffineCoreg):
     "centroid". The translation parameters are also stored individually in the keys "shift_x", "shift_y" and "shift_z"
     (in georeferenced units for horizontal shifts, and unit of the elevation dataset inputs for the vertical shift).
 
-    Requires 'opencv'. See opencv doc for more info: https://docs.opencv.org/master/dc/d9b/classcv_1_1ppf__match__3d_1_1ICP.html
+    Requires 'opencv'. See opencv doc for more info:
+    https://docs.opencv.org/master/dc/d9b/classcv_1_1ppf__match__3d_1_1ICP.html
     """
 
     def __init__(
--- a/xdem/coreg/affine.py
+++ b/xdem/coreg/affine.py
@@ -1032,7 +1032,9 @@ projected CRS. First, reproject your DEMs in a local projected CRS, e.g. UTM, an
 
         applied_epc = gpd.GeoDataFrame(
             geometry=gpd.points_from_xy(
-                x=elev.geometry.x.values + self._meta["shift_x"], y=elev.geometry.y.values + self._meta["shift_y"], crs=elev.crs
+                x=elev.geometry.x.values + self._meta["shift_x"],
+                y=elev.geometry.y.values + self._meta["shift_y"],
+                crs=elev.crs,
             ),
             data={z_name: elev[z_name].values + self._meta["shift_z"]},
         )
--- a/xdem/coreg/affine.py
+++ b/xdem/coreg/affine.py
@@ -226,14 +226,14 @@ class AffineCoreg(Coreg):
         super().__init__(meta=meta)
 
         # Define subsample size
-        self._meta["subsample"] = subsample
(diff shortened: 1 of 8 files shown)


rhugonnet:
@adehecq @erikmannerfelt  Just want to quickly check that you agree with the new names before merging. Also, are we sure about `meta`? Fine with me, but could use something else if you have ideas.

rhugonnet:
Merging to document in #502, can still make changes on your comments there before releasing

xdem/coreg/affine.py
```diff
@@ -385,27 +390,30 @@ def _apply_pts(
...
     """
-    Iterative Closest Point DEM coregistration.
-    Based on 3D registration of Besl and McKay (1992), https://doi.org/10.1117/12.57955.
+    Iterative closest point registration, based on Besl and McKay (1992), https://doi.org/10.1117/12.57955.
+
+    Estimates a rigid transform (rotation + translation) between two elevation datasets.
 
-    Estimates a rigid transform (rotation + translation) between two DEMs.
+    The transform is stored in the `self.meta` key "matrix", with rotation centered on the coordinates in the key
+    "centroid". The translation parameters are also stored individually in the keys "shift_x", "shift_y" and "shift_z"
```
adehecq, line 412: But these may be misleading, if there is a rotation, no? Should there also be an item "rotation" in the meta?

xdem/coreg/affine.py
```diff
@@ -677,16 +693,20 @@ def _to_matrix_func(self) -> NDArrayf:
...
 
         return empty_matrix
 
 
 class NuthKaab(AffineCoreg):
     """
-    Nuth and Kääb (2011) DEM coregistration: iterative registration of horizontal and vertical shift using slope/aspect.
+    Nuth and Kääb (2011) coregistration, https://doi.org/10.5194/tc-5-271-2011.
+
+    Estimate horizontal and vertical translations by iterative slope/aspect alignment.
```
adehecq, line 705: I would write "Estimate horizontal translations iteratively using a relationship between elevation difference, slope and aspect".
Also one thing that we need to discuss/clarify. The Nuth & Kaab (2011) papers suggest to correct the vertical misalignment using the parameters of the sinusoid fit, which we don't. We use a median of the elevation difference. We should maybe add this to the description. And do we want to make the reduction function optional or we assume median is the best? The latter would be easier and is probably fine. But this means that there is no need to follow the NuthKaab method with a VerticalShift.

adehecq:
> @adehecq @erikmannerfelt Just want to quickly check that you agree with the new names before merging. Also, are we sure about `meta`? Fine with me, but could use something else if you have ideas.

I'm fine with the names used. But before it was clear that the unit of offsets was pixels, because it was in the name, now it's not as clear. I'm wondering if we should have a description somewhere of the possible items in "meta". Would it make sense to have a panda's dataframe instead with a column containing the value and another containing a description? 
I don't know if "meta" is the clearest name... By itself it means "beyond, above, transcending". Here I assume it was meant as a short name for metadata. I can't find a good alternative. We could use "params" but it is ambiguous with the parameters of the function. Or "results", but could be ambiguous with the coregistered DEM? Maybe "transformation", to indicate this is the applied transformation? I don't have a definite answer here sorry...

xdem/coreg/affine.py
rhugonnet: A rotation here is quite complex (3D) and already contained in the affine matrix. Maybe we can have a note pointing at `pytransform3d` that has routines to convert the affine matrix into its rotation/translation/scaling components? (it's really not trivial)
We could also include the components as `rotation_x`, `rotation_y`, `rotation_z`.
Or just do both!

PR #527, opened 2024-05-23
Add `Coreg.fit_and_apply()` to run everything in one step

### LINKED ISSUE #513: Add function `Coreg.run()` or `Coreg.fit_and_apply()` to run both steps simultaneously
As we discussed several times before, but we did not have a separate issue for it. 
Quite an important though easy change: it'll minimize user errors (inverting order of input parameters, or passing different georeferencing info to both), and also make it easier to consistently write plotting functions!

rhugonnet:
Resolves #513

### DIFF
--- a/tests/test_fit.py
+++ b/tests/test_fit.py
@@ -25,7 +25,7 @@ class TestRobustFitting:
     def test_robust_norder_polynomial_fit(self, pkg_estimator: str) -> None:
 
         # Define x vector
-        x = np.linspace(1, 50, 1000)
+        x = np.linspace(-50, 50, 10000)
         # Define exact polynomial
         true_coefs = [-100, 5, 3, 2]
         y = np.polyval(np.flip(true_coefs), x).astype(np.float32)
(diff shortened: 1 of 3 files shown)


tests/test_fit.py
```diff
@@ -25,14 +25,15 @@ class TestRobustFitting:
     def test_robust_norder_polynomial_fit(self, pkg_estimator: str) -> None:
 
         # Define x vector
-        x = np.linspace(1, 10, 1000)
+        x = np.linspace(-50, 50, 10000)
```
rhugonnet, line 28: This is to reduce random errors triggered on Mac sometimes (strange because `random_state` is fixed... Couldn't find a logical explanation).

xdem/coreg/base.py
```diff
@@ -1351,6 +1351,140 @@ def apply(
         else:
             return applied_elev, out_transform
 
+    @overload
+    def fit_and_apply(
```
adehecq, line 1355: Simple ! 
I think 99% of the users will only need to use that function. Maybe we could just add a note in the other functions `fit` and `apply` that this one should be used in priority. The only reasons I can see to split the two is 1) if making changes to the coregistration or 2) to run the same coregistration to several DEMs. Do you see any other reason? 
It would almost make sense to hide the two step functions...

adehecq: I added an issue to make sure we update the examples in the doc to encourage using this function: https://github.com/GlacioHack/xdem/issues/532

Issue #139, opened 2021-06-03
Enable optional plots for coregistration functions


adehecq:
As mentioned in #135, it is useful for understanding what the script does and for debugging, to plot figures of intermediate steps.
See for example how this was done in old geoutils: https://github.com/GeoUtils/geoutils/blob/5c41b2a343d1f0fc161eed4426154656c06841a4/geoutils/dem_coregistration.py#L126

PR #530, opened 2024-05-24
Re-structure coregistration affine methods, point-raster pre-processing, binning and fitting, and add pixel interpretation support

### LINKED ISSUE #234: Speed-up CoregPipeline fit function
For speed up, this line should not be called on the last iteration, since tba_dem_mod is not used afterwards.
https://github.com/GlacioHack/xdem/blob/2564f5f1c4654f1b153a88193851f40a55f09226/xdem/coreg.py#L1140

### LINKED ISSUE #327: Re-organization of `coreg.py`

### LINKED ISSUE #485: Improve tests in `test_affine` once core class restructured with `BiasCorr` parameters
The logic in `test_biascorr` can be copied to test all types of inputs (raster-raster, point-raster, point-point), but for efficiency we should only do this once the function class parameters are fixed (fit optimizer, binning, etc) to avoid rewriting the tests twice. See #435.

### LINKED ISSUE #546: Bug with raster-point coregistration
Hi,
I was about to co register some IceSat2 data with a reference raster using the NuthKaab Method. When I used the apply method of the co-registration object it seems to shift the points in the opposite direction. 
Syntax I used looks like this:
```python
    NuthKaab=xdem.coreg.NuthKaab()
    coreg_result=NuthKaab.fit(ref_dem, points, z_name ='z', verbose=True)
    points2=NuthKaab.apply(points)
```

Unfortunately, test_coreg_example_shift in test_affine .py does not capture this error, so edited it a little to make the bug reproducible:
```python
    def test_coreg_point_shift(self, shift_px, z_shift=2, coreg_class=coreg.NuthKaab, verbose=False, subsample=5000):
        """
        For comparison of coreg algorithms:
        Shift a ref_dem on purpose, e.g. shift_px = (1,1), and then applying coreg to shift it back.
        """
        res = self.ref.res[0]

        # shift DEM by shift_px
        shifted_ref = self.ref.copy()
        shifted_ref.shift(shift_px[0] * res, shift_px[1] * res, inplace=True)

        shifted_points = shifted_ref.to_pointcloud(
            subsample=subsample, force_pixel_offset="center", random_state=42
        ).ds
        points_on_ref = self.ref.to_pointcloud(
            subsample=subsample, force_pixel_offset="center", random_state=42
        ).ds
        
        shifted_points["E"] = shifted_points.geometry.x
        shifted_points["N"] = shifted_points.geometry.y
        shifted_points.rename(columns={"b1": "z"}, inplace=True)
        points_on_ref.rename(columns={"b1": "z"}, inplace=True)
        shifted_points["z"]=shifted_points["z"]+z_shift
        
        kwargs = {} if coreg_class.__name__ != "GradientDescending" else {"subsample": subsample}

        coreg_obj = coreg_class(**kwargs)

        coreg_obj.fit(self.ref, shifted_points, verbose=verbose, random_state=42)
        corrected_points = coreg_obj.apply(shifted_points)
            

        if ((corrected_points.geometry.x-points_on_ref.geometry.x)[0] == pytest.approx(0, abs=0.1)) and ((
                corrected_points.geometry.y-points_on_ref.geometry.y)[0] == pytest.approx(0, abs=0.1)) and ((
                    corrected_points['z']-points_on_ref['z'])[0]== pytest.approx(0, abs=0.1)):
            return
        east_diff = (corrected_points.geometry.x-points_on_ref.geometry.x)[0]
        north_diff = (corrected_points.geometry.y-points_on_ref.geometry.y)[0]
        print(f"Diffs are too big. east: {east_diff:.2f} m, north: {north_diff:.2f} m")
```

I traced the cause of this behavior to coreg/affine.py in the method : NuthKaab._fit_rst_pts.
In line 1014 to 1015 there is the following code:
```python
self._meta["shift_x"] = offset_east * resolution if ref == "point" else -offset_east 
self._meta["shift_y"] = offset_north * resolution if ref == "point" else -offset_north 
```

I don't see a reason why the shift for points is stored in pixels. From what I see, the if condition here is unnecessary and deleting it will let the test I created run smoothly. Also it is a little confusing with the signs. They actually should be there so it makes sense. But, since the signs are opposite in the _apply_pts function they don't need to be taken care off in _fit_rst_pts.
It was a bit hard to get my head around this and it would be probably better to have the apply method consistent for points and rasters as well as for height and lateral shift.

This could also concern other co registration methods, where the exact same lines are found.

I hope I didn't get anything wrong with the syntax and the behavior was intended as I assumed.

### LINKED ISSUE #547: Rename `Coreg.meta` into a clearer naming before documenting it
See https://github.com/GlacioHack/xdem/pull/526#issuecomment-2134784437

### LINKED ISSUE #561: Add a single consistent function for iterating methods to a tolerance
Addressing this in #530 right now.
Right now we use different code for `NuthKaab`, `GradientDescending`, and we will need this Python code for `ICP` (#483).

### LINKED ISSUE #571: Make mypy happy when initializing TypeDicts from coregistration methods
See for instance here: https://github.com/GlacioHack/xdem/blob/02c20268ecfececc2be75e07597ec66997902743/xdem/coreg/biascorr.py#L122
A lot of other instances added in #530.

The TypedDict are super useful to declare the expected type of keys and values, but initializing them without this error is a hassle. Maybe I'm simply doing something wrong...

I found some related posts, but nothing that helped me solving this issue:
https://stackoverflow.com/questions/73242556/typeddict-class-is-an-incompatible-type-for-function-expecting-dict
https://github.com/python/mypy/issues/13122

**An important note:**
Most users rely on the fact that TypedDict always have a fixed number of keys, but here we don't (argument `total=False` of `TypedDict`).

### LINKED ISSUE #573: Organize coregistration metadata
Several recent PRs including in particular #530 brought up the fact that we need better organization of coregistration metadata.
Currently, every input/output is saved in the same `Coreg.meta` dictionary, which is getting messy and hard to manipulate. Its current keys are summarized in the `CoregDict` typing class: https://github.com/GlacioHack/xdem/blob/02c20268ecfececc2be75e07597ec66997902743/xdem/coreg/base.py#L1157

Below I'm proposing a structure to try to remedy this, and bring more clarity (and practicality?) on the parameters for both users and developers.

First, we would have a nested structure for the `CoregDict`, which could be for instance:
```python
class CoregDict(TypedDict, total=False):

   # Inputs = parameters
    input_random: InRandomDict
    input_fitorbin: InFitOrBinDict
    input_iterative: InIterativeDict
    input_specific: InSpecificDict

    # Outputs
    output_random: OutRandomDict
    output_fitorbin: OutFitOrBinDict
    output_iterative: OutIterativeDict
    output_affine: OutAffineDict
```
where we'd have for instance:
```python
class InRandomDict(TypedDict, total=False):
    # Subsample size input by user, and final size available from data
    subsample: int | float
    # Random state (for subsampling, but also possibly for some fitting methods)
    random_state: int | np.random.Generator | None

class OutRandomDict(TypedDict, total=False):
   # Final subsample size used, depending on amount of valid data
    subsample_final: int
```
Similarly, the `FitOrBin` dictionaries would have user parameters (`fit_optimizer`, `bin_sizes`, etc) on input side, and outputs (`fit_params`, `bin_df`) on the other side.

The advantage of this for the developing side, despite the added structure, is that we can easily extract a group of parameter saved at different stages (usually initialization) for a specific function.
For example:
```python
class MyNewCoreg:
    def _fit_rst():
         xxx = self._subsampling(yyy)
    def _subsampling():
             inputs_random = self.meta["inputs_random"]  # This contains "subsample" and "random_state"
```
or for plotting in the future:
```python
class Coreg:
   def plot_fit():
        outputs_fitorbin = self.meta["outputs_fitorbin"] # Contains everything necessary to plot the fit/bin
```
The problem with the nested structure would be for users to navigate the nested dictionary... But this should become clear quickly with a simple print (and a nice guide in the documentation):
```python
print({"random": {"subsample": 1000, "random_state": 42}})
```
```
{'random': {'subsample': 1000, 'random_state': 42}}
```
Additionally, we could code a `Coreg.info()` method that summarizes everything nicely in words:
```python
Coreg.info()
```
```
Summary of coregistration metadata
Inputs:
    Randomization:
         Subsample size required by user: 1000
         Random state for randomization: 42
         ...
```

### LINKED ISSUE #574: Speed-up ICP-related tests
Related to #455.
Here's an overview of our test lengths that are longer than 10s:
```
83.76s call     tests/test_doc.py::TestDocs::test_build
49.62s call     tests/test_coreg/test_base.py::TestAffineManipulation::test_apply_matrix__raster_realdata
41.60s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[ICP-ICP]
26.23s call     tests/test_coreg/test_base.py::TestCoregClass::test_fit_and_apply[ICP]
22.69s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[DirectionalBias-ICP]
22.17s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[ICP-DirectionalBias]
21.69s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[ICP-NuthKaab]
21.44s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__biasvar[coreg2_init_kwargs1-ICP]
21.43s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[TerrainBias-ICP]
21.11s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[NuthKaab-ICP]
21.02s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[ICP-VerticalShift]
20.97s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[ICP-TerrainBias]
20.65s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[VerticalShift-ICP]
20.64s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[ICP-Deramp]
20.60s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__nobiasvar[Deramp-ICP]
20.55s call     tests/test_coreg/test_base.py::TestCoregPipeline::test_pipeline_combinations__biasvar[coreg2_init_kwargs0-ICP]
17.81s call     tests/test_coreg/test_base.py::TestCoregClass::test_subsample[ICP]
14.05s call     tests/test_coreg/test_affine.py::TestAffineCoreg::test_icp_opencv
11.94s call     tests/test_coreg/test_affine.py::TestAffineCoreg::test_gradientdescending
```
Almost all of them come from using ICP in a pipeline, or on its own.
We'd save 6min (half of test durations), by finding a way to speed up those. I'll look into it.

rhugonnet:
This PR restructures affine coregistration methods and many generic subfunctions to be consistent and modular. 
Primarily, it makes 1/ "point-raster" subsampling, 2/ any binning and/or fitting, and 3/ iteration over methods, the same across all `Affine` and `BiasCorr` methods with extensive modularity. 
Additionally, this PR updates horizontal shift reprojection from using a different implementation in every method (e.g. Nuth and Kääb used `scipy.RectBivariateSpline`) to using consistently `interp` from GeoUtils which supports any resampling method (for which details can be chosen in global `config` parameters), with consistent nodata propagation built on top, and preserves computational efficiency by allowing to return a single interpolator object re-used (https://github.com/GlacioHack/geoutils/pull/560). This also allows to avoid sub-pixel errors of Rasterio reprojection.
It also re-structures the code of affine methods, defining core functions outside the classes' `_fit_rst_rst` or `_fit_rst_pts`, which allows to have consistent pre-processing steps, and clear input/outputs.
Finally, this PR adds support for pixel interpretation for coregistration methods (important for point-raster operations).

Resolves #234
Resolves #574 
Resolves #561
Resolves #546
Resolves #573
Resolves #571 
Resolves #547 
Resolves #327 (except plot but those are mentioned in https://github.com/GlacioHack/xdem/issues/139)
Big step forward for #435

Opens #572
Opens #575 
Opens #579 

### 1/ New core processing functions

**For consistent binning/fitting**:
Restructures the `coreg/affine` module to support the `fit_or_bin` logic previously added in `coreg/biascorr`: However, as in the case of coregistration, the fit function is fixed (that of the method; for instance dh/slopetan = aspect for Nuth and Kääb), the "bin" option alone is impossible and only "fit" or "bin_and_fit" are supported. Therefore, the mirrored structure simply allows to pass a boolean argument `bin_before_fit` to perform binning before fitting, as well as any fit optimizer, bin statistic, and bin sizes.
To do this, this PR modifies the bin and fit of `BiasCorr` into a single `_bin_or_and_fit_nd` function in `coreg.base` used by all `Affine` and `BiasCorr` methods.

**For consistent point-raster/raster-raster subsampling:**
This pre-processing step is different from that running directly in `Coreg.fit()`, which simply extracts the `transform`, `crs`, and others. This step revolves around subsampling the point-raster (raster-raster, or point-raster) at the same points for a valid subsample (no NaNs) of a certain size. This cannot be done consistently in `fit()` before `_fit_xxx`, because auxiliary variables might need to be derived by the method on the full arrays (such as slope, aspect, etc), which propagate NaNs. In particular, the tricky aspect is the case of point-raster, where we need to first interpolate the raster at the coordinates of the points to get an idea of the available valid subsample (depending here also on NaN propagation).
To do this, this PR modifies the subsampling that existed in `BiasCorr` into a single `_preprocess_pts_rst_subsample` function in `coreg.base` relying on two steps: `_get_subsample_mask_pts_rst` and `_subsample_on_mask`. In order to allow to return a dh interpolator that can be re-used iteratively for some affine methods, another function `_preprocess_pts_rst_subsample_with_dhinterpolator` in `coreg.affine` relies on a different second step `_subsample_on_mask_with_dhinterpolator`. 

**For consistent iterative methods**:
Methods such as ICP, Nuth and Kääb, etc, all work based on iterations. To have consistent treatment of inputs/outputs and iterative arguments (tolerance to reach, maximum number of iterations), those are now grouped through an `_iterate_method` function.

**Ensuring consistent argument across these core functions and other parameters:**
To ensure the argument passed to these three core functions (subsampling, binning/fitting, and iteration) are consistent, new dictionary types are defined for each which defines their key names and value types: `InFitOrBinDict` ("In" for input), `InRandomDict` and `InIterativeDict`. Additionally `InAffineDict` for affine-related parameters, and `InSpecificDict` for any method-specific parameters that does not fit other categories.
Same is done for outputs: `OutRandomDict`, etc.
See #573 for details on the new organization.

### 2/ Robust and consistent horizontal shift reprojection
 
The core of `fit_` functions of horizontal methods such as `NuthKaab` and `GradientDescending` has been modify to rely on the `_preprocess_pts_rst_subsample_with_dhinterpolator` mentioned above.
This function relies on `geoutils.interp_points` (directly for point-raster, and through a new function `_reproject_horizontal_shift_samecrs` for raster-raster) to perform a consistent 2D grid interpolation with nodata propagation supporting any SciPy method, see details here: https://github.com/GlacioHack/geoutils/pull/560.
This PR adds a test checking that the result of this functionality matches GDAL reprojection, which ensures Rasterio sub-pixel reprojection errors are gone.

### 3/ Re-structuration of affine classes

All code from affine classes is moved out of the `_fit_xxx` functions into functions that rely on the same steps where possible.
For instance, the `NuthKaab` method is divided into a `_nuthkaab_fit_func` to define the function to optimize, a `_nuthkaab_fit` which wraps `_bin_or_and_fit_nd` with some initial guess, a `_nuthkaab_aux_vars` that derives auxiliary variables required for the fit (slope, aspect), and a `_nuthkaab_iteration_step` that is passed to `_iterate_method`. 

All the `_apply_xxx` affine functions are removed to avoid unnecessary redundancies/inconsistencies, as the `apply_matrix` is the same across all affine methods, and since #531 supports `resample=False` that modifies only the geotransform for horizontal shifts.

### 4/ Add support for pixel interpretation

Dealing with `area_or_point` was previously done arbitrarily in each subclass. Using the above interpolation/reprojection (through `geoutils.interp_points`), we can now simply pass the argument to GeoUtils who accounts for it when comparing point-raster, or raises a warning for raster-raster having a different interpretation.
It is only required for the `fit` part of coregistration, and thus added to all related functions.

### 5/ Removal of `Tilt`, and other subfunctions

With the above changes, the affine `Tilt` function corresponds exactly to `Deramp(order=1)` (instead of having its own, separate implementation). Additionally, it is not truly affine (cannot be modelled by an affine matrix), and thus has no place in affine methods. We can simply remove, and users can rely on `Deramp` instead.
All sub-functions that were created to perform some processing steps in coregistration modules (many times duplicated from another method, or from a GeoUtils functionality) for a specific affine method are now removed. This is largely thanks to https://github.com/GlacioHack/geoutils/pull/560 which re-structures GeoUtils raster functionalities to be called without having to build a Raster object (simply passing array, transform, CRS, etc).
Until now, we were deconstructing/reconstructing Raster objects everywhere to call simple functions (`interp_points`, `coords`, etc), which made us lose performance. This is now largely solved too!
A couple exceptions remains: `warp_dem` used only by `BlockwiseCoreg` and `_mask_as_array` used only by the `dDEM` class, but their use seems more specific and I'm not sure it is covered by anything in GeoUtils right now.

## Tests

I did not write new tests on purpose in this PR to ensure I could check that all these changes do not affect our current methods significantly. Improving tests (in particular for affine methods) will be addressed in a next PR.

And indeed, all tests are passing with minor modifications! :partying_face: 

Additionally, I increased the speed of the tests following #574 (ICP tests were extremely slow and cumulated to 5min+, or half of xDEM's test length). Now nearly all the tests of `test_coreg/` run on a cropped version of the reference/tba example DEMs.
**And all of xDEM's test run in 5min instead of 12min+.** :smiley:  

## To-do-list

- [x] Fix resolution error in `NuthKaab` of #546,
- [x] Finalize https://github.com/GlacioHack/geoutils/pull/560,
- [x] Finalize same-CRS SciPy interpolator reprojection to avoid Rasterio sub-pixel errors,
- [x] Move subsampling out of the "bin_and_fit" functionality,
- [x] Make "bin_and_or_fit" a generic function applicable to other functions (not only `BiasCorr` subclasses),
- [x] Split preprocessing into consistent functions: subsampling supporting raster-raster or point-raster accounting for invalid data in other variables (bias vars, slope, etc), interpolators supporting raster-raster and point-raster inputs,
- [x] Add a generic iterator function for iterative methods (Nuth and Kääb, etc),
- [x] Migrate `Affine` functions out of the classes, and adapt them to work with the new generic core functions,
- [x] Add `is_translation` property to trigger error for unsupported `resample=False`,
- [x] Consistently pass "area_or_point" to Coreg subfunctions,
- [x] Finalize mirroring BiasCorr parameters for modular bin and fit,
- [x] Add nested, typed dictionary structure to organize coregistration parameters and ensure error-free implementations.
- [x] Add `Coreg.info()` to summarize coregistration information, inputs and outputs.
- [x] Save final subsample size for affine classes (pass through preproc function),
- [x] Use `_check_bin_or_fit` function for affine classes (opened #575).

## Will do in separate PRs
- Add robust tests as in BiasCorr, see list in #485,
- Document in #502.


rhugonnet:
@adehecq @erikmannerfelt This PR is ready for your review!

I have two main aspects on which I'm hesitating on how to move forward, and would especially like your opinion on:
1. Currently I defined a `nuth_kaab` main function called by `NuthKaab._fit_xxx` directly (and this for all methods), which is itself made of some of the new core subfunctions (listed in the main description, point 1/) and a method-specific one. This ensure things are consistent, but is annoying to save variables to the `Coreg` metadata that are only available at a given step. For instance, for an easy one: What is the final subsample size on valid values? Not available right now. A harder one: What is the binning/parameters at a given iteration step? (might not be necessary, but could become a need in the future for plotting or other)
I see two solutions here to make that easier:
a) Either we move the content of `nuth_kaab` to `NuthKaab ._fit_xxx`, and wrap the subfunctions called in `nuth_kaab` as methods of `Coreg` which always pass/save the metadata consistently when they run (you can see I already did this with `_bin_or_and_fit_nd` for `BiasCorr`, it is duplicated as a `Coreg` method in order to pass/write the metadata). It would make it impossible to call the same `nuth_kaab` separately than through the `Coreg` class (but I'm not sure that's a need?), but make those new class methods consistent.
b) Or we define new function inputs/outputs (dictionaries?) for each of these subfunctions, to pass metadata of interest through the functions back to the class (could become a hassle very fast).

2. The `CoregDict` is getting busy and all over the place. It seems like it could become a combination of the new dicts created in this PR: `RandomDict`, `FitOrBinDict`, `IterativeDict` (and maybe an output dict on the results?)...
How should we structure this? Having a nested dictionary would make it harder for users to access a specific key easily, even though it's all there. If we did this, we could add an `info()` method that prints the metadata nicely to partially remedy that. Any other ideas?

rhugonnet:
@adehecq @erikmannerfelt After quite a bit of thinking, I've addressed the above points to finalize this PR completely.

**For 2:** I implemented the structure proposed in #573, it is quite satisfactory to extract/write back argument to the Coreg class. See details in that issue. But the nested dictionary are a tiny bit long, I propose an easy solution further below when discussing the renaming of `Coreg.meta`.

**For 1:** I left the new structure as is. For affine functions, it now passes the `subsample_final` from the functions back to the class. We can discuss in more details how to further refine this structure when we need to add stats/plots. It shouldn't be much work in any case now that the base functions are consistent. It would be simply copy/pasting in a subclass method if we chose the class way, or simply adding some more function outputs if we stayed fully function-based. In the second case, we could simply add outputs grouped into the typed dictionary defined for "outputs" **in point 2 above**, which would keep things nicely organized, checked, and wouldn't be too much effort (which is also why I decided not to change the current structure).

**The last big question is: what name should we use for the Coreg dictionary (currently `meta`)?** See #547 for previous discussions about renaming it.
**My latest idea would be to have 2 dictionaries, `Coreg.inputs` and `Coreg.outputs`**! This would reduce one nested layer in the current dictionary `Coreg.meta` which splits into `inputs` and `outputs` in the structure introduced in this PR, which would more practical to access keys. And I think they'd be the clearest names for both developers and users! :smile: 

Ready to merge! Need a review ASAP to use this to finalize the documentation of #502 :wink:

xdem/ddem.py
```diff
@@ -15,6 +17,47 @@
 from xdem._typing import MArrayf, NDArrayf
 
 
+def _mask_as_array(reference_raster: gu.Raster, mask: str | gu.Vector | gu.Raster) -> NDArrayf:
```
adehecq, line 20: I'm really wondering if this is worth having a separate function for that... It does not make much sense that it lies in ddem. But if it's only used here... Not of much help here 😉

adehecq:
Trying to quickly react to your questions before the end of the day:
- regarding 2 (nested dictionary), I like that all input and output parameters are saved and easy to access. I'm fine with `Coreg.inputs` and `Coreg.outputs` as these are clear (although a little bit confusing with the input and output DEM but that's acceptable I think).
- regarding point 1... I didn't manage to understand sorry 😅

xdem/coreg/base.py
```diff
@@ -638,98 +497,448 @@ def _postprocess_coreg_apply(
...
+        )
+
+        # Now, we need to pass this new data to the fitting function and optimizer
+        # We use only the N-D binning estimates (maximum dimension, equal to length of variable list)
+        df_nd = df[df.nd == len(bias_vars)]
+
+        # We get the middle of bin values for variable, and statistic for the diff
+        new_vars = [pd.IntervalIndex(df_nd[var_name]).mid.values for var_name in bias_vars.keys()]
+        new_diff = df_nd[params_fit_or_bin["bin_statistic"].__name__].values
+        # TODO: pass a new sigma based on "count" and original sigma (and correlation?)?
```
adehecq, line 920: is sigma used for weighting the fit later? Of course if the bin counts are very different, it would be great to take that into account during the fit by giving different weights, but probably difficult to implement.

xdem/coreg/base.py
```diff
@@ -1154,57 +1363,152 @@ class NotImplementedCoregApply(NotImplementedError):
...
 
 
-class CoregDict(TypedDict, total=False):
-    """
-    Defining the type of each possible key in the metadata dictionary of Process classes.
-    The parameter total=False means that the key are not required. In the recent PEP 655 (
-    https://peps.python.org/pep-0655/) there is an easy way to specific Required or NotRequired for each key, if we
-    want to change this in the future.
-    """
+class InRandomDict(TypedDict, total=False):
```
adehecq, line 1366: what's the meaning of "total=False"?

xdem/coreg/base.py
```diff
@@ -1154,57 +1363,152 @@ class NotImplementedCoregApply(NotImplementedError):
...
+class OutIterativeDict(TypedDict, total=False):
+    """Keys and types of outputs associated with iterative methods."""
+
+    # Iteration at which the algorithm stopped
+    last_iteration: int
+    # Tolerances of each iteration until threshold
+    all_tolerances: list[float]
+
+
+class InSpecificDict(TypedDict, total=False):
```
adehecq, line 1428: I'm wondering if all these dictionaries should not live in a separate file. This "base.py" contains a lot of different definitions now.

xdem/coreg/base.py
adehecq: Nevermind, I knew I saw the explanation somewhere!

xdem/coreg/base.py
```diff
@@ -1224,18 +1528,52 @@ class Coreg:
...
+    def __init__(self, meta: dict[str, Any] | None = None) -> None:
         """Instantiate a generic processing step method."""
-        self._meta: CoregDict = meta or {}  # All __init__ functions should instantiate an empty dict.
+
+        # Automatically sort input keys into their appropriate nested level using only the TypedDicts defined
+        # above which make up the CoregDict altogether
+        dict_meta = CoregDict(inputs={}, outputs={})
+        if meta is not None:
+            # First, we get the levels ("random", "fitorbin", etc)
+            list_input_levels = list(InputCoregDict.__annotations__.keys())
```
adehecq, line 1543: what is the meaning of "__annotations__" here? Is `InputCoregDict.keys()` not sufficient?

xdem/coreg/base.py
```diff
@@ -1224,18 +1528,52 @@ class Coreg:
...
+            # Then the list of keys per level
+            keys_per_level = [
+                list(globals()[InputCoregDict.__annotations__[lv].__forward_arg__].__annotations__.keys())
```
adehecq, line 1546: similar question about "__forward_arg__".

xdem/coreg/base.py
```diff
@@ -1259,57 +1597,233 @@ def is_affine(self) -> bool:
...
+
+    @overload
+    def info(self, verbose: Literal[False]) -> str:
+        ...
+
+    def info(self, verbose: bool = True) -> None | str:
+        """Summarize information about this coregistration."""
+
+        # Map each key name to a descriptor string
+        dict_key_to_str = {
```
adehecq, line 1635: Should we not have a separate list outside of this class? If one adds a new argument to an existing coreg method or a new coreg method, one has to add the argument here and also in the corresponding dictionary no? I don't know if there is a simple way to combine both...

xdem/coreg/affine.py
```diff
@@ -54,147 +54,728 @@
...
-    elevation_difference: NDArrayf, slope: NDArrayf, aspect: NDArrayf, min_count: int = 20
-) -> tuple[float, float, float]:
+        # Apply method and get new statistic to compare to tolerance, new inputs for next iterations, and
+        # outputs in case this is the final one
+        new_inputs, new_statistic = method(new_inputs, *constant_inputs)
+
+        # Print final results
+        # TODO: Allow to pass a string to _iterate_method on how to print/describe exactly the iterating input
+        if verbose:
+            pbar.write(f"      Iteration #{i + 1:d} - Offset: {new_inputs}; Magnitude: {new_statistic}")
```
adehecq, line 194: I would need to test it to have a better idea of what the prints look like, e.g. for NuthKaab, but indeed generic statements can make it less easy to understand for the user.

xdem/coreg/affine.py
```diff
@@ -54,147 +54,728 @@
...
+        else:
+            rst_elev = ref_elev
+
+        # Derive slope and aspect from the raster dataset
+        slope_tan, aspect = _calculate_slope_and_aspect_nuthkaab(rst_elev)
+
+    return slope_tan, aspect
+
+
+def _nuth_kaab_iteration_step(
```
adehecq, line 490: Would be good to briefly describe the input arguments (in case one needs to do some debugging in the future :wink:)

xdem/coreg/affine.py
```diff
@@ -54,147 +54,728 @@
...
+
+    :return: Final estimated offset: east, north, vertical (in georeferenced units).
+    """
+    if verbose:
+        print("Running Nuth and Kääb (2011) coregistration")
+
+    # Check that DEM CRS is projected, otherwise slope is not correctly calculated
+    if not crs.is_projected:
+        raise NotImplementedError(
+            f"NuthKaab coregistration only works with in a projected CRS, current CRS is {crs}. Reproject "
```
adehecq, line 571: typo "with in a"

xdem/coreg/affine.py
```diff
@@ -54,147 +54,728 @@
...
+    """
+    Fitting function of gradient descending method, returns the NMAD of elevation residuals.
+
+    :returns: NMAD of residuals.
+    """
+
+    # Calculate the elevation difference
+    dh = dh_interpolator(coords_offsets[0], coords_offsets[1])
+    vshift = -np.nanmedian(dh)
+    dh += vshift
```
adehecq, line 637: Removing the median bias should not change the value of the nmad, so these two lines are not needed I believe.

tests/test_coreg/test_base.py
```diff
@@ -888,7 +910,8 @@ def test_blockwise_coreg_large_gaps(self) -> None:
         ddem_post = (aligned - self.ref).data.compressed()
         ddem_pre = (tba - self.ref).data.compressed()
         assert abs(np.nanmedian(ddem_pre)) > abs(np.nanmedian(ddem_post))
-        assert np.nanstd(ddem_pre) > np.nanstd(ddem_post)
+        # TODO: Figure out why STD here is larger since PR #530
```
adehecq, line 913: Ouch 🤕

tests/test_coreg/test_affine.py
```diff
@@ -36,14 +102,53 @@ class TestAffineCoreg:
         inlier_mask=inlier_mask,
         transform=ref.transform,
         crs=ref.crs,
-        verbose=False,
+        verbose=True,
```
adehecq, line 105: still needed or was for testing/debugging?

tests/test_coreg/test_affine.py
```diff
@@ -139,38 +244,49 @@ def test_coreg_example(self, verbose: bool = False) -> None:
...
         # Check the output .metadata is always the same
-        shifts = (nuth_kaab.meta["shift_x"], nuth_kaab.meta["shift_y"], nuth_kaab.meta["shift_z"])
-        res = self.ref.res[0]
-        assert shifts == pytest.approx((-0.463 * res, -0.1339999 * res, -1.9922009))
+        shifts = (
+            nuth_kaab.meta["outputs"]["affine"]["shift_x"],
+            nuth_kaab.meta["outputs"]["affine"]["shift_y"],
+            nuth_kaab.meta["outputs"]["affine"]["shift_z"],
+        )
+        assert shifts == pytest.approx((-9.200801, -2.785496, -1.9818556))
```
adehecq, line 261: I see that the shifts change at the 2nd decimal (nearly 1st decimal when rounded). I'm curious if this could be related to the different bin size used previously and now (72 vs 80)?

tests/test_coreg/test_affine.py
```diff
@@ -139,38 +244,49 @@ def test_coreg_example(self, verbose: bool = False) -> None:
...
 
-        res = self.ref.res[0]
-        shifts = (gds.meta["shift_x"], gds.meta["shift_y"], gds.meta["shift_z"])
-        assert shifts == pytest.approx((0.03525 * res, -0.59775 * res, -2.39144), abs=10e-5)
+        shifts = (
+            gds.meta["outputs"]["affine"]["shift_x"],
+            gds.meta["outputs"]["affine"]["shift_y"],
+            gds.meta["outputs"]["affine"]["shift_z"],
+        )
+        assert shifts == pytest.approx((-10.625, -2.65625, 1.940031), abs=10e-5)
```
adehecq, line 289: why are the shifts completely different compared to the old tests?

tests/test_coreg/test_affine.py
```diff
@@ -209,25 +319,19 @@ def test_coreg_example_shift(self, shift_px, coreg_class, points_or_raster, verb
...
-        best_north_diff = coreg_obj.meta["shift_y"] - shift_px[1]
-
-        raise AssertionError(f"Diffs are too big. east: {best_east_diff:.2f} px, north: {best_north_diff:.2f} px")
+        assert coreg_obj.meta["outputs"]["affine"]["shift_x"] == pytest.approx(-shift_px[0] * res, rel=precision)
+        assert coreg_obj.meta["outputs"]["affine"]["shift_y"] == pytest.approx(-shift_px[0] * res, rel=precision)
 
     def test_nuth_kaab(self) -> None:
 
-        nuth_kaab = coreg.NuthKaab(max_iterations=10)
+        nuth_kaab = coreg.NuthKaab(max_iterations=50)
```
adehecq, line 334: is 50 really needed? Usually the algorithm converges in 2-3 iterations

tests/test_coreg/test_affine.py
```diff
@@ -268,34 +372,12 @@ def test_nuth_kaab(self) -> None:
 
         # Check that the x shift is close to the pixel_shift * image resolution
         assert all(
-            abs((transformed_points.geometry.x.values - self.points.geometry.x.values) - pixel_shift * self.ref.res[0])
+            abs((transformed_points.geometry.x.values - self.points.geometry.x.values) + pixel_shift * self.ref.res[0])
```
adehecq, line 375: was it a mistake in the older test?

adehecq (dismissed):
Phew this is a massive PR !! 🥵 😆 
Thanks a lot for this impressive effort and for explaining all the changes in detail!

I could not reasonably check all lines of code, so I had to just trust the tests and go for a macroscopic review. This means I also did not check whether all docstrings and comments were sufficient to understand the code, but most of them are. We'll see when we have to get back to the code 😉 

Overall, I'm quite positive about the changes and the fact that each coregistration method is more consistent with the others. I can see the benefits for example when trying to compare different methods while keeping full control on each step.

On the other hand, I see some drawbacks of going even more in depth into a generalized framework:
- it is certainly more difficult to understand when reading the code, maintain and debug. Except maybe for you who's got a good sense of the different subfunctions.
- it also means it is more difficult to customize some aspects of the methods. I'm thinking about print statements, possibly figures in the future etc. Hopefully we can come up with generic ways that fit all methods.
- I don't recognize my initial Nuth & Kaab implementation anymore 😢 😉 
- I'm curious if the changes make any difference in term of computation time. Is this something you were able to assess?
I think the first 2 points would call for a wiki page or document somewhere describing how the Coreg and BiasCorr classes are structured, and what needs to be done in case one want to add a new one (also linked to the discussions with demcompare). Not needed for this PR, but probably easier to write while it's still fresh?

Now I have to test it all !

In short, it's always a fine balance between writing generic code and readable code. I think we've moved a step further in complexity 😅

adehecq:
> 1. a) Either we move the content of `nuth_kaab` to `NuthKaab ._fit_xxx`, and wrap the subfunctions called in `nuth_kaab` as methods of `Coreg` which always pass/save the metadata consistently when they run (you can see I already did this with `_bin_or_and_fit_nd` for `BiasCorr`, it is duplicated as a `Coreg` method in order to pass/write the metadata). It would make it impossible to call the same `nuth_kaab` separately than through the `Coreg` class (but I'm not sure that's a need?), but make those new class methods consistent.

And I don't know if this point is still needed, but now that I get a better sense of the question, I would be in favor of option a I think, i.e. the metadata dictionary gets updated at each iteration if needed.

xdem/ddem.py
rhugonnet: Yes... I guess we can leave this to be cleaned up when we revamp the `ddem` module.

xdem/coreg/base.py
```diff
@@ -638,98 +497,448 @@ def _postprocess_coreg_apply(
...
+            values=values,
+            list_var=list(bias_vars.values()),
+            list_var_names=list(bias_vars.keys()),
+            list_var_bins=bin_sizes,
+            statistics=(params_fit_or_bin["bin_statistic"], "count"),
+        )
+
+        # Now, we need to pass this new data to the fitting function and optimizer
+        # We use only the N-D binning estimates (maximum dimension, equal to length of variable list)
+        df_nd = df[df.nd == len(bias_vars)]
```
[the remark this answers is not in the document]
rhugonnet: Yes, I've been thinking of this too, opening an issue.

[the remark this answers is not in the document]
rhugonnet: Opened #578

xdem/coreg/base.py
rhugonnet: Yes, I'm thinking of adding this for the coreg study (propagating uncertainties while binning to pass to the later fit), it would be one of the last thing left to implement to have error-aware coregistrations! :smiley:

rhugonnet: For now leaving aside

xdem/coreg/base.py
rhugonnet: Added a comment to explain. Those are typed dict classes, so need to get the keys in `__annotations__`.

xdem/coreg/base.py
rhugonnet: Also added a comment here. This is to get the value of a Typed Dict class key (so get the nested dict class).

xdem/coreg/affine.py
rhugonnet: Yes, opening an issue to get that done.

rhugonnet: Opened #579

xdem/coreg/affine.py
```diff
@@ -54,147 +54,728 @@
...
+
+            diff_rst_pts = pts_elev[z_name][sub_mask].values - rst_elev_interpolator(
+                (sub_coords[1] + shift_y, sub_coords[0] + shift_x)
+            )
+
+            # Always return ref minus tba
+            if ref == "point":
+                return diff_rst_pts
+            else:
+                return diff_rst_pts
```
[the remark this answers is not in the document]
rhugonnet: Good catch! That really shows we need to update the test in `test_affine` (to be done in a separate PR).

xdem/coreg/affine.py
rhugonnet: Added! I leave out the ones copied from `fit()` or `apply()` to avoid redundancy, but forgot these new ones!

xdem/coreg/affine.py
rhugonnet: Fixed

xdem/coreg/affine.py
rhugonnet: Good point! I had just copied the original implementation blindly :sweat_smile:

xdem/coreg/affine.py
```diff
@@ -709,19 +1272,53 @@ class NuthKaab(AffineCoreg):
...
     """
 
-    def __init__(self, max_iterations: int = 10, offset_threshold: float = 0.05, subsample: int | float = 5e5) -> None:
+    def __init__(
+        self,
+        max_iterations: int = 10,
+        offset_threshold: float = 0.05,
+        bin_before_fit: bool = True,
+        fit_optimizer: Callable[..., tuple[NDArrayf, Any]] = scipy.optimize.curve_fit,
+        bin_sizes: int | dict[str, int | Iterable[float]] = 80,
```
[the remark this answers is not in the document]
rhugonnet: Yes, I had rounded to 80, but let's leave 72 :)

xdem/coreg/affine.py
```diff
@@ -709,19 +1272,53 @@ class NuthKaab(AffineCoreg):
...
+        # Input checks
+        _check_inputs_bin_before_fit(
+            bin_before_fit=bin_before_fit, fit_optimizer=fit_optimizer, bin_sizes=bin_sizes, bin_statistic=bin_statistic
+        )
+
+        # Define iterative parameters
+        meta_input_iterative = {"max_iterations": max_iterations, "tolerance": offset_threshold}
+
+        # Define parameters exactly as in BiasCorr, but with only "fit" or "bin_and_fit" as option, so a bin_before_fit
+        # boolean, no bin apply option, and fit_func is preferefind
```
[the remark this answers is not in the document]
rhugonnet: That's a big typo :laughing:

tests/test_coreg/test_base.py
rhugonnet: :sweat:

tests/test_coreg/test_affine.py
rhugonnet: I think we can leave as is, because `pytest` only returns prints when the test fails, so it won't pollute the output and is always useful even for later.

tests/test_coreg/test_affine.py
rhugonnet: There's several possible reasons, in order of possible importance I would say:
1. We used to have forced filters in the Nuth and Kääb function (larger than a certain dh, I think 100 meters, and flat or very high slopes), those are now removed; To have those filters one can use the `dem_coregistration` workflow right now, and we should add the planned `CoregFilters` classes to add this directly in a pipeline soon,
2. Bin size as you mention,
3. We use to have a rounding + default tolerance parameters for the SciPy algorithm to make the result the same exactly on all OS... I might have to deal with this again once I get the CI running (now that GeoUtils PR is merged),
4. Interpolation is now very slightly different.

rhugonnet: It probably all comes from 1.

tests/test_coreg/test_affine.py
rhugonnet: There's a half pixel shift (10 meters) because the `GradientDescending` used to force a pixel interpretation that was not the one in the DEM.
Now the shifts match that of `NuthKaab`, so it's more consistent!

tests/test_coreg/test_affine.py
rhugonnet: Good catch, that was me playing around while debugging :sweat_smile:

rhugonnet: It does converge in 2/3 iterations still

tests/test_coreg/test_affine.py
rhugonnet: That or the missing sign you noticed above, I will check.

xdem/coreg/base.py
rhugonnet: Yes, now that the module is cleaned up, it's much more generic.
I think those sub-dictionaries could remain an exception simply to be more readable (right next to the main `CoregDict`, and I can add this in a xDEM wiki page for implementing new coregistrations!

xdem/coreg/base.py
rhugonnet: Agreed! Did this, and added a test in `test_base` that will raise an error if a key in the `info()` dictionary has not been added :smile:

rhugonnet:
Agreed on all.
Except maybe for "it is certainly more difficult to understand when reading the code, maintain and debug.". I think in time (getting used to the functions) and maybe with a Wiki page, it will be much easier to maintain and debug a couple consistent core functions, rather than each `Coreg` subclass doing its own thing :wink:.

For computation times: Similar for Raster-Raster, and faster for Point-Raster (due to avoiding rebuilding a Raster to call `interp_points`, now this is done from the array/transform directly :smile:).

Good idea, I will try to write a short Wiki page why this is fresh!

rhugonnet:
Merging like this!
Want to know @erikmannerfelt's opinion before changing `Coreg.meta` into `Coreg.inputs` and `Coreg.outputs`. I actually don't dislike `meta` too much (data about other data: the coregistration algorithm in this case), as long as it's organized inside.

rhugonnet:
FYI: I've added a Wiki page here on the `Coreg` structure and steps to implement a new method: https://github.com/GlacioHack/xdem/wiki/Things-to-know-on-coregistration-class-structure

Issue #483, opened 2024-03-04
Add Python-based ICP to support weights


rhugonnet:
To account for DEM uncertainty during coregistration, we need our own "weight-supporting" ICP functionality.

It seems fairly straightforward to code, here are some implementations we could base ourselves on and give credit to:
Basic implementation: https://github.com/ClayFlannigan/icp
Implementation of CODEM (looks like it supports weight): https://github.com/NCALM-UH/CODEM/blob/main/src/codem/registration/icp.py
SimpleICP existing in multiple languages: https://github.com/pglira/simpleICP/blob/master/python/simpleicp/simpleicp.py

PR #531, opened 2024-05-24
Rework affine transformation re-gridding (from 3D point cloud to 2.5D DEM) for accuracy and performance

### LINKED ISSUE #110: Validate new `apply_matrix()` function further.

### LINKED ISSUE #499: Add option to modify only input `transform` for shifts in `apply_matrix()`?
This is mostly code architecture, it would allow to get rid of redundant `apply_rst` functions for `Affine` classes that only perform shifts (`NuthKaab`, `VerticalShift`, etc...). And ensure the `apply` operation is fully consistent across affine classes.

rhugonnet:
This PR updates `apply_matrix` to use efficient and accurate 3D affine transformation for rasters, for any transformation.

It also modifies the exact affine transformation of a point cloud to use only NumPy instead of OpenCV, which brings us one step closer to not depend on OpenCV (which is a pretty heavy dependency to have). The only remaining functionality relying on OpenCV is the ICP algorithm, which will be addressed in #483.

# Problem and summary of previous changes

In `apply_matrix` for rasters, after applying the affine transformation on the 3D point cloud converted from the DEM (which is an exact conversion, in this direction), we need to then approximate a re-gridded DEM from the transformed point cloud (for which the coordinates do not align with any grid anymore). This is an irregular gridding problem typically done using triangulation (Delauney triangles), and interpolating linearly in each (as done for point clouds by https://gdal.org/programs/gdal_grid.html#gdal-grid, PDAL through GDAL, and in Python by `scipy.interpolate.griddata`). 

In #87 we removed `scipy.interpolate.griddata` as being too slow for `apply_matrix` for rasters, and replaced it by a 2-step routine: (1) correct translations with `scipy.interpolate.BivariateSpline` + (2) correct small rotations with 2D deramping. This increased the speed a lot for everything, but deramping creates a lot of artefacts when approximating a rotation (as explained by @erikmannerfelt), and does not support scaling.

# New logic for affine transformation of rasters

This PR keeps the logic of #87 to differentiate the case of having only translations (to accelerate horizontal/vertical grid shifts), but implements a new iterative method to accurately and quickly resolve the regridding for small rotations, and re-introduces griddata (a bit slower but accurate) for the general case of any rotations/scaling. The artefacts @erikmannerfelt previously noticed from the `griddata` implementation were due to a bad propagation of NaNs during griddata (need to be derived as a function of the distance to the transformed points), so I also fixed this here by implementing a tested point cloud gridding function in GeoUtils: https://github.com/GlacioHack/geoutils/pull/553.

I also optimized the performance and added more flexibility to of a couple point cloud utilities: https://github.com/GlacioHack/geoutils/pull/549, https://github.com/GlacioHack/geoutils/pull/546, https://github.com/GlacioHack/geoutils/pull/547, https://github.com/GlacioHack/geoutils/pull/554.

## Schematic of new implementation

The iterative method for small rotations aims to find the (unknown) elevation Z at (unknown) irregular coordinates X, Y which transform exactly onto the (known) regular grid X0, Y0 with (unknown = objective) elevation Z.

Because we know everything is linked by an exact affine transformation and that the initial DEM can be interpolated at any 2D coordinates X,Y to get Z, we can go faster than `griddata`. We search for coordinates X,Y derived from the inverse affine transform of X0, Y0 and an (unknown) Z, initialized with an initial closest-neighbor search, and then refined iteratively by interpolating the original DEM at coordinates X,Y (and we can re-use the same interpolator, which is quite fast).
The algorithm typically converges (to X0,Y0 within 10e-5) independently for 99% of points in 4/5 iterations, and the last 1% of points take 4/5 more to reach the tolerance.
All grid points converges independently, and this works well even on a 1,000,000 point real-world DEM with outliers, noise, etc (our Longyearbyen example in this case).

xdem_apply_matrix_iterate

## Tests

This PR adds a lot of new tests for both implementations (griddata and the iterative one), some that live in GeoUtils with the underlying functionality.
Both methods are able to reproduce the real transformed elevations of the transformed DEM at a precision of 10e-5 for a constant slope DEM :slightly_smiling_face:. For varying slopes, more differences arise as expected (near 90° slopes have multiple solutions), but mostly because we cannot check as robustly (2D linear interpolation is not equivalent before/after transformation anymore). The tests also check the spreading of NaNs through both methods to ensure they behave as expected.

## Performance

Independently of the nearest neighbour search (detailed below) and data conversions (which we could try to minimize mroe generally at the package scale), the iterative algorithm is much faster than griddata, about 7 times faster for the Longyearbyen DEM! :partying_face: 
```
Elapsed for iterative algo: 1.36470627784729
Elapsed for griddata: 9.616541385650635
```

Running a nearest neighbour search between two points clouds is necessary in both methods (to initialize the Z for the iterative affine method, and to get the locations of NaNs for the griddata method), and is currently one of the limiting factors (takes about 5 seconds on the Longyearbyen dataset, so almost as slow as `griddata` by just a factor 2). 

But using `scipy.KDTree` or refining the `max_distance` parameter of `geopandas.sjoin_nearest` could both improve speed a lot for this part of the methods: https://gis.stackexchange.com/questions/222315/finding-nearest-point-in-other-geodataframe-using-geopandas.

**UPDATE:** Neither using the `max_distance` parameter of `sjoin_nearest`, nor using `scipy.spatial.KDTree` improved the speed at all, so I'm not including any additional changes on this in this PR. There might be a better way to initialize Z than with a nearest neighbour search (also utilizing knowledge from the affine transformation), which would make the "iterative" algorithm much faster overall, but I'm not inspired right now :thinking:.

**UPDATE 2:** Actually using the transformed elevation associated with the original point (even if not aligned very well in X/Y) works as well as a nearest neighbour to initialize, and only takes a couple more iterations (which are ~100x faster than a nearest neighbour search)! Problem solved, and we officially have a re-gridding for affine transforms around 10x faster than the usual irregular point re-gridding and as accurate! :partying_face: :love_you_gesture: 

# Details

This PR:

- [x] Replaces OpenCV by NumPy when applying a 4x4 matrix to a 3D point cloud to reduce the need for the optional dependency (tested locally, not added to the tests yet; gives exactly the same result :tada:),
- [x] Adds a new implementation that finds a fast nearly exact solution by iteration for small rotations,
- [x] Adds a schematic to explain that implementation,
- [x] Adds tests to check nodata propagation of "iterative" method,
- [x] Optimizes speed of iterative method, 
- [x] Speeds-up `to_pointcloud` for `subsample=1`, see https://github.com/GlacioHack/geoutils/pull/549,
- [x] Adds the `griddata` implementation for large rotations, see https://github.com/GlacioHack/geoutils/pull/553,
- [x] Adds tests checking the new implementations of `apply_matrix`.

I will soon add a Python version of ICP, which will reduce even more the need for having OpenCV as an optional dependency (only if using ICP + asking for method="opencv").

Resolves #110
Resolves #499

### DIFF
--- a/tests/test_coreg/test_base.py
+++ b/tests/test_coreg/test_base.py
@@ -1030,15 +1030,28 @@ class TestAffineManipulation:
         trans_epc = apply_matrix(epc, matrix=matrix, centroid=centroid)
 
         # Interpolate transformed DEM at coordinates of the transformed point cloud, and check values are very close
-        z_points = trans_dem_it.interp_points(points=(trans_epc.geometry.x.values, trans_epc.geometry.y.values))
+        z_points_it = trans_dem_it.interp_points(points=(trans_epc.geometry.x.values, trans_epc.geometry.y.values))
+        z_points_gd = trans_dem_gd.interp_points(points=(trans_epc.geometry.x.values, trans_epc.geometry.y.values))
 
-        valids = np.isfinite(z_points)
-        diff_valids = z_points[valids] - trans_epc.z.values[valids]
-
+        valids = np.logical_and(np.isfinite(z_points_it), np.isfinite(z_points_gd))
         assert np.count_nonzero(valids) > 0
-        # Because of outliers, noise and slope near 90°, checking against the interpolated points does not work as well
-        # some points will have large elevation differences, but 90% of those should remain relatively small
-        assert np.percentile(np.abs(diff_valids), 90) < 1
+
+        diff_it = z_points_it[valids] - trans_epc.z.values[valids]
+        diff_gd = z_points_gd[valids] - trans_epc.z.values[valids]
+
+        # Because of outliers, noise and slope near 90°, several solutions can exist
+        # Additionally, contrary to the check in the __raster test which uses a constant slope DEM, the slopes vary
+        # here so the interpolation check is less accurate so all values can vary a bit
+        assert np.percentile(np.abs(diff_it), 90) < 1
+        assert np.percentile(np.abs(diff_it), 50) < 0.2
+        assert np.percentile(np.abs(diff_gd), 90) < 1
+        assert np.percentile(np.abs(diff_gd), 50) < 0.2
+
+        # But, between themselves, the two re-gridding methods should yield much closer results
+        # (no errors due to 2D interpolation for checking)
+        diff_it_gd = z_points_gd[valids] - z_points_it[valids]
+        assert np.percentile(np.abs(diff_it_gd), 99) < 1  # 99% of values are within a meter (instead of 90%)
+        assert np.percentile(np.abs(diff_it_gd), 50) < 0.02  # 10 times more precise than above
 
 
 def test_warp_dem() -> None:
--- a/tests/test_coreg/test_base.py
+++ b/tests/test_coreg/test_base.py
@@ -948,7 +948,7 @@ class TestAffineManipulation:
 
         # Transform point cloud back to array
         trans_numpy = np.array([trans_epc.geometry.x.values, trans_epc.geometry.y.values, trans_epc["z"].values]).T
-        assert np.array_equal(trans_numpy, trans_cv2_arr)
+        assert np.allclose(trans_numpy, trans_cv2_arr)
 
     @pytest.mark.parametrize("regrid_method", [None, "iterative", "griddata"])  # type: ignore
     @pytest.mark.parametrize("matrix", list_matrices)  # type: ignore
@@ -1011,8 +1011,7 @@ class TestAffineManipulation:
         # All smallest mask value should exist in the mask of griddata
         assert np.array_equal(np.logical_or(smallest_mask, mask_nodata_gd), mask_nodata_gd)
 
-    @pytest.mark.parametrize("regrid_method", ["iterative", "griddata"])  # type: ignore
-    def test_apply_matrix__raster_realdata(self, regrid_method: str) -> None:
+    def test_apply_matrix__raster_realdata(self) -> None:
         """Testing real data no complex matrix only to avoid all loops"""
 
         # Use real data
@@ -1026,11 +1025,12 @@ class TestAffineManipulation:
         centroid = (np.mean(epc.geometry.x.values), np.mean(epc.geometry.y.values), 0.0)
 
         # Apply affine transformation to both datasets
-        trans_dem = apply_matrix(dem, matrix=matrix, centroid=centroid, force_regrid_method=regrid_method)
+        trans_dem_it = apply_matrix(dem, matrix=matrix, centroid=centroid, force_regrid_method="iterative")
+        trans_dem_gd = apply_matrix(dem, matrix=matrix, centroid=centroid, force_regrid_method="griddata")
         trans_epc = apply_matrix(epc, matrix=matrix, centroid=centroid)
 
         # Interpolate transformed DEM at coordinates of the transformed point cloud, and check values are very close
-        z_points = trans_dem.interp_points(points=(trans_epc.geometry.x.values, trans_epc.geometry.y.values))
+        z_points = trans_dem_it.interp_points(points=(trans_epc.geometry.x.values, trans_epc.geometry.y.values))
 
         valids = np.isfinite(z_points)
         diff_valids = z_points[valids] - trans_epc.z.values[valids]
--- a/tests/test_coreg/test_base.py
+++ b/tests/test_coreg/test_base.py
@@ -16,6 +16,7 @@ import pytransform3d.rotations
 import rasterio as rio
 from geoutils import Raster, Vector
 from geoutils.raster import RasterType
+from scipy.ndimage import binary_dilation
 
 import xdem
 from xdem import coreg, examples, misc, spatialstats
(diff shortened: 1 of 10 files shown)


rhugonnet:
@erikmannerfelt @adehecq I want to check if you both agree on the way forward to finalize the `apply_matrix` function to support any affine transform matrix without introducing approximation biases (minimizing them at least).

Currently, in `apply_matrix`, after applying the affine transformation on the 3D point cloud converted (exactly, in this direction) from the DEM, we need to approximate a re-gridded DEM from that new point cloud (as the coordinates do not align with any grid anymore). And we use `deramping()` for this. However, as @erikmannerfelt pointed out, this is just an approximation for very small rotation angles, which actually introduces some errors. Plus, it does not account for any scaling that could be passed in the affine matrix.

Instead, I think we should re-grid using `scipy.interpolate.griddata`, which is a delauney triangle-based interpolation, same as used in `gdal_grid` (https://gdal.org/programs/gdal_grid.html#gdal-grid, and used by PDAL through GDAL too for lidar point clouds). I'm not sure we could do better than that! The interpolation in the triangles can be done as "nearest", "linear" or "cubic", and should always perform well when re-gridding a dense point cloud from a DEM back into another DEM, and for any type of affine transformation (translation, rotation, scaling).

And we would maintain the case of an horizontal shift separate for computational efficiency.

The only delicate part of the new routine should be to re-create NaNs where we want to have NaNs after interpolation (any pixel that does not have a shifted data point), but I think I can figure that out.

adehecq:
I totally agree that we should use an exact regridding for the `apply_matrix` function and not a workaround. But in the end, you plan to use the same function as before #87, so it will be very slow for other affine transformations than horizontal/vertical shifts, right? Any chance there would be a more efficient option than griddata 3 years later?
And what about [scipy.interpolate.LinearNDInterpolator](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.LinearNDInterpolator.html) ? Of course, it provides only linear interpolation.

rhugonnet:
So `gridddata` is actually a wrapper around `LinearNDInterpolator`, so both are the same thing!

I looked into the issue in details over the weekend, and it looks like there is no "exact" solution for the reprojection to a regular 2D grid (for instance, rotating the DEM by 90°, so essentially turning it vertical along one axis, many possible points could intersect a similar regular grid point).

However, using `griddata` still seemed like a bit of a computational waste, because we know the exact affine transform function + we have points on a regular grid initially.
Using this, I was able to write a little inversion algorithm that converges fast towards the solution for small rotations (which is what we have in 95% of the cases).
I finished testing it today on a synthetic DEM + the Longyearbyen DEM, and it works for both! All pixels (that is, more than 1 million pixel for the Longyearbyen DEM) converge independently for rotations smaller than 20°, in just 4/5 iterations that only involve affine transformations (exact matrix multiplication) and 2D interpolation on regular grid (very fast)! :partying_face: 

In the next days I will try to make a couple plots to characterize convergence/divergence + estimate the speed-up compared to `griddata` (that I will still add for large rotations) :wink:

rhugonnet:
@erikmannerfelt @adehecq This PR is finished and ready for your review! :smile:

rhugonnet:
Alright, I had to fix a couple issues in the documentation + optimize the speed of some functions. Now all done and ready to merge! :grin:

xdem/coreg/base.py
```diff
@@ -748,47 +748,223 @@ def invert_matrix(matrix: NDArrayf) -> NDArrayf:
...
+        x=epc.geometry.x.values, y=epc.geometry.y.values, z=epc.z.values, matrix=matrix, centroid=centroid
+    )[2]
+
+    # We need to find the elevation Z of a transformed DEM at the exact grid coordinates X,Y
+    # Which means we need to find coordinates X',Y',Z' of the original DEM that, after the exact affine transform,
+    # fall exactly on regular X,Y coordinates
+
+    # 1/ The elevation of the original DEM, Z', is simply a 2D interpolator function of X',Y' (bilinear, typically)
+    xycoords = dem_rst.coords(grid=False)
+    z_interp = scipy.interpolate.RegularGridInterpolator(
```
adehecq, line 850: Why not use Raster.interp here?

tests/test_coreg/test_base.py
```diff
@@ -889,116 +891,167 @@ def test_blockwise_coreg_large_gaps(self) -> None:
...
+    matrix_vertical[2, 3] = 1
+
+    # Vertical and horizontal shifts
+    matrix_translations = matrix_identity.copy()
+    matrix_translations[:3, 3] = [0.5, 1, 1.5]
+
+    # Single rotation
+    rotation = np.deg2rad(5)
+    matrix_rotations = matrix_identity.copy()
+    matrix_rotations[1, 1] = np.cos(rotation)
```
adehecq, line 912: Not a big deal, but I guess you could use pytransform3d's function [`matrix_from_euler`](https://dfki-ric.github.io/pytransform3d/_apidoc/pytransform3d.rotations.matrix_from_euler.html#pytransform3d.rotations.matrix_from_euler) to do the math?

adehecq: Ha I see you use it for the next example

tests/test_coreg/test_base.py
```diff
@@ -889,116 +891,167 @@ def test_blockwise_coreg_large_gaps(self) -> None:
...
+        trans_dem_gd = apply_matrix(dem, matrix=matrix, centroid=centroid, force_regrid_method="griddata")
+
+        # Get nodata mask
+        mask_nodata_it = trans_dem_it.data.mask
+        mask_nodata_gd = trans_dem_gd.data.mask
+
+        # The iterative mask should be larger and contain the other (as griddata interpolates up to 1 pixel away)
+        assert np.array_equal(np.logical_or(mask_nodata_gd, mask_nodata_it), mask_nodata_it)
+
+        # Verify nodata masks are located within two pixels of each other
```
adehecq, line 1007: Why 2 and not 1 pixel, since griddata interpolates up to 1 pixel away.

tests/test_coreg/test_base.py
```diff
@@ -889,116 +891,167 @@ def test_blockwise_coreg_large_gaps(self) -> None:
...
+
+        valids = np.logical_and(np.isfinite(z_points_it), np.isfinite(z_points_gd))
+        assert np.count_nonzero(valids) > 0
+
+        diff_it = z_points_it[valids] - trans_epc.z.values[valids]
+        diff_gd = z_points_gd[valids] - trans_epc.z.values[valids]
+
+        # Because of outliers, noise and slope near 90°, several solutions can exist
+        # Additionally, contrary to the check in the __raster test which uses a constant slope DEM, the slopes vary
+        # here so the interpolation check is less accurate so all values can vary a bit
```
adehecq, line 1044: Maybe for the real case example, you could also try with very small rotations. Something more than 0.1 degree seems unrealistic. In this case your errors should be much less no?

adehecq:
Fantastic !! Nice work ! 🙌 
Very smart algorithm and fancy scheme! 😄 
Although I get what you do (trying to find from which point of the original DEM each regularly gridded point comes from), I still struggle to follow the steps in the code. Maybe an illustration with a 4-5 points line and a rotation, showing what is done at each step, would help but it's also ok if I don't fully understand everything at the moment.
The tests look robust and I trust the outputs anyway.
Just a few minor comments above.

adehecq (dismissed):
Forgot to approve!

xdem/coreg/base.py
rhugonnet: For computational speed: better to create the interpolator only once, and re-use it everywhere, then re-do it at every step!
Maybe we can have this done in GeoUtils directly in the future (when we move the function contents outside of the `Raster` class)

rhugonnet: Added a comment to specify this

tests/test_coreg/test_base.py
rhugonnet: Griddata can add 1 pixel around the original data, while the other methods can remove 1 pixel (due to 2D interpolation), so it's a 2 pixel diff. Added this to the comments.
I'm not sure if there'd be a way to have griddata give a closer NaN propagation to what we have with regular grid interpolation...

rhugonnet: For now I've added a comment to clarify

tests/test_coreg/test_base.py
rhugonnet: Yes good point! Adding this to the tests in a `pytest.mark.parametrize` with different checks

PR #559, opened 2024-08-02
Fix pip install and add new tests

### LINKED ISSUE #557: Missing dependencies after `pip` install
Strangely we have an import error after `pip install xdem`: 
```ModuleNotFoundError: No module named 'fiona'```

Even though we use the same environment file as for conda, and the pip build + import passes in tests.

Looking into it...

rhugonnet:
Ongoing

Resolves #557


adehecq (commented):
Thanks for fixing it! Maybe just update the description of the PR to match that of geoutils (I will do it).

PR #581, opened 2024-09-05
Add ICP in Python

### LINKED ISSUE #483: Add Python-based ICP to support weights

rhugonnet:
Following #530, we can add a Python-based ICP relying on the same core functions as other coregs.

This will allow us to control the method in more details (weights, type of ICP), and to remove the optional OpenCV dependency that is a heavy one.

Resolves #483 

Ongoing


erikmannerfelt:
Interesting!! Is opencv an essential requirement for xdem if this method ends up being on-par with or outperforming the current ICP?

rhugonnet:
Closing as changes of this PR are applied in #701 instead (shared core functions with CPD).

PR #585, opened 2024-09-10
Homogenize sign of translation coregistration and add affine tests

### LINKED ISSUE #422: `xdem.coreg.ICP` fails very underscriptively when not enough points are available
I tried wrap my head around an error I got which even took me to the OpenCV C++ source code. Ooga booga!

When running `.fit_pts()` on a dataset that eventually didn't have enough points for registration (I'm still trying to figure out where the problem is; might be another issue coming up), ICP fails with a very strange error:
```
error: OpenCV(4.7.0) /build/source/modules/core/src/matrix.cpp:749: error: (-215:Assertion failed) m.dims >= 2 in function 'Mat'
```

Some debugging later and I realized one of the point clouds were empty, which the above error is **obviously** trying to tell me. 

There's already an exception handler (see below) for this but for some reason the exception is new. Easy fix is to just catch both exceptions and print the nicer error message.
https://github.com/GlacioHack/xdem/blob/3956073ef6c1846ffc6de3ac5368269f45131f42/xdem/coreg/affine.py#L543-L553

### LINKED ISSUE #423: `xdem.coreg.ICP.fit_pts` fails for unknown reason with a regular dataset
As promised in #422 , here's the other issue hehe.

This is a strange one. I already have a fix, but I don't understand why it helps!

## Problem description
When running `ICP.fit_pts()` I ran into the issue that led to #422; my points were all filtered away for some reason. After some debugging, I found where this happens, but I still don't know **why**.

First, rasters containing information on the normals are sampled using regular `Raster.interp_points` calls:
https://github.com/GlacioHack/xdem/blob/3956073ef6c1846ffc6de3ac5368269f45131f42/xdem/coreg/affine.py#L525-L529

Then, when the rasters have been sampled, a convoluted nan removal line removes the unsuccessful samples:
https://github.com/GlacioHack/xdem/blob/3956073ef6c1846ffc6de3ac5368269f45131f42/xdem/coreg/affine.py#L536-L537

This led to my ref_dem being empty, as the above `interp_points` for some reason only returned nans! I figured out a fix for this problem, but I still don't understand the symptom...

## The solution

Instead of the sampling in the code (shown above), I changed it to:
```python
normal_east.tags["AREA_OR_POINT"] = "Area"
rowcol = np.round(np.dstack(normal_east.xy2ij(ref_dem["E"], ref_dem["N"], shift_area_or_point=True)).squeeze()).astype(int)

if any(col not in ref_dem for col in ["nx", "ny", "nz"]):
    for key, raster in [("nx", normal_east), ("ny", normal_north), ("nz", normal_up)]:
        ref_dem[key] = raster.data[rowcol[:, 0], rowcol[:, 1]].filled(np.nan)
```
Basically, I make a more efficient nearest neighbour sampling by first determining the row and column indices, then directly indexing the underlying raster. This should be much more efficient, as that part is only done once instead of three times, and for some reason it fixes my problem with the sampling. 

## What's the actual problem?
Without a minimal (non-)working example, it's hard for me to figure out exactly what's wrong... Honestly, I think it boils down to one of two potential issues:

1. There's something wrong with `Raster.interp_points ` in geoutils
2. There's something wrong with **my use** of the functionality

Honestly, and argument for making a PR to "fix" this can simply be motivated by performance. If `Raster.interp_points` is not used in the method, no `Raster`s have to be instantiated and I can just use the rasterio coords to ij functions. That'd reduce the amount of copying that might happen and it will probably make the code look cleaner. I might open a PR to improve the code and then add an associated geoutils issue for tackling the underlying problem.

Still, it would be great to get to the bottom of why this actually fails... @rhugonnet or @adehecq, have you experienced something similar before with this function?

### LINKED ISSUE #485: Improve tests in `test_affine` once core class restructured with `BiasCorr` parameters

### LINKED ISSUE #586: Inconsistent sign of translations in `AffineCoreg` methods since restructuration
Due to old code being inconsistent in `_apply_rst` and `_to_matrix_func` of `AffineCoreg` methods, since the #530 restructuration which makes everything use the same `apply_matrix` function, the direction of a translation is not stored/applied consistently resulting in bugs.

Fixing this in #585 with more rigorous tests...

rhugonnet:
This PR fixes and homogenizes the sign of translation parameters in coregistrations, and adds consistent tests mirrored from `test_biascorr` to ensure that affine coregistrations behave as intended for all types of input (raster-raster, raster-point, and point-raster).

In short: the new tests introduce synthetic shifts, then check that estimated coregistration shifts are exactly the same as those introduced within 1% (10% for ICP), and that applying the coregistration then corrects more than 99% of the variance in the initial elevation difference (95% for ICP). This ensures that the direction of coregistration is the right one, and that there is no other errors introduced in the process.

## More background

Before #530, the `_apply_rst` of different method such as `NuthKaab` and `ICP` relied on different code to apply the translation. In some occasions, it called the same `apply_matrix` function, but even then it relied on a different `to_matrix_func` to derive the matrix despite depending on the same parameters (`shift_x`, `shift_y`, `shift_z`). 

This meant that the signs of shifts stored in `shift_x`, `shift_y`, and `shift_y` could be inconsistent between methods. Since #530 relies on the same `apply_matrix` everywhere, this resulted in a erroneous coregistration direction for certain input types (point or raster) and certain method, which weren't capture by the older tests.

## Additional fixes

Now that ICP is tested for all combinations of raster-raster, point-raster and raster-point with a subsample, we can indeed conclude that #422 and #423 indeed came from the `prefilter` issue of our old `geoutils.interp_points` that was removing data by default, which caused issues for point runs, see https://github.com/GlacioHack/geoutils/pull/484 for details.

Resolves #422
Resolves #423
Resolves #485
Resolves #586

### DIFF
--- a/tests/test_coreg/test_affine.py
+++ b/tests/test_coreg/test_affine.py
@@ -277,7 +277,6 @@ class TestAffineCoreg:
         tol = 0.01 if coreg_method != coreg.ICP else 0.05
         assert np.nanvar(dh / np.nanstd(init_dh)) < tol
 
-
     @pytest.mark.parametrize(
         "coreg_method__shift",
         [
@@ -388,9 +387,7 @@ class TestAffineCoreg:
         assert vshift == pytest.approx(expected_vshift)
 
     @pytest.mark.parametrize("fit_args", all_fit_args)  # type: ignore
-    @pytest.mark.parametrize(
-        "shifts_rotations", [(20, 5, 0, 0.02, 0.05, 0.1), (-50, 100, 0, 10, 5, 4)]
-    )  # type: ignore
+    @pytest.mark.parametrize("shifts_rotations", [(20, 5, 0, 0.02, 0.05, 0.1), (-50, 100, 0, 10, 5, 4)])  # type: ignore
     @pytest.mark.parametrize("coreg_method", [coreg.ICP])  # type: ignore
     def test_coreg_rigid__synthetic(self, fit_args, shifts_rotations, coreg_method) -> None:
         """
--- a/tests/test_coreg/test_affine.py
+++ b/tests/test_coreg/test_affine.py
@@ -7,8 +7,6 @@ import geopandas as gpd
 import numpy as np
 import pytest
 import rasterio as rio
-
-import geoutils
 from geoutils import Raster, Vector
 from geoutils._typing import NDArrayNum
 from geoutils.raster import RasterType
@@ -237,7 +235,9 @@ class TestAffineCoreg:
 
         # For a point cloud output, need to interpolate with the other DEM to get dh
         if isinstance(elev_fit_args["to_be_aligned_elev"], gpd.GeoDataFrame):
-            init_dh = ref.interp_points((ref_shifted.geometry.x.values, ref_shifted.geometry.y.values)) - ref_shifted["z"]
+            init_dh = (
+                ref.interp_points((ref_shifted.geometry.x.values, ref_shifted.geometry.y.values)) - ref_shifted["z"]
+            )
             dh = ref.interp_points((coreg_dem.geometry.x.values, coreg_dem.geometry.y.values)) - coreg_dem["z"]
         else:
             init_dh = ref - ref_shifted.reproject(ref)
--- a/tests/test_coreg/test_affine.py
+++ b/tests/test_coreg/test_affine.py
@@ -7,6 +7,8 @@ import geopandas as gpd
 import numpy as np
 import pytest
 import rasterio as rio
+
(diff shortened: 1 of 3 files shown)


rhugonnet:
@adehecq Merging this to use it to finalize #502, as it's mostly tests and your comments can easily be accounted for in a separate PR!

PR #595, opened 2024-09-21
Refactor `GradientDescending` to `DhMinimize` with generic inputs

### LINKED ISSUE #231: Add coregistration by minimization of NMAD
A nice suggestion by Simon Gascoin (@sgascoin). An example of a very simple algorithm that finds a horizontal shift between two DEMs by minimizing the NMAD. A good alternative to the Nuth & Kaab.

```
import cv2
import numpy as np
import scipy.optimize as optimize

def shiftDEM(Z,shiftX,shiftY):
    M = np.float32([[1, 0, shiftX], [0, 1, shiftY]])
    shifted = cv2.warpAffine(Z, M, (Z.shape[1], Z.shape[0]),borderValue=np.nan)
    return shifted

def nmad(x):
    return 1.4826*np.nanmedian(abs(x - np.nanmedian(x)))

def errNmad(params, Zslave, Zmaster):
    shiftX, shiftY = params
    M = np.float32([[1, 0, shiftX], [0, 1, shiftY]])
    shifted = shiftDEM(Zslave,shiftX,shiftY)
    return nmad(shifted-Zmaster)

initial_guess = [1, 1] # shift in pixels
result = optimize.minimize(errNmad, initial_guess, args=(Zwinter, Zsummer),method='Nelder-Mead')
if result.success:
    print("[shiftX, shiftY]", result.x)
else:
    raise ValueError(result.message)
```

### LINKED ISSUE #594: Homogenize `GradientDescending` coregistration with other methods
By using the same generic fitting options, see #231

### LINKED ISSUE #599: OSGeo module not found in tests
The module `osgeo` (`gdal`, `ogr`, etc) is not found anymore. It is only used in tests involving comparison to GDAL.

I thought this was due to another change in Rasterio 1.4 (#597), but it still fails on 1.3.X.

Trying to understand the cause...

rhugonnet:
This PR performs the renaming and generalization as discussed in #231

@liuh886 @erikmannerfelt I indeed found a fairly big difference in performance between `scipy.optimize.minimize()` and `noisyopt.minimizeCompass()`. For SciPy, the "Nelder-Mead" method examplified by @sgascoin in #231 performed best of all, but it still had some trouble consistently correcting an artificial shift introduced in our Longyearbyen example data at different shift magnitudes and subsample sizes (even when passing bounds and various tolerances), while `minimizeCompass` performed OK on all cases. So between the default `minimize()` and `minimizeCompass()`, there is a big difference in performance, but the difference is much smaller between  `minimize(method="Nelder-Mead")` and `minimizeCompass`.

So, to avoid imposing the `noisopt` dependency on users that want to try this function, I was thinking of having the new function `DhMinimize` with default to `minimize(method="Nelder-Mead")`, and adding a "Tip" box in the documentation (Coregistration page, new "DhMinimize" section) to remind that `noisyopt` might perform better. 
And a similar note in the function description.

Does that sound good to all?

Opens #596 
Opens #597 
Resolves #231
Resolves #594
Resolves #599

### DIFF
--- a/tests/test_coreg/test_affine.py
+++ b/tests/test_coreg/test_affine.py
@@ -17,7 +17,7 @@ from scipy.ndimage import binary_dilation
 
 from xdem import coreg, examples
 from xdem.coreg.affine import AffineCoreg, _reproject_horizontal_shift_samecrs
-
+from noisyopt import minimizeCompass
 
 def load_examples(crop: bool = True) -> tuple[RasterType, RasterType, Vector]:
     """Load example files to try coregistration methods with."""
@@ -232,7 +232,11 @@ class TestAffineCoreg:
         warnings.filterwarnings("ignore", message="Covariance of the parameters*")
 
         # Initiate coregistration
-        horizontal_coreg = coreg_method()
+        if coreg_method == coreg.DhMinimize:
+            kwargs = {"fit_minimizer": minimizeCompass}
+        else:
+            kwargs = {}
+        horizontal_coreg = coreg_method(**kwargs)
 
         # Copy dictionary and remove inlier mask
         elev_fit_args = fit_args.copy()
@@ -253,7 +257,7 @@ class TestAffineCoreg:
         fit_shifts = [-horizontal_coreg.meta["outputs"]["affine"][k] for k in ["shift_x", "shift_y", "shift_z"]]
 
         # ICP can be less precise than other methods
-        rtol = 10e-2 if coreg_method != coreg.ICP else 10e-1
+        rtol = 10e-2 if coreg_method == coreg.NuthKaab else 10e-1
         assert np.allclose(shifts, fit_shifts, rtol=rtol)
 
         # For a point cloud output, need to interpolate with the other DEM to get dh
@@ -278,14 +282,14 @@ class TestAffineCoreg:
 
         # Check applying the coregistration removes 99% of the variance (95% for ICP)
         # Need to standardize by the elevation difference spread to avoid huge/small values close to infinity
-        tol = 0.01 if coreg_method != coreg.ICP else 0.05
+        tol = 0.01 if coreg_method == coreg.NuthKaab else 0.05
         assert np.nanvar(dh / np.nanstd(init_dh)) < tol
 
     @pytest.mark.parametrize(
         "coreg_method__shift",
         [
             (coreg.NuthKaab, (9.202739, 2.735573, -1.97733)),
-            (coreg.DhMinimize, (10.0, 2.5, -1.964539)),
+            (coreg.DhMinimize, (10.125, 2.875, -1.943813)),
             (coreg.ICP, (8.73833, 1.584255, -1.943957)),
         ],
     )  # type: ignore
@@ -304,7 +308,11 @@ class TestAffineCoreg:
         coreg_method, expected_shifts = coreg_method__shift
 
         # Run co-registration
-        c = coreg_method(subsample=50000)
+        if coreg_method == coreg.DhMinimize:
+            kwargs = {"fit_minimizer": minimizeCompass}
+        else:
+            kwargs = {}
+        c = coreg_method(subsample=50000, **kwargs)
         c.fit(ref, tba, inlier_mask=inlier_mask, verbose=verbose, random_state=42)
 
         # Check the output translations match the exact values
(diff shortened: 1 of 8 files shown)


sgascoin:
Thank you Romain, regarding the Nelder-Mead optimizer I found that it is sensitive to the initial vector, in particular I noticed that it can fail if the initial components are zero. That is why I always initialize with [1,1].

rhugonnet:
Wow, great tip! :star_struck: This indeed solved the issues that Nelder-Mead had in some cases. It now performs similarly as `noisyopt`. And so I've also made it the default in tests. 
Thanks a lot @sgascoin!

xdem/coreg/affine.py
```diff
@@ -648,81 +647,83 @@ def _gradient_descending_fit_func(
...
+
+    # Initial offset near zero
+    init_offsets = (0, 0)
+
+    # Default parameters depending on optimizer used
+    if params_fit_or_bin["fit_minimizer"] == scipy.optimize.minimize:
+        if "method" not in kwargs.keys():
+            kwargs.update({"method": "Nelder-Mead"})
+            kwargs.update({"options": {"xatol": 10e-6, "maxiter": 500}})
+            # This method has trouble when initialized with 0,0, so defaulting to 1,1 (tip from Simon Gascoin)
```
erikmannerfelt, line 685: Perhaps give a permalink to the PR comment

tests/test_coreg/test_affine.py
```diff
@@ -227,7 +229,8 @@ def test_coreg_translations__synthetic(self, fit_args, shifts, coreg_method) ->
         to be the right one; and that there is no other errors introduced in the process).
         """
 
-        # Initiate coregistration
+        warnings.filterwarnings("ignore", message="Covariance of the parameters*")
```
erikmannerfelt, line 232: I'm pretty sure this should be done within a context manager. But please correct me if there's a source that says it's not needed. The problem is that this setting might propagate to other subsequent tests.

```python
    warnings.filterwarnings("ignore", message="Covariance of the parameters*")
    # here goes the warning-triggering code

```

erikmannerfelt (commented):
Very nice, and as you can see from my tiny comments, I basically approve of everything!

There's only one thing I wish and that's for some backward compatibility. As far as I understand (but please correct me if I'm wrong), this removes the GradientDescending class entirely from the repo. I know a few people here using it so it's a pretty breaking change. I suggest to make this a very simple subclass:

```python
class GradientDescending(DhMinimize):
    def __init__(
        self,
        x0: tuple[float, float] = (0, 0),
        bounds: tuple[float, float] = (-10, 10),
        deltainit: int = 2,
        deltatol: float = 0.004,
        feps: float = 0.0001,
        subsample: int | float = 5e5
)
        """The old docstring here."""
        cls = super().__init__(subsample=subsample, fit_minimizer=minimizeCompass)
        cls.meta.update({
            "x0": x0,
            "bounds": bounds,
            "deltainit": deltainit,
            "feps": feps,
        })
        return cls
```

As far as I can tell, these relatively few lines are enough to retain the previous functionality which I know is used. But now we have the alternative to (through `DhMinimize`) use new minimizers/loss functions thanks to you!

tests/test_coreg/test_affine.py
rhugonnet: Yes, you have to do the context manager to target a specific bit of code. Here it applies to the entire test function. Because this function only has one call that can raise this type of warning, it's equivalent. And as the best practice to test something new is to write another small test function (something we didn't do so much at the beginning), it shouldn't become a problem! :smile:

xdem/coreg/affine.py
rhugonnet: Done!

rhugonnet:
Perfect, thanks!

Actually, thinking about implementing your proposition of backwards-comp above, I realize we can't easily make an equivalent. All optimizer arguments have to be passed to the `kwargs` of `Coreg.fit()` now, and not its `__init__`. And there is no dictionary to save those temporarily, we'd have to overwrite the `_fit_xxx()` and create a new attribute specific to this subclass as well.

As the `GradientDescending` method was neither shown in the API, or documented, these changes should affect very few users. I was simply planning on listing breaking changes in the `0.1` release. (especially as we have a warning on the landing page that states that xDEM is very much in-development, and to note the version number to reproduce results.)

So either updating the code, or forcing an older xDEM version both seem like good solutions here. 
Is that OK?

For clarity the update would just be:
```python
c = GradientDescending(my_optimizer_args)
c.fit()
```
to
```
c = DhMinimize(fit_optimizer=minimizeCompass)
c.fit(my_optimizer_args)
```

rhugonnet:
Merging this to finalize #502, can add backwards-comp in separate PR if still feels necessary.

PR #610, opened 2024-10-22
Replace print statements and verbose flag with logging and log levels

### LINKED ISSUE #565: [POC] use logging module
# Context

When using xdem errors and other types of information are displayed via `print`. The use of the logging module is more relevant, especially with the arrival of the CLI. This ticket, therefore proposes to replace all prints of xdem in logging. 

First, it is good to read the [documentation] (https://docs.python.org/en/3/howto/logging.html) in order to master the different levels and constraints related to the use 

# Code

The files affected by the changes are as follows:
-xdem/examples.py
- xdem/affine.py
-xdem/base.py
-xdem/biascorr.py
-xdem/workflows.py
- xdem/fit.py
-xdem/misc.py
-xdem/spatialstats.py
- xdem/volume.py 
- example/plot_nuth_kaab.py
- example/plot_block_wise_coreg.py
- example/plor_deramp.py
- example/plot_heterosc_estimation_modelling.py
- example/plot_standardization.py
- example/plot_variogram_estimation_modelling.py
- example/plot_dem_substraction.py
- example/plot_infer_heterosc.py
- example/plot_infer_spatial_correlation.py
- example/plot_spatial_error_propagation.py
- test/test_coreg_pipeline.py
- test/test_demcollection.py
- test/test_doc.py

 - [ ] In xdem at the moment there is a verbose variable that if enabled then prints are displayed. Therefore, this variable must be removed from the APIs. 
 - [ ] replace the prints with adequate logging.debug, logging.info, etc.
:warning: f-string doesn’t pass pylint with logging we recommand using the percent formatting. (%s :string, %d int, %f float, ...)

- [ ] Add some logs in run and tests it 

## For CLI
- [ ] Add a new key in config file : verbose_level : info, debug, warning, error
- [ ] add a configuration for cli inspired by: 
In xdem_cli.py add the function for setup the log https://github.com/CNES/demcompare/blob/master/demcompare/log_conf.py
and instanciate a file for saving https://github.com/CNES/demcompare/blob/master/demcompare/__init__.py#L92

# Doc 

 - [ ] Add a paragraph to tell the user how to enable this verbosity in CLI mode but also in API mode 

## Some help for an API use
```py
import logging  
  
# Configuration 
logging.basicConfig(level=logging.DEBUG,  
                    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',  
                    datefmt='%Y-%m-%d %H:%M:%S',  
                    handlers=[  
                        logging.FileHandler('app.log'),  # saving in a file  
  logging.StreamHandler()  # Print in console  
  ])  
  
  
logger = logging.getLogger('mylogger')  
  
def division(a, b):  
    try:  
        result = a / b  
  except ZeroDivisionError:  
        logger.error("Error : division by zero")  
    else:  
        logger.info("The %s / %s = %s", a, b, result)  
        return result  
  
division(10, 2)  
division(10, 0)
```

# Tests

- [ ] verify that the logging file from CLI has been created 

### Estimation 3d

vschaffn:
# Replace `print` statements with `logging` 
Resolves #565 

## Summary
This pull request addresses the transition from using `print()`statements to the Python `logging` module. The update improves flexibility for managing output messages.

## Changes made:

- Replaced all occurences of `print` with appropriate logging levels (`debug`, `info`, `warning`, `error`)
- Removed the `verbose` flag in favor of log levels
- Replaced all `f-strings` with `%` formatting in logging statements to comply with `pylint` requirements

### Note on `tdqm` and `logging` integration in `xdem.correg.affine.py`
Since `tdqm` directly manages output to the console, mixing its output with standard logging (especially at lower verbosity levels) can lead to overlapping or garbled output.

To addess this, I ensured that:
- The `pbar`(from `tdqm.trange`) is conditionnally displayed depending on the current logging level. When the logging level is set to `INFO` or higher, the progress bar is visible.
- `pbar.write()` is used for logging messages within the loop, after checking the logging level.

## Documentation
Updated the documentation to include instructions on how users can configure logging verbosity.

## Example
Added an example showing how a users can use and configure logging

### DIFF
--- a/doc/source/logging.md
+++ b/doc/source/logging.md
@@ -1,0 +1,29 @@
+---
+file_format: mystnb
+jupytext:
+  formats: md:myst
+  text_representation:
+    extension: .md
+    format_name: myst
+kernelspec:
+  display_name: xdem-env
+  language: python
+  name: xdem
+---
+# Logging configuration
+ To configure logging for xDEM, you can utilize Python's built-in `logging` module. Begin by setting up the logging
+ configuration at the start. This involves specifying the logging level, format, and handlers. For example :
+ ```python
+import logging
+
+# Configuration
+logging.basicConfig(level=logging.INFO,
+                    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+                    datefmt='%Y-%m-%d %H:%M:%S',
+                    handlers=[
+                        logging.FileHandler('app.log'),  # Log messages will be saved to this file
+                        logging.StreamHandler()           # Log messages will also be printed to the console
+                    ])
+```
+This configuration will log messages with a severity level of `INFO` and above, including timestamps, logger names, and
+log levels in the output. You can change the logging level to `DEBUG`, `WARNING`, `ERROR` or `CRITICAL` as needed.
--- a/examples/basic/logging_configuration.py
+++ b/examples/basic/logging_configuration.py
@@ -1,0 +1,47 @@
+"""
+Configuring Logging in xDEM
+===============================
+
+This example demonstrates how to configure logging in xDEM by using the Nuth and Kääb
+(`2011 <https:https://doi.org/10.5194/tc-5-271-2011>`_) coregistration method.
+Logging can be customized to various levels, from `DEBUG` for detailed diagnostic output, to `INFO` for general
+updates, `WARNING` for potential issues, and `ERROR` or `CRITICAL` for serious problems.
+
+We will demonstrate how to set up logging and show logs while running a typical xDEM function.
+"""
+
+import logging
+
+import xdem
+
+# Step 1: Configure logging
+logging.basicConfig(
+    level=logging.INFO,  # Change this level to DEBUG or WARNING to see different outputs.
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+    datefmt="%Y-%m-%d %H:%M:%S",
+    handlers=[
+        logging.FileHandler("../xdem_example.log"),  # Save logs to a file
+        logging.StreamHandler(),  # Print logs to the console
+    ],
+)
+
+# Step 2: Load example DEMs (Digital Elevation Models) to work with
+reference_dem = xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
+dem_to_be_aligned = xdem.DEM(xdem.examples.get_path("longyearbyen_tba_dem"))
+
+# Step 3: Perform Nuth & Kaab coregistration (for alignment of DEMs)
+coreg = xdem.coreg.NuthKaab()
+
+# Step 4: Demonstrate logging at various levels
+logging.info("Starting Nuth & Kaab coregistration...")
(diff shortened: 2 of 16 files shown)


adebardo (commented):
@adehecq @rhugonnet We reviewed it together before, what do you think of this first contribution?

duboise-cnes:
> @adehecq @rhugonnet @duboise-cnes We reviewed it together with @vschaffn, what do you think of this first contribution?

Seems ok from my view of the xdem code. Maybe not useful to put the example from now and maybe not at this place ? 
It would be put into a pipeline code but don't know if it is necessary. @adehecq @rhugonnet your opinon ? 
Otherwise, i let Glaciohack say if it is ok for xdem. 

As said before, the goal is to merge CNES/CS contributions into this branch-poc branch to aggregate contributions to xdem. @adehecq @rhugonnet feel free to propose other contribution way

rhugonnet:
@vschaffn @adebardo This is great, thanks! :smile: 

Good initative for the example, and documentation page! I think your new documentation page (that is not listed in the `index.md` right now) would fit very well in the "Reference" section of the doc (with API/Background currently, and where there would be CLI/Config section mirroring the doc of GeoUtils: https://geoutils.readthedocs.io/en/stable/), if this organization makes sense to you all as well?

For the example (not sure I'm commenting on the same commits), I agree with @duboise-cnes that it might not be the best spot, or maybe too generic. I could see it live well in the current "Basic" examples if re-structured into a Sphinx-Gallery format as the others, and calling a typical xDEM function (instead of `divide`) to show the logging levels (like `NuthKaab`?).
There is a bit of redundancy between documentation pages and Sphinx-Gallery examples, but this is intended as different users like to learn in different ways.

In terms of the code modifications: It all looks great! I would add a note in the PR description about how you dealt with the interface between `tqdm` and `logging`, which doesn't seem trivial, for posterity.

Happy with the organization of merging to `GlacioHack:branch-poc`. Given the nature of the POC changes, I think there should be minimal conflicts with other changes ongoing on the `main` branch this fall. But if you'd prefer merging each distinct contribution (like the "logging" one, which is fully documented) individually to `main` to avoid accumulating too much on `branch-poc` and potentially get more merge conflicts, this would work as well. :wink:

adebardo:
Thank you for your feedback, @duboise-cnes  and @rhugonnet .

- Indeed, this example seems problematic; we will propose another solution.
- We will also add a note in the PR regarding `tqdm`.
- Perhaps we could do a mix of both: we can continue working on `branch-poc`, and if it becomes too complicated, we will do more frequent merges into `main`.

vschaffn:
@rhugonnet @duboise-cnes I have taken your feedbacks into account:

- I have changed the interactions between `tqdm` and `logging` in `xdem.coreg.affine.py` to avoid any conflict between the two in the console and added a note about this in the PR.
- I have modified the logging example to be more xdem-specific (using `NuthKaab` coregistration) and restructured it to match the Sphinx-Gallery format.

adebardo (approved):

PR #615, opened 2024-10-24
Unpin Rasterio<1.4 now that 2D xy/rowcol fixed

### LINKED ISSUE #597: Unpin Rasterio 1.3 once 2D xy/rowcol fixed in 1.4.1
Pinned in #595.
See https://github.com/rasterio/rasterio/issues/3191

Currently breaks our `_reproject_samecrs_` function in `coreg/affine`.

rhugonnet:
Resolves #597


rhugonnet:
Ran this again with Rasterio 1.4.1, still failing. Need to look into it in more details...

rhugonnet:
FYI: The error that was raised in Rasterio 1.4.0 was fixed in 1.4.1.
However, the functions `xy` and `rowcol` do not support any other input than 1D anymore (previously ND), and this is intended behaviour. 
So this PR reshapes the input/output to solve this (and still be backwards-compatible with Rasterio pre1.4).

PR #618, opened 2024-10-24
Fix pip tests since Python 3.13 release

### LINKED ISSUE #616: Pip checks breaking with Python 3.13
Mirroring https://github.com/GlacioHack/geoutils/issues/614. Tests in CI will currently fail because of this.

We need to fix the pinning of Python versions for `pip` (lost in translation of `environment.yml` to `requirements.txt`), so only works properly for `conda` right now. Working on it.

rhugonnet:
Resolves #616


rhugonnet:
@adebardo @vschaffn Problem solved! Merging your fork branches with "upstream"/main (or however you named "GlacioHack/xdem.git") should solve the CI issues.

Issue #623, opened 2024-10-28
[POC] simple CLI  nuth and kaab


adebardo:
Given that Valentin is making good progress this month, we will set up a simple CLI that allows for the coregistration of two DEMs with Nuth and Kaab.

# Context

The goal of this ticket is to implement coregistration from the xdem run.

- [ ] add the right function to veirfy both path exists
- [ ] Load the reference and secondary DEMs using the accessor:
`reference_elev, to_be_aligned_elev = geoutils.rasters.load_multiple_rasters([config["inputs"]["reference_elev"], config["inputs"]["to_be_aligned_elev"]) `

- [ ] Execute a coregistration:
```py
from worflows import dem_coregistration

coreg_dem, coreg_method, out_stats, inlier_mask = dem_coregistration(reference_elev, to_be_aligned_elev, config["outputs"]["path"])
coreg_dem.save("aligned_dem.tif")
print(coreg.meta)
print(out_stats)
inlier_mask.save("inlier_mask.tif")

```

# For the tests

- [ ] Retrieve test data and ground truth data
- [ ] Verify that the outputs matches the ground truth from [here](https://github.com/GlacioHack/xdem/blob/main/tests/test_coreg/test_affine.py#L228) in the file `tests/test_cli.py`
- [ ] add tests for new entry 

# For the documentation

- [ ] Update the quick start guide in the documentation

PR #629, opened 2024-10-31
Implement Coregistration Functionality for CLI

### LINKED ISSUE #622: [POC] simple CLI
Given that @vschaffn is making good progress this month, we will set up a simple CLI that allows for the coregistration of two DEMs with Nuth and Kaab.

# Context

The purpose of this ticket is to set up the initial file that will allow xDEM to be executed via a CLI command: `xdem path_ref path_sec -v`. Initially, the command will not execute any functionality but will simply display output.

 - [ ] Add the file `xdem_cli.py`
 - [ ] Add the function `get_parser`

```py
def get_parser():
    """
    ArgumentParser for xdem

    :return: parser
    """
    parser = argparse.ArgumentParser(
        description=("Compare Digital Elevation Models"),
        fromfile_prefix_chars="@",
    )

    parser.add_argument(
        "reference_dem",
        help=(
            "path to a reference dem"
        ),
    )
    
    parser.add_argument(
        "dem_to_be_aligned",
        help=(
            "path to a second dem"
        ),
    )
    
	 parser.add_argument(
	 "--loglevel",
	 default="INFO",
	 choices=("DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"),
	 help="Logger level (default: INFO. Should be one of "
	 "(DEBUG, INFO, WARNING, ERROR, CRITICAL)",
	 )

    parser.add_argument(
        "--version",
        "-v",
        action="version",
        version=f"%(prog)s {xdem.__version__}",
    )

	return parser
```

 - [ ] add main function

```py
def main():
    """
    Call demcompare's main
    """
    parser = get_parser()
    argcomplete.autocomplete(parser)
    args = parser.parse_args()
    # use a global try/except to cath
    xdem.run(args.reference_dem, args.dem_to_be_aligned, args.verbose)
    
if __name__ == "__main__":
    main()
```

 - [ ] add imports in setup.py file
 - [ ] in file __init__.py add function `run(configfile)`

```py
def run(reference_dem: str, dem_to_be_aligned: str, verbose: bool):  
	"""
	docstring
	"""

	print("hello world")
```

 - [ ] in setup.cfg add 

```py
# xdem entry points cli scripts
# xdem : main cli Program
[options.entry_points]
console_scripts =
    xdem = xdem.xdem:main
```

# Tests

 - [ ] "Add a test that runs the execution and checks the output string in a file named tests/test_cli.py"

# Documentation

In folder doc/source/quickstart.md : 
 - [ ] Add a section about the CLI

vschaffn:
Resolves #622 and #623

# Description

This PR implements the coregistration functionality for the CLI. The aim is to enhance the xdem run by adding the necessary functions to verify the existence of both input paths and executing the coregistration of DEMs.

## **Key Changes**:
1. **Path Verification:**
   - Implemented functionality to verify that the paths for both the reference and secondary DEMs exist before processing.

2. **Loading DEMs:**
   - Utilized the `geoutils.rasters.load_multiple_rasters` accessor to load the reference and secondary DEMs:
     ```python
     reference_dem, to_be_aligned_dem = geoutils.raster.load_multiple_rasters([ref_dem_path, tba_dem_path])
     ```

3. **Coregistration Execution:**
   - Integrated the `dem_coregistration` function from the `workflows` module to perform the coregistration:
     ```python
     coreg_dem, coreg_method, out_stats, inlier_mask = dem_coregistration(
        to_be_aligned_dem, reference_dem, "aligned_dem.tiff"
    )
     ```
   - The aligned DEM is saved as `aligned_dem.tiff`, and the inlier mask is saved as `inlier_mask.tiff`.

4. **Output Verification:**
   - Added print statements to output the metadata of the coregistered DEM and the resulting statistics.

## **Testing**:
Developed tests to retrieve test data and validate the outputs against ground truth data as specified in `tests/test_cli.py`.

## **Documentation**
Updated the Quick Start Guide in the documentation to reflect the new coregistration capabilities and usage instructions.

 
## **Future Work**:
As #617 is not merged yet, the changes made do not take account of the change of licence. After #617 merge, this PR should be update with a header for `xdem_cli.py` and a description of the package `argcomplete` in the NOTICE file.


rhugonnet:
Great! :slightly_smiling_face:  

I don't have much to comment on the code, my main remark at this stage is rather on the structure: Should we already anticipate other CLI usage than coregistration (e.g., terrain attributes, uncertainty)?

If yes, we could potentially separate into different namings, for instance by adding a first parameter to the parsing:  `xdem coregister path_ref path_sec`, which could be extended to other modules to run for instance `xdem terrain "slope"`, `xdem uncertainty "method1"` etc.
But this is a structure that works only if the functionalities are independent. We could also think of another structure where all these steps can be run sequentially (for instance all users run the uncertainty after coregistration), and everything is combined into a single call like `xdem pipeline *params`. 
Or we could have both.
The structure(s) will have to be reflected in the JSON config file(s), so worth thinking ahead I think!

Small remarks:
- Should we modify "path_sec" -> "path_tba" to match the Python API? Or we can change the Python API.
- To save `inlier_mask`, can also write more shortly (duplicates the boolean data, which is not ideal even if it's pretty light; we'll have the `_save()` function callable on the NumPy array in GeoUtils to avoid that soon):
```python
inlier_rst = coreg_dem.copy(new_array=inlier_mask)
inlier_rst.save("xxx.tif")
```

vschaffn:
Thanks for your feedback @rhugonnet.

- I agree with you for the anticipation of other CLI usages, then I have restructured the code to allow other uses to be integrated.
Now, to perform coregistration, the command is :
`xdem [options] coregister path_ref path_tba`

- I have changed the path name in the documentation to match with Python API

- Thanks for the tip to save the inlier_mask in a simpler way, even if it is not perfect 👍

rhugonnet:
@adebardo @belletva @marinebcht I assume we can close this PR now.

CLI (building on top of this first implementation) was added in #764.

PR #630, opened 2024-10-31
Makefile initial installation

### LINKED ISSUE #582: [POC] Makefile installations
# Context

The purpose of this ticket is to set up the first makefile, which will aim to simplify the installation of xdem for developers using Linux.

To begin, we will draw inspiration from the [makefile](https://github.com/CNES/demcompare/blob/master/Makefile) of demcompare.

# Example

```Makefile 
# Autodocumented Makefile
# see: https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
# Dependencies : python3 venv
# Some Makefile global variables can be set in make command line
# Recall: .PHONY  defines special targets not associated with files

############### GLOBAL VARIABLES ######################
.DEFAULT_GOAL := help
# Set shell to BASH
SHELL := /bin/bash

# Set Virtualenv directory name
# Example: VENV="other-venv/" make install
ifndef VENV
	VENV = "venv"
endif

# Browser definition for sphinx and coverage
define BROWSER_PYSCRIPT
import os, webbrowser, sys

from urllib.request import pathname2url

webbrowser.open("file://" + pathname2url(os.path.abspath(sys.argv[1])))
endef
export BROWSER_PYSCRIPT
BROWSER := python -c "$$BROWSER_PYSCRIPT"

# Python global variables definition
PYTHON_VERSION_MIN = 3.9
# Set PYTHON if not defined in command line
# Example: PYTHON="python3.10" make venv to use python 3.10 for the venv
# By default the default python3 of the system.
ifndef PYTHON
	PYTHON = "python3"
endif
PYTHON_CMD=$(shell command -v $(PYTHON))

PYTHON_VERSION_CUR=$(shell $(PYTHON_CMD) -c 'import sys; print("%d.%d"% sys.version_info[0:2])')
PYTHON_VERSION_OK=$(shell $(PYTHON_CMD) -c 'import sys; cur_ver = sys.version_info[0:2]; min_ver = tuple(map(int, "$(PYTHON_VERSION_MIN)".split("."))); print(int(cur_ver >= min_ver))')

############### Check python version supported ############

ifeq (, $(PYTHON_CMD))
    $(error "PYTHON_CMD=$(PYTHON_CMD) not found in $(PATH)")
endif

ifeq ($(PYTHON_VERSION_OK), 0)
    $(error "Requires python version >= $(PYTHON_VERSION_MIN). Current version is $(PYTHON_VERSION_CUR)")
endif

################ MAKE targets by sections ######################

help: ## this help
	@echo "      XDEM MAKE HELP"
	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'

.PHONY: venv
venv: ## create virtualenv in "venv" dir if not exists
	@test -d ${VENV} || $(PYTHON_CMD) -m venv ${VENV}
	@touch ${VENV}/bin/activate
	@${VENV}/bin/python -m pip install --upgrade wheel setuptools pip # no check to upgrade each time

.PHONY: install
install: venv  ## install environment for development target (depends venv)
	@test -f ${VENV}/bin/xdem || echo "Install xdem package from local directory"
	@test -f ${VENV}/bin/xdem || ${VENV}/bin/pip install -e .[dev]
	@test -f .git/hooks/pre-commit || echo "Install pre-commit"
	@test -f .git/hooks/pre-commit || ${VENV}/bin/pre-commit install -t pre-commit
	@test -f .git/hooks/pre-push || ${VENV}/bin/pre-commit install -t pre-push
	@chmod +x ${VENV}/bin/register-python-argcomplete
	@echo "Xdem installed in dev mode in virtualenv ${VENV} with Sphinx docs"
	@echo "Xdem venv usage : source ${VENV}/bin/activate; xdem -h"

## Clean section

.PHONY: clean
clean: clean-venv clean-build clean-precommit clean-pyc clean-test clean-lint clean-docs clean-notebook ## clean all

.PHONY: clean-venv
clean-venv: ## clean venv
	@echo "+ $@"
	@rm -rf ${VENV}

.PHONY: clean-build
clean-build: ## clean build artifacts
	@echo "+ $@"
	@rm -fr build/
	@rm -fr dist/
	@rm -fr .eggs/
	@find . -name '*.egg-info' -exec rm -fr {} +
	@find . -name '*.egg' -exec rm -f {} +

.PHONY: clean-precommit
clean-precommit: ## clean precommit hooks in .git/hooks
	@rm -f .git/hooks/pre-commit
	@rm -f .git/hooks/pre-push

.PHONY: clean-pyc
clean-pyc: ## clean Python file artifacts
	@echo "+ $@"
	@find . -type f -name "*.py[co]" -exec rm -fr {} +
	@find . -type d -name "__pycache__" -exec rm -fr {} +
	@find . -name '*~' -exec rm -fr {} +
```

# Tests

- [ ] Test installations with Python 3.9, 3.10, 3.11

vschaffn:
Resolves #582.

## Description
This PR introduces the initial Makefile for the `xdem` project, aimed at streamlining the installation process for developers on Linux systems. By implementing a Makefile, we provide a simple command-line interface for common setup tasks, enhancing the developer experience and reducing the potential for errors during installation.

### Key Changes:

1. **Makefile Structure**: The Makefile has been created based on the existing structure of the `demcompare` project. 

2. **Targets Implemented**:
   - **`help`**: Displays help information about available targets.
   - **`venv`**: Creates a virtual environment in the specified directory if it does not exist.
   - **`install-gdal`**: Installs the version of GDAL corresponding to that of the system.
   - **`install`**: Installs the `xdem` package in editable mode along with necessary dependencies and sets up pre-commit hooks for Git.
   - **`clean`**: A comprehensive clean target to remove build artifacts, virtual environment, and pre-commit hooks.
   - **`test`**: Run pytest (if GDAL is in the virtual environment).

3. **Update dependencies**:
The dependencies used in each sections (required, optionnal, test, doc) have been updated to match with the projet.

## Documentation
The documentation https://xdem.readthedocs.io/en/stable/how_to_install.html, README.md and CONTRIBUTING.md have been updated with the `make install` command for contributors.

## Constraints
- GDAL is needed on the system to be installed on the virtual environment and run the tests.
- For now `richdem` is not stable enough to build a wheel on python > 3.10. For the moment, the `make install` command only works on python 3.10 and fails if this version is not installed on the system.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,7 @@ be reported by CoverAlls.
 
 ### Setup
 
+#### With `mamba`
 Clone the git repo and create a `mamba` environment (see how to install `mamba` in the [mamba documentation](https://mamba.readthedocs.io/en/latest/)):
 
 ```bash
@@ -37,6 +38,15 @@ cd xdem
 mamba env create -f dev-environment.yml  # Add '-n custom_name' if you want.
 mamba activate xdem-dev  # Or any other name specified above
 ```
+
+#### With `pip`
+```bash
+git clone https://github.com/GlacioHack/xdem.git
+cd xdem
+make install
+```
+
+Please note: pip installation is currently only possible under python3.10.
 
 ### Tests
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,88 @@
+# Autodocumented Makefile for xDEM
+# see: https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
+# Dependencies : python3 venv
+
+############### GLOBAL VARIABLES ######################
+.DEFAULT_GOAL := help
+SHELL := /bin/bash
+
+# Virtualenv directory name (can be overridden)
+ifndef VENV
+	VENV = "venv"
+endif
+
+# Python version requirement
+PYTHON_VERSION_MIN = 3.9
+
+ifndef PYTHON
+	PYTHON = "python3"
+endif
+PYTHON_CMD=$(shell command -v $(PYTHON))
+
+PYTHON_VERSION_CUR=$(shell $(PYTHON_CMD) -c 'import sys; print("%d.%d" % sys.version_info[0:2])')
+PYTHON_VERSION_OK=$(shell $(PYTHON_CMD) -c 'import sys; cur_ver = sys.version_info[0:2]; min_ver = tuple(map(int, "$(PYTHON_VERSION_MIN)".split("."))); print(int(cur_ver >= min_ver))')
+
+############### Check python version supported ############
+
+ifeq (, $(PYTHON_CMD))
+    $(error "PYTHON_CMD=$(PYTHON_CMD) not found in $(PATH)")
+endif
+
+ifeq ($(PYTHON_VERSION_OK), 0)
+    $(error "Requires python version >= $(PYTHON_VERSION_MIN). Current version is $(PYTHON_VERSION_CUR)")
+endif
+
+################ MAKE Targets ######################
+
+help: ## Show this help
+	@echo "      XDEM MAKE HELP"
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: venv
+venv: ## Create a virtual environment in 'venv' directory if it doesn't exist
+	@test -d ${VENV} || $(PYTHON_CMD) -m venv ${VENV}
+	@touch ${VENV}/bin/activate
+	@${VENV}/bin/python -m pip install --upgrade wheel setuptools pip
+
+.PHONY: install
+install: venv ## Install xDEM for development (depends on venv)
+	@test -f ${VENV}/bin/xdem || echo "Installing xdem in development mode"
+	@test -f ${VENV}/bin/xdem || ${VENV}/bin/pip install -e .[dev]
+	@test -f .git/hooks/pre-commit || echo "Installing pre-commit hooks"
+	@test -f .git/hooks/pre-commit || ${VENV}/bin/pre-commit install -t pre-commit
+	@test -f .git/hooks/pre-push || ${VENV}/bin/pre-commit install -t pre-push
+	@echo "XDEM installed in development mode in virtualenv ${VENV}"
+	@echo "To use: source ${VENV}/bin/activate; xdem -h"
+
+#.PHONY: test
+#test: ## run tests
+#	@${VENV}/bin/pytest
+
+## Clean section
+
+.PHONY: clean
+clean: clean-venv clean-build clean-pyc clean-precommit ## Clean all
+
+.PHONY: clean-venv
+clean-venv: ## Clean the virtual environment
+	@echo "+ $@"
+	@rm -rf ${VENV}
+
+.PHONY: clean-build
+clean-build: ## Remove build artifacts
+	@echo "+ $@"
+	@rm -rf build/ dist/ .eggs/
+	@find . -name '*.egg-info' -exec rm -rf {} +
+	@find . -name '*.egg' -exec rm -f {} +
+
+.PHONY: clean-precommit
+clean-precommit: ## Remove pre-commit hooks from .git/hooks
+	@rm -f .git/hooks/pre-commit
+	@rm -f .git/hooks/pre-push
+
+.PHONY: clean-pyc
+clean-pyc: ## Remove Python cache and artifacts
+	@echo "+ $@"
+	@find . -type f -name "*.py[co]" -exec rm -rf {} +
+	@find . -type d -name "__pycache__" -exec rm -rf {} +
+	@find . -name '*~' -exec rm -rf {} +
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,16 @@ ifndef VENV
 endif
 
 # Python version requirement
-PYTHON_VERSION_MIN = 3.9
+PYTHON_VERSION_REQUIRED = 3.10
 
 ifndef PYTHON
-	PYTHON = "python3"
+	# Try to find python version required
+	PYTHON = "python$(PYTHON_VERSION_REQUIRED)"
 endif
 PYTHON_CMD=$(shell command -v $(PYTHON))
 
 PYTHON_VERSION_CUR=$(shell $(PYTHON_CMD) -c 'import sys; print("%d.%d" % sys.version_info[0:2])')
-PYTHON_VERSION_OK=$(shell $(PYTHON_CMD) -c 'import sys; cur_ver = sys.version_info[0:2]; min_ver = tuple(map(int, "$(PYTHON_VERSION_MIN)".split("."))); print(int(cur_ver >= min_ver))')
+PYTHON_VERSION_OK=$(shell $(PYTHON_CMD) -c 'import sys; req_ver = tuple(map(int, "$(PYTHON_VERSION_REQUIRED)".split("."))); cur_ver = sys.version_info[0:2]; print(int(cur_ver == req_ver))')
 
 ############### Check python version supported ############
 
@@ -29,7 +30,7 @@ ifeq (, $(PYTHON_CMD))
 endif
 
 ifeq ($(PYTHON_VERSION_OK), 0)
-    $(error "Requires python version >= $(PYTHON_VERSION_MIN). Current version is $(PYTHON_VERSION_CUR)")
+    $(error "Requires Python version == $(PYTHON_VERSION_REQUIRED). Current version is $(PYTHON_VERSION_CUR)")
 endif
 
 ################ MAKE Targets ######################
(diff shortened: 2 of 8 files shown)


rhugonnet:
I am not very familiar with Makefile builds (@adehecq is much more). If my understanding is correct, this would replace/improve the current developer installation via `mamba` + `dev-environment.yml` shown here: https://xdem.readthedocs.io/en/stable/how_to_install.html#installing-for-contributors?

For us (even before xDEM, and on Python 2), using the `conda-forge` channel (with `conda` or `mamba`) to solve for GDAL + anything else was always the successful recipe. Is that possible within a Makefile build?
But even through this, GDAL + RichDEM did occasionally break in CI from time to time in the past 4 years :sweat_smile: (once or twice, for instance https://github.com/GlacioHack/xdem/issues/376; but has been passing on all OS and Python versions recently).

We could also coordinate with [Richard Barnes](https://github.com/r-barnes) to solve the issue.

vschaffn:
@rhugonnet the aim of this makefile is to install easily xdem in develoment mode with pip, so it would not replace/improve `mamba` installation but an alternative without using conda

adebardo:
Hello ! 
- The 'make test' command doesn't work for me. If it's because of GDAL, the best thing to do is delete the line in the Makefile for the moment.

adebardo:
@duboise-cnes Could you give us your opinion on the dependances issues?

Makefile
```diff
@@ -0,0 +1,88 @@
...
+
+.PHONY: venv
+venv: ## Create a virtual environment in 'venv' directory if it doesn't exist
+	@test -d ${VENV} || $(PYTHON_CMD) -m venv ${VENV}
+	@touch ${VENV}/bin/activate
+	@${VENV}/bin/python -m pip install --upgrade wheel setuptools pip
+
+.PHONY: install
+install: venv ## Install xdem for development (depends on venv)
+	@test -f ${VENV}/bin/xdem || echo "Installing xdem in development mode"
```
adebardo, line 49: Could you replace with xDEM ? :)

Makefile
```diff
@@ -0,0 +1,88 @@
+# Autodocumented Makefile for xdem
```
adebardo, line 1: xDEM :)

adebardo:
@rhugonnet Would you agree to update the readme by indicating that a pip install is viable?

adebardo:
We also should update documentation : https://xdem.readthedocs.io/en/stable/how_to_install.html

vschaffn:
@adebardo For now I have commented the `make test`. The documentation https://xdem.readthedocs.io/en/stable/how_to_install.html has been updated with the `make install`.

rhugonnet:
Yes, let's add the `pip` option to the README and the makefile option to the "How to install" page!

duboise-cnes:
> I am not very familiar with Makefile builds (@adehecq is much more). If my understanding is correct, this would replace/improve the current developer installation via `mamba` + `dev-environment.yml` shown here: https://xdem.readthedocs.io/en/stable/how_to_install.html#installing-for-contributors?
> 
> For us (even before xDEM, and on Python 2), using the `conda-forge` channel (with `conda` or `mamba`) to solve for GDAL + anything else was always the successful recipe. Is that possible within a Makefile build? But even through `conda-forge`, GDAL + RichDEM did occasionally break in CI from time to time in the past 4 years 😅 (once or twice, for instance #376; but has been passing on all OS and Python versions recently).
> 
> We could also coordinate with [Richard Barnes](https://github.com/r-barnes) to solve the issue.

Makefile is not really "needed" absolutely but it is common tool in linux development and a simple wrapper for development practise. 
I think for support for other developers, the make clean is really great to go back to a clean state for instance. I don't know how to do that with conda for instance. It also suppose to be on a python3 pip pure install and therefore is more general than conda environment to handle possible dependencies problems. Anyway, in our internal chains, we don't use conda and having a developer environment way without it ensure compatibility. And we have generalized the makefile to organize common way to understand a dev python repository. Just clone and do a make help and you have the supported dev recipes of the project...

I understand that conda ease the way to have GDAL (even if you are not root on your pc). But in our experience, we have tried to remove GDAL direct dependency because of difficulty of maintenance and use it through rasterio directly in python3 or other way. 

So I would propose two steps: 
- finish this PR adding dependency of GDAL for the make install in documentation and in Makefile doing a test for GDAL presence and crash if not present 
- second step, discuss possibilities in next meeting for getting rid of gdal  direct dependency if ok or possible. (use through rasterio, have reference generated outside the code). I think even for contributors that it would be better not having direct GDAL dependency for maintenance in XDEM. My opinon, but to be exchanged. For developers, it is not so difficult to get GDAL (we have it on the cnes cluster) and there is anyway the conda approach in parallel. 

As for richdem, if not so maintained for different python, can the dependency be removed, included, updated as discussed (if i remember well) ? Again, maybe in another PR, not sure 

I have several comments on the code line by line in following comments.

Makefile
```diff
@@ -0,0 +1,88 @@
...
+	@test -f ${VENV}/bin/xdem || echo "Installing xdem in development mode"
+	@test -f ${VENV}/bin/xdem || ${VENV}/bin/pip install -e .[dev]
+	@test -f .git/hooks/pre-commit || echo "Installing pre-commit hooks"
+	@test -f .git/hooks/pre-commit || ${VENV}/bin/pre-commit install -t pre-commit
+	@test -f .git/hooks/pre-push || ${VENV}/bin/pre-commit install -t pre-push
+	@echo "XDEM installed in development mode in virtualenv ${VENV}"
+	@echo "To use: source ${VENV}/bin/activate; xdem -h"
+
+#.PHONY: test
+#test: ## run tests
```
duboise-cnes, line 58: I would keep the test command and add a test if GDAL is installed or not, for the test to work (if gdal is needed)

doc/source/how_to_install.md
```diff
@@ -28,9 +28,18 @@ Updating packages with `pip` (and sometimes `mamba`) can break your installation
...
 
+### With ``mamba``
+
 ```bash
 git clone https://github.com/GlacioHack/xdem.git
 mamba env create -f xdem/dev-environment.yml
 ```
 
+### With ``pip``
+
```
duboise-cnes, line 39: I would add documentation for this PR saying that GDAL is needed for development. Maybe in a section with make test ?

doc/source/how_to_install.md
```diff
@@ -28,9 +28,18 @@ Updating packages with `pip` (and sometimes `mamba`) can break your installation
...
+```bash
+git clone https://github.com/GlacioHack/xdem.git
+make install
+```
+
 After installing, you can check that everything is working by running the tests: `pytest`.
```
duboise-cnes, line 45: maybe separate a way with conda and a way with Makefile (as commented, to be discussed in next meeting for future way...)

Makefile
```diff
@@ -0,0 +1,88 @@
...
+
+.PHONY: install
+install: venv ## Install xDEM for development (depends on venv)
+	@test -f ${VENV}/bin/xdem || echo "Installing xdem in development mode"
+	@test -f ${VENV}/bin/xdem || ${VENV}/bin/pip install -e .[dev]
+	@test -f .git/hooks/pre-commit || echo "Installing pre-commit hooks"
+	@test -f .git/hooks/pre-commit || ${VENV}/bin/pre-commit install -t pre-commit
+	@test -f .git/hooks/pre-push || ${VENV}/bin/pre-commit install -t pre-push
+	@echo "XDEM installed in development mode in virtualenv ${VENV}"
+	@echo "To use: source ${VENV}/bin/activate; xdem -h"
```
duboise-cnes, line 55: the xdem cli doesn't exist maybe python3 -c "import xdem" or nothing for now

setup.cfg
```diff
@@ -49,17 +49,19 @@ include =
...
+    pyyaml
 test =
     pytest
     pytest-xdist
-    pyyaml
     flake8
     pylint
+    scikit-learn
     richdem
+    gdal
```
duboise-cnes, line 64: cf my global remark, It would be better to remove gdal if possible, in my opinon... in the future ?

rhugonnet:
Thanks @duboise-cnes, it's much clearer!

For GDAL/RichDEM: They are not dependencies for the users, but they are for developers (only used in tests). 

They are quite important as they verify all types of terrain attributes (slope, aspect, TPI, TRI, etc), custom reprojection in the same CRS (to avoid sub-pixel errors of Rasterio that haven't been fixed in more than a year), etc. Same for RichDEM to check curvatures.

I'd advocate for keeping them. But if we want to remove them, that's doable too :slightly_smiling_face:. 
We could generate output files from GDAL/RichDEM (slopes, curvatures, etc) in the `xdem-data` repo instead, then retrieve those and continue testing in `xdem` even without the dependencies.

vschaffn:
@adebardo @rhugonnet @duboise-cnes The PR has been updated, I have found a way to solve the GDAL configuration problem. Now the makefile works if the user has GDAL on the system (if not the makefile gives instructions on how to install it). We still have constraints on richdem with python > 3.10, so for the moment the makefile is only configured for python3.10.
A check has been added to `make test` so that it does not run if GDAL is not in the virtual environment.
I have kept "To use: source ${VENV}/bin/activate; xdem -h" to anticipate the CLI (#629)

CONTRIBUTING.md
```diff
@@ -38,6 +39,15 @@ mamba env create -f dev-environment.yml  # Add '-n custom_name' if you want.
...
 ```
 
+#### With `pip`
+```bash
+git clone https://github.com/GlacioHack/xdem.git
+cd xdem
+make install
+```
+
+Please note: pip installation is currently only possible under python3.10.
```
rhugonnet, line 49: Should we leave this README section to be only for users (for which `pip install xdem` should be sufficient?), and point towards the documentation for developer install?

rhugonnet: My bad, I just noticed this is already the dev installation in CONTRIBUTING!

Makefile
```diff
@@ -0,0 +1,112 @@
...
+.PHONY: install
+install: venv ## Install xDEM for development (depends on venv)
+	@test -f ${VENV}/bin/xdem || echo "Installing xdem in development mode"
+	@test -f ${VENV}/bin/xdem || ${VENV}/bin/pip install -e .[dev]
+	@test -f .git/hooks/pre-commit || echo "Installing pre-commit hooks"
+	@test -f .git/hooks/pre-commit || ${VENV}/bin/pre-commit install -t pre-commit
+	@test -f .git/hooks/pre-push || ${VENV}/bin/pre-commit install -t pre-push
+	@echo "Attempting to install GDAL..."
+	@make install-gdal
+	@echo "XDEM installed in development mode in virtualenv ${VENV}"
```
rhugonnet, line 71: "xdem" in lower-case to match syntax above

rhugonnet:
@vschaffn Perfect for the solve! Not much to comment on my side, looks great :)

setup.cfg
```diff
@@ -72,8 +73,9 @@ doc =
     myst-nb
     numpydoc
 dev =
+    pre-commit
```
rhugonnet, line 76: Wondering if we should move `pre-commit` to test dependencies (both here and in `dev-environment.yml`). 
I've had some feedback that is unpractical to have to add it separately when doing the `conda` dev installation `mamba env create -f dev-environment.yml`. Would be more practical?
I don't know if there's a recommended practice for this.

adebardo (approved):

adehecq:
I would love to have a meeting to discuss how to use this make file and how we can reconcile the conda vs pip ways. I like the idea of the make file to make it easier for developers to install and also to run tests, maybe build documentation etc. But at the moment I'm really used to conda and not pip + venv...

PR #650, opened 2024-11-15
Add Initial shift Capability to DEM Coregistration

### LINKED ISSUE #603: [POC]: Add an initial shift for coregistration
### Context

The CNES wants to integrate the ability to instantiate an initial offset for the co-registration calculation. Because the `dem_coregistration` function allows for easy addition of this feature and we will recommend its use to CNES members, we propose simple modifications at this level.

### Implementation

1. **Using the `translate` function from Geoutils**:

   The `translate` function from Geoutils allows for translating a Digital Elevation Model (DEM). We will call this function in the `dem_coregsitration` method .

```diff
def dem_coregistration(
    src_dem_path: str | RasterType,
    ref_dem_path: str | RasterType,
    out_dem_path: str | None = None,
    coreg_method: Coreg | None = NuthKaab() + VerticalShift(),
    grid: str = "ref",
    resample: bool = False,
    resampling: rio.warp.Resampling | None = rio.warp.Resampling.bilinear,
    shp_list: list[str | gu.Vector] | tuple[str | gu.Vector] | tuple[()] = (),
    inout: list[int] | tuple[int] | tuple[()] = (),
    filtering: bool = True,
    dh_max: Number = None,
    nmad_factor: Number = 5,
    slope_lim: list[Number] | tuple[Number, Number] = (0.1, 40),
    plot: bool = False,
    out_fig: str = None,
+   estimated_initial_shift : list = [0, 0] 
) -> tuple[DEM, Coreg, pd.DataFrame, NDArrayf]:
+	"""
+	explain estimated_initial_shift + affine only
+	"""

	    if isinstance(ref_dem_path, str):
	        if grid == "ref":
	            ref_dem, src_dem = gu.raster.load_multiple_rasters([ref_dem_path, src_dem_path], ref_grid=0)
	        elif grid == "src":
	            ref_dem, src_dem = gu.raster.load_multiple_rasters([ref_dem_path, src_dem_path], ref_grid=1)
	    else:
	        ref_dem = ref_dem_path
	        src_dem = src_dem_path
	        if grid == "ref":
	            src_dem = src_dem.reproject(ref_dem, silent=True)
	        elif grid == "src":
	            ref_dem = ref_dem.reproject(src_dem, silent=True)
+      if estimated_initial_shift != [0, 0] :
+          logging.warning("Initial shift in affine mode only")
+          estimated_initial_shift = [estimated_initial_shift[0] * pixel_resolution_x, estimated_initial_shift[0] * pixel_resolution_y]
+          src_dem.translate(estimated_initial_shift[0], estimated_initial_shift[1], inplace=True)
	.
	.
	.
    # Coregister to reference - Note: this will spread NaN
    coreg_method.fit(ref_dem, src_dem, inlier_mask)
    dem_coreg = coreg_method.apply(src_dem, resample=resample, resampling=resampling)
+   if estimated_initial_shift != [0, 0] :
+   	calculated_shift_x = dem_coreg.meta["shift_x"]
+       calculated_shift_y = dem_coreg.meta["shift_y"]

+		dem_coreg.meta["shift_x"] = calculated_shift_x + estimated_initial_shift[0] 
+		dem_coreg.meta["shift_y"] = calculated_shift_x + estimated_initial_shift[1]

```

### Tests
-  Set up a test for the `dem_coregistration` function with an initial offset value different from 0.  
[Here](https://github.com/GlacioHack/xdem/blob/main/tests/test_coreg/test_workflows.py#L226), a new paragraph should be added to test this updated version.

### Documentation

- Update the documentation to include examples on using the new `estimated_initial_shift` 

/estimation 3d

vschaffn:
Resolves #603.

## Description:
This PR introduces the ability to specify an initial shift (in X and Y) for DEM co-registration. The key changes include the integration of an `estimated_initial_shift` parameter in the `dem_coregistration()` function, allowing users to apply a custom translation (affine-based) to the source DEM before coregistration.

## Key Modifications:
1. **`dem_coregistration()` function:**
   - A new `estimated_initial_shift` parameter is added, allowing users to input a list representing the initial shifts along the X and Y axes.
   - If a shift is provided, the source DEM is translated using `geoutils.translate()` based on the calculated pixel resolution.
   - The `shift_x` and `shift_y` values in the coregistration method’s metadata (`meta['outputs']['affine']`) are updated to include the initial shifts.

2. **Raise `TypeError`** if the coregistration method or one of the coregistration method in the coregistration pipeline is not affine when an initial shift is provided.

2. **Recursive Shift Update Logic:**
   - A recursive `update_shift()` function is implemented to handle both `Coreg` and `CoregPipeline` objects, ensuring that the shift is propagated across all coregistration steps in a pipeline.
   - Metadata handling ensures shifts are correctly updated in all pipeline steps and for standalone `Coreg` objects.

3. **Unit Tests:**
   - Added a test to verify the correct behavior of the `estimated_initial_shift` parameter.
   - Added a test to ensure `TypeError` is raised an initial shift is provided and the coregistration method is not affine.

## How It Works:
- The `estimated_initial_shift` input, given as a list `[x_shift, y_shift]`, is multiplied by the DEM’s pixel resolution to calculate the translation in map units.
- This shift is then applied before coregistration, and the resulting `shift_x` and `shift_y` in the `meta` are updated accordingly.
- For `CoregPipeline`, this process is repeated for each coregistration step to ensure consistency.

## Example:
```python
dem_coregistration(
    src_dem_path="path/to/src_dem.tif",
    ref_dem_path="path/to/ref_dem.tif",
    estimated_initial_shift=[10, 5]
)
```
In this case, the source DEM is initially translated by 10 pixels in the X direction and 5 pixels in the Y direction before the coregistration process begins.

## Documentation
- Updated the function docstring to reflect the new `estimated_initial_shift` parameter and its usage.
- Update API documentation.

### DIFF
--- a/tests/test_coreg/test_workflows.py
+++ b/tests/test_coreg/test_workflows.py
@@ -320,3 +320,19 @@ class TestWorkflows:
                 estimated_initial_shift=test_shift_tuple,
                 random_state=42,
             )
+
+        # Check if the appropriate exception is raised with a wrong type initial shift
+        with pytest.raises(ValueError, match=r".*two numerical values.*"):
+            dem_coregistration(
+                tba_dem,
+                ref_dem,
+                estimated_initial_shift=["2", 2],
+                random_state=42,
+            )
+        with pytest.raises(ValueError, match=r".*two numerical values.*"):
+            dem_coregistration(
+                tba_dem,
+                ref_dem,
+                estimated_initial_shift=[2, 3, 5],
+                random_state=42,
+            )
--- a/tests/test_coreg/test_workflows.py
+++ b/tests/test_coreg/test_workflows.py
@@ -256,9 +256,67 @@ class TestWorkflows:
         out_fig.close()
 
         # Testing different coreg method
-        dem_coreg, coreg_method, coreg_stats, inlier_mask = dem_coregistration(
+        dem_coreg2, coreg_method2, coreg_stats2, inlier_mask2 = dem_coregistration(
             tba_dem, ref_dem, coreg_method=xdem.coreg.Deramp()
         )
-        assert isinstance(coreg_method, xdem.coreg.Deramp)
-        assert abs(coreg_stats["med_orig"].values) > abs(coreg_stats["med_coreg"].values)
-        assert coreg_stats["nmad_orig"].values > coreg_stats["nmad_coreg"].values
+        assert isinstance(coreg_method2, xdem.coreg.Deramp)
+        assert abs(coreg_stats2["med_orig"].values) > abs(coreg_stats2["med_coreg"].values)
+        assert coreg_stats2["nmad_orig"].values > coreg_stats2["nmad_coreg"].values
+
+        # Testing with initial shift
+        test_shift_list = [10, 5]
+        tba_dem_origin = tba_dem.copy()
+        coreg_pipeline = xdem.coreg.affine.NuthKaab() + xdem.coreg.affine.VerticalShift()
+
+        dem_coreg2, coreg_method2, coreg_stats2, inlier_mask2 = dem_coregistration(
+            tba_dem, ref_dem, coreg_method=coreg_pipeline, estimated_initial_shift=test_shift_list, random_state=42
+        )
+        dem_coreg3, coreg_method3, coreg_stats3, inlier_mask3 = dem_coregistration(
+            tba_dem, ref_dem, coreg_method=coreg_pipeline, random_state=42
+        )
+        assert tba_dem.raster_equal(tba_dem_origin)
+        assert isinstance(coreg_method2, xdem.coreg.CoregPipeline)
+        assert isinstance(coreg_method3, xdem.coreg.CoregPipeline)
+        assert isinstance(coreg_method2.pipeline[0], xdem.coreg.AffineCoreg)
+        assert isinstance(coreg_method3.pipeline[0], xdem.coreg.AffineCoreg)
+        assert (
+            coreg_method2.pipeline[0].meta["outputs"]["affine"]["shift_x"]
+            == coreg_method3.pipeline[0].meta["outputs"]["affine"]["shift_x"]
+        )
+        assert (
+            coreg_method2.pipeline[0].meta["outputs"]["affine"]["shift_y"]
+            == coreg_method3.pipeline[0].meta["outputs"]["affine"]["shift_y"]
+        )
+
+        # Testing without coreg pipeline
+        test_shift_tuple = (-5, 2)  # tuple
+        coreg_simple = xdem.coreg.affine.DhMinimize()
+
+        dem_coreg2, coreg_method2, coreg_stats2, inlier_mask2 = dem_coregistration(
+            tba_dem, ref_dem, coreg_method=coreg_simple, estimated_initial_shift=test_shift_tuple, random_state=42
+        )
+        dem_coreg3, coreg_method3, coreg_stats3, inlier_mask3 = dem_coregistration(
+            tba_dem, ref_dem, coreg_method=coreg_simple, random_state=42
+        )
+        assert isinstance(coreg_method2, xdem.coreg.AffineCoreg)
+        assert isinstance(coreg_method3, xdem.coreg.AffineCoreg)
+        assert coreg_method2.meta["outputs"]["affine"]["shift_x"] == coreg_method3.meta["outputs"]["affine"]["shift_x"]
+        assert coreg_method2.meta["outputs"]["affine"]["shift_y"] == coreg_method3.meta["outputs"]["affine"]["shift_y"]
+
+        # Check if the appropriate exception is raised with an initial shift and without affine coreg
+        with pytest.raises(TypeError, match=r".*affine.*"):
+            dem_coregistration(
+                tba_dem,
+                ref_dem,
+                coreg_method=xdem.coreg.Deramp(),
+                estimated_initial_shift=test_shift_tuple,
+                random_state=42,
+            )
+        with pytest.raises(TypeError, match=r".*affine.*"):
+            dem_coregistration(
+                tba_dem,
+                ref_dem,
+                coreg_method=xdem.coreg.Deramp() + xdem.coreg.TerrainBias(),
+                estimated_initial_shift=test_shift_tuple,
+                random_state=42,
+            )
--- a/xdem/coreg/workflows.py
+++ b/xdem/coreg/workflows.py
@@ -31,6 +31,7 @@ from geoutils._typing import Number
 from geoutils.raster import RasterType
 
 from xdem._typing import NDArrayf
+from xdem.coreg import CoregPipeline
 from xdem.coreg.affine import NuthKaab, VerticalShift
 from xdem.coreg.base import Coreg
 from xdem.dem import DEM
@@ -148,7 +149,7 @@ def dem_coregistration(
     src_dem_path: str | RasterType,
     ref_dem_path: str | RasterType,
     out_dem_path: str | None = None,
-    coreg_method: Coreg | None = None,
+    coreg_method: Coreg | CoregPipeline | None = None,
     grid: str = "ref",
     resample: bool = False,
     resampling: rio.warp.Resampling | None = rio.warp.Resampling.bilinear,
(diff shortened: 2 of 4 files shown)


xdem/coreg/workflows.py
```diff
@@ -243,6 +259,27 @@ def dem_coregistration(
     coreg_method.fit(ref_dem, src_dem, inlier_mask)
     dem_coreg = coreg_method.apply(src_dem, resample=resample, resampling=resampling)
 
+    # Add the initial shift to the calculated shift
+    if estimated_initial_shift:
+
+        def update_shift(
```
adebardo, line 265: We need a unit test for this function.

xdem/coreg/workflows.py
```diff
@@ -201,16 +205,28 @@ def dem_coregistration(
 
     if isinstance(ref_dem_path, str):
         if grid == "ref":
-            ref_dem, src_dem = gu.raster.load_multiple_rasters([ref_dem_path, src_dem_path], ref_grid=0)
+            ref_dem, src_dem = gu.raster.load_multiple_rasters([ref_dem_path, src_dem_path])
```
adebardo, line 208: Could you just tell me why you're removing the ref_grid?

vschaffn: with a ref_grid in parameter, the function load_multiple_rasters perform the reprojection of one of the other DEM. If a shift is given, it has to be perform before any reprojection. That's why I removed the ref_grid so the DEM are just loaded, then the shift is applied to src_dem if applicable, and then I added the reprojection step.

rhugonnet:
Quick note, the tests you are adding are in a `test_` function currently skipped because of an old segfault:
https://github.com/GlacioHack/xdem/blob/22dbe7c6a3020a5f556065008862842463933a4c/tests/test_coreg/test_workflows.py#L186
I see some code in this test that is now deprecated (for example using `==` to check raster equality was an old choice, now changed to `raster_equal()` to mirror NumPy and leave `==` for element-wise equality, as for arrays). I will update that quickly in a separate PR and see if I can make the original test pass :wink:

rhugonnet:
@vschaffn I managed to update the code for the test and unskip it. I merged it to main so that you can update with upstream and run it in this PR :wink:: https://github.com/GlacioHack/xdem/pull/651. 
(The segfault was likely not from us but a dependency, so I didn't have to do anything on this).

xdem/coreg/workflows.py
adebardo: great thanks

xdem/coreg/workflows.py
```diff
@@ -220,16 +224,28 @@ def dem_coregistration(
...
         ref_dem = ref_dem_path
         src_dem = src_dem_path
-        if grid == "ref":
-            src_dem = src_dem.reproject(ref_dem, silent=True)
-        elif grid == "src":
-            ref_dem = ref_dem.reproject(src_dem, silent=True)
+
+    # If an initial shift is provided, apply it before coregistration
+    if estimated_initial_shift:
+        logging.warning("Initial shift in affine mode only")
```
adebardo, line 236: Instead of this warning, should we rather add a raise in case we are not in an affine coregistration?

adebardo (commented):
I can't remember if we talked about adding a section in the documentation.

vschaffn:
@adebardo I added xdem.coreg.workflows.dem_coregistration in API documentation, and a raise when an estimated shift is provided but the coregistration method is not affine

xdem/coreg/workflows.py
```diff
@@ -173,7 +175,7 @@ def dem_coregistration(
     :param src_dem_path: Path to the input DEM to be coregistered
     :param ref_dem_path: Path to the reference DEM
     :param out_dem_path: Path where to save the coregistered DEM. If set to None (default), will not save to file.
-    :param coreg_method: Coregistration method or pipeline.
+    :param coreg_method: The xdem coregistration method, or pipeline.
```
rhugonnet, line 178: I think we can state simply "Coregistration method" here, it's implicit we're working with xDEM :wink:

xdem/coreg/workflows.py
```diff
@@ -263,6 +292,27 @@ def dem_coregistration(
     coreg_method.fit(ref_dem, src_dem, inlier_mask, random_state=random_state)
     dem_coreg = coreg_method.apply(src_dem, resample=resample, resampling=resampling)
 
+    # Add the initial shift to the calculated shift
+    if estimated_initial_shift:
+
+        def update_shift(
```
rhugonnet, line 298: Clever the recursive function to take care of the pipelines!

tests/test_coreg/test_workflows.py
```diff
@@ -256,9 +256,58 @@ def test_dem_coregistration(self) -> None:
...
+        assert coreg_stats2["nmad_orig"].values > coreg_stats2["nmad_coreg"].values
+
+        # Testing with initial shift
+        test_shift_list = [10, 5]
+        tba_dem_shift = tba_dem.copy()
+        dem_coreg2, coreg_method2, coreg_stats2, inlier_mask2 = dem_coregistration(
+            tba_dem_shift, ref_dem, estimated_initial_shift=test_shift_list
+        )
+        tba_dem_shift_test = tba_dem.translate(test_shift_list[0] * tba_dem.res[0], test_shift_list[1] * tba_dem.res[1])
+        assert tba_dem_shift.raster_equal(tba_dem_shift_test)
```
rhugonnet, line 273: Reading this test (and looking again at the code in `workflows`), I realize that the input `tba_dem` is shifted in-place, which thus modifies the input from the user.
Until now our practice has been to "never authorize the modification of any user input", unless the function has a clear `inplace` argument (when returning same type as self) or is called `set_xxx` (when re-setting an attribute).

If that makes sense to you as well, we should return a copy in the `workflows` function instead of modifying the input, I'll tag where :slightly_smiling_face:

xdem/coreg/workflows.py
```diff
@@ -216,21 +220,46 @@ def dem_coregistration(
...
+
+    # If an initial shift is provided, apply it before coregistration
+    if estimated_initial_shift:
+
+        # convert shift
+        shift_x = estimated_initial_shift[0] * src_dem.res[0]
+        shift_y = estimated_initial_shift[1] * src_dem.res[1]
+
+        # Apply the shift to the source dem
+        src_dem.translate(shift_x, shift_y, inplace=True)
```
rhugonnet, line 257: As stated above, here we'd need to remove the `inplace` op to not modify user input.

tests/test_coreg/test_workflows.py
```diff
@@ -256,9 +256,58 @@ def test_dem_coregistration(self) -> None:
...
+        )
+
+        # Testing without coreg pipeline
+        test_shift_tuple = (-5, 2)  # tuple
+        coreg_simple = xdem.coreg.affine.DhMinimize()
+
+        tba_dem_shift = tba_dem.copy()
+        dem_coreg2, coreg_method2, coreg_stats2, inlier_mask2 = dem_coregistration(
+            tba_dem_shift, ref_dem, coreg_method=coreg_simple, estimated_initial_shift=test_shift_tuple
+        )
```
rhugonnet, line 293: I think we might have to pass a `random_state=42` everywhere (all the `dem_coregistration` calls) to avoid possible random failures in CI in the future... :thinking: 

This is because the output shifts depend on both the initial shift and the one optimized by `DhMinimize`, which doesn't use all the data but uses a random subsample instead, and can lead to slightly different output. Same for all other methods.
Here it is falling on the exact same value probably because of the internal tolerance of the algorithm, but that might change with a different subsample, or under a different OS.

rhugonnet:
This is great, thanks! Just a couple comments above that can be solved quickly :smile:  

(Can also remove "xdem" in the PR title, to echo the comment on the argument description!)

vschaffn:
@rhugonnet many thanks for your feedbacks. 
- As copying the DEM at translation step would have triggered refactoring issues, I simply copied the DEM at the begging of the function (if a DEM is provided, if it is a path it is not applicable as the raster created in the function will not go through), so that the DEM is not modified by the function (I added a test to ensure that).
- For the tests I have noticed the random aspect in coregistration methods, but as I instantiate them before use, I have no trouble with that. But if the test does not pass with you I agree.
- It seems logical to keep only one coregistration function, as xdem is moving towards object-oriented design, it would make sense to keep `DEM.coregister_3d()`, but it means that it won't be possible anymore to compute coregistrations from paths only

rhugonnet:
@vschaffn Perfect!

> For the tests I have noticed the random aspect in coregistration methods, but as I instantiate them before use, I have no trouble with that.

Actually the `random_state` does not depend on instantiation, it's defined during `fit()`. So you will have a different randomized output with the same `Coreg` instance for different `fit()` calls! (or with different `Coreg` instances too; basically all the time except if you pass `random_state` to `fit()`).

There was a different small issue with `dem_coregistration` due the way the default argument was defined (that this discussion reminded me of), I used the opportunity to fix it: https://github.com/GlacioHack/xdem/pull/662

adebardo (approved):

PR #656, opened 2024-11-21
Add `dem` Xarray accessor and `epc` Pandas accessor mirroring `DEM`/`EPC` classes

### LINKED ISSUE #370: Overridde `DEM.crs` with `DEM.ccrs` to respect 3D CRSs
For instance useful if saving to file. Not a priority, but could be a good feature in the long term when 3D CRSs get more popular.

We'd have to test thoroughly if all types of 3D CRS can be processed (I don't think saving will work for grids, for example, we'd have to raise a `UserWarning` then).

> Regarding your suggestion to override crs with vcrs, do you mean to override crs with ccrs, or am I missing something? I'm still a bit new to vcrs so it might be a dumb question!

Yes, exactly, overriding it with `ccrs` so that it's always 3D! (then I guess we could remove `ccrs` and have it just be `crs`).

@adehecq: Curious about your feedback on this PR as well! 
FYI: It's a problem also of interest to the ICESat-2 SlideRule team that wants to integrate vertical transforms in their cloud processing, so we're discussing it a bit there too in parallel.

_Originally posted by @rhugonnet in https://github.com/GlacioHack/xdem/issues/350#issuecomment-1522538814_

### LINKED ISSUE #392: Add Xarray accessor `dem`
Big task here also! Mirroring https://github.com/GlacioHack/geoutils/issues/383.

Looks like the accessor name `dem` is available! :smile: 
See discussions started in: https://github.com/pydata/xarray/issues/8040, https://github.com/corteva/rioxarray/issues/687 and https://github.com/makepath/xarray-spatial/issues/790.

Important to coordinate with `xarray-spatial` to avoid too much redundancy.

### LINKED ISSUE #933: Add Dask support for terrain attributes
Mirroring the Multiproc implementation, doing it in #656

### LINKED ISSUE #934: Add Dask/Multiproc support for `to_vcrs` transformation
Doing it in #656

### LINKED ISSUE #935: Wrong unit shown in `workflow` plots
@belletva @marinebcht While revisiting this today, I realized that `crs.linear_units` is the horizontal units, not the vertical. The function cannot allow to differentiate vertical/horizontal axes. We need to use `axis.unit_name` after isolating the vertical axis (if it exists).

I'm fixing this in #656 with the 3D CRS override.

### LINKED ISSUE #999: Add Pandas accessor mirroring `EPC`
Mirrors https://github.com/GlacioHack/geoutils/issues/916

rhugonnet:
This PR adds the `dem` Xarray accessor mirroring the `DEM` class and `epc` Pandas accessor mirroring the `EPC` class, since the finalization of the `rst` accessor in https://github.com/GlacioHack/geoutils/pull/446 and `pc` accessor in https://github.com/GlacioHack/geoutils/pull/912, and Dask/Multiproc support for terrain attributes and vertical CRS transformation.
It adds Dask and Dask-GeoPandas as optional dependencies. LAS/LAZ/COPC support through EPC/epc in out-of-core manner is also supported thanks to #912.

More complex Dask/MP implementations (coregistration, uncertainty analysis) will be done in separate PRs (such as #759).

The accessor allows to access all attributes and run all methods already implemented for DEMs from a `xarray.DataArray` object (e.g., `ds.dem.slope()`, `ds.dem.coregister_3d()`), including functionalities from rasters inherited from the `rst` accessor (e.g., `ds.dem.reproject()`).
The setup follows a double inheritance between the `RasterAccessor` and `DEMBase`, so that the `dem` accessor can also run `reproject()`, etc.
Same for elevation point clouds.

Test structure is mirrored from GeoUtils, checking exact equality of outputs and respect of Dask/MP/eagerness of operations (loading status, laziness, etc), and in addition also checks correct inheritance.

Resolves #392
Resolves #370
Resolves #933 
Resolves #934 
Resolves #935 
Resolves #999

## Summary of changes

To understand the changes below, first see changes related to the `rst` accessor in GeoUtils: https://github.com/GlacioHack/geoutils/pull/446

As for the `rst` accessor, most of the contents of the `DEM` class are moved into a `DEMBase` parent class, inherited by both the `DEM` class and `dem` accessor class.

In this case, however, inheritance is more delicate because `DEM` also needs to inherit functionalities from `Raster` (that are not in `RasterBase`) and the `dem` accessor needs to inherit functionalities from `rst` (that are not in `RasterBase` either).
Therefore, the `DEM` class has dual-inheritance of `Raster` and `DEMBase`, while the `dem` accessor has dual inheritance of `rst` and `DEMBase`. This makes things slightly more complex when explicitly calling `super()`, which is only needed in a few occurences of core functions or `__init__`, but otherwise easy for other functionalities!

Similar as for the GeoUtils PR, the `from_array` functions allow to return a same-class object. In this case, only `DEM` needs to override with its own `from_array` (as `dem` already returns a `xarray.DataArray` through `rst`).

Same logic for `EPC` and `epc`.

### Code

- [x] Ensure dual support of masked arrays and NaN arrays for all DEM methods,
- [x] Add Dask support for all terrain attributes,
- [x] Add Dask/Multiprocessing support for `to_vcrs`,
- [x] Add tests specific to `DEMBase`,
- [x] Add tests specific to `DEMAccessor`,
- [x] Add tests specific to chunked implementation in terrain and vcrs modules.


marinebcht:
Yes, you right, as the `info()` function in xDEM is only used to add a layer to handle `.vcrs` when they are (were!) separate from `.crs` objects, it is not really usefull now.

Plus, the idea to correct the error reported in https://github.com/GlacioHack/geoutils/issues/805 (dataset 2) is to move the componed crs treatment into geoutils to handle CRSs when have directly a vertical axis.

My only suggestion s about your last point (crs with no vertical axis that I have mention in #805), either we print classical EPSG:XXXX for exemple or we print [ EPSG:XXXX, None ]. Perhaps, we could add a `xxx.crs_to_string()` function (and call directly it in `infos()`) in geoutils without the explicit `; NONE]`, and with it in xDEM, since we are dealing with actual elevation data. The only parameter I see is if the use of the VCRS become the norm in the futur.

rhugonnet:
@marinebcht Thanks! I'll do that then: Update GeoUtils, have it inherited in xDEM. We can still adjust later if required.

Reminder: About 2 weeks to give feedback here. I'll consolidate and merge this PR soon after consolidating GeoUtils' mirror Xarray PR and releasing there.

rhugonnet:
This PR is pretty much finished. I need to merge https://github.com/GlacioHack/geoutils/pull/925 first, on it.

@marinebcht: For info, once this PR is merged, we are temporarily testing against GeoUtils' main branch in CI. This way we can ensure all is compatible up to the moment of release.

rhugonnet:
Should be mergeable now with https://github.com/GlacioHack/geoutils/pull/925 done

rhugonnet:
All passing locally, merging.

adehecq:
Amazing! 😍

PR #657, opened 2024-11-21
Move license header files to avoid GitHub misrecognizing license

rhugonnet:
This tiny PR moves the license header files to a different folder.

Since #617, GitHub confuses the license file due to the license header file being in root, and shows a weird message as a result.

Screenshot from 2024-11-20 16-15-18

Solving it by simply moving the header + script to `.github/scripts` with other dev resources used for `pre-commit` or CI!


rhugonnet:
FYI @vschaffn for the small change :wink:

Issue #584, opened 2024-09-06
Fix `BlockwiseCoreg` errors


rhugonnet:
Since #530, it looks like `BlockwiseCoreg` has an issue:  `apply` seems wrong, in the documentation built in #502 example it creates more errors than it corrects (wrong direction?).
Additionally, it fails with `BiasCorr` method during the `apply` step.

rhugonnet:
Additionally, would be a good moment to add more tests to avoid this happening again in the future.

rhugonnet:
OK, the cause is indeed that `BlockwiseCoreg` either applies or estimates the horizontal shift in the wrong direction contrary to classic `apply` (vertical unaffected), I don't understand why yet...

rhugonnet:
Even when the coregistration is applied in the right direction (which is not the case by default for `BlockwiseCoreg` somehow, even after #585, so probably an error in the original implementation), the `warp_dem` function creates weird artefacts...
See example below that I had written in the documentation (temporarily removed for now):

> ## Dividing coregistration in blocks
> 
> ### The {class}`~xdem.coreg.BlockwiseCoreg` object
> 
> Sometimes, we want to split a coregistration across different spatial subsets of an elevation dataset, running that 
> method independently in each subset. A {class}`~xdem.coreg.BlockwiseCoreg` can be constructed for this:
> 
> ```{code-cell} ipython3
> blockwise = xdem.coreg.BlockwiseCoreg(xdem.coreg.NuthKaab(), subdivision=16)
> ```
> 
> The subdivision corresponds to an equal-length block division across the extent of the elevation dataset. It needs 
> to be a number of the form 2{sup}`n` (such as 4 or 256).
> 
> It is run the same way as other coregistrations:
> 
> ```{code-cell} ipython3
> # Run 16 block coregistrations
> aligned_dem = blockwise.fit_and_apply(ref_dem, tba_dem_shifted)
> ```
> 
> ```{code-cell} ipython3
> :tags: [hide-input]
> :mystnb:
> :  code_prompt_show: "Show plotting code"
> :  code_prompt_hide: "Hide plotting code"
> 
> # Plot before and after
> f, ax = plt.subplots(1, 2)
> ax[0].set_title("Before block NK")
> (tba_dem_shifted - ref_dem).plot(cmap='RdYlBu', vmin=-30, vmax=30, ax=ax[0])
> ax[1].set_title("After block NK")
[...]

PR #665, opened 2024-12-06
Fix `BlockwiseCoreg` shift direction

rhugonnet:
Partially solves #584.
It looks like the `warp_dem` function was simply in the wrong horizontal direction. This is because the `NuthKaab` metadata used to be the opposite before the homogenization in #530, and it is only tested on that method.

Re-adds `BlockwiseCoreg` to the documentation, with a warning about the functionality being experimental.
This PR does not improve tests (only did visual checking on our example output), this should be done when solving the rest of #584.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,5 +153,6 @@ doc/source/gen_modules/
 doc/source/sg_execution_times.rst
 examples/basic/temp.tif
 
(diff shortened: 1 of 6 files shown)


rhugonnet:
This small fix and warning are sufficient to release `0.1`, we can dive into the small artefacts of #584 separately.

PR #670, opened 2024-12-10
Add `apply_z_correction` Parameter to `warp_dem` Function

### LINKED ISSUE #643: [POC]: Configuration of the z offset applied during coregistration : possibility to deactivate the vertical shift
# Context

The purpose of this ticket is to propose a solution that allows the user to control altitude adjustment after coregistration.

# Code
The choice of whether to apply vertical correction or not is made [here](https://github.com/GlacioHack/xdem/blob/main/xdem/coreg/base.py#L3476), and its application is performed [here](https://github.com/GlacioHack/xdem/blob/main/xdem/coreg/base.py#L3538). However, we would like this condition to be user-configurable.

In practice, this condition rarely, if ever, fails, meaning that the Z-offset is almost always applied. However, in the context of QA studies, this adjustment is not always relevant. The simplest proposal is to add an argument to the function containing this condition, which can be toggled using a boolean. Depending on its state, the Z-adjustment will either be applied or skipped. This boolean can default to `True`.

# Testing

Ensure to update any tests that may be affected and test with both boolean states.

# Documentation

Update the documentation accordingly.

# Extra

If possible, create a notebook demonstrating the use of this adjustment, including a before-and-after altitude comparison.

# Estimation

2 days

vschaffn:
Resolves #643.

## Description
This pull request introduces a new parameter, `apply_z_correction`, to the `warp_dem` function. This enhancement allows users to control whether vertical correction (Z-offset) is applied during the DEM warping process. The default behavior remains the same, with the Z-correction applied. However, users can now disable this correction by setting `apply_z_correction=False`.

## Changes
- `apply_z_correction` parameter: A boolean parameter was added to the `warp_dem` function and the `BlockwiseCoreg` class to allow toggling of Z-offset application. When set to False, the Z-adjustment step is skipped during the warping process.

## Test Update
- The test for `warp_dem` has been updated to ensure that the function works as expected when `apply_z_correction=False`.
-  The test also verifies that the function correctly undoes the transformation when the Z-correction is disabled, ensuring the output remains consistent with the original DEM.

### DIFF
--- a/tests/test_coreg/test_base.py
+++ b/tests/test_coreg/test_base.py
@@ -1204,6 +1204,37 @@ def test_warp_dem() -> None:
     # Due to the randomness, the threshold is quite high, but would be something like 10+ if it was incorrect.
     assert spatialstats.nmad(dem - untransformed_dem) < 0.5
 
+    # Test with Z-correction disabled
+    transformed_dem_no_z = coreg.base.warp_dem(
+        dem=dem,
+        transform=transform,
+        source_coords=source_coords,
+        destination_coords=dest_coords,
+        resampling="linear",
+        apply_z_correction=False,
+    )
+
+    # Try to undo the warp by reversing the source-destination coordinates with Z-correction disabled
+    untransformed_dem_no_z = coreg.base.warp_dem(
+        dem=transformed_dem_no_z,
+        transform=transform,
+        source_coords=dest_coords,
+        destination_coords=source_coords,
+        resampling="linear",
+        apply_z_correction=False,
+    )
+
+    # Validate that the DEM is now more or less the same as the original, with Z-correction disabled.
+    # The result should be similar to the original, but with no Z-shift applied.
+    assert spatialstats.nmad(dem - untransformed_dem_no_z) < 0.5
+
+    # The difference between the two DEMs should be the vertical shift.
+    # We expect the difference to be approximately equal to the average vertical shift.
+    expected_vshift = np.mean(dest_coords[:, 2] - source_coords[:, 2])
+
+    # Check that the mean difference between the DEMs matches the expected vertical shift.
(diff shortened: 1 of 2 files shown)


tests/test_coreg/test_base.py
```diff
@@ -1204,6 +1204,30 @@ def test_warp_dem() -> None:
...
+        transform=transform,
+        source_coords=dest_coords,
+        destination_coords=source_coords,
+        resampling="linear",
+        apply_z_correction=False,
+    )
+
+    # Validate that the DEM is now more or less the same as the original, with Z-correction disabled.
+    # The result should be similar to the original, but with no Z-shift applied.
+    assert spatialstats.nmad(dem - untransformed_dem_no_z) < 0.5
```
rhugonnet, line 1229: The assertion from this transformation/untransformation would likely work without `apply_z_correction=False` as well? It's still a good check to have for the reversibility, but it doesn't show that `apply_z_correction` alone works.
For this, we could compare existing outputs with the same horizontal transformation but `apply_z_correction` as `False` or `True`.
The difference between the two DEMs should then be only the vertical shift, and close to the average of vertical shift?
Something like:
```python
# Not sure if I if we can do better than the mean here? Maybe I have the sign wrong too
expected_vshift = np.mean(dest_coords[:, 2] - source_coords[:, 2])  
assert np.mean(transformed_dem - transformed_dem_no_z) == pytest.approx(expected_vshift, rel=0.1)
```

rhugonnet:
Great! Just a small comment on the tests! :slightly_smiling_face: 

On the topic of blockwise: Did you get any new idea about the edge artefacts that it creates while working on it, [see example here](https://github.com/GlacioHack/xdem/issues/584#issuecomment-2345024069)?
I wonder if we should try to find the source of those (that are probably from `warp_dem`) before improving the functionality further (in case we have to scrap existing code).

adebardo (approved):

PR #674, opened 2024-12-20
[WIP] Implement linting with Ruff

### LINKED ISSUE #550: [POC] linting with pylint
# Context

The goal of this ticket is to implement linting on the code. We have noticed that certain rules applied to demcompare are not applied in the files of xDEM. 

Examples of errors present in the test files:
- `tests/test_volume.py:252:11: W0125: Using a conditional statement with a constant value (using-constant-test)`
- `tests/test_spatialstats.py:563:12: W0612: Unused variable 'ny' (unused-variable)`
- `tests/test_spatialstats.py:509:0: C0115: Missing class docstring (missing-class-docstring)`
- `tests/test_spatialstats.py:547:4: R0914: Too many local variables (27/15) (too-many-locals)`
- `tests/test_vcrs.py:22:8: C0206: Consider iterating with .items() (consider-using-dict-items)`

We cannot guarantee that the same code quality is applied to all the code. Therefore, we propose to add the pylint tool to the code dependencies and correct the errors highlighted.

# Tasks

- [ ] Explain why pylint is in the "test" dependencies in the [setup](https://github.com/GlacioHack/xdem/blob/main/setup.cfg) file and see if it's the most appropriate place for it
- [ ] Add a .pylintrc file
    - [ ] Use the one from demcompare (Amaury, if you can take a look and see if you want to modify anything)
```diff
# Maximum number of characters on a single line.
- max-line-length=80
+ max-line-length=120
- # Maximum number of lines in a module.
- max-module-lines=700
```
- [ ] Add a pylint dependency in the .pre-commit-config.yaml file (same config as demcompare)
- [ ] Correct the errors returned
    - [ ] If possible
    - [ ] Otherwise, add the # ignore and list them in a document
- [ ] Add pylint to CI github actions

quinto-clara:
Resolves [#550](https://github.com/GlacioHack/xdem/issues/550)

## Description : 

The Pull Request implements linting on xDEM in order to guarantee the same code quality as demcompare. 
The idea was to limit the number of tools and to apply the same rules as demcompare. 
That's why we choose to implement Ruff in Xdem, a fast python linter and code formatter gathering a lot of functionalities and replacing pylint, black, flake8 and isort. 

## Key changes : 

- Ruff configuration : 

    - [x] ruff.toml file addded including the following rules : 

       - Target version of python : 3.10 
       - The formatter will wraps lines at a length of 120 and use 4 spaces per tabulation when a line exceed 120 characters. 
       - Ruff will ignore 46 lint rules (for now)

    - [x] Ruff added in makefile 
    - [x] Ruff added in CI file 
    - [x] Ruff added in setup.cfg file 
    

- Code correction : 

    Ruff returned a lot of errors related to 108 lint rules : 

    - [x] 62 of them has been fixed, the remaining is ignored and can be fixed later

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,58 +30,39 @@ repos:
           hooks:
                 - id: absolufy-imports
 
-        # Format the code aggressively using black
-        - repo: https://github.com/psf/black
-          rev: 24.10.0
+        # Ruff linter and code formatter
+        - repo: https://github.com/astral-sh/ruff-pre-commit
+          # Ruff version.
+          rev: v0.8.6
           hooks:
-                  - id: black
-                    args: [--line-length=120]
+            # Run the linter.
+            - id: ruff
+            # Run the formatter.
+            - id: ruff-format
 
-        # Lint the code using flake8
-        - repo: https://github.com/pycqa/flake8
-          rev: 7.1.1
+        # Enable lint fixes with ruff
+        - repo: https://github.com/astral-sh/ruff-pre-commit
+          # Ruff version.
+          rev: v0.8.6
           hooks:
-                - id: flake8
-                  # More than one argument in the second list, so need to pass arguments as below (and -- to finish)
-                  args: [
-                          '--max-line-length', '120',  # we can write dicts however we want
-                          '--extend-ignore', 'E203,C408,B028', # flake8 disagrees with black, so this should be ignored.
-                          '--'
-                  ]
-                  additional_dependencies:
-                          - flake8-comprehensions
-                          - flake8-bugbear
-                  files: ^(xdem|tests)
+            # Run the linter.
+            - id: ruff
+              args: [ --fix ]
+            # Run the formatter.
+            - id: ruff-format
 
-        # Lint the code using mypy
-        - repo: https://github.com/pre-commit/mirrors-mypy
-          rev: v1.13.0
+        # To run ruff over Jupyter Notebooks
+        - repo: https://github.com/astral-sh/ruff-pre-commit
+          # Ruff version.
+          rev: v0.8.6
           hooks:
-                - id: mypy
-                  args: [
-                        --config-file=mypy.ini,
-                        --strict,
-                        --implicit-optional,
-                        --ignore-missing-imports,  # Don't warn about stubs since pre-commit runs in a limited env
-                        --allow-untyped-calls,  # Dynamic function/method calls are okay. Untyped function definitions are not okay.
-                        --show-error-codes,
-                        --no-warn-unused-ignores,  # Ignore 'type: ignore' comments that are not used.
-                        --disable-error-code=attr-defined,  # "Module has no attribute 'XXX'" occurs because of the pre-commit env.
-                        --disable-error-code=name-defined,  # "Name 'XXX' is not defined" occurs because of the pre-commit env.
-                        --disable-error-code=var-annotated,
-                        --disable-error-code=no-any-return
-
-                  ]
-                  additional_dependencies: [tokenize-rt==3.2.0, numpy==1.26]
-                  files: ^(xdem|tests|doc/code)
-
-
-        # Sort imports using isort
-        - repo: https://github.com/PyCQA/isort
-          rev: 5.13.2
-          hooks:
-                  - id: isort
-                    args: ["--profile", "black"]
+            # Run the linter.
+            - id: ruff
+              types_or: [ python, pyi, jupyter ]
+              args: [ --fix ]
+            # Run the formatter.
+            - id: ruff-format
+              types_or: [ python, pyi, jupyter ]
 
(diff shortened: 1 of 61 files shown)


vschaffn:
In the tests, the protected attributes (like `_need_vars`) were changed to non-protected attributes (for example, `need_vars`). This is causing issues because the non-protected attributes may not actually exist in the class in the same form, which leads to errors. Protected attributes (with a leading underscore) are intentionally kept as internal members of the class and might not have corresponding non-protected versions.

It seems that these changes could have been made in response to warnings from Ruff or due to the use of the --fix option. However, the removal of the leading underscore breaks the logic, since the non-protected version of these attributes may not exist.

To fix this:
- The protected attributes should be restored with their leading underscores.
- We may also want to review the Ruff configuration to ensure it's not suggesting incorrect automatic changes.

vschaffn:
To apply ruff in pre-commits and avoid conflicts with other linters, you would need to add ruff to `.pre-commit-config.yaml` and remove isort, pylint, flake8 and black.

xdem/spatialstats.py
```diff
@@ -826,20 +816,19 @@ def infer_heteroscedasticity_from_stable(
...
     """
-
     # Create placeholder variables names if those don't exist
     if list_var_names is None:
         list_var_names = ["var" + str(i + 1) for i in range(len(list_var))]
 
     # Get the arrays for proxy values and explanatory variables
     list_all_arr, gsd = _preprocess_values_with_mask_to_array(
-        values=[dvalues] + list_var, include_mask=stable_mask, exclude_mask=unstable_mask, preserve_shape=False
+        values=[dvalues, list_var], include_mask=stable_mask, exclude_mask=unstable_mask, preserve_shape=False,
```
vschaffn, line 825: This modifications triggered an issue : it creates a nested list where `list_var` is treated as a single element, resulting in [dvalues, [list_var]].

xdem/spatialstats.py
```diff
@@ -1007,7 +988,7 @@ def _aggregate_pdist_empirical_variogram(
         # Define subsampling parameters
         list_inside_radius = []
         list_outside_radius: list[float | None] = []
-        binned_ranges = [0.0] + pdist_multi_ranges
+        binned_ranges = [0.0, pdist_multi_ranges]
```
vschaffn, line 991: same here

xdem/spatialstats.py
```diff
@@ -3104,7 +3027,7 @@ def plot_variogram(
                 first_xmin = np.min(df.lags) / 2
             else:
                 first_xmin = 0
-            xscale_range_split = [first_xmin] + xscale_range_split
+            xscale_range_split = [first_xmin, xscale_range_split]
```
vschaffn, line 3030: same here

xdem/spatialstats.py
```diff
@@ -3135,7 +3058,7 @@ def plot_variogram(
         ax0.set_xticks([])
 
         # Plot the histogram manually with fill_between
-        interval_var = [0] + list(df.lags)
+        interval_var = [0, list(df.lags)]
```
vschaffn, line 3061: same here

xdem/spatialstats.py
```diff
@@ -3160,7 +3083,7 @@ def plot_variogram(
         ax1 = ax.inset_axes(grid[3:, xgridmin[k] : xgridmax[k]].get_position(fig).bounds)
 
         # Get the lags bin centers
-        bins_center = np.subtract(df.lags, np.diff([0] + df.lags.tolist()) / 2)
+        bins_center = np.subtract(df.lags, np.diff([0, df.lags.tolist()]) / 2)
```
vschaffn, line 3086: same here

rhugonnet:
Great! :star_struck: 
Tell me if you need more feedback on this PR :)

Question: I heard that Ruff was a "drop-in replacement" for black, flake8, etc (so producing almost exactly the same output, just much much faster). So most of the changes should arise mostly from updating black and other to more recent versions, or different rules, right?
(Out of curiosity, did you try to mirror the current configuration to check that there were almost no changes?)

.pre-commit-config.yaml
```diff
@@ -30,58 +30,39 @@ repos:
...
-                          '--extend-ignore', 'E203,C408,B028', # flake8 disagrees with black, so this should be ignored.
-                          '--'
-                  ]
-                  additional_dependencies:
-                          - flake8-comprehensions
-                          - flake8-bugbear
-                  files: ^(xdem|tests)
-
-        # Lint the code using mypy
-        - repo: https://github.com/pre-commit/mirrors-mypy
```
roelofvandijkO, line 57: `Ruff` does not replace `mypy`

roelofvandijkO:
This PR would become much easier to merge if you disable most ruff rules and re-enable them one by one.

rhugonnet:
Closing, we'll re-do in a new PR.

PR #676, opened 2025-01-09
Update documentation for ROI and stats

### LINKED ISSUE #602: [POC] Simplify the creation of an ROI for a user.
# Context

The goal of this task is to enable the creation of an ROI (Region of Interest) in **PIXEL** when loading a DEM. Although this ticket is proposed in the xDEM repository, the developments are intended for implementation in the geoUtils project.

To achieve this, we propose adding a parameter to the DEM creation [[here](https://github.com/GlacioHack/xdem/blob/main/xdem/dem.py#L94)](https://github.com/GlacioHack/xdem/blob/main/xdem/dem.py#L94). Using rasterio's `Window` option, we will load only the desired area. The resulting API would look like this:  
```python
dem = xdem.DEM("path/to/dem", roi={"left": "xmin", "bottom": "ymin", "right": "xmax", "top": "ymax"})
```

# Code

1. **Modify the `RASTER` class**
   - Add the parameter `self._ROI`, which can be either a dictionary or `None`.
   - Handle cases where the file is on disk: [here](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L426).
   - Case 2 : DEM as dictionnary, add  ROI if existing and adapt  `Raster.from_array` .
   - Case 3 : If already a raster, add  ROI if existing and adapt.

2. **Update data loading**  
   - Modify this function: [here](https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L166)(https://github.com/GlacioHack/geoutils/blob/main/geoutils/raster/raster.py#L166).
   - Add an ROI parameter.
   - Adjust the management of the `window` variable.

   Example of window creation:
   ```python
   w = src.read(1, window=Window(0, 0, 512, 256))
   ```

# Tests + Documentation Update**

Tests and documentation must be updated to reflect these changes.

vschaffn:
Resolves #602.
Completes GlacioHack/geoutils#642.
Completes GlacioHack/geoutils#638.

## Description
This PR introduces the ability to use a Region of Interest (ROI) when working with `xdem.DEM`, which is a subclass of `geoutils.Raster`. This functionality allows users to focus on specific regions within the raster data without having to load or process the entire dataset.

## Documentation
- A section about statistics has been added to the `DEM` class documentation.
- A section about the ROI has been added to the `DEM` class documentation.

## Example
1. Statistics:
- Get all statistics in a dict:
```{code-cell} ipython3
dem.get_stats()
```

- Get a single statistic (e.g., 'mean') as a float:
```{code-cell} ipython3
dem.get_stats("mean")
```

- Get multiple statistics in a dict:
```{code-cell} ipython3
dem.get_stats(["mean", "max", "rmse"])
```

- Using a custom callable statistic:
```{code-cell} ipython3
def custom_stat(data):
    return np.nansum(data > 100)  # Count the number of pixels above 100
dem.get_stats(custom_stat)
```

Note: as `get_stats()` is a raster method, it can also be used for terrain attributes:
```{code-cell} ipython3
dem.slope().get_stats()
```

## Note
- This PR has to be merged after the next release of geoutils containing GlacioHack/geoutils#642, GlacioHack/geoutils#638.
- The geoutils version in the virtual environment configuration files (`environment.yml`, `dev-environment.yml`, ...) would have to be updated.

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -13,7 +13,7 @@ dependencies:
   - tqdm
   - scikit-image=0.*
   - scikit-gstat>=1.0.18,<1.1
-  - geoutils=0.1.12
+  - geoutils=0.1.13
   - affine
   - pandas
   - pyogrio
--- a/doc/source/dem_class.md
+++ b/doc/source/dem_class.md
@@ -136,6 +136,38 @@ slope.plot(cmap="Reds", cbar_title="Slope (°)")
 For the full list of terrain attributes, see the {ref}`terrain-attributes` page.
 ```
 
+## Statistics
+The `get_stats()` method allows to extract key statistical information from a raster in a dictionary.
+Supported statistics are : mean, median, max, mean, sum, sum of squares, 90th percentile, nmad, rmse, std.
+Callable functions are supported as well.
+
+- Get all statistics in a dict:
+```{code-cell} ipython3
+dem.get_stats()
+```
+
+- Get a single statistic (e.g., 'mean') as a float:
+```{code-cell} ipython3
+dem.get_stats("mean")
+```
+
+- Get multiple statistics in a dict:
+```{code-cell} ipython3
+dem.get_stats(["mean", "max", "rmse"])
+```
+
+- Using a custom callable statistic:
+```{code-cell} ipython3
+def custom_stat(data):
+    return np.nansum(data > 100)  # Count the number of pixels above 100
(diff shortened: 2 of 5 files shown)


rhugonnet:
All good for me! I don't have any other comments than the ones that are in the GeoUtils' PR.

adebardo (approved):

PR #680, opened 2025-02-06
fix: makefile test section

adebardo:
# Context  

We realized that we forgot to remove the test for the GDAL installation in the Makefile during our last PR.  

This one fixes that oversight.


rhugonnet:
Good catch! I hadn't noticed GDAL was also in the test section :sweat_smile:

(As the PR title will become the name of the commit to the main during squash-merge, don't forget to rename it into something easy to grasp like "Remove GDAL from Makefile tests", to facilitate navigation of commits on the main branch! :wink:)

adebardo:
Thanks @rhugonnet, is it possible for you to add your review ?

rhugonnet:
Ah yes forgot to approve, sorry, done!

PR #682, opened 2025-02-06
Add Governance and code of conduct files

adebardo:
Now that CNES and GlacioHack are working together to maintain xDEM, it is necessary to propose a governance model to ensure that the current collaboration continues smoothly, and if possible, to increase the number of contributors who would feel reassured with the establishment of a structured working framework.

To this end, we have added two files: one for governance and one for the code of conduct.

These files are inspired by the MVG (Minimum Viable Governance) project [https://github.com/github/mvg](https://github.com/github/mvg) and [https://www.contributor-covenant.org/](https://www.contributor-covenant.org/).

We can use this PR to discuss any changes that need to be made.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -11,34 +11,43 @@ contributions are subject to the project's copyright under the terms of the
 
 This file keeps track of authors contributions.
 
+## Maintainers / Steering Comittee
+
+---
+
+- **Romain Hugonnet** [@rhugonnet](https://github.com/rhugonnet) <romain.hugonnet@gmail.com>
+- **Amaury Dehecq** [@adehecq](https://github.com/adehecq)
+- **Erik Schytt Mannerfelt** [@erikmannerfelt](https://github.com/erikmannerfelt)
+- **Emmanuel Dubois** [@duboise-cnes](https://github.com/duboise-cnes) <emmanuel.dubois@cnes.fr>
+- **Valentine Bellet** [@Valentine-Bellet](https://github.com/valentine-bellet) <valentine.bellet@cnes.fr>
+- **Alice de Bardonnèche-Richard** [@adebardo](https://github.com/adebardo) <alice.de-bardonneche-richard@cs-soprasteria.com>
+
+
 ## Development Lead
 
 ---
 
 - **Romain Hugonnet** [@rhugonnet](https://github.com/rhugonnet) <romain.hugonnet@gmail.com>
-- **Amaury Dehecq** [@adehecq](https://github/adehecq) <amaury.dehecq@univ-grenoble-alpes.fr>
-- **Erik Schytt Mannerfelt** [@erikmannerfelt](https://github/erikmannerfelt)
+- **Amaury Dehecq** [@adehecq](https://github.com/adehecq)
+- **Erik Schytt Mannerfelt** [@erikmannerfelt](https://github.com/erikmannerfelt)
+
 
 ## Contributors
 
 ---
 
-- **Friedrich Knuth** [@friedrichknuth](https://github/friedrichknuth)
-- **Andrew Tedstone** [@atedstone](https://github/atedstone)
-- **Zhihao LIU** [@liuh886](https://github/liuh886)
-- **Diego Cusicanqui** [@cusicand](https://github/cusicand)
-- **Alessandro Gentilini** [@alessandro-gentilini](https://github/alessandro-gentilini)
-- **Ferdinand Schenck** [@fnands](https://github/fnands)
-- **Johannes Landmann** [@jlandmann](https://github/jlandmann)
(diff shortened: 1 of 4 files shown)


rhugonnet:
Perfect! It looks good to me :slightly_smiling_face: 

I have two remarks/questions about the Governance file:
- The "Organization's Steering Committee" is mentioned twice, but not defined (I see that https://github.com/github/mvg?tab=readme-ov-file has a special file `STEERING_COMMITTEE.md` for it). In our case, is it meant to be defined somewhere, or to be removed entirely (given that we have few maintainers, it might be somewhat redundant?),
- We define the list of "Maintainers" from the `AUTHORS.md` file here (while https://github.com/github/mvg?tab=readme-ov-file as a specific `MAINTAINERS.md` file). This makes sense, we don't want to duplicate `AUTHORS.md` in another file! Should we then update/add a header in our `AUTHORS.md` to make it clear who are the "Maintainers". Right now we have "Development lead", "Contributors", and "Original developers". We could make "Development Lead" be "Maintainers" (which would be the occasion to move CNES/CS there), or keep those three separate and create "Maintainers" as a fourth category?

adebardo:
Hi!  
Thanks for the feedback. I've added the two minimum files recommended by Sébastien. For now, given our small team, the **STEERING_COMMITTEE.md** would indeed be redundant. However, we can indicate that the maintainers form this committee.  

Along the same lines of avoiding having to many files to maintain, adding more sections within the **Authors** file seems like a good approach. I'll take care of it :)

duboise-cnes:
Thanks for the work. I think it is a good first "simple" organization and ok for me. Thanks @rhugonnet for the clarification in authors.md, i agree with that. So ok for me to merge, to be tested and updated if something must evolve ;)
Cheers

AUTHORS.md
```diff
@@ -11,14 +11,26 @@ contributions are subject to the project's copyright under the terms of the
...
 This file keeps track of authors contributions.
 
+## Maintainers / Steering Comittee
+
+---
+
+- **Romain Hugonnet** [@rhugonnet](https://github.com/rhugonnet) <romain.hugonnet@gmail.com>
+- **Amaury Dehecq** [@adehecq](https://github/adehecq)
+- **Erik Schytt Mannerfelt** [@erikmannerfelt](https://github/erikmannerfelt)
+- **Emmanuel Dubois** [@duboise-cnes](https://github/duboise-cnes) <emmanuel.dubois@cnes.fr>
```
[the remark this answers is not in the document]
duboise-cnes: thanks

CODE-OF-CONDUCT.md
```diff
@@ -0,0 +1,132 @@
...
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+[INSERT CONTACT METHOD].
```
atedstone, line 63: The contact method requires updating.

adebardo: Thanks :)

doc/source/authors.md
adehecq, line 1: Instead of completely removing the file, should we simply point towards the repository file?

adehecq:
Thanks for adding the files and making a first proposition! I think it is fine as it is for now.

But looking at the MVG project, this also raises a few questions/issues, especially regarding the parent organization (currently GlacioHack):
- should we add a governance file and define a steering committee at the organization level? I see that the governance model is slightly different with a vote in case no consensus is found for example.
- in that case, the steering committee should be removed from xdem. The same would also apply to geoutils. A project should only have maintainers and contributors.
- should we do some cleaning in the organization? There are several empty/deprecated repositories.
- should we rename the organization or create a new one?

doc/source/authors.md
adebardo: Indeed, it has changed.

adebardo:
@adehecq @rhugonnet okay for you ?

rhugonnet:
All good for me! Tagging @erikmannerfelt as well, and another reminder by email.

erikmannerfelt:
Hi all. Sorry for my slow reply. If you all think that this is the right approach then I support it.

adehecq:
> @adehecq @rhugonnet okay for you ?

Yes, ok for me too. We can address the issues of the organization later.

adehecq (approved):
Formally approve

duboise-cnes:
Thanks for the work of the team and the acceptance of this first governance status, conjoint work of the study. A good and needed point for the project in this synergy with CNES, in my opinion. Great ! to be updated with future real organization questions for this project !

PR #688, opened 2025-02-21
Cluster Classes and Testing Implementation

### LINKED ISSUE #681: [POC] cluster module
# Context  

The goal of this ticket is to set up a cluster in xDEM to enable distributed computing.  
This ticket focuses only on creating a multiprocessing cluster, but it is designed in a way that will allow Dask integration in the future if needed.  

# Implementation  

This cluster is simple.  
First, we propose creating a `cluster.py` module at the root of the source code, where we will add the internal proof-of-concept (PoC) code.  

```python
class ClusterGenerator:
    def __new__(self, name, nb_workers=2):
        if name == "basic":
            self.cluster = BasicCluster
        else:
            self.cluster = MpCluster(conf={"nb_workers": nb_workers})
        return self.cluster

class AbstractCluster:
    def __init__(self):
        self.pool = None
    
    def __enter__(self):
        return self
    
    def __exit__(self, exc_type, exc_val, exc_tb):
        self.close()
    
    def close(self):
        pass
    
    def launch_task(self, fun, args=[], kwargs={}):
        pass
    
    def get_res(self, future):
        return future
    
    def return_wrapper(self):
        pass
    
    def tile_retriever(self, res):
        pass

class BasicCluster(AbstractCluster):
    def launch_task(self, fun, args=[], kwargs={}):
        return fun(*args, **kwargs)
        
class MpCluster(AbstractCluster):
    def __init__(self, conf=None):
        nb_workers = 1
        if conf is not None:
            nb_workers = conf.get("nb_workers", 1)
        ctx_in_main = multiprocessing.get_context("forkserver")
        self.pool = ctx_in_main.Pool(processes=nb_workers, maxtasksperchild=10)

    def close(self):
        self.pool.terminate()
        self.pool.join()

def launch_task(self, fun, args=[], kwargs={}):
    return self.pool.apply_async(fun, args=args, kwds=kwargs)

def get_res(self, future):
    return future.get(timeout=5000)
```

# Tests  

It is possible to refer to the tests done here: [https://github.com/CNES/cars/blob/master/tests/orchestrator/cluster/test_cluster.py](https://github.com/CNES/cars/blob/master/tests/orchestrator/cluster/test_cluster.py)

vschaffn:
Resolves #681.

## Overview
This PR introduces a set of classes designed to handle the execution of tasks in either a basic or a multi-processing cluster. It also includes a series of tests that verify the correct behavior of the cluster functionalities.

## Key changes
- `ClusterGenerator`: A factory class that decides which type of cluster to create based on the given configuration (`basic` or `multiprocessing`). It can initialize a `BasicCluster` or `MpCluster` with a specified number of worker processes.
- `AbstractCluster`: The base class for both `BasicCluster` and `MpCluster`. It defines methods to launch tasks, retrieve results, and manage resources such as opening and closing the cluster.
- `BasicCluster`: A subclass of `AbstractCluster` where tasks are run synchronously (i.e., in a single process). The launch_task method simply executes the provided function immediately.
- `MpCluster`: A subclass of `AbstractCluster` that utilizes Python’s `multiprocessing` module for parallel task execution. The cluster pool is initialized with a defined number of worker processes, and tasks are executed asynchronously using `apply_async`. The results of tasks can be retrieved via the `get_res` method.

## Test Suite
`TestClusterGenerator`: A test suite implemented with `pytest` to validate the functionality of the cluster system.
- `test_basic_cluster`: Verifies that tasks are executed synchronously in the `BasicCluster`.
- `test_mp_cluster_task`: Confirms that tasks are launched asynchronously in the `MpCluster`.
- `test_mp_cluster_parallelism`: Tests that multiple tasks can be executed concurrently with multiple workers in `MpCluster`.
- `test_mp_cluster_termination`: Ensures that the cluster terminates correctly and no tasks can be launched after termination.

### DIFF
--- a/tests/test_cluster.py
+++ b/tests/test_cluster.py
@@ -1,0 +1,58 @@
+""" Functions to test the clusters."""
+
+import time
+
+import pytest
+
+from xdem.cluster import BasicCluster, ClusterGenerator, MpCluster
+
+
+# Sample function for testing
+def sample_function(x: float, y: float) -> float:
+    return x + y
+
+
+# Function to simulate a long-running task
+def long_running_task(x: float) -> float:
+    time.sleep(1)
+    return x * 2
+
+
+class TestClusterGenerator:
+    def test_basic_cluster(self) -> None:
+        # Test that tasks are run synchronously in BasicCluster
+        cluster = ClusterGenerator(name="basic")
+        assert isinstance(cluster, BasicCluster)
+
+        result = cluster.launch_task(sample_function, args=[2, 3])
+        assert result == 5
+
+    def test_mp_cluster_task(self) -> None:
+        # Test that tasks are launched asynchronously in MpCluster
+        cluster = ClusterGenerator("multiprocessing", nb_workers=2)
+        assert isinstance(cluster, MpCluster)
+
+        future = cluster.launch_task(sample_function, args=[2, 3])
+        result = cluster.get_res(future)
+        assert result == 5
+
+    def test_mp_cluster_parallelism(self) -> None:
+        # Test that multiple tasks are run in parallel
+        cluster = ClusterGenerator("multiprocessing", nb_workers=2)
+        assert isinstance(cluster, MpCluster)
+
+        futures = [cluster.launch_task(long_running_task, args=[i]) for i in range(4)]
+        results = [cluster.get_res(f) for f in futures]
+        assert results == [0, 2, 4, 6]
+
+    def test_mp_cluster_termination(self) -> None:
+        # Test that the pool terminates correctly after closing
+        cluster = ClusterGenerator("multiprocessing", nb_workers=2)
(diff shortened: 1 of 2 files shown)


rhugonnet:
Looks good, thanks! :slightly_smiling_face:  

I only have two generic comments:
1. Won't we need this in GeoUtils for `delayed_multiprocessing`?
2. Adding short comments for each class method could make the code easier to read. For instance, I suppose the `AbstractCluster` methods with `pass` are meant to always be subclassed? This could be described in a short function description "Meant to be subclassed only", or by raising something like a `NotImplementedError("This function should never be called, meant to be subclassed")` instead of `pass`, which would also make it more directly readable.

vschaffn:
@adebardo @rhugonnet I added comments for each class method, and moved PR to geoutils

GlacioHack/geoutils#656

PR #701, opened 2025-03-11
Add CPD and make ICP Python-based

### LINKED ISSUE #100: Add rigid CPD coregistration
Paper (1200+ citations in 10 years, I think this method is a winner): https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=5432191
Python implementation: https://github.com/siavashk/pycpd

### LINKED ISSUE #483: Add Python-based ICP to support weights

rhugonnet:
This PR adds Coherent Point Drift (CPD) coregistration, and modifies Iterative Closest Point (ICP) coregistration to be fully Python-based. This allows to remove the (heavy) OpenCV dependency, and call consistent optimizer/parameters.

As both methods are point-point, a new preprocessing has to be introduced, which is why both methods are in the same PR.

This PR takes over #581, which is closed. It is possible thanks to #530 to use the same core functions as other coregs.

Resolves #100
Resolves #483


rhugonnet:
@adehecq @erikmannerfelt @adebardo @belletva @vschaffn This PR is ready for your review.

doc/source/coregistration.md
adehecq, line 1: Just by going through the doc, I'm thinking that when we describe "VerticalShift", maybe we should point readers to the "Vertical referencing" section to avoid having people using this method to correct for differences in vertical referencing?

doc/source/coregistration.md
```diff
@@ -305,20 +311,17 @@ _ = ax[1].set_yticklabels([])
...
 
-(icp)=
-### Iterative closest point
+(lzd)=
+### Least Z-difference
 
-{class}`xdem.coreg.ICP`
+{class}`xdem.coreg.LZD`
 
 - **Performs:** Rigid transform transformation (3D translation + 3D rotation).
```
adehecq, line 319: Should "Rigid transform transformation" be "Rigid transformation"? It appears several times.

doc/source/coregistration.md
adehecq: Maybe one needs to update the "recommended pipelines" section? As it does not include the new methods.

doc/source/coregistration.md
```diff
@@ -476,6 +550,8 @@ For both **inputs** and **outputs**, four consistent categories of metadata are
 
 **4. Affine metadata (common to all affine methods)**:
 
+- An input `only_translation` to define if a coregistration should solve only for translations,
+- An input `standardize` to define if the input data should be standardized to the unit sphere before coregistration,
```
adehecq, line 554: This is not super clear. Is it better described somewhere, that you could point to?

xdem/coreg/base.py
```diff
@@ -981,17 +1040,124 @@ def _bin_or_and_fit_nd(
...
         warnings.filterwarnings("ignore", message="`np.float` is a deprecated alias for the builtin `float`")
 
+        # return np.linalg.inv(matrix)
+
         checked_matrix = pytransform3d.transformations.check_transform(matrix)
         # Invert the transform if wanted.
         return pytransform3d.transformations.invert_transform(checked_matrix)
 
 
+def _apply_matrix_pts_mat(
```
adehecq, line 1133: I don't understand what this function does. Could you add a quick explanation in the docstring?

xdem/coreg/affine.py
```diff
@@ -740,282 +782,1289 @@ def vertical_shift(
...
+    # Point-to-point is simply the difference, from Besl and McKay (1992), https://doi.org/10.1117/12.57955
+    if method == "point-to-point":
+        diffs = (trans_tba - ref) ** 2
+    # Point-to-plane used the normals, from Chen and Medioni (1992), https://doi.org/10.1016/0262-8856(92)90066-C
+    # A priori, this method is faster based on Rusinkiewicz and Levoy (2001), https://doi.org/10.1109/IM.2001.924423
+    elif method == "point-to-plane":
+        assert norm is not None
+        diffs = (trans_tba - ref) * norm
+    else:
+        raise ValueError("ICP method must be 'point-to-point' or 'point-to-plane'.")
```
adehecq, line 838: Ideally this should be caught at the beginning of the function

xdem/coreg/affine.py
```diff
@@ -740,282 +782,1289 @@ def vertical_shift(
...
+    # Build an affine matrix for 3D translations and rotations
+    matrix = matrix_from_translations_rotations(t1, t2, t3, alpha1, alpha2, alpha3, use_degrees=False)
+
+    # Apply affine transformation
+    trans_tba = _apply_matrix_pts_mat(mat=tba, matrix=matrix)
+
+    # Define residuals depending on type of ICP method
+    # Point-to-point is simply the difference, from Besl and McKay (1992), https://doi.org/10.1117/12.57955
+    if method == "point-to-point":
+        diffs = (trans_tba - ref) ** 2
```
adehecq, line 831: I'm confused about this part. These are point clouds right? I.e. an array of Nx3, but why would N be the same in both the tba and ref? 
I always assumed that ICP requires finding the nearest neighbor in the other point cloud, just just taking the difference point to point. 
But I guess it must work since your results compare to other implementations. I just don't understand why... 🤔

adehecq: Or I assume this only works for DEMs sampled on the same grid that are just converted to a flat 3D array?

xdem/coreg/affine.py
```diff
@@ -740,282 +782,1289 @@ def vertical_shift(
...
-        rots = pytransform3d.rotations.euler_from_matrix(matrix, i=0, j=1, k=2, extrinsic=True, strict_check=True)
-        rots = np.rad2deg(np.array(rots))
-        return rots[0], rots[1], rots[2]
+    # Group inputs into a single array
+    inputs = (ref, tba, norms)
 
-    def centroid(self) -> tuple[float, float, float] | None:
-        """Get the centroid of the coregistration, if defined."""
-        meta_centroid = self._meta["outputs"]["affine"].get("centroid")
+    # For this type of method, the procedure can only be fit
```
adehecq, line 933: to be deleted?

rhugonnet:
@adehecq made the remark that it is hard to get into the structure of coregistration function such as `nuth_kaab()`, `icp()` when one hasn't seen it in a while.
I made a flowchart summary, that I will also add to the Wiki so we have a reminder of it! :wink: 

xdem_coreg_iterative

xdem/coreg/affine.py
```diff
@@ -740,282 +782,1289 @@ def vertical_shift(
...
+    standardize: bool = True,
+) -> tuple[NDArrayf, tuple[float, float, float], int]:
+    """
+    Main function for Coherent Point Drift coregistration.
+    See Myronenko and Song (2010), http://dx.doi.org/10.1109/TPAMI.2010.46.
+
+    This function subsamples input data, then runs CPD iteration steps to optimize its expectation-minimization until
+    convergence or a maximum of iterations is reached.
+
+    The function assumes we have a two DEMs, or DEM and an elevation point cloud, in the same CRS.
```
adehecq, line 1381: typo "a two" -> "two"

xdem/coreg/affine.py
```diff
@@ -740,282 +782,1289 @@ def vertical_shift(
...
+    Deriving gradient in X/Y expected by the Least Z-difference coregistration.
+
+    :return: Gradient in X/Y, scaled based on the DEM resolution.
+    """
+
+    # If inputs are both point clouds, raise an error
+    if isinstance(ref_elev, gpd.GeoDataFrame) and isinstance(tba_elev, gpd.GeoDataFrame):
+
+        raise TypeError(
+            "The Nuth and Kääb (2011) coregistration does not support two point clouds, one elevation "
```
adehecq, line 1451: This should be "LZD" not Nuth & Kaab"

xdem/coreg/affine.py
```diff
@@ -740,282 +782,1289 @@ def vertical_shift(
...
+    weight_cpd: float,
+    sigma2: float,
+    sigma2_min: float,
+    scale: bool = False,
+    only_translation: bool = False,
+) -> tuple[NDArrayf, float, float]:
+    """
+    Fit step of Coherent Point Drift by expectation-minimization, with variance updating.
+
+    See Fig. 2 of Myronenko and Song (2010), https://arxiv.org/pdf/0905.2635.pdf for equations below.
```
adehecq, line 1222: Should use the DOI: [10.1109/TPAMI.2010.46](https://doi.org/10.1109/TPAMI.2010.46)

xdem/coreg/affine.py
```diff
@@ -1052,44 +2101,73 @@ def _to_matrix_func(self) -> NDArrayf:
...
+     - Point-to-point of Besl and McKay (1992), https://doi.org/10.1117/12.57955, where the loss function is computed
+       on the 3D distances of closest points (original method),
+     - Point-to-plane of Chen and Medioni (1992), https://doi.org/10.1016/0262-8856(92)90066-C where the loss function
+       is computed on the 3D distances of closest points projected on the plane normals (generally faster and more
+       accurate, particularly for point clouds representing contiguous surfaces).
+
+
+    Other ICP options are:
+
+    - Linearized approximation of the point-to-plane least-square optimization of Low (2004),
```
adehecq, line 2119: I don't understand how this method can be set by the user here...

adehecq: Nevermind, I see it now!

xdem/coreg/affine.py
```diff
@@ -1160,102 +2214,150 @@ def _fit_rst_pts(
...
-            raise ValueError(
-                "Not enough valid points in input data."
-                f"'reference_dem' had {points['ref'].size} valid points."
-                f"'dem_to_be_aligned' had {points['tba'].size} valid points."
-            )
+    Estimate translations and rotations.
+
+    The estimated transform is stored in the `self.meta["outputs"]["affine"]` key "matrix", with rotation centered
+    on the coordinates in the key "centroid". The translation parameters are also stored individually in the
+    keys "shift_x", "shift_y" and "shift_z" (in georeferenced units for horizontal shifts, and unit of the
```
adehecq, line 2261: Maybe the answer lies in the code, but if the rotation is not 0, this probably means translation of the centroid?

adehecq (approved):
Really impressive !!! 🤯 Thanks for the work! 🙌 🙌 
I was not able to check everything thoroughly, but tried to focus more on the logic and structure. It all look goods to me and if the tests are conclusive, it's better than any review 😉

rhugonnet:
Thanks a lot @adehecq! Good catch for all these docstring changes :sweat_smile:
Indeed, it's very hard to verify these types of code other than through tests. Even testing the internal functions is a bit useless, because it's the overall algorithm convergence that really matters.
This is why I focused mostly on the synthetic misalignment tests in `test_affine.py`, and the independent benchmarking comparison with other packages (OpenCV, Open3D, PyCPD), showing that we get very similar results in terms of accuracy/convergence.

@adebardo @vschaffn @belletva If you want to review this PR or already have notes, it would be great to have your feedback rapidly (within 2 weeks?). Otherwise just tell me so I can move forward :wink:

doc/source/coregistration.md
rhugonnet: I was thinking of updating the recommended pipelines once my study is out. For now it's non published so hard to justify.

doc/source/coregistration.md
rhugonnet: Tried to clarify as well.

xdem/coreg/base.py
rhugonnet: For speed, description added.

xdem/coreg/affine.py
rhugonnet: Yes that's it. It was hinted in the function description but not extensively. I've added a series of comments and descriptions to make it immediately clearer.

xdem/coreg/affine.py
```diff
@@ -740,282 +782,1289 @@ def vertical_shift(
...
+    max_iterations: int,
+    tolerance: float,
+    params_random: InRandomDict,
+    params_fit_or_bin: InFitOrBinDict,
+    only_translation: bool,
+) -> tuple[NDArrayf, tuple[float, float, float], int]:
+    """
+    Least Z-differences coregistration.
+    See Rosenholm and Torlegård (1988),
+    https://www.asprs.org/wp-content/uploads/pers/1988journal/oct/1988_oct_1385-1389.pdf.
```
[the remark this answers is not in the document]
rhugonnet: :sweat_smile:

xdem/coreg/affine.py
rhugonnet: Done, this was a duplicate check.

rhugonnet:
All comments resolved. Merging!

PR #702, opened 2025-03-20
Save file with driver COG

belletva:
- save file with driver `COG` (`xdem/coreg/workflows.py`)


adebardo:
I'm sorry, @belletva , but after discussing with @vschaffn , it might be more relevant to pass a string 'driver' str or None directly to dem_coregistration and then to save.

adebardo:
Maybe driver and compression can be pass as kwargs ?

belletva:
Thanks for the comment @adebardo . However, the function `dem_coregistration` has already plenty of arguments. If we pass `driver `and `compression` as kwargs, maybe we should do the same for the other arguments? What do you think @rhugonnet and @adehecq ?

rhugonnet:
Thanks! My thoughts on this join this previous discussion: https://github.com/GlacioHack/xdem/pull/650#issuecomment-2494831960.
In short, I think we should deprecate `dem_coregistration` in favor of `DEM.coregister_3d()` (and very soon `ds.dem.coregister_3d()` through the Xarray accessor). :sweat_smile: 

This would solve the above problem by allowing to pass all arguments linked to I/O separately (which is the norm, whether in Xarray, Pandas or other):
```python
dem1 = DEM(fn1)
dem2 = DEM(fn2)
# Case 1: In-memory
dem_coreg = dem1.coregister_3d(dem2)
dem_coreg.save(outfile, driver=, compression=)
# Case 2: Out-of-memory Multiprocessing
mp_config = MultiprocConfig(outfile, driver=, compression=)
dem_coreg = dem1.coregister_3d(dem2, mp_config=mp_config)
# Case 3: Out-of-memory Dask
dem_coreg = dem1.coregister_3d(dem2)
writer = dem_coreg.to_raster(outfile, encoding={driver=, compression=}, compute=False)
writer.compute()
```
I think having a one-liner convenience function that can take all these arguments at once would belong better in a CLI where we cannot chain I/O operations easily.
What do you think?

rhugonnet:
In the meantime, while we keep `dem_coregistration` a bit longer, we could simply add the arguments.

adebardo:
Thanks for your feedback, @rhugonnet .
You're right, I'll adjust my habits when using coregister_3d, it's important, so thanks for pointing it out.
I'm merging as-is for this one anyway.

PR #704, opened 2025-03-21
Add multiprocessing for terrain attribute calculation

### LINKED ISSUE #703: Add multiprocessing support for terrain attribute calculation with large DEMs
## Description
The current `terrain.get_terrain_attribute` function loads the entire DEM into memory, which can be inefficient for large datasets. There is a need to implement a multiprocessing version that handles large DEMs by processing them in tiles, reducing memory usage, and improving performance through parallel computation. This function should allow users to calculate multiple terrain attributes efficiently using multiprocessing while keeping DEMs on disk.

vschaffn:
Resolves #703 

## Context

This PR introduces the multiprocessing possibility in `get_terrain_attribute`, improving memory efficiency and processing speed when calculating terrain attributes on large DEMs by leveraging tiled processing and multiprocessing.

## Changes
- `get_terrain_attribute` has been renamed `_get_terrain_attribute`.
- A new function `get_terrain_attribute` has been created with an additional parameter `multiproc_config`. If `multiproc_config` is not None, then `_get_terrain_attribute` is run through the `map_overlap_multiproc` function of geoutils (see GlacioHack/geoutils#669), and the result is saved in "multiproc_config.outfile". Otherwise `_get_terrain_attribute` is run as before.
- The `multiproc_config` parameter has been passed to the individual terrain attribute functions, in `terrain` and `dem`.

## Tests
- A test similar to `test_get_terrain_attribute` has been added for the multiprocessed version.
image

## Documentation
- A section about multiprocessing has been added to `terrain.md`.

## Example

```{code-cell} ipython3
from geoutils.raster.distributed_computing import ClusterGenerator, MultiprocConfig

mp_config = MultiprocConfig(
            chunk_size=200,
            outfile="slope.tif",
            cluster=ClusterGenerator("mp", nb_workers=4),
        )

# Compute slope out-of-memory
xdem.terrain.get_terrain_attribute(dem, "slope", mp_config)
# same
dem.slope(mp_config)
# Slope saved under "slope.tif"
```

## Note
- This PR needs GlacioHack/geoutils#661


rhugonnet:
Great! :star_struck: 

I'm sure many users will find this extremely useful! (I know several have been waiting for it)

I don't have comments on the functionality itself, just about slightly changing the structure to better prepare for next steps, and keep consistent syntax with Dask to facilitate user understanding long-term :slightly_smiling_face: 

(Below I'm using "chunk" instead of "tile", as exactly the same concept)

1. **Chunk mapping with overlap:** The function `multiproc.get_terrain_attribute_multiproc()` is fully generic, nearly all arguments are called by `cluster.launch_task()` except `window=` and `out_dir=`. Shoul we call it instead `multiproc.map_overlap(func, *args, depth=, out_dir=,)` (mirroring https://docs.dask.org/en/latest/generated/dask.array.overlap.map_overlap.html) that we can re-use for all other chunked multiprocessing with overlap we'll need? (`window` would be passed to `depth`, and `out_file` would be the only other extra argument existing for multiproc compared to Dask, as here we need to write the file out). Then `get_terrain_attribute_multiproc()` would be a short wrapper that calls that function with `terrain_attribute_by_block()`.
2. **Repository for the chunk mappings:** From the above, we should probably add these generic `map_overlap`/`map_block` functions based on `multiprocessing` in GeoUtils (useful there for filters, rasterization, etc)? And actually, I think we should make them **public functions in the API** :smile:  (that a user can call himself with any custom-function they desire to run on chunks!).
3. **Trigger the multiproc call in a function:** I think we could integrate `get_terrain_attribute_multiproc()` within `get_terrain_attribute()` and have it be triggered by an additional argument `multiproc_chunk_sizes=` (that defines the chunk sizes the user wants, depending on his desired memory usage). This would be the same behaviour as Xarray/Dask that trigger out-of-memory operations when loading with a `chunk_sizes=` argument. We woud just need an optional extra argument `multiproc_out_file=` that can be passed to `out_file=` if the user wants to specify the output file path. I think this approach blends better in, and avoids duplicating our API with multiproc versions of functions. For documenting it, we could add a main documentation page in GeoUtils & xDEM about how to trigger out-of-memory multiprocessing for all functions in general (when available).

For 3., with the way `get_terrain_attribute` is currently called by `terrain_attribute_by_block`, we have a circular problem. We'd need to rename  `get_terrain_attribute()` into a `_get_terrain_attribute()`, and create a new "front" wrapper `get_terrain_attribute` with the function description, and that either calls `_get_terrain_attribute()` or `_get_terrain_attribute_multiproc()` :wink:

vschaffn:
@rhugonnet thanks for your feedback !
Hoping I have understood your comments, I have opened a ticket and a PR on the subject on geoutils :
https://github.com/GlacioHack/geoutils/issues/670
https://github.com/GlacioHack/geoutils/pull/669

If these changes are approved, I will adapt this PR (which will be simplified a lot !) to the new geoutils multiprocessing structure 😄

vschaffn:
@rhugonnet this PR is ready to be merged. Is it all right for you ? 😃

rhugonnet:
Yes! Just two small remarks: 
1/ Should we name the argument `mp_config` for short, given that `mp` is the official short-name for importing `multiprocessing` from the Python documentation?
2/ Should we move the `mp_config` argument at the end of the argument list (after `method`, `window_size`, etc)?
As `mp_config` is going to be an argument that is not method-specific (available for all functions eventually) and described as such in the documentation, users will know it is available and it doesn't have to appear immediately. So I think we should keep method-specific arguments at the top for clarity :wink:.

doc/source/terrain.md
```diff
@@ -405,6 +405,25 @@ fractal_roughness = dem.fractal_roughness()
 fractal_roughness.plot(cmap="Reds", cbar_title="Fractal roughness (dimensions)")
 ```
 
+## Generating attributes in multiprocessing (Soon available, waiting next GeoUtils release)
```
rhugonnet, line 408: Still need to update the doc (or comment it for now!) :wink:

rhugonnet (dismissed):
Great, thanks! :smile:

rhugonnet (approved):

PR #707, opened 2025-03-27
Refactor blockwise coregistration method

adebardo:
Lots of work to do but we will use this PR to debate tomorrow

### DIFF
--- a/xdem/coreg/blockwise.py
+++ b/xdem/coreg/blockwise.py
@@ -38,10 +38,10 @@ from geoutils.raster.distributed_computing import (
     map_overlap_multiproc_save,
 )
 from geoutils.raster.tiling import compute_tiling
-from xdem._typing import MArrayf, NDArrayb, NDArrayf
-from xdem.coreg import Coreg, CoregPipeline
-from xdem.coreg.base import Coreg, CoregPipeline, CoregType
+
 import xdem
+from xdem._typing import MArrayf, NDArrayb, NDArrayf
+from xdem.coreg.base import Coreg, CoregPipeline, CoregType
 
 
 class BlockwiseCoreg(Coreg):
@@ -66,27 +66,23 @@ class BlockwiseCoreg(Coreg):
         """
         if isinstance(step, type):
             raise ValueError(
-                "The 'step' argument must be an instantiated Coreg subclass. "
-                "Hint: write e.g. ICP() instead of ICP"
+                "The 'step' argument must be an instantiated Coreg subclass. " "Hint: write e.g. ICP() instead of ICP"
             )
         self.procstep = step
         self.tile_size = tile_size
         self.apply_z_correction = apply_z_correction
         self.output_path = output_path
-        self.x_coords = []
-        self.y_coords = []
-        self.shifts_x = []
-        self.shifts_y = []
-        self.shifts_z = []
 
         tile_size = 500
         self.mp_config = MultiprocConfig(tile_size)
 
         super().__init__()
 
+        print("toto")
+
     @staticmethod
     def coreg_wrapper(
-        ref_dem_tiled, tba_dem, coreg_method, inlier_mask
+        ref_dem_tiled: RasterType, tba_dem: RasterType, coreg_method: Coreg | CoregPipeline, inlier_mask: RasterType
     ) -> Coreg | CoregPipeline:
         """
         Wrapper function to apply Nuth & Kääb coregistration on a tile pair.
@@ -104,12 +100,8 @@ class BlockwiseCoreg(Coreg):
         :return: Reprojected secondary elevation dataset.
         """
 
-        mp_config = MultiprocConfig(
-            chunk_size=self.tile_size, outfile=self.output_path + "SEC_reprojected.tif"
-        )
-        reproject_dem = sec.reproject(
-            ref=ref, resampling="cubic", multiproc_config=mp_config, silent=True
-        )
+        mp_config = MultiprocConfig(chunk_size=self.tile_size, outfile=self.output_path + "SEC_reprojected.tif")
+        reproject_dem = sec.reproject(ref=ref, resampling="cubic", multiproc_config=mp_config, silent=True)
 
         return reproject_dem
 
@@ -129,14 +121,10 @@ class BlockwiseCoreg(Coreg):
         **kwargs: Any,
     ) -> CoregType:
 
-        if isinstance(reference_elev, gpd.GeoDataFrame) and isinstance(
-            to_be_aligned_elev, gpd.GeoDataFrame
-        ):
-            raise NotImplementedError(
-                "Blockwise coregistration does not yet support two elevation point cloud inputs."
-            )
+        if isinstance(reference_elev, gpd.GeoDataFrame) and isinstance(to_be_aligned_elev, gpd.GeoDataFrame):
+            raise NotImplementedError("Blockwise coregistration does not yet support two elevation point cloud inputs.")
 
-        self.reproject_dem = to_be_aligned_elev.reproject(
+        self.reproject_dem = to_be_aligned_elev.reproject(  # type: ignore
             ref=reference_elev, multiproc_config=self.mp_config, silent=True
         )
 
@@ -152,12 +140,8 @@ class BlockwiseCoreg(Coreg):
             return_tile=True,
         )
 
-        shape_tiling_grid = compute_tiling(
-            self.tile_size, reference_elev.shape, to_be_aligned_elev.shape
-        ).shape
-        rows_cols = list(
-            itertools.product(range(shape_tiling_grid[0]), range(shape_tiling_grid[1]))
-        )
+        shape_tiling_grid = compute_tiling(self.tile_size, reference_elev.shape, to_be_aligned_elev.shape).shape
+        rows_cols = list(itertools.product(range(shape_tiling_grid[0]), range(shape_tiling_grid[1])))
 
         for idx, (coreg, tile_coords) in enumerate(outputs_coreg):
             tile_str = f"{rows_cols[idx][0]}_{rows_cols[idx][1]}"
@@ -165,8 +149,8 @@ class BlockwiseCoreg(Coreg):
                 shift_x = coreg.meta["outputs"]["affine"]["shift_x"]
                 shift_y = coreg.meta["outputs"]["affine"]["shift_y"]
                 shift_z = coreg.meta["outputs"]["affine"]["shift_z"]
-                self.meta["outputs"][tile_str] = {
-                    "shift_x": shift_y,
+                self.meta["outputs"][tile_str] = {  # type: ignore
+                    "shift_x": shift_x,
                     "shift_y": shift_y,
                     "shift_z": shift_z,
                 }
@@ -176,7 +160,7 @@ class BlockwiseCoreg(Coreg):
                 self.shifts_z.append(shift_z)
 
             except KeyError:
-                self.meta["outputs"][tile_str] = {
+                self.meta["outputs"][tile_str] = {  # type: ignore
                     "shift_x": np.nan,
                     "shift_y": np.nan,
                     "shift_z": np.nan,
@@ -215,9 +199,7 @@ class BlockwiseCoreg(Coreg):
         try:
             (a, b, c, d), _ = plane.fit(points, thresh=threshold, minPoints=min_inliers)
         except ValueError:
-            raise ValueError(
-                "Not enough inliers, please increase the size of your tiles. "
-            )
+            raise ValueError("Not enough inliers, please increase the size of your tiles. ")
 
         # Convert plane ax + by + cz + d = 0 to z = f(x, y)
         # z = -(a*x + b*y + d) / c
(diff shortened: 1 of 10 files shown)


adebardo:
image

rhugonnet:
Great! :slightly_smiling_face:

I went through all the code as we mentioned in our discussion yesterday. All good on the functionality. In terms of the structure and consistency with the rest of the package, here are my remarks (difficulty level of the change between parenthesis):
1. (Easy) As mentioned in the call, we should likely add `plane_fit_ransac` as an option of `apply()`, keeping the behaviour consistent with `BlockwiseCoreg` otherwise,
2. (Easy-Moderate) I don't understand why `reference_dem` and `to_be_aligned_dem` need to be `__init__`. The `tiling_grid` could be defined during `fit()` as it only relates to `reference_dem` (and not the coregistration method itself), and the `profile` saved there too?
3. (Moderate) After reviewing the functions, it would be a bit of effort but not as difficult as I anticipated to integrate these changes into `BlockwiseCoreg` directly, and trigger the out-of-memory behaviour by passing a `MultiprocConfig` to `fit()` and `apply()`.

For 3.: Why? I realized we actually don't need to mirror the full complexity of #525 for `BlockwiseCoreg`. This is thanks to `BlockwiseCoreg` having its own `fit()/apply()` function independent of `Coreg.fit()/apply()`.
As `BlockwiseCoreg` splits the input array very early on, we don't need as many complex out-of-memory processing steps during the rest of the coregistration, as we need for the generic `Coreg.fit()/apply()` steps. For instance, `coreg.NuthKaab()` requires out-of-memory subsampling, then out-of-memory interpolation of that subsample that re-runs at each iteration of the algorithm depending on the shift iteration.
Here, we only need two one-time steps to work in the beginning:
- Out-of-memory reprojection (almost done in https://github.com/GlacioHack/geoutils/pull/661),
- Out-of-memory masking to define `inlier_data` (that we could run with `map_overlap_multiproc` of https://github.com/GlacioHack/geoutils/pull/669, or a very close version).

**In short:** We could pass down the `MultiprocConfig` to the `preprocess_fit/apply` functions, and define filenames in `MultiprocConfig` for these two "break points" (that will create a file for the "reprojected_dem" and "inlier_mask"). Then we just continue the rest of the coregistration with these two files!

So, in short, I think 1-2. are doable. 
For 3., it's a bit more effort but not as much as though, and would be ideal to avoid splitting into two classes. I'm happy to help out on this to finalize it in time :wink:

**Finally, I also have one question:** Is there any core difference between our `interp_points()` and `resample_dem()`?
We have a grid resample wrapper based on `interp_points()` (tested to match exactly GDAL.warp in a same CRS), which would be easy to run with Multiprocessing:
 https://github.com/GlacioHack/xdem/blob/6e9db15795d41466c4a6a87041876f0ee5f7d281/xdem/coreg/base.py#L1306
 
We generally prefer to use this `interp_points()` implementation for consistency due to:
- Its integrated nodata propagation scheme depending on the resampling, to avoid unexpected outliers,
- Its understanding of pixel interpretation (center or corner of pixel), to avoid unexpected shifts,
- It is the resamping used by the coregistration algorithms, so it ensures the resulting DEM alignment is fully consistent (otherwise, with a different resampling here, if we re-ran the coregistration algorithm, we would likely find another shift again instead of (0,0,0)).

Some issues I mention above sound like the ones you described in the call, so this might part of the solution.
**In terms of performance of our resampling:** The default "bilinear" sampling relies on `scipy.ndimage.map_coordinates` (to harness the regular-grid nature of a raster = X and Y axis have equal spacing), and therefore is **much faster** than Xarray-SciPy's `interpn` (that works on a rectilinear-grid = X and Y can have any coordinates).
It's likely not as fast as `cars-resample`, but shouldn't be too far off :wink:

rhugonnet:
@adebardo I thought about the `apply` step again.

Performing the resampling we talked about using `map_overlap` (more likely a modified version resembling the original implementation of `reproject_multiproc` from @vschaffn that gets the right shifted block with overlap) coupled with `reproject_samecrs()` should work well.
My only worry is that we might get tiling artefact at the boundary of tiles. Having smaller tiles would solve this, but then the Nuth and Kääb algorithm would not have enough points to converge, so we'd be blocked.

Thinking ahead on this, I see two solutions:
- We could run `apply()` on a different (smaller) tiling than `fit()`, which solves the issue above,
- Otherwise, we could run it on every point to avoid any deformation at all: We apply the 2D RANSAC plan function of X/Y shifts to every pixel of the DEM. This wouldn't yield a grid anymore, but a point cloud, that we need to re-grid. We could do this easily out-of-memory by combining the same `map_overlap` function as above with GeoUtils' `PointCloud.grid()` function: https://github.com/GlacioHack/geoutils/blob/1bbfddde0aa36415fa9d75a8239652c1500e877d/geoutils/interface/gridding.py#L31, which is already what we do in `apply_matrix()` for large rotations here: https://github.com/GlacioHack/xdem/blob/6e9db15795d41466c4a6a87041876f0ee5f7d281/xdem/coreg/base.py#L1277

So I think we should have it covered in any case :wink:

adebardo:
Hi @rhugonnet , I know it's not perfect, but could you take a look? I imagine there will be quite a few changes, and I'd like to get everything aligned before I properly finish the tests.

rhugonnet:
Looking good! Shorter and better code :smile:

I don't have any generic comment, so will just add a couple line-by-line!

xdem/coreg/blockwise.py
```diff
@@ -17,84 +17,91 @@
...
-        self.subdivision = subdivision
-        self.success_threshold = success_threshold
-        self.n_threads = n_threads
-        self.warn_failures = warn_failures
+        self.tile_size = tile_size
         self.apply_z_correction = apply_z_correction
+        self.output_path = output_path
+
+        tile_size = 500
+        self.mp_config = MultiprocConfig(tile_size)
```
rhugonnet, line 77: I had trouble conceptualizing the API for a minute here. I think that's because, technically, the user-chosen `chunk_size` (how much is loaded in memory at once from the raster) and the `block_size` (size of subdivision to fit/apply the coregistration) are two different arguments.

For now we have to check/force that the `chunk_size` is the same or larger than the `block_size` (because normal Coregs don't support chunking, so we always load the full block, but this could change in time).

To finalize the API here, we could move `mp_config` as an argument of `__init__` (so that we can remove `output_path`, and that `chunk_size` can also be passed independently there by the user). And we leave the current `tile_size` (I prefer `block_size` to echo `BlockwiseCoreg` :wink:) for the subdivision?

xdem/coreg/blockwise.py
```diff
@@ -17,84 +17,91 @@
...
+        ref_dem_tiled: RasterType, tba_dem: RasterType, coreg_method: Coreg | CoregPipeline, inlier_mask: RasterType
+    ) -> Coreg | CoregPipeline:
+        """
+        Wrapper function to apply Nuth & Kääb coregistration on a tile pair.
+        """
+        coreg_method = coreg_method.copy()
+        tba_dem_tiled = tba_dem.crop(ref_dem_tiled)
+        return coreg_method.fit(ref_dem_tiled, tba_dem_tiled, inlier_mask)
+
+    def preprocess(self, ref: RasterType, sec: RasterType) -> RasterType | None:
```
rhugonnet, line 92: I don't think `preprocess` is used anywhere

xdem/coreg/blockwise.py
```diff
@@ -115,331 +122,142 @@ def fit(
...
-
-        progress_bar = tqdm(
-            total=indices.size, desc="Processing chunks", disable=logging.getLogger().getEffectiveLevel() > logging.INFO
+        outputs_coreg = map_multiproc_collect(
+            self.coreg_wrapper,
+            reference_elev,
+            self.mp_config,
+            self.reproject_dem,
+            self.procstep,
+            inlier_mask,
```
rhugonnet, line 137: Perfect!
We should be able to add a `map_overlap` with `depth=0` to compute the `inlier_mask` out-of-memory fairly easily (the only missing input here).
It's essentially just: 
```python
np.logical_and.reduce((np.isfinite(reference_elev), np.isfinite(reproject_dem), inlier_mask))
```
We can use the same as logic as in `coreg_wrapper` that you wrote, with `crop` on the 3 inputs.

xdem/coreg/blockwise.py
```diff
@@ -115,331 +122,142 @@ def fit(
...
+            (a, b, c, d), _ = plane.fit(points, thresh=threshold, minPoints=min_inliers)
+        except ValueError:
+            raise ValueError("Not enough inliers, please increase the size of your tiles. ")
+
+        # Convert plane ax + by + cz + d = 0 to z = f(x, y)
+        # z = -(a*x + b*y + d) / c
+
+        return -a / c, -b / c, -d / c
+
+    def wrapper_apply_epc(
```
rhugonnet, line 207: Also a static method?

xdem/coreg/blockwise.py
```diff
@@ -449,213 +267,37 @@ def _apply_rst(
...
-            method=resampling,
-            fill_value=np.nan,
+        _ = map_overlap_multiproc_save(
+            self.wrapper_apply_epc,
+            self.reproject_dem,
+            apply_config,
+            coeff_x,
+            coeff_y,
+            coeff_z,
+            depth=10,
```
rhugonnet, line 297: Here the depth will have to be `max(all_x_shifts, all_y_shifts)`. This will explode a bit the memory usage if the DEM is shifted by a lot (let's say +100, +100 across the entire DEM).
I'm not sure if there's another solution than writing a custom `map_overlap`-style function that reads the shifted chunks, like @vschaffn's initial reprojection, this way we're sure to keep memory usage to a minimum.

rhugonnet:
It's great, thanks a lot for all the work! 
Much more concise, readable and efficient code :slightly_smiling_face: 

I have an idea that might solve the efforts you showed to optimize memory usage (in case you don't know the I/O behaviour with chunks, if you know, sorry for explaining again!):
There are chunks "on disk", which are different from chunks "in memory". On file (TIF, COG), a raster is written with certain chunk sizes defined, let's say (300, 300). If you then read a window with Rasterio, let's say (20, 20), it's actually the full (300, 300) chunk that gets loaded, before subsetting to (20, 20) later.
So, if you match your disk chunks in memory (300, 300), and you use a non-zero `depth` with `map_overlap`, you'll load 9 tiles (the central one + 8 surrounding) of (300, 300) all at once just to compute the overlap. That's a big memory peak.

In short, you want to have two things: 1/ Use memory chunks that are a large multiple (x4-x10) of your disk chunks, and 2/ Ensure your disk chunks aren't too big (many rasters only have 4 chunks, so almost everything gets loaded all the time, no matter the memory chunks).
Sometimes you need to re-encode the disk chunks (rewrite the TIF/COG) to your desired sizes to get the optimal memory usage. I do this in GeoUtils' out-of-memory test (for netCDFs, but it's similar for TIF/COG) here: https://github.com/GlacioHack/geoutils/blob/10519e99f9f8ec8ee76a72ac9702e2eac8467d90/tests/test_raster/test_distributing_computing/test_delayed_dask.py#L273
Maybe that helps for the memory tests :wink:

xdem/coreg/blockwise.py
adebardo: indeed ^^

xdem/coreg/blockwise.py
adebardo: I will open an issue about that

xdem/coreg/blockwise.py
adebardo: I forgot to delete those test with test_tile = 500, actually chunk_size = blocksize

xdem/coreg/blockwise.py
adebardo: Isn't it already aply here : https://github.com/GlacioHack/xdem/blob/main/xdem/coreg/base.py#L250

xdem/coreg/blockwise.py
```diff
@@ -17,645 +17,359 @@
...
+                threshold_ransac,
+                min_inliers_ransac,
+                max_iterations_ransac,
             )
+        else:
+            coeff_z = (0, 0, 0)
 
-        if dilate_mask:
-            new_mask = scipy.ndimage.binary_dilation(new_mask, iterations=order[resampling]).astype(new_mask.dtype)
+        self.mp_config.outfile = self.output_path_aligned
```
rhugonnet, line 360: I guess these two `self.mp_config.outfile=...` calls will make the operations fail when Multiprocessing is not used (`mp_config=None`).
Not sure how we make the behaviour consistent between the two cases... Maybe just adding `if mp_config is not None:` before the above call?

rhugonnet:
Looking good, not much to say on the code side!
I'm mostly curious about how the shifted DEM looks like with this new `apply` using `grid()`, and how well the memory usage performs! :blush:

vschaffn (approved):

rhugonnet:
Great! :smile: 
@adebardo Would be interesting to see if we still have the edge effects with the new apply method by updating the plot here: https://xdem.readthedocs.io/en/stable/coregistration.html#the-blockwisecoreg-object).
I see it's not there in the `latest` doc that updated after merging this PR: https://xdem.readthedocs.io/en/latest/coregistration.html#the-blockwisecoreg-object

rhugonnet:
There are some rendering issues with the changes in the doc too, we'll have to fix that in another PR.

PR #710, opened 2025-04-03
Switching terrain attribute calculation to `float32`

### LINKED ISSUE #709: Memory optimization for terrain attributes (float64 -> float32)
Currently, terrain attributes are computed using `float64`, but `float32` would be sufficient. Switching to `float32` would reduce memory usage and improve performance.

vschaffn:
Resolves #709 

This PR optimizes memory usage in the `_get_quadric_coefficients` function by changing the data type of the output array from `float64` to `float32`, which is sufficient for calculating terrain attributes.


rhugonnet:
Perfect!

For the failing test, maybe it's sufficient to invert the doctest example DEM to yield an aspect of 180 instead of 0/360, otherwise the test modulo 360 is more ambiguous.

We could also think of having a "dtype" argument to return the desired dtype. Several terrain attributes are often returned as integers (hillshade for example is often `int8`).

PR #711, opened 2025-04-04
Add min, max, mean and median filters

quinto-clara:
# Context

This PR adds to the available filters minimum, maximum, median and mean filters useful for the CNES QI team. 
Each filter's function has a 2D/3D input array (`array`) and the size of the kernel (`kernel_size`) as parameters. We get the filtered array in output, with the same shape as the input array. The filters are robust to NaN values.

These additions are temporary in xDEM and will be moved to GeoUtils in the future.

# Changes

- New functions added into `filters.py` :

    - `median_filter_scipy()` for scipy median filter
    - `median_filter()` for mean filter
    - `min_filter_scipy()` for scipy minimum filter
    - `max_filter_scipy()` for scipy maximum filter

- Tests added into `test_filters.py` :

    - `test_median_filter()`
    - `test_mean_filter()`
    - `test_min_filter()`
    - `test_max_filter()`

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -48,7 +48,7 @@ This file keeps track of authors contributions.
 - **Enrico Mattea** [@MatteaE](https://github.com/MatteaE)
 - **Amelie Froessl** [@ameliefroessl](https://github.com/ameliefroessl)
 - **Simon Gascoin** [@sgascoin](https://github.com/sgascoin)
-
+- **Clara Quinto** [@quinto-clara](https://github.com/quinto-clara) <clara.quinto@cs-soprasteria.com>
 
 ## Original Developers/Designers/Supporters
 
--- a/tests/test_filters.py
+++ b/tests/test_filters.py
@@ -65,6 +65,131 @@ class TestFilters:
         pytest.raises(ValueError, xdem.filters.gaussian_filter_scipy, data, sigma=5)
         pytest.raises(ValueError, xdem.filters.gaussian_filter_cv, data, sigma=5)
 
+    def test_median_filter(self) -> None:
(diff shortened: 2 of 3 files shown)


rhugonnet:
Great, thanks a lot! :smile: 
(as the comments in the file showed, we've been wanting to add those for a while!)

I only have small comments:
- As the tests seem to all check the same things (if I didn't miss anything), we could probably merge them into a single test function with a `@pytest.mark.parametrize("filter_func", [xdem.filters.median_filter_scipy, xdem.filters.mean_filter_scipy, ...])`,
- We could wrap the very useful `ndimage.generic_filter` function in the same manner. I remember using it a lot with DEM data for robustness to outliers, it allows to leave the exact filtering function to the user.
For instance:
```python
def robust_min(a):
    return np.nanpercentile(a, 10)
min_arr = filters.generic_filter(arr, robust_min)
```
- The above point actuallly reminds me that we can also avoid NaNs by using `np.nanfunc` through `generic_filter`, I wonder which approach is the fastest
- We could pass **kwargs to all calls, I know a lot of people who like to use a disk-shaped `footprint` instead of `size`. Actually we have a function to get that disk footprint shape buried in xDEM for similar purposes, in [spatialstats](https://github.com/GlacioHack/xdem/blob/6e9db15795d41466c4a6a87041876f0ee5f7d281/xdem/spatialstats.py#L863).

quinto-clara:
**New features :**

- Combines some filter tests into one by using : `@pytest.mark.parametrize("name, filter_func", [...])`
- New function `generic_filter()` added into `filters.py`
- `**kwargs` added to filter functions using scipy filters

adebardo (approved):

PR #716, opened 2025-04-09
Switch the `nmad` of xDEM to the `nmad` of GeoUtils

### LINKED ISSUE #715: Delete `spatialstats.nmad` and use the one in GeoUtils
There are currently 2 `nmad` implementations in use (one in GeoUtils, `geoutils.stats.nmad` and one in xDEM, `xdem.spatialstats.nmad`). To simplify maintenance and avoid duplication, I propose to keep only the one in GeoUtils.
This requires:
- A small adaptation of the `nmad` of GeoUtils compared to the `nmad` of xDEM.
- Deleting the xDEM `nmad`, and changing the path (replacing `xdem.spatialstats.nmad` by `geoutils.stats.nmad`) wherever it is used in xDEM.

vschaffn:
Resolves #715.

- Deletes `spatialstats.nmad`.
- Switches `xdem.spatialstats.nmad` to `geoutils.stats.nmad`.

## Note
- Needs GlacioHack/geoutils#678

### DIFF
--- a/xdem/spatialstats.py
+++ b/xdem/spatialstats.py
@@ -28,6 +28,7 @@ import warnings
 from typing import Any, Callable, Iterable, Literal, TypedDict, overload
 
 import geopandas as gpd
+import geoutils as gu
 import matplotlib
 import matplotlib.colors as colors
 import matplotlib.pyplot as plt
@@ -36,9 +37,9 @@ import numpy as np
 import pandas as pd
 from geoutils.raster import Mask, Raster, RasterType, subsample_array
 from geoutils.raster.array import get_array_and_mask
-from geoutils.stats import nmad
 from geoutils.vector.vector import Vector, VectorType
 from numpy.typing import ArrayLike
+from packaging.version import Version
 from scipy import integrate
 from scipy.interpolate import RegularGridInterpolator, griddata
 from scipy.optimize import curve_fit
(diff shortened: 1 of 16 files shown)


rhugonnet:
@vschaffn Ah yes, good thinking, I almost forgot this!
As this function was highly used, we need to keep `xdem.spatialstats.nmad` active a bit longer (pointing towards `geoutils.stats.nmad`) and raise a deprecation warning with our `@deprecate` decorator (maybe until two/three minor versions, so let's put `0.4`?): https://github.com/GlacioHack/xdem/blob/82307bd3212bd66a9ea6d916766259beedec1ba7/xdem/misc.py#L94

vschaffn:
@rhugonnet Yes of course if it is highly used it makes sense, it has been done 😃

xdem/spatialstats.py
```diff
@@ -47,13 +49,17 @@
...
+from xdem.misc import deprecate
 
 with warnings.catch_warnings():
     warnings.filterwarnings("ignore", category=DeprecationWarning)
     import skgstat as skg
 
 
-def nmad(data: NDArrayf | RasterType, nfact: float = 1.4826) -> np.floating[Any]:
+@deprecate(
+    removal_version=Version("0.4"), details="xDEM's nmad will be replaced by the new nmad implementation in GeoUtils"
```
rhugonnet, line 60: I think we should point them directly towards the new signature so they don't have to look for it, something along the lines of: "xdem.spatialstats.nmad is being deprecated in favor of geoutils.stats.nmad"

rhugonnet (approved):

adebardo (approved):

Issue #636, opened 2024-11-06
Move glacier-specific tools to a separate package/module


rhugonnet:
Four years later, we've moved forward a lot, and some tools that made sense in the beginning don't fit so well in the package anymore! They should not be discarded, rather moved somewhere that makes more sense, and where we can keep building on top of what we already have :slightly_smiling_face: 

The tools concerned are the gap-filling methods of `volume.py`, with a closely linked API in `dDEM`, and several related issues (all pretty old now):
- #254,
- #247,
- #184,
- #109,
- #97,
- #72,
- #41.

## Where should we make the distinction?

After many discussions, I see **two distinct categories of applicative tools in relation to xDEM**:
1. **Tools requiring only one or several DEMs** and **deriving terrain-related outputs**, even when being more useful in a given application, such as the different depression-filling algorithms of RichDEM (used mostly in hydro).
2. **Tools requiring auxiliary data specific to the application** (e.g., glacier outlines) and **deriving application-related outputs** (such as glacier volume change).

Tools like 1/ could fit directly in xDEM, because they are generalizable, while 2/ would fit better in a separate repo.

## What would this new repo look like?

We also bumped into this question many times in the past years. (coding @adehecq KH9 repo, and others)
If the nature of the data (elevation) is not the center of this new repo (because you have packages like xDEM for different data types), what should it be?

After a lot of thinking and looking around, there seems to be the need for a "statistical estimation" package for EO surface data, which could combine all types of surface data (whether elevation, spectral, velocity, model estimates such as ice thickness, etc).
In the case of volume change, it happens that we "only" need glacier outlines + DEMs. But most of the other variables estimated for glaciers (SMB, frontal ablation, area change, etc) all require data of different nature.

In short, maybe the right way to organize this in a logical unit would be to create a package focused on the statistical estimation for surface applications (focusing mostly on cryo or glaciers?):
- Estimating glacier volume change and mass balance from elevation differences and glacier outlines (including hypsometric gap-filling, density conversion, etc),
- Estimating surface mass balance from elevation differences, velocity and ice thickness,
- Estimate surface albedo based on spectral data,
- Estimating glacier area change from multi-temporal glacier outlines,
- Estimating frontal ablation/discharge from velocity fields, fluxgates and glacier outlines,
- Delineating glacier outlines from elevation/velocity/spectral data.
- Identifying crevasses from spectral+elevation imagery,
- etc...

[...]

PR #717, opened 2025-04-13
Add NumPy 2 support

rhugonnet:
This PR updates xDEM to support NumPy 2.0.

Follows the same update in GeoUtils: https://github.com/GlacioHack/geoutils/pull/605. And also requires a default nodata defined for int64 types: https://github.com/GlacioHack/geoutils/pull/680.

Two main changes:

1. As NumPy 2.0 now returns typed scalars (e.g., `np.float32(1.0)` instead of `1.0`), it made doctests fail in `terrain.py`. This was the occasion to also catch that attributes had inconsistent output data types (sometimes `np.float32`, sometimes `np.float64`).

As a fix, a new `out_dtype` argument can be passed to `get_terrain_attibute` to define which data type should have the output during calculation, which defaults to the `dtype` of the input DEM. All Numba-optimized functions now have arrays initialized on those data types, which propagate all the way to the final output.

Adds a test to check this behaviour.

@vschaffn This complements your modif in #710.

2. NumPy 2.0 also triggered changes linked to the manual binning in `volume.py`. I managed to fix the main issue in `hypsometric_binning`, but a last test failing remains, and is impossible to track (tests are so scarce)... 
I think we should anyway rewrite that hypsometric code from scratch using Pandas/SciPy (in another package, potentially, as discussed in #636). It's getting quite old and hard to maintain...

### DIFF
--- a/tests/test_terrain.py
+++ b/tests/test_terrain.py
@@ -394,3 +394,19 @@ class TestTerrainAttribute:
         # When using edge wrapping, all coefficients should be finite.
         coefs = xdem.terrain.get_quadric_coefficients(dem, resolution=1.0, edge_method="wrap")
         assert np.count_nonzero(np.isfinite(coefs[0, :, :])) == 9
+
+    def test_get_terrain_attribute__out_dtype(self):
+
+        # Get one attribute using quadratic coeff, and one using windowed indexes
+        slope, tpi = xdem.terrain.get_terrain_attribute(self.dem, attribute=["slope", "topographic_position_index"])
+
+        assert slope.dtype == self.dem.dtype
+        assert tpi.dtype == self.dem.dtype
+
+        # Using a different output dtype
+        out_dtype = np.float64
+        slope, tpi = xdem.terrain.get_terrain_attribute(self.dem, attribute=["slope", "topographic_position_index"], out_dtype=out_dtype)
+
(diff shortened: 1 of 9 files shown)


tests/test_terrain.py
```diff
@@ -394,3 +394,21 @@ def test_get_quadric_coefficients(self) -> None:
...
+
+        # Using a different output dtype
+        out_dtype = np.float64
+        slope, tpi = xdem.terrain.get_terrain_attribute(
+            self.dem, attribute=["slope", "topographic_position_index"], out_dtype=out_dtype
+        )
+
+        assert self.dem.dtype != out_dtype
+        assert np.dtype(slope.dtype) == out_dtype
+        assert np.dtype(tpi.dtype) == out_dtype
```
adebardo, line 414: Are we still consistant to numpy 1.X now we are working with 2.X? 

with numpy 1.26.4 with main geoutils I got failures
image

rhugonnet: We are not consistent with tests (I don't think it's possible to be in some instances, in particular for these doctests).
But we're consistent with the functionalities themselves, where we are lucky there were no impacting changes :slightly_smiling_face: (maybe not the ones in `xdem/volume.py`, but those are not so maintained and will be removed in time).

PR #723, opened 2025-04-29
Fixing blockwise documentation and overlap

adebardo:


doc/source/conf.py
```diff
@@ -146,6 +146,9 @@ def setup(app):
     app.registry.source_suffix.pop(".ipynb", None)
 
 
+# To avoid warning in doc about standard error
+suppress_warnings = ["mystnb.stderr"]
```
rhugonnet, line 150: For what warning was this added for?

To manage the notebook warning/showing an output stderr in the code, we had been using this:
https://github.com/GlacioHack/geoutils/blob/8a88c8467c53e3f280fa0b18a3558c192f9fc215/doc/source/conf.py#L57
And for cells that show a warning on purpose (that we want the users to see in the doc), we override with this:
https://github.com/GlacioHack/geoutils/blob/8a88c8467c53e3f280fa0b18a3558c192f9fc215/doc/source/quick_start.md?plain=1#L63

I'm not sure, but I think the suppress argument might silence all cells regardless of we want to show them or not?

PR #725, opened 2025-04-30
Replace logging.warning to logging.info

adebardo:


rhugonnet:
Sorry if see this just now, I had no idea what the warnings that were suppressed were in the other PR so it was hard to grasp the picture for me (which is why I asked). 
Now I'm seeing it here: https://github.com/GlacioHack/xdem/blob/cba9474e863b4cfe42fb5c11fe3f3245e0644202/xdem/coreg/blockwise.py#L150
It seems like this is linked to normal behaviour and appears every time when running blockwise with `mp_config`, not during an exception. Should we simply change those to `logging.info()` then?

That and rolling back the `suppress_warnings` should solve the issue in a couple lines :wink:.

We can do the same for the one in GeoUtils (but we don't have to re-release just for having this in xDEM, it's not that important, we can wait until we naturally do another release in a few weeks).
https://github.com/GlacioHack/geoutils/blob/8a88c8467c53e3f280fa0b18a3558c192f9fc215/geoutils/raster/distributed_computing/multiproc.py#L256

adebardo:
I add the `---
mystnb:
  output_stderr: show
---` and it works for doc compilation. I think this information is more important than just an info in the log 
Don't you agree ?

vschaffn (approved):

Issue #154, opened 2021-06-24
Improve management of optional imports


rhugonnet:
Yes, I thought about the same thing (actually Fabien did!). I was quickly browsing to see if there's a good practice on how to handle optional imports. The only improvement I could find is to use importlib to check if a module exists or not: https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder.find_spec

_Originally posted by @adehecq in https://github.com/GlacioHack/xdem/pull/151#discussion_r656272585_

rhugonnet:
Forgot this before releasing: we should move `noisyopt` to optional imports

rhugonnet:
After all this time, this is done rigourously!
I'll make a Wiki that I'll link in https://github.com/GlacioHack/geoutils/pull/810

PR #727, opened 2025-05-02
Use sklearn instead of pyransac3d

adebardo:
# Context

This PR includes the change from pyransac3d to the sklearn.linear\_model module.

# Dependencies

We have removed all calls to and references to pyransac3d.

# Code

We have refactored the ransac function, which now appears less non-deterministic than before.

# Tests

We have updated the unit tests.

### DIFF
--- a/.github/workflows/pip-checks.yml
+++ b/.github/workflows/pip-checks.yml
@@ -29,7 +29,6 @@ jobs:
       - name: Install project
         run: |
           python -m pip install . -vv
-          pip install pyransac3d
 
       # Check import works
       - name: Check import works with base environment
--- a/NOTICE
+++ b/NOTICE
@@ -125,7 +125,3 @@ noisyopt: Library for optimizing noisy functions in Python.
 Copyright (c) 2016-2017 Andreas Mayer.
 Website: https://pypi.org/project/noisyopt/
 License: MIT.
-
-pyransac3d: Implementation of Random sample consensus (RANSAC) method
-Website: https://pypi.org/project/pyransac3d/
-License: Apache License 2.0
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -53,7 +53,6 @@ dependencies:
     - noisyopt
     # "Headless" needed for opencv to install without requiring system dependencies
     - opencv-contrib-python-headless
-    - pyransac3d
 
     # To run CI against latest GeoUtils
 #    - git+https://github.com/rhugonnet/geoutils.git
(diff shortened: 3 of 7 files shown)


rhugonnet:
Perfect, thanks a lot! :slightly_smiling_face: 

Only thing left is to update the imports to be optional like this: https://github.com/GlacioHack/xdem/blob/e1e2065ea52a7db3c8e88045d743aa1f6efd08bb/xdem/fit.py#L47

And raise an error during `BlockwiseCoreg.apply()` if the import failed, like this: 
https://github.com/GlacioHack/xdem/blob/e1e2065ea52a7db3c8e88045d743aa1f6efd08bb/xdem/fit.py#L420C10-L421C88

Is the 2/3D plane fit the only `apply()` method right now? If yes, we could even raise the warning during `__init__` because the users don't have a choice. (And we can move it to apply later, if we add another `apply()` method :wink:)

rhugonnet:
Right now the entire module requires SciKit-Learn to be imported. From experience this can create two problems:
- If another module imports something from `blockwise.py` in the future, and the user doesn't have this optional import, it will make the other module (and potentially entire package) fail on import,
- If the user wants to use another `apply_method` than the RANSAC plane fitting in `BlockwiseCoreg`, it won't work either, even though he won't need SciKit-Learn.

So I think we probably should save a `_has_sklearn` variable without raising an error, and test it + raise error within the method only, exactly like on the example I linked. :slightly_smiling_face: 
If you know of a better practice for this, we're happy to hear about it, we were thinking about formalizing how to manage optional imports at some point early in the package development:
- https://github.com/GlacioHack/xdem/issues/154

rhugonnet (approved):
Perfect!
I'm not exactly sure why you also removed `robust_ransac` and kept only `_ransac`. By "less non-deterministic" you mean "more deterministic"?
For all randomized functions in GeoUtils/xDEM, we pass a `random_state` argument to pin the behaviour during tests (it's saved in the `Coreg` metadata), which is quite useful for both tests and reproducibility. 
We should also pass it to `RANSACRegressor`, what do you think? Maybe we can open a small issue (or add it here if it's little effort, as you want).

adebardo:
Hi @rhugonnet :)
With pyransac, it could sometimes happen that `_ransac` would fail 1 out of 10 times without any clear reason. We had `robust_ransac` to avoid that, but we haven’t encountered this issue with the new RANSAC, so I removed it.
 
I'm not sure I fully understand the idea behind random_state, sorry.

rhugonnet:
All done, thanks! :smile: 
Don't forget to add "Resolves #XXX" in the PR descriptions in the future, so that issues are closed/linked automatically :wink: (here I had to do it manually, and edited the PR description for the record)

PR #732, opened 2025-05-07
Finish replacing `nmad` by version in GeoUtils

adebardo:
I'm waiting for the CI to see the warnings of pytest. 

I think the 400 warnings come from the install

### DIFF
--- a/xdem/spatialstats.py
+++ b/xdem/spatialstats.py
@@ -2902,7 +2902,7 @@ def patches_method(
     """
     Monte Carlo patches method that samples multiple patches of terrain, square or circular, of a certain area and
     computes a statistic in each patch. Then, another statistic is computed between all patches. Typically, a statistic
-    of central tendency (e.g., the mean) is computed for each patch, then a statistic of spread (e.g.,the gu.stats.nmad)
+    of central tendency (e.g., the mean) is computed for each patch, then a statistic of spread (e.g., the NMAD)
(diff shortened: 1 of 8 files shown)


rhugonnet:
I think the warnings (still 4000 of them in the CI) come mostly from the default parameters in `xdem.spatialstats` that use the deprecated `nmad()` defined right above, like here: https://github.com/GlacioHack/xdem/blob/e1e2065ea52a7db3c8e88045d743aa1f6efd08bb/xdem/spatialstats.py#L82

If we replace those instances with `gu.stats.nmad`, all the warnings should go away :slightly_smiling_face:

adebardo:
image
 warnings summary actuel :)

xdem/spatialstats.py
```diff
@@ -49,37 +48,18 @@
...
 
 with warnings.catch_warnings():
     warnings.filterwarnings("ignore", category=DeprecationWarning)
     import skgstat as skg
 
 
-@deprecate(
-    removal_version=Version("0.4"), details="xdem.spatialstats.nmad is being deprecated in favor of geoutils.stats.nmad"
-)
-def nmad(data: NDArrayf, nfact: float = 1.4826) -> np.floating[Any]:
```
rhugonnet, line 62: This function + decorator need to stay until version 0.4 (see message of the deprecation). Otherwise users will have an unexplained breaking change.

rhugonnet:
Thanks, we just need to add back the original function as explained above, then we can merge this :slightly_smiling_face:.

PR #735, opened 2025-05-20
Remove `dem_coregistration` in favor of `DEM.coregister_3d`

### LINKED ISSUE #722: Merge dem_coregistration and coregistration_3d
# Context

Currently in `xDEM`, there are two ways to perform a coregistration:  
- `dem_coregistration(dem_a, dem_b)`  
- `dem_a.coregister_3d(dem_b)`  

# Development Plan

- [ ] Make the `dem_coregistration` function **private**
- [ ] Call the (now private) `dem_coregistration` function from `coregister_3d`, and **bring it up to date**
- [ ] **Remove references** to `dem_coregistration` in the documentation
- [ ] **Update tests** to reflect the changes
- [ ] **Relax input validation** to allow coregistration of **identical DEMs**  
  → Show a **warning** rather than throwing an error

adebardo:
# Context
In this PR, we propose to unify the dem_coregistration and coregister_3d functions in order to simplify the work for developers as well as for users who wish to automate their coregistration process.

# Code Changes
- Removed dem_coregistration and its references
- Simplified the automatic coregistration pipeline
- Updated tests

# Outstanding Questions
- The purpose of workflows.py
- What to do with create_inlier_mask

Resolves #722


adebardo:
If you want some informations about the coregistration method parameters and results (x, y and z shift shifts for example)

adebardo:
@rhugonnet @adehecq it's ready :)

rhugonnet:
Thanks @adebardo! 
Not much from me as we already iterated in https://github.com/GlacioHack/xdem/issues/722. I expect more for @adehecq, let's tag him on Slack.

Two comments:

1. I don't think we need to add `coreg_method` as an output, it's already updated by the call:
```python
# Define the method
coreg_method = xdem.coreg.NuthKaab()
# Pass it
dem_aligned = dem.coregister_3d(ref_dem, coreg_method)
# The method was updated with the metadata
coreg_method.info()
``` 
2. The `aligned_reprojected` and `resample` arguments seem to be the same thing. Let's use only one, the same one as for `Coreg.fit_and_apply` for consistency (which is `resample`).

Regarding your questions: If we later need workflows for the CLI+API, we can always implement them based on that old code. For `create_inlier_mask`, it was meant to be a high-level convenience function to facilitate things for users. Given all the examples we have in the documentation that show how to manipulate masks in a couple lines liked I demonstrate in #722 (coregistration or uncertainty examples), I'm not sure it's essential. I think the usefulness would mostly be about optimizing the logic for a non-expert users (in which order things are reprojected, rasterized, combined logically).

Finally: Don't forget to link the issue in the PR description each time :wink:, I had to search for it because I remembered but @adehecq or @belletva might have missed it (I've added it now). Plus it doesn't automatically close the issue when we merge later! I had to do that manually after the last couple of PRs.

xdem/dem.py
```diff
@@ -497,24 +509,105 @@ def coregister_3d(
...
             bias_vars=bias_vars,
             **kwargs,
         )
-        return coreg_method.apply(self)  # type: ignore
+
+        aligned_dem = coreg_method.apply(src_dem, resample=resample, resampling=resampling)
+
+        if aligned_reprojected:
+            # Reprojection needed for statistics and plots
+            aligned_dem = aligned_dem.reproject(reference_elev, silent=True)
```
adehecq, line 587: as mentioned by @rhugonnet, this is already done when option `resample=True`.

xdem/dem.py
```diff
@@ -497,24 +509,105 @@ def coregister_3d(
...
+
+        # # Add the initial shift to the calculated shift
+        if estimated_initial_shift:
+
+            def update_shift(
```
adehecq, line 592: why is it needed to have a function for this? It could be directly in the code without a function.

tests/test_dem.py
```diff
@@ -345,7 +345,7 @@ def test_to_vcrs__grids(self, grid_shifts: dict[str, Any]) -> None:
         # Compare the elevation difference
         z_diff = 100 - trans_dem.data[0, 0]
 
-        # Check the shift is the one expect within 10%
+        # Check the shift is the one expecting within 10%
```
adehecq, line 348: "one expecting" -> "expected one"

tests/test_dem.py
```diff
@@ -360,21 +360,100 @@ def test_terrain_attributes_wrappers(self, terrain_attribute: str) -> None:
...
+                [xdem.coreg.AffineCoreg],
+                id="Simple affine method: DhMinimize without initial shift",
+            ),
+        ],
+    )
+    def test_coregister_3d(coreg_method, initial_shift, expected_pipeline_types) -> None:  # type: ignore
+        """
+        Test coregister_3d functionality
+        """
         fn_ref = xdem.examples.get_path("longyearbyen_ref_dem")
```
adehecq, line 404: As pointed out in the TODO of the original tests, one issue we have is that both input DEMs are already on the same grid, so we never tests that it works with two DEMs on a different projection/grid. This is a bit of work as it requires adding a new test dataset. Should we add an issue about it?

adehecq:
> I don't think we need to add `coreg_method` as an output, it's already updated by the call:

Actually, since `coreg_method` is an optional parameter, yes we need to output it... We could consider not making this argument optional but force the user to define it. This would make the function output simpler, but require a bit more thinking from the user.

adehecq (approved):
Except for a few minor points to be addressed easily (and one bigger point about test to be likely moved to a new issue), this looks good to me! 
Thanks! 🙌

rhugonnet:
> > I don't think we need to add `coreg_method` as an output, it's already updated by the call:
> 
> Actually, since `coreg_method` is an optional parameter, yes we need to output it... We could consider not making this argument optional but force the user to define it. This would make the function output simpler, but require a bit more thinking from the user.

I think those options are not necessarily mutually exclusive: The user could use the default if he doesn't need the `Coreg` object back. If he does, he has to pass the default method himself.

But if we want them to be mutually exclusive still, I would rather we remove the default method and force the user to always pass a `Coreg` method. A large part of the purpose of having `Coreg` as classes is to efficiently update method metadata without requiring to always return the object.

tests/test_dem.py
adebardo: https://github.com/GlacioHack/xdem/issues/737

xdem/coreg/workflows.py
```diff
@@ -1,377 +0,0 @@
...
-        src_dem.translate(shift_x, shift_y, inplace=True)
-
-    if grid == "ref":
-        src_dem = src_dem.reproject(ref_dem, silent=True)
-    elif grid == "src":
-        ref_dem = ref_dem.reproject(src_dem, silent=True)
-
-    # Convert to DEM instance with Float32 dtype
-    # TODO: Could only convert types int into float, but any other float dtype should yield very similar results
-    ref_dem = DEM(ref_dem.astype(np.float32))
```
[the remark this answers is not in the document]
rhugonnet: I think this would belong better directly in `Coreg.fit()` and `Coreg.apply()`.
`DEM.coregister_3d()` should only wrap `Coreg.fit_and_apply()` for convenience, but all functionalities/checks generic to all coregs should live there.

Have we tried running a coregistration with an integer type? NumPy typically converts the output to float when needed, so I wouldn't suprised if it worked nonetheless.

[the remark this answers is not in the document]
adebardo: Yes we can open an issue and take it in the next sprint :)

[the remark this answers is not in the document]
rhugonnet: OK, yes let's remove `grid` temporarily. I'll open the issue.

rhugonnet:
Thanks a lot @adebardo! :slightly_smiling_face: 

Looks good for removing the unnecessary arguments.

For the function return: As discussed above, we should not output the `coreg_method` in `DEM.coregister_3d()`, but leave as it previously was. The `Coreg` object can be defined by the user separately and is then updated naturally without the need for a return (as demonstrated in my short code example above) . This is why the original function did not return it, on purpose.

The comment of @adehecq that went in the other direction was only due to the special case of using the "default" argument where it would be inaccessible (even though, in practice, the default method could be defined separately to get the metadata). I see you've now removed the default argument as we proposed, so this special case is solved. We can now remove the `coreg_method` output without any doubt!

doc/source/coregistration.md
```diff
@@ -67,21 +67,15 @@ Then, coregistering a pair of elevation data can be done by calling {func}`xdem.
...
 
-```{code-cell} ipython3
-# Coregister by calling the dem_coregistration function
-aligned_dem = dem_coregistration(tba_dem, ref_dem, coreg_method=my_coreg_pipeline)
-```
-
 ```{code-cell} ipython3
 # Coregister by calling the DEM method
-aligned_dem = tba_dem.coregister_3d(ref_dem, my_coreg_pipeline)
+aligned_dem, coreg_method = tba_dem.coregister_3d(ref_dem, my_coreg_pipeline)
```
rhugonnet: This return needs to be removed

doc/source/biascorr.md
```diff
@@ -92,14 +92,14 @@ Once defined, they can be applied the same two ways as for coregistration (using
 
 ```{code-cell} ipython3
 # Coregister with bias correction by calling the DEM method
-corrected_dem = tba_dem.coregister_3d(ref_dem, biascorr)
+corrected_dem, _ = tba_dem.coregister_3d(ref_dem, biascorr)
```
rhugonnet: Here too

doc/source/api.md
```diff
@@ -236,7 +236,7 @@ To build and pass your coregistration pipeline to {func}`~xdem.DEM.coregister_3d
 .. autosummary::
     :toctree: gen_modules/
 
-    coreg.workflows.dem_coregistration
+    DEM.coregister_3d
```
rhugonnet: This entire subsection "Quick coregistration" can be removed, as `DEM.coregister_3d()` is already listed further above in the API.

xdem/dem.py
```diff
@@ -481,14 +486,17 @@ def fractal_roughness(self, window_size: int = 13, mp_config: MultiprocConfig |
...
+        coreg_method: coreg.Coreg,
         inlier_mask: Mask | NDArrayb = None,
         bias_vars: dict[str, NDArrayf | MArrayf | RasterType] = None,
-        **kwargs: Any,
-    ) -> DEM:
+        estimated_initial_shift: list[Number] | tuple[Number, Number] | None = None,
+        random_state: int | np.random.Generator | None = None,
+        resample: bool = False,
+        **kwargs,
+    ) -> DEM | GeoDataFrame:
```
rhugonnet: The output type here can't be a `GeoDataFrame`, only a `DEM`?

xdem/dem.py
```diff
@@ -499,22 +507,84 @@ def coregister_3d(
...
+        :param estimated_initial_shift: List containing x and y shifts (in pixels). These shifts are applied before \
+            the coregistration process begins.
+        :param random_state: Random state or seed number to use for subsampling and optimizer.
+        :param resample: If set to True, will reproject output Raster on the same grid as input. Otherwise, only \
+            the array/transform will be updated (if possible) and no resampling is done. \
+            Useful to avoid spreading data gaps.
         :param kwargs: Keyword arguments passed to Coreg.fit().
 
-        :return: Coregistered DEM.
+        :return: Coregistered DEM as a xdem.DEM instance
```
rhugonnet: Not sure we need to specify "as a `xdem.DEM` instance" in the text, it's the main object of the package and thhe type is already shown by the typing declaration :wink:

rhugonnet (approved):

PR #741, opened 2025-05-28
Fix `DEM` casting to `Mask` by removing `from_array` overloading

### LINKED ISSUE #739: Automatic casting to Mask does not work with DEM instances
For a geoutils.Raster instance, when performing comparisons, the output is [automatically casted to a Mask instance](https://geoutils.readthedocs.io/en/stable/feature_overview.html#casting-to-mask-indexing-and-overload). This does not work with an xdem.DEM instance. In that case, the output remains a DEM instance.
See MWE below:
```
import geoutils as gu
import xdem
rst = gu.Raster(gu.examples.get_path("everest_landsat_b4"))
mask = rst>200
# output `mask` of type geoutils.raster.raster.Mask
dem = xdem.DEM(gu.examples.get_path("exploradores_aster_dem"))
mask = dem>200
# output `mask` of type xdem.dem.DEM...
```

rhugonnet:
Replaces #740 

This is an easier fix than #740: Simply an `if` statement depending on the output of `Raster.from_array`. 

Another way to do this for a future re-work (if we want users to more easily subclass `Raster`, for example) would be to separate `Raster.from_array()` into a non-public function that does the casting under-the-hood for all arithmetic ops (`Raster._from_array()`, which wouldn't be overloaded by `DEM.from_array()`) and a public function that works only for building a Raster as intended (refuses a `boolean` input).
Actually, we might have to already do this for the Xarray accessors, for consistency in the behaviour.

Resolves #739

### DIFF
--- a/tests/test_dem.py
+++ b/tests/test_dem.py
@@ -158,6 +158,15 @@ class TestDEM:
         )
         assert dem.vcrs == CRS("EPSG:5773")
 
+    def test_from_array__cast_mask(self) -> None:
+        """Test that DEMs are cast into mask for a logical operation."""
+
+        transform = rio.transform.from_bounds(0, 0, 1, 1, 5, 5)
+        dem = DEM.from_array(data=np.ones((5, 5)), transform=transform, crs=CRS("EPSG:4326"), nodata=None)
+
+        mask_dem = dem > 1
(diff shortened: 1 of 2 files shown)


rhugonnet:
As the fix is just 2 lines, should be good to merge.

PR #749, opened 2025-06-13
736 adapt size for apply blockwise

adebardo:
WIP


xdem/coreg/blockwise.py
belletva, line 1: Add also this NB in the doc ?

xdem/coreg/blockwise.py
```diff
@@ -167,7 +182,14 @@ def fit(
...
         )
 
-        self.shape_tiling_grid = compute_tiling(self.block_size, reference_elev.shape, to_be_aligned_elev.shape).shape
+        self.shape_tiling_grid = compute_tiling(
+            self.block_size_fit, reference_elev.shape, to_be_aligned_elev.shape
+        ).shape
+
+        # RANSAC interpolator need minimum 5 tiles
+        if self.shape_tiling_grid[0] * self.shape_tiling_grid[1] < 5:
+            raise ValueError("Not enough block, need at least 5, please reduce block_size_fit parameter")
```
rhugonnet: Hm, it's a bit counter-intuitive that a small number of tiles (= large size tiles) would trigger an error... It feels like it should rather be the opposite: When there is a large number of tiles (= small size tiles), there probably aren't enough points to perform a coregistration correctly, and a warning/error could be raised.

But users should be able to use simple tilings (for instance, 4 or 2 tiles) without problems.

In terms of the fit, this should not be an issue:
- To have a unique solution at fitting a 2D plan, I think one needs 3 point minimum.
- If there are only 2 tiles, then the fit in not 2D anymore, it's either only in X, or only in Y. So 2 points are sufficient to constrain it.
- If there is only 1 tile, it's the equivalent of running a normal coregistration.

In terms of the RANSAC steps specifically: Maybe we don't need to run RANSAC below X number of tiles (where does the number 5 come from?), because there aren't enough tiles to detect outliers anyway (and with such large tile sizes, there shouldn't be any outliers).
We could move the RANSAC to trigger only above a certain number of tiles (and for the rest, default to a simple 2D plan fit with `scipy.optimize.curve_fit()`). This could also be the occasion to make the RANSAC step optional altogether to keep modularity for users.

adebardo: I removed everything related to min_inlier. I think that option came from the old RANSAC implementation and no longer makes sense now, as you explained.

I'm opening a ticket for using scipy.optimize.curve_fit() when there are only 2 tiles.

rhugonnet: Perfect!
For 2 tiles: This is not related to `scipy.optimize.curve_fit()`, it's also an issue for the current `RANSACRegressor`. And not only an issue for 2 tiles.

If the tiles are in distributed 1D (only along X or in Y, for 2 tiles it's the only option, but could be 10 tiles as well if the user splits a (5000, 1000) DEM in (500, 1000) chunks for instance), the regression problem needs to be re-defined to 1D as well or it will (very likely) fail every time due to lack of constraints.

The adjustment is fairly easy. For `RANSACRegressor`, the regression step here: https://github.com/GlacioHack/xdem/blob/aad0d2c5a49769d4c4c02b1f4359fb5f153fd392/xdem/coreg/blockwise.py#L242 would need to pass only `points[:,0]` or `points[:, 1]` to the regression depending on the dimension, get a single coefficient out, and apply only that coefficient later (we can then set the other coefficient with a zero value to avoid having to change anything in the rest of the code :wink:).

rhugonnet: Thinking more about it, because `points` is squeezable to 1D in the above mentioned cases, I guess this could actually be supported by `RANSACRegressor` (that might accept that the second coefficient can't be constrained anyway... Unsure but it's a possibility). In that case we wouldn't need to change anything (except for the edge case of a single tile maybe?).
We could add a small test to verify if those cases work (1D tiles, single tile)!

adebardo: Hi @rhugonnet,

Sorry for digging up this ticket 😅

I understand your feedback, but handling non-square tiles seems a bit more complex to implement at this point (especially since it would involve changing the MultiprocConfig signature). I’ll open a separate ticket to address that aspect properly.

Thanks for your understanding!

rhugonnet: The tiles don't necessarily have to be non-square for this problem, it's just about the number of dimensions in which there are tiles.
For example: 
DEM of size (1000, 500). 
Tile size of (500, 500). 
There are two tiles, but only changing coordinates in the X dimensions.

I don't immediately see how this would affect the Multiproc call (if it loops over 1 or N coordinates in Y should work the same?), but I may be missing something here. 

If that's still problematic with the clarification above, OK to open an issue in the meantime. But this should be priority, because it'll probably open fairly often for users. (many DEMs are not square and might be chunked only in 1D).

rhugonnet: And it should be (if I didn't miss anything) a tiny change to support it, or even already supported by our current code (and we just need to check it in `test_blockwise`).

doc/source/coregistration.md
```diff
@@ -579,17 +579,27 @@ methods, and create edge artefacts.
...
+Sometimes, we want to split a coregistration across different spatial blocks of an elevation dataset, running that
 method independently in each subset.
 It can also happen that memory requirements exceed the capacity of the computers.
 In such cases, it may be necessary to process the data in blocks and then aggregate the results afterward.
 
+> **Note:**
+> The `block_size_fit` parameter adjusts the size of the tiles over which the coregistration methods are computed.
+>
+> The `block_size_apply` parameter allows the DEM to be aligned in blocks to prevent memory leaks. In such case
+> reduce block_size_apply value.
```
rhugonnet: We should write "to optimize memory usage" instead of "to prevent memory leaks". 

Memory leaks is incorrect memory build-up in a program, e.g. when memory is allocated and not released later. Here it's all correctly allocated, we just decrease the peak memory usage by running smaller tiles :slightly_smiling_face: 

Then, instead of "In such case reduce block_size_apply value", we can explain the trade-off: "Smaller blocks during the `apply` step reduce memory usage but increase computing time".

xdem/coreg/blockwise.py
```diff
@@ -214,7 +232,6 @@ def _ransac(
         y_coords: tuple[float, float, float],
```
rhugonnet: I think wrong type for `x_coords`, `y_coords` and `shifts`, should be `NDArrayf`?

tests/test_coreg/test_blockwise.py
```diff
@@ -153,3 +146,40 @@ def test_blockwise_coreg_pipeline(self, blockwise_coreg, example_data, tmp_path)
 
         valid = (expected.data.data != expected.nodata) & (aligned.data.data != aligned.nodata)
         assert np.allclose(expected.data.data[valid], aligned.data.data[valid], atol=20)
+
+    def test_ransac_on_horizontal_tiles(self, blockwise_coreg) -> None:
+        """Test case where RANSAC works on horizontal tiles."""
+        x = np.linspace(0, 100, 50)
+        y = np.full_like(x, 50)
```
rhugonnet, line 153: Not sure this is testing 1D: Both `x` and `y` are of size 50? I think one of them should be of size 1.
Same for vertical below.

rhugonnet:
Thanks! Some small comments above. :wink: 
For testing the robustness of different tile shapes (1D horizontal, 1D vertical and 2D): Right now the tests only call `_ransac()`, we might as well modify those to call directly the entire process `blockwise.fit_and_apply()`, as there are good chances that other things might break the edge cases outside of `_ransac()`.

adebardo:
@rhugonnet as seen today(yesterday) @pytest.mark.parametrize("block_size", [500, 985, 1332], ids=["2d_shifts", "1d_shifts_x", "monotile"]) :)

rhugonnet (approved):
Perfect, thanks a lot! :smile:

PR #759, opened 2025-07-16
Improve coregistration modularity and revamp uncertainty module

### LINKED ISSUE #300: Move `convolution` methods to `geoutils` once mature
Those are introduced in #292 and completely generic, only used for the `patches_method`, and maybe soon for vectorizing functions of `terrain.py`.
We should move these functionalities to `geoutils` once they have reached maturity.

### LINKED ISSUE #378: Re-structure `spatialstats.py`

### LINKED ISSUE #596: Merge options `fit_optimizer` and `fit_minimizer` + `fit_loss_func` for coregistration methods
Right now, we have a `fit_optimizer` which defaults to `scipy.optimize.curve_fit()` solving a problem y = f(x, *args), with the loss function included in the function kwargs.

However, this generic implementation does not work for certain methods such as `DhMinimize`, because of a technical problem with NaNs: 
- For `DhMinimize`, in order to use `scipy.optimize.curve_fit`, we can simple define the function `f(x, *args)` as the function returning the flattened vector of elevation differences between the two datasets (which are the constant `x`, doesn't even need to be passed) for a certain shift of X/Y (the `args`), and define `y` as zero, which makes us fall back on a classic minimization problem with residuals `y - f(x, *args)`.
- However, after some testing, it appears that the function `scipy.optimize.curve_fit()` expects that `f(x, *args)` always returns **a vector of fixed length with only finite values**. When applying a shift in X/Y evaluated at the same subsampled points (re-subsampling at every iteration would take ages and prevent the RAM savings of subsampling), we create more and more NaNs, which makes the algorithm fail. When returning only the finite values, the algorithm fails because `f` changes vector length between iterations...

Hence, in #595, this method is implemented using `scipy.optimize.minimize` instead (define as a `fit_minimizer`), that expects directly the residuals already reduced by a loss function `fit_loss_func` (`nmad` by default).

This is annoying, because we now have two different coregistration parameters doing roughly the same things (`fit_optimizer` and `fit_minimizer` + `fit_loss_func`). We should homogenize this...
I see two solutions:
1/ (Long shot) Open an issue in SciPy to allow either vector of different length or NaN values within `curve_fit` methods,
2/ Use `fit_minimize` + `fit_loss_func` everywhere.

I'm not sure yet if 2/ makes sense on an mathematical equivalent point-of-view: `scipy.optimize.curve_fit()` relies on `least_squares`, and not `minimize`. It is also a minimization, but does not support the same methods...
Is there actually a way to merge those two types (curve fitting, and generic minimization) consistently together? I'm at the limit of my expertise here.

### LINKED ISSUE #947: Revamp of uncertainty module
The aim of this thread is to **conceptualise what would be our ideal uncertainty module API**, for longevity, ease-of-use and modularity. 😄 
This is assuming we already have scalable variography tools consolidated somewhere (SciKit-GStat or internal); which is a separate topic.

In coordination with @Cassandra-Brigham
Also very relevant to the coregistration uncertainty API we've been recently discussing with @dshean
It would be good to have a broader opinion on the API @belletva @adehecq @marinebcht 

Linked discussions: https://github.com/GlacioHack/xdem/issues/378, https://github.com/GlacioHack/xdem/issues/588

## Short reminder of current structure and issues

The uncertainty module contains 3 functionalities that rely on different tools under-the-hood: heteroscedasticity estimation, error correlation estimator, and uncertainty propagation.

**Part 1: Estimation and modelling of heteroscedasticity**, which consists of 5 steps:
- Subsampling at similar valid (finite) points for any raster-raster or point-raster and explanatory variables (point or raster), 
- Elevation differencing between the two datasets,
- Binning a spread statistic in N-dimensions,
- Empirical or parametric function fit on this N-D binning,
- Applying the function to explanatory variables to infer errors everywhere.

**Part 2: Estimation and modelling of error correlation**, which consists of 4 steps:
- Elevation differencing of the two datasets,
- (Optional) Standardization by heteroscedasticity, if passed by user (just a division),
- Empirical variogram estimation and model fit,
- Conversion to correlation function.

**Part 3: Uncertainty propagation**, which is currently limited and generally has 2 options:
- Theoretical (only supported for area averages with specific optimizations),
- Numerical (left to the user currently).

Part 1 currently returns a `pd.DataFrame` of binned statistics, a `Callable` function with the N-D function, and a `Raster/PointCloud` error field.
Part 2 currently returns a `pd.DataFrame` with empirical variogram, a `pd.DataFrame` with model fit parameters and a `Callable` with final correlation function.
Part 3 returns the uncertainty value.
All these objects are volative: Separated, difficult to plot (need to know the right functions hidden in xDEM) and relate.

## Proposed API

Here's my idea for an API: We store every relevant parameter in an `ErrorStructure` lightweight class that contains the error covariance structure from  heteroscedasticity + error correlation (a bit like the `Coreg` class contains only input/output parameters of coregistration optimization), which can be passed by users to a new uncertainty propagation function `propagate_uncertainty` which would work on any method such as `coregister()`, `terrain_attribute()` or new `difference()`/`integrate_area()` (We would add an "Elevation differencing" feature section in the package with this type of methods).

The `ErrorStructure` class could look like this:

```
ErrorStructure
- Attributes
  - hetsc  # Store all heteroscedasticity parameters
  - corr  # Store all correlation parameter
  - metadata

- Initialization  # Define default parameters for minimal user inputs
  - init(hetsc=True, hetsc_predictor={"slope", "max_curvature"}, spread_estimator="nmad", corr_nested_ranges=3...

- Inference
  - infer_heteroscedasticity()  # Only "dh" would be required and new parameters would override defaults
  - infer_correlation()  # Same here
  - infer_error_structure()  # Both steps at once

- Plotting
  - plot_heteroscedasticity()  # Binning + fit
  - plot_correlation()  # Variogram + fit
  - plot()  # Both steps in multiple panels
 
- Refit model only
  - refit_heteroscedasticity()
  - refit_correlation()

- Conversions and theoretical propagation
  - predict_std()
  - predict_variance()
  - predict_correlation()
  - to_covariance_matrix()

- For numerical simulation
  - generate_random_field()
```

For instance, the minimal interface would be:
```python
err = ErrorStructure()
err.infer_error_structure(ref_dem=dem, other_dem=other_dem, spread_estimator=np.nanstd)
err.plot()

# Then the user could iterate based on results
err.refit_correlation(nested_models=["Gaussian", "Spherical"], bounds=...)
```
And would also be wrapped in object-based function for those who desire less modularity, exactly like for the mirrored `Coreg.fit_and_apply` and `DEM.coregister()`:

```python
dem = xdem.DEM(path)
err = dem.infer_error_structure()  # Or can pass a predefined object
err.plot()
```

The uncertainty propagation section is where I'm the least clear.
The best might be to have a single `propagate` function that can wrap any method for numerical uncertainty propagation (and some for theoretical, where supported) which would only require to pass an elevation object, a method and the `ErrorStructure`:
```
xdem.uncertainty.propagate(elev: DEM | EPC, method: Coreg | Literal["terrain_attribute", "differencing"], err: ErrorStructure, nsim: int =  50, **kwargs_method) -> Coreg | Raster
```
which would also be available through the objects directly through `DEM.propagate_uncertainty()`.

I think it would be impractical to add a `uncertainty=...` argument to all existing methods like `terrain_attribute`, etc, except when they support theoretical uncertainty propagation (very few cases for now).

## Internal function changes

The `ErrorStructure` class would be mostly for modularity and ease-of-use.

Internally, for scalability and consistency: almost all the `spatialstats` module would be stripped down (sampling, binning, variography 😅) to consolidate modernized functions that are scalable and consistent into a new `uncertainty` module.
Most aspects are done already:
- A new `cosample()` function supports the subsampling at valid values of any point/raster-raster pair of objects including external variables (and be used internally in both coregistration and uncertainty propagation), I've consolidated it in #759,
- The new `grouped_stats()` of GeoUtils should be practical to have consistent N-D binning supporting masking, and remove the old `nd_binning()` of xDEM; A simple subsampling argument can make it scalable without needing chunked support for binning (but that can also be added, we can rely on existing tools there as this is common),
- Variography will be dealt separately in SciKit-GStat or internally, but should be wrapped in a `variogram()` function in GeoUtils working on raster/point clouds objects directly (like Xarray wraps SciPy's `interpn()` to simplify the steps for users), that is scalable (@Cassandra-Brigham's work + Dask support added in #759),
- A `random_field()` function could also be added which would support a lightweight `Variogram` object input that can be converted to any other (running either GSTools/GPyTorch in the background), we already have code for this.

Voilà, that's it for my ideas!
Curious of your thoughts :wink:
Once we've agreed on the API, we can just go for it!

rhugonnet:
This PR adds support for many new parametrizations for coregistration, and performs a revamp of the uncertainty module. Both are done together because some coregistrations now support analytical uncertainty propagation (requiring the new uncertainty objects), while numerical uncertainty propagation is introduced for all features.

Resolves #596 
Resolves #378 
Resolves #947 
Resolves #300

## Context and previous discussions

See discussion on reworking the old `spatialstats` module in #378, and the plan of revamping into an `uncertainty` in #947, and the move of convolutions used in patch methods in #300.
We now rely on generic, scalable binning/variography tools implemented in GeoUtils (https://github.com/GlacioHack/geoutils/pull/925) and supporting Dask/MP.

### New parametrizations for coregistration

This PR adds several options and optimizations to affine coregistrations, namely:
- The `tolerance` argument is replaced by `tolerance_translation`, `tolerance_rotation` (and `tolerance_objective` for `CPD`), and (internally) the coregistration algorithms now all compute the associated statistics at each iteration (magnitude of change in translation, rotation, objective function) and the iteration stops only when **all** tolerances are reached,
- A `trim_residuals` option is added to all methods for outlier filtering, including four sub-options: `trim_central_statistic`, `trim_spread_statistic`, `trim_spread_coverage`, which describe how the trimming is done (for instance median +/- 3NMAD), and `trim_iterative` to either perform the trimming only once, or by re-computing the trimming statistic iteratively until all outliers are removed,
- The `fit_optimizer` is removed in favor of `fit_minimizer` (more generic for optimizer supports, can vary loss function details more easily for instance) and the default consequently changes from `scipy.optimize.curve_fit()` to `scipy.optimize.least_squares()` (takes residuals = Y - f(X) as input, instead of X and Y independently). For this adaptation to work, we mirror some defaults of `curve_fit()`: for example, if undefined, the initial values are set as an array of ones (required argument for `least_squares`),
- The `ICP` old support for a "lsq_linearized" `fit_minimizer` is moved to an argument `linearized=True`, which corresponds the linearization (Taylor expansion) of the "point-to-plane" optimisation from Low (2004). This linearization now also supports any `fit_minimizer` (previously a plain NumPy least-square matrix calculation),
- The `DhMinimize` method is deprecated in favor of a `linearized=False` argument in `LZD` (and, to keep the old behaviour, would have to pass `only_translation=True` and original optimizer/loss functions, because the new `LZD(linearized=False)` now supports rotations as well!). They are actually the same method (both doing a "Least Z-difference" i.e. LZD), with the Rosenholm and Torlegard (1988) formulation being the linearized version; this greatly improves the consistency of naming/arguments in the package,
- A new variant of `CPD` named "local surface geometry" (LSG) is added through the argument `lsg=True`. The LSG variant uses the local normals to better constrain the optimization (like "point-to-plane" does for ICP).
- A new `anisotropic` argument is added to `ICP` and `CPD` to optionally perform automated weighting along the axes X/Y/Z, with strategies "xy_vs_z" and "per_axis". This is very useful to better constrain all rotations/translations for data that has very different extents in X/Y and Z (for instance a DEM that is 50km long, but with only 1km of elevation range),
- A `sampling_strategy` argument is added for point-point methods (`ICP`, `CPD`) which describes how a raster is converted to point cloud before coregistration: It can be set to "same_xy" (both inputs are subsampled at the same X/Y), or "independent" (both inputs are independently subsampled at different random points) or "iterative_same_xy" (both inputs are re-subsampled at the same X/Y every iteration of the algorithm by interpolating the raster input). The new "independent" sampling option required the addition of a new `_subsample` function logic which works on all inputs,
- An internal `reproj_same_grid` argument is added for preprocessing the input datasets in `Coreg.fit()`, it is set as `True` for bias-correction, but `False` for affine coregistrations (because point-point registration can work even if the datasets don't overlap in Z coordinate, so reprojection should not be enforced). This change adds the requirement to pass independently a `ref_transform` and `tba_transform` to subfunctions (previously a single `transform` valid for both grids, which were always reprojected), so this is modified for all `Coreg._fit_rst_rst` subfunctions.
- The iterative output metadata (`last_iteration` saving the last iteration number, and `iteration_stats` saving the iteration statistics compared to each tolerance) were derived but not saved in the final `.meta` until now. We now store these statistics, which can be printed in `info()` by users.

### New uncertainty module

The implementation generally follows what was discussed in #947.

To continue...

## TO-DOs

- [x] Add tests for `reproj_same_grid` (and change boolean to actual value in `fit()`).
- [ ] Merge `DhMinimize` into `LZD`?
- [x] Add support for `LSG-CPD` (uses normals) for a fair comparison of the three algorithms,
- [x] Add support to X/Y vs Z anisotropic weighting for ICP/CPD,
- [x] Add support for general minimizers in CPD,
- [x] Make linearization of `ICP` solve with same minimizers,
- [ ] Add formatting of statistics in `_iterate_method` prints following https://github.com/GlacioHack/xdem/issues/760#issuecomment-3950546429,


rhugonnet:
@belletva @marinebcht The changes of this PR, specifically the addition of the internal `reproj_same_grid` argument in `Coreg.fit()` (details above), made me think of https://github.com/GlacioHack/xdem/pull/873.

In short: Point-point coregistrations (such as `ICP` and `CPD`) work even if datasets are not on the same grid, and it's actually essential to align datasets that have little vertical overlap. So we shouldn't prevent the user from using coregistration when datasets are not on the same grid in #873.

Maybe we just leave the work to `Coreg()` (for instance `NuthKaab` or `LZD` should fail without overlap) and catch/adapt errors to make them more digestible for CLI input/users? What do you think?

PR #765, opened 2025-07-28
Make blockwise robust to block errors

adebardo:
This PR allows the blockwise process to continue even if on tile doesn't converge in fit.


adebardo:
@adehecq can u look at it

PR #768, opened 2025-08-11
Add texture shading to available DEM terrain attributes 

friedrichknuth:
Hi all :wave:

This PR adds `xdem.DEM.texture_shading()` to the available DEM terrain attributes.

The method is based on [Brown (2010)](https://mountaincartography.icaci.org/activities/workshops/banff_canada/papers/brown.pdf) and adapted from the Python implementation by @fasiha at https://github.com/fasiha/texshade-py. The corresponding [license](https://github.com/fasiha/texshade-py/blob/main/UNLICENSE) is quite permissive, so hopefully adapting it is welcomed and OK! 🙂 

## Files changed
- `xdem/terrain.py`
- `xdem/dem.py`
- `tests/test_terrain.py`
- `doc/source/terrain.md`

## Functions added
- `xdem.terrain._nextprod_fft()` - Helper function for optimal FFT sizes
- `xdem.terrain._texture_shading_fft()` - Core fractional Laplacian implementation using scipy.fft
- `xdem.terrain.texture_shading()` - Main texture shading function with parameter validation
- `xdem.DEM.texture_shading()` - DEM class method with alpha=0.8 default and parameter forwarding

## Tests added
- `test_texture_shading()` - Basic functionality and parameter validation
- `test_texture_shading_via_get_terrain_attribute()` - Integration with terrain attribute system
- `test_texture_shading_larger_dem()` - Performance testing on larger dataset
- `test_nextprod_fft()` - Helper function validation

## Documentation added
- Texture shading section in terrain attributes documentation
- Formulation of fractional Laplacian operator
- Comparison and description of complementary use with hillshading
- Basic usage example
- References to Brown (2010) and fasiha/texshade-py implementation

### Disclaimer
I am not an expert of FFT processing and used the Claude Sonnet 4 co-pilot to facilitate integration with `xdem`. There might be another way to handle `_nextprod_fft()` (e.g. `scipy.fft.next_fast_len(n)`?), but kept it close to the original python implementation for now.

### DIFF
--- a/doc/source/terrain.md
+++ b/doc/source/terrain.md
@@ -106,6 +106,9 @@ compilation is cached and can later be re-used in the same Python environment.
    * - {ref}`fractrough`
      - Fractal dimension number (1 to 3)
      - [Taud and Parrot (2005)](https://doi.org/10.4000/geomorphologie.622)
+   * - {ref}`texture-shading`
+     - Unitless
+     - [Brown (2010)](https://mountaincartography.icaci.org/activities/workshops/banff_canada/papers/brown.pdf)
 ```
 
 ```{note}
@@ -406,6 +409,32 @@ DEM pixels. Its unit is that of a dimension, and is always between 1 (dimension 
 ```{code-cell} ipython3
 fractal_roughness = dem.fractal_roughness()
 fractal_roughness.plot(cmap="Reds", cbar_title="Fractal roughness (dimensions)")
+```
+
+(texture-shading)=
+## Texture shading
+
+{func}`xdem.DEM.texture_shading`
+
+The texture shading technique produces an isotropic relief visualization that emphasizes the drainage network structure of terrain, including ridges and canyons. Unlike traditional hillshading, texture shading exhibits scale invariance and orientation independence, making it particularly effective for revealing the hierarchical structure of mountainous topography without directional bias.
+
+Texture shading is computed by applying a fractional Laplacian operator to the elevation data, which acts as a scale-independent high-pass filter. The technique produces relative elevation values where positive values correspond to ridges and peaks (higher than surrounding terrain) and negative values correspond to valleys and canyons (lower than surrounding terrain). The method includes a detail parameter α (typically 0.5-1.0) that controls the emphasis given to fine terrain features versus major landscape structure.
+
+The fractional Laplacian operator L^α is based on [Brown (2010)](https://mountaincartography.icaci.org/activities/workshops/banff_canada/papers/brown.pdf) and computed in the frequency domain as:
+
+L^α(f_x, f_y) = (2π)^α (f_x² + f_y²)^(α/2)
+
+where f_x and f_y are spatial frequencies, and α is the fractional order parameter.
+
+Unlike hillshading, texture shading maintains visual hierarchy across different scales - smaller terrain features automatically have lower contrast relative to larger features, creating a "self-generalizing" effect. This makes texture shading particularly valuable for displaying terrain at multiple zoom levels or for combining with traditional hillshading to enhance the visibility of drainage networks.
+
+Unlike curvature, which highlights local convexity and concavity based on the second derivative of elevation, texture shading emphasizes the multi-scale drainage network structure through a fractional Laplacian that preserves relationships across different spatial frequencies. While curvature is optimal for identifying specific geomorphological features like ridges and valley bottoms at a single scale, texture shading provides a scale-invariant visualization that simultaneously reveals both fine-scale terrain details and broad landscape patterns.
+
+For more information on texture shading, see the [official texture shading website](https://www.textureshading.com/Home.html) and the [Python implementation reference](https://github.com/fasiha/texshade-py).
+
+```{code-cell} ipython3
+texture_shading = dem.texture_shading(alpha=0.8)
+texture_shading.plot(cmap="Greys_r", cbar_title="Texture shading")
 ```
 
 ## Generating attributes in multiprocessing
--- a/tests/test_terrain.py
+++ b/tests/test_terrain.py
@@ -576,3 +576,110 @@ class TestTerrainAttribute:
         assert self.dem.dtype != out_dtype
         assert np.dtype(slope.dtype) == out_dtype
         assert np.dtype(tpi.dtype) == out_dtype
+
+    def test_texture_shading(self) -> None:
+        """Test the texture_shading function."""
+        
+        # Test with a simple DEM
+        dem_simple = np.array([[1, 1, 1], [1, 2, 1], [1, 1, 1]], dtype="float32")
+        
+        # Test basic functionality
+        result = xdem.terrain.texture_shading(dem_simple, alpha=0.8)
+        
+        # Check output properties
+        assert result.shape == dem_simple.shape
+        assert np.issubdtype(result.dtype, np.floating)
+        assert np.all(np.isfinite(result))  # No NaN values for simple case
+        
+        # Test different alpha values
+        result_low = xdem.terrain.texture_shading(dem_simple, alpha=0.5)
+        result_mid = xdem.terrain.texture_shading(dem_simple, alpha=0.8)
+        result_high = xdem.terrain.texture_shading(dem_simple, alpha=1.5)
+        
+        # Results should be different for different alpha values
+        assert not np.array_equal(result_low, result_mid)
+        assert not np.array_equal(result_mid, result_high)
+        
+        # Test with NaN values
+        dem_with_nan = dem_simple.copy()
+        dem_with_nan[0, 0] = np.nan
+        
+        result_nan = xdem.terrain.texture_shading(dem_with_nan, alpha=0.8)
+        assert result_nan.shape == dem_with_nan.shape
+        assert np.isnan(result_nan[0, 0])  # NaN should be preserved
+        
+        # Test error handling
+        with pytest.raises(ValueError, match="Alpha must be between 0 and 2"):
+            xdem.terrain.texture_shading(dem_simple, alpha=-0.1)
+            
+        with pytest.raises(ValueError, match="Alpha must be between 0 and 2"):
+            xdem.terrain.texture_shading(dem_simple, alpha=2.1)
+            
+        with pytest.raises(ValueError, match="Only 'fft' method is supported"):
+            xdem.terrain.texture_shading(dem_simple, method="invalid")
+
+    def test_texture_shading_via_get_terrain_attribute(self) -> None:
+        """Test texture_shading via the get_terrain_attribute interface."""
+        
+        # Test with a simple DEM
+        dem_simple = np.array([[1, 1, 1], [1, 2, 1], [1, 1, 1]], dtype="float32")
+        
+        # Test via get_terrain_attribute
+        result = xdem.terrain.get_terrain_attribute(dem_simple, "texture_shading")
+        
+        # Check output properties
+        assert result.shape == dem_simple.shape
+        assert np.issubdtype(result.dtype, np.floating)
+        assert np.all(np.isfinite(result))
+        
+        # Test with multiple attributes including texture_shading
+        slope, texture = xdem.terrain.get_terrain_attribute(
+            dem_simple, ["slope", "texture_shading"], resolution=1.0
+        )
+        
+        assert slope.shape == dem_simple.shape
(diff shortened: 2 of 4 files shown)


xdem/terrain.py
```diff
@@ -2258,3 +2283,208 @@ def fractal_roughness(
...
+        >>> textured = texture_shading(dem, alpha=0.8)
+        >>> textured.shape == dem.shape
+        True
+        >>> # Higher alpha enhances fine details
+        >>> textured_enhanced = texture_shading(dem, alpha=1.5)
+
+    :returns: Texture shaded array with same shape as input DEM
+    """
+    # Get array from input
+    dem_arr = gu.raster.get_array_and_mask(dem)[0]
```
rhugonnet: We should be able to call `get_terrain_attribute(attribute="texture_shading")` here to avoid repeating the Raster unwrapping/wrapping :)

friedrichknuth: Nice! Yes - that is simpler :) We can now call `get_terrain_attribute()` here and pass along the `alpha` parameter accordingly.

doc/source/terrain.md
```diff
@@ -106,6 +106,9 @@ compilation is cached and can later be re-used in the same Python environment.
    * - {ref}`fractrough`
      - Fractal dimension number (1 to 3)
      - [Taud and Parrot (2005)](https://doi.org/10.4000/geomorphologie.622)
+   * - {ref}`texture-shading`
+     - Unitless
+     - [Brown (2010)](https://mountaincartography.icaci.org/activities/workshops/banff_canada/papers/brown.pdf)
```
adehecq: A bit annoying that there is no proper publication/doi for this. Let's hope the URL remains valid a long time!

doc/source/terrain.md
```diff
@@ -408,6 +411,32 @@ fractal_roughness = dem.fractal_roughness()
...
 ```
 
+(texture-shading)=
+## Texture shading
+
+{func}`xdem.DEM.texture_shading`
+
+The texture shading technique produces an isotropic relief visualization that emphasizes the drainage network structure of terrain, including ridges and canyons. Unlike traditional hillshading, texture shading exhibits scale invariance and orientation independence, making it particularly effective for revealing the hierarchical structure of mountainous topography without directional bias.
+
+Texture shading is computed by applying a fractional Laplacian operator to the elevation data, which acts as a scale-independent high-pass filter. The technique produces relative elevation values where positive values correspond to ridges and peaks (higher than surrounding terrain) and negative values correspond to valleys and canyons (lower than surrounding terrain). The method includes a detail parameter α (typically 0.5-1.0) that controls the emphasis given to fine terrain features versus major landscape structure.
```
adehecq, line 421: This makes me think a bit about the TPI, also implemented in xDEM. I'm curious how both would compare on the same terrain, but TPI is definitely not scale-independent.

tests/test_terrain.py
```diff
@@ -576,3 +576,105 @@ def test_get_terrain_attribute__out_dtype(self) -> None:
...
+
+        assert slope.shape == dem_simple.shape
+        assert texture.shape == dem_simple.shape
+        assert not np.array_equal(slope, texture)  # Should be different attributes
+
+    def test_texture_shading_larger_dem(self) -> None:
+        """Test texture_shading with a larger, more realistic DEM."""
+
+        # Create a synthetic DEM with some terrain features
+        x = np.linspace(-5, 5, 50)
```
adehecq: This synthetic DEM is still not very large. How long do the tests rake to run? If short (<1 s), you could almost skip the other tests and go directly with that one, no?

tests/test_terrain.py
```diff
@@ -576,3 +576,105 @@ def test_get_terrain_attribute__out_dtype(self) -> None:
...
+        # Check output properties
+        assert result.shape == dem_simple.shape
+        assert np.issubdtype(result.dtype, np.floating)
+        assert np.all(np.isfinite(result))  # No NaN values for simple case
+
+        # Test different alpha values
+        result_low = xdem.terrain.texture_shading(dem_simple, alpha=0.5)
+        result_mid = xdem.terrain.texture_shading(dem_simple, alpha=0.8)
+        result_high = xdem.terrain.texture_shading(dem_simple, alpha=1.5)
+
```
rhugonnet, line 598: Are there known values we can test here? 
For example, if the DEM is flat (e.g., `np.ones((3, 3))`), the texture shading should be a special value? (zero?)
Same for a fixed slope terrain? `np.array([[0, 0, 0], [1, 1, 1], [2, 2, 2]], dtype="float32")`

tests/test_terrain.py
```diff
@@ -576,3 +576,105 @@ def test_get_terrain_attribute__out_dtype(self) -> None:
...
+        assert result.shape == dem_synthetic.shape
+        assert np.issubdtype(result.dtype, np.floating)
+        assert np.all(np.isfinite(result))
+
+        # Check that texture shading enhances features (should have higher variance than flat areas)
+        flat_region = np.ones((50, 50), dtype=np.float32) * 100
+        flat_result = xdem.terrain.texture_shading(flat_region, alpha=1.0)
+
+        # Textured terrain should have higher variance than flat terrain
+        assert np.var(result) > np.var(flat_result)
```
rhugonnet: Are texture values bounded (like hillshade between 0 and 255?)?
If yes, we could add a test on the bounds on the values for this larger test case.

xdem/terrain.py
```diff
@@ -2258,3 +2295,185 @@ def fractal_roughness(
...
+
+    :raises ValueError: If alpha is not between 0 and 2
+
+    :examples:
+        >>> import numpy as np
+        >>> # Create a simple test DEM with a ridge
+        >>> dem = np.zeros((50, 50))
+        >>> dem[20:30, :] = np.sin(np.linspace(0, np.pi, 50)) * 10
+        >>> textured = texture_shading(dem, alpha=0.8)
+        >>> textured.shape == dem.shape
```
rhugonnet, line 2467: Same remark here as above in the tests: If we know the value for a flat case, let's test it here!

xdem/terrain.py
```diff
@@ -2258,3 +2295,185 @@ def fractal_roughness(
...
+    dem_fft = fft.fft2(dem_padded)
+
+    # Apply fractional Laplacian in frequency domain
+    result_fft = dem_fft * laplacian_filter
+
+    # Transform back to spatial domain
+    result_padded = np.real(fft.ifft2(result_fft))
+
+    # Extract the original size from padded result
+    result = result_padded[pad_rows : pad_rows + rows, pad_cols : pad_cols + cols]
```
rhugonnet: In terms of memory usage, we're creating 4 different variables the size of the DEM array above: `dem_fft`, `result_fft`, `result_padded` and `result`.
Should we overwrite the `result` variable in-place 4 times instead?

xdem/terrain.py
```diff
@@ -2258,3 +2295,185 @@ def fractal_roughness(
...
+
+    # For small sizes, use powers of 2
+    if n <= 1024:
+        return int(2 ** np.ceil(np.log2(n)))
+
+    # For larger sizes, find the smallest m >= n such that m = 2^a * 3^b * 5^c * 7^d
+    factors = [2, 3, 5, 7]
+    candidate = n
+
+    while True:
```
rhugonnet, line 2320: I've rarely seen a `while True` haha, is it a previous parameter that disappeared? :stuck_out_tongue: 
Should we remove this?

tests/test_terrain.py
rhugonnet: If we can use the smaller tests above to check for texture values on DEMs we know (e.g., completely flat, fixed slope; see my comment above), then I would leave them. Otherwise, yes, remove them.

To time the news tests, you can run `pytest tests/test_terrain.py -k "texture" --durations=0"` :wink:

xdem/terrain.py
```diff
@@ -2258,3 +2295,185 @@ def fractal_roughness(
...
+    fft_cols = _nextprod_fft(cols)
+
+    # Pad the array for FFT
+    pad_rows = (fft_rows - rows) // 2
+    pad_cols = (fft_cols - cols) // 2
+
+    # Use symmetric padding to reduce edge effects
+    dem_padded = np.pad(
+        dem_work, ((pad_rows, fft_rows - rows - pad_rows), (pad_cols, fft_cols - cols - pad_cols)), mode="symmetric"
+    )
```
rhugonnet, line 2377: Same comment as below on memory usage, we could overwrite `dem_work` in-place to save RAM.

xdem/terrain.py
rhugonnet: (you might have to ignore Mypy with a `type: ignore`, it doesn't always like in-place ops)

tests/test_terrain.py
friedrichknuth: Indeed, this "larger" DEM isn't really that large and the test runs just as fast (0.0s) as on the tiny 3x3 DEMs. I changed the test to operate on the real DEM provided in the test suite and skip the flat terrain validation, since that is now handled by a dedicated small test.

`test_texture_shading_large_dem()` --> `test_texture_shading_real_dem()`

The new test on the real DEM takes 0.07s.

tests/test_terrain.py
friedrichknuth: Added the following tests to check for expected outcomes:

flat surface --> 0 everywhere
`test_texture_shading_flat_surface()` 

 planar ramp --> monotonic increase row to row, constant rows, no variation between columns
`test_texture_shading_planar_ramp()` 

adding vertical shift --> always signed and ~no effect on output
`test_texture_shading_offset_invariance_and_signed()` 

 increased alpha --> increased power in high frequencies
`test_texture_shading_spectral_shift_with_alpha()`

tests/test_terrain.py
friedrichknuth: No, there are no fixed theoretical bounds that are independent of the input DEM. If you scale the DEM by factor c, the texture shading output scales by the same factor. This is now captured by a new test.

linear scaling --> T(c * DEM) ≈ c * T(DEM)
`test_texture_shading_linear_scaling()`

xdem/terrain.py
friedrichknuth: Added flat DEM example to the docstring. 

```
:examples:
       ...
        >>> # Flat surface returns no texture (all 0)
        >>> dem_flat = np.ones((32, 32), dtype=float)
        >>> dem_flat_ts = texture_shading(dem_flat, alpha=0.8)
        >>> np.allclose(dem_flat_ts, 0.0)
        True
       ...
```

xdem/terrain.py
friedrichknuth: To reduce memory usage I now:

1. Pass along and replace `result` where possible
2. Switched to working with the real components of the FFT only
3. Added `overwrite_x=True` flags to reduce peak memory usage
4. Apply the fractional Laplacian in-place

xdem/terrain.py
friedrichknuth: It is mainly a stylistic choice (newly adopted 😄) and there are [other ways to handle it](https://realpython.com/python-walrus-operator/#while-loops), but we need the while loop and there is a clear exit condition that will always be met (integers with only factors 2,3,5,7 occur infinitely often) so I think it is safe 😇.

xdem/terrain.py
friedrichknuth: Done! :)

doc/source/terrain.md
friedrichknuth: Found and added [Allmendinger and Karabinos (2023)](https://doi.org/10.1130/GES02531.1) to the documentation. It has a permanent DOI and describes Leland Brown's method. At least this way we still have something else to point to in case the original work disappears! 😅

xdem/terrain.py
rhugonnet: Ah yes, didn't see the exit!

tests/test_terrain.py
rhugonnet: Amazing, thanks!

rhugonnet:
The new tests are amazing :star_struck:. Thanks for all the work.
One last point I realized: We also need to add texture shading + references to this table: https://xdem.readthedocs.io/en/stable/citation.html#terrain-attributes

Then, we can merge! :smile:

rhugonnet:
Thank you! :tada: 
Merging now :slightly_smiling_face:

PR #772, opened 2025-08-19
Pin setuptools_scm max version to `9.0` to avoid build failures

adebardo:
# Context:

Currently, the CI is failing to find the xdem installations. This is due to a recent release of [setuptools_scm](https://github.com/pypa/setuptools-scm/releases/tag/v9.2.0) : 

While waiting for a 100% python CI without conda to be implemented, I suggest pinning the version of setuptools.

EDIT from @rhugonnet: See issue here: https://github.com/pypa/setuptools-scm/issues/1213

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 # Minimum requirements for the build system to execute.
 requires = [
     "setuptools>=64",
-    "setuptools_scm[toml]>=8",
+    "setuptools_scm[toml]>=8,<9.0.0",
     "wheel",
 ]
 build-backend = "setuptools.build_meta"


pyproject.toml
```diff
@@ -2,7 +2,7 @@
 # Minimum requirements for the build system to execute.
 requires = [
     "setuptools>=64",
-    "setuptools_scm[toml]>=8",
+    "setuptools_scm[toml]>=8,<9.2",
```
belletva: Releases 9.0.0, 9.0.1, 9.0.2, 9.0.3, 9.1.0, 9.1.1 are yanked. I propose: <9.0.0

rhugonnet:
Nice job finding this! I think it's failing in GeoUtils for the same reason: https://github.com/GlacioHack/geoutils/pull/721

adebardo:
Hi @rhugonnet, 
Marine will pin the version in her next PR for geoutils. 

From the error logs, it seemed to me that the installations were not set up and managed to support this new version. Conflicts in the paths were probably due to our installation setup choices.

Issue #661, opened 2024-11-22
Proposal: terrain attributes based on 5x5 polynomial fit


trchudley:
Posted here following discussion with @rhugonnet 

Currently, `xdem` [calculates terrain coefficients](https://github.com/GlacioHack/xdem/blob/main/xdem/terrain.py) over a 3x3 window following Zevenbergen and Thorne (1987) and Horn (1981). This is suitable approach for the vast majority of scenarios, but there are potential alternatives which have advantages in specific situations.

An alternative approach by [Florinsky (2009)](https://doi.org/10.1080/13658810802527499) calculates derivatives by fitting a third-order partial polynomial to a 5x5 window. This larger window can be better suited to high-resolution photogrammetry data as it leads to a local denoising effect. 

I have implemented this method in a `numba`-friendly way in [my own work for ArcticDEM/REMA strips](https://github.com/trchudley/pdemtools/blob/main/src/pdemtools/_geomorphometry.py) and would be happy to help porting this functionality to `xdem` if desired - if only because in the medium term I would like to move my coregistration functions to use `xdem` in the backend!

P.S. Elsewhere in Florinsky's work, there are also a [number of further curvature attributes](https://doi.org/10.1177/0309133317733667) that might be desirable to port into `xdem` and are easy to implement. This includes horizontal, vertical, and gaussian curvature as well as unsphericity.

rhugonnet:
Amazing, your implementations are much shorter, probably also more computationally efficient than ours (I wonder if our "if" loops slow down the process or not at all, I don't have a good enough understanding of `numba`)! :sweat_smile:

For this type of kernel/array size relation, I remember reading a blogpost showing that convolution was typically faster than numba, which is why we have this ongoing: https://github.com/GlacioHack/xdem/pull/486. 
But we need to do some more testing before actually implementing it. It could be the perfect occassion to compare with your numba implementation too, then move forward from here :slightly_smiling_face:

rhugonnet:
Hey @trchudley,

Revisiting this now as I finished optimizing the terrain attributes in xDEM, see #486. 
[...]

PR #777, opened 2025-09-04
New geomorphometry options - 'Florinsky' surface fit & refreshed geometric/directional curvatures

trchudley:
EDIT from @rhugonnet to shortly described the finalized PR:
- **This PR adds support for 5x5 window surface fit from [Florinsky (2009)](https://doi.org/10.1080/13658810802527499)**, in addition to "Zevenberg and Thorne" or "Horn" that were previously implemented (3x3 windows, "Horn" only worked for first derivatives). Internally, we made the derivation of attributes consistent depending on these various `surface_fit` definitions (name that replaces the old `slope_method` parameter, now deprecated), this new structure should work to easily support other surface fits in the future should the need arise,
- **This PR adds rigorous implementation of curvatures following [Minár et al. (2020)](https://doi.org/10.1016/j.earscirev.2020.103414)**. We picked the 6 main curvatures (others can be derived from a sum/difference of those): "profile", "tangential", "planform", "flowline", "maximal/maximum" and "minimal/minimum". All of those can have either a "geometric" or "directional derivative" definition, and we thus include both through a new parameter `curv_method`. In terms of backwards-comp, previous `planform_curvature` corresponds to the new `tangential_curvature` with a "directional" definition, and our previous `profile_curvature` is still `profile_curvature` with a "directional" definition. The old "curvature" is deprecated.

And here's the new doc page: https://xdem-rhugonnet.readthedocs.io/en/geomorpho_docs/terrain.html

## Original description

Work-in-progress following #661. @rhugonnet recommended opening a PR early rather than later!

Currently, have added a "Florinsky" option for surface fit, with surface fit parameters calculated following the 5x5 third-order partial derivatives (https://doi.org/10.1080/13658810802527499).

Now defaults to Florinsky, but alternatives can be requested (e.g. `dem.planform_curvature(surface_fit="ZevenbergThorne")`). Previously, `"ZevenbergThorne"` or `"Horn"` were requested using either `method` or `slope_method`, depending on the function. Now, `surface_fit` is the universal and preferred parameter: requesting `method`/`slope_method` raises a deprecation warning. All done as I was getting used to the structure of the package, so no doubt some very hacky approaches that need tightening up.

Currently, no new geometric/directional derivatives are included, as discussed in #661 - I've simply added Florinsky as an option to existing curvatures. I think more thought is needed about which standard curvatures to choose and how to offer it - hopefully we can have a discussion below.

Yet to update docstrings, documentation, tests, etc etc.

Hope this is a good start!

### DIFF
--- a/examples/advanced/plot_slope_methods.py
+++ b/examples/advanced/plot_slope_methods.py
@@ -2,31 +2,9 @@
 Slope and aspect methods
 ========================
 
-Within `xdem`, slope is always calculated as
-
-$$
-\alpha = \arctan\\!\\left(
-  \\sqrt{
-    \\left(\frac{\\partial z}{\\partial x}\right)^{2}
-    +
-    \\left(\frac{\\partial z}{\\partial y}\right)^{2}
-  }
-\right),
-$$
-
-Where $z$ is elevation, and $x$ and $y$ are the Cartesian co-ordinates.
-
-Aspect is always calculated as
-
-$$
-\theta =
-\\left( -\tan^{-1} \\left(
-    \frac{\\partial z/\\partial x}{\\partial z/\\partial y}
-  \right) - \\pi
-\right) \bmod (2\\pi).
-$$
-
-However, calculating terrain attributes - not only slope and aspect but also curvatures - requires estimating the elevation derivatives of the surface (e.g. $\frac{\\partial z}{\\partial x}$). `xdem` offers three different ways to calculate elevation derivatives, which can result in slightly different results.
+Calculating terrain attributes—not only slope and aspect but also curvatures—requires estimating the
+elevation derivatives of the surface. xDEM offers three different ways to
+calculate elevation derivatives, which can result in slightly different results.
 
 Here is an example of how to generate the two with each method, and understand their differences.
 
@@ -66,6 +44,8 @@ def plot_attribute(attribute, cmap, label=None, vlim=None):
 
 # %%
 # Slope with method of Horn (1981).
+
+# %%
 # .. note::  (GDAL default), based on a refined approximation of the gradient (page 18, bottom left, and pages 20-21).
 
 slope_horn = xdem.terrain.slope(dem, surface_fit="Horn")
@@ -142,4 +122,5 @@ plot_attribute(
 # differences for areas with nearly flat slopes, owing to the high sensitivity of orientation estimation
 # for flat terrain.
 
+# %%
 # .. note:: The default aspect for a 0° slope is 180°, as in GDAL.
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -26,7 +26,7 @@ import xdem
 # -- Project information -----------------------------------------------------
 
 project = "xDEM"
-copyright = "2024, xDEM developers"
+copyright = "2025, xDEM developers"
 author = "Romain Hugonnet, Erik Mannerfelt, Amaury Dehecq and others"
 
 # The full version, including alpha/beta/rc tags


xdem/dem.py
```diff
@@ -89,8 +89,17 @@ class DEM(Raster):  # type: ignore
 
     def __init__(
         self,
-        filename_or_dataset: str | RasterType | rio.io.DatasetReader | rio.io.MemoryFile,
-        vcrs: Literal["Ellipsoid", "EGM08", "EGM96"] | VerticalCRS | str | pathlib.Path | int | None = None,
+        filename_or_dataset: (
+            str | RasterType | rio.io.DatasetReader | rio.io.MemoryFile
+        ),
+        vcrs: (
+            Literal["Ellipsoid", "EGM08", "EGM96"]
```
rhugonnet: Looks like your IDE might be linting the files you modified?
If you install `pre-commit` (pip or conda) and run `pre-commit run --all` from the root of the package (more details in CONTRIBUTING) , it will apply all the linting rules of the repo, so that we only see changes that are not due to linting. :wink:

xdem/terrain.py
```diff
@@ -95,11 +95,116 @@
 h2 = np.array([[-1, -2, -1], [0, 0, 0], [1, 2, 1]])
 horn_coefs = {"h1": h1, "h2": h2}
 
-# Florinsky (2009) coefficients: ASK TOM?
+# Florinsky (2009) coefficients, equations 12-20
```
rhugonnet, line 168: Haha you found this comment :stuck_out_tongue:

xdem/terrain.py
```diff
@@ -344,52 +540,110 @@ def _make_attribute_from_coefs(
...
     if make_curvature:
 
-        # Curvature is the second derivative of the surface fit equation.
-        # (URL in get_quadric_coefficients() docstring)
-        # Curvature = -2(D + E) * 100, see Moore et al. (1991) Equation 16 based on Zevenberg and Thorne (1987)
-        attrs[curv_idx] = -2.0 * (C[zt_d_idx] + C[zt_e_idx]) * 100
+        # TODO: No action for `if surface_fit_id == 0` (invalid). Cannot raise
+        # ValueError here because numba cannot parralelise assertions.
+        # Multiple checks exist before this point, but is there another
+        # thing we can do at this point?
```
rhugonnet: Not sure we can. Relying on previous checks should be sufficient, we can keep this in mind for the tests

xdem/terrain.py
```diff
@@ -344,52 +540,110 @@ def _make_attribute_from_coefs(
...
-            * (
-                C[zt_d_idx] * C[zt_g_idx] ** 2
-                + C[zt_e_idx] * C[zt_h_idx] ** 2
-                + C[zt_f_idx] * C[zt_g_idx] * C[zt_h_idx]
+        # TODO: No action for `if surface_fit_id == 0` (invalid). Cannot raise
+        # ValueError here because numba cannot parralelise assertions.
+        # Multiple checks exist before this point, but is there another
+        # thing we can do at this point?
+
+        if surface_fit_id == 1:
```
rhugonnet: Maybe we can simplify this section by defining:
```python
if surface_fit_id == 1:
    p = zt_g_idx
    q = zt_h_idx
else:
    p = fl_p_idx
    q = fl_q_idx
```
Then use `p` and `q` in a single equation? (otherwise it'll be even more copying once we add the "geometric" vs "directional derivative" definitions)

Hopefully Numba understands this well when compiling... This is the type of change that can break down Numba speed significantly but is hard to track down. 
I have some local scripts I started in a `xdem-benchmark` repo to verify this (not on GitHub yet). If the change doesn't break anything, Numba speed should be ~2 times faster than SciPy for any array size. Like here: https://github.com/GlacioHack/xdem/pull/486#issue-2180613123. 
```python
import os
import xdem
import numpy as np
import pandas as pd
from time import time

rnd = np.random.default_rng(42)

# List of sizes and iterations
size = [100, 200, 500, 1000, 2000, 5000]
nb_it = 10

# Initialize numba JIT compile for both sub-routines (surface fits and windowed indexes)
_ = xdem.terrain.get_terrain_attribute(np.ones((10, 10)), "slope", resolution=1, engine="numba")
_ = xdem.terrain.get_terrain_attribute(np.ones((10, 10)), "roughness", resolution=1, engine="numba")

list_df = []
for attr in xdem.terrain.available_attributes:
    for s in size:
        dem = rnd.normal(size=(1, s, s))

        # Get computation times

        t1 = time()
        for i in range(nb_it):
            _ = xdem.terrain.get_terrain_attribute(dem, [attr], resolution=1, engine="scipy")
        t2 = time()
        for i in range(nb_it):
            _ = xdem.terrain.get_terrain_attribute(dem, [attr], resolution=1, engine="numba")
        t3 = time()

        # Save
        elapsed_scipy = t2 - t1
        elapsed_numba = t3 - t2

        # elapsed_scipy_cv = np.mean(benchmark(xdem.spatialstats.convolution, (dem, "slope", 1, "scipy"), n_repeat=nb_it).gpu_times)
        # elapsed_numba_cv = np.mean(benchmark(xdem.spatialstats.convolution, (dem, "slope", 1, "numba"), n_repeat=nb_it).gpu_times)

        df = pd.DataFrame()
        df = df.assign(elap_numba=[elapsed_numba], elap_scipy=[elapsed_scipy])
        df /= nb_it
        df["s"] = s
        df["attr"] = attr

        print(f"Attribute: {attr}")
        print(f"Size: {s}")
        print(f"Elapsed SciPy: {elapsed_scipy}")
        print(f"Elapsed Numba: {elapsed_numba}")

        list_df.append(df)

# Concatenate
df_all = pd.concat(list_df)
```

xdem/terrain.py
```diff
@@ -595,16 +913,30 @@ def _get_surface_attributes(
...
 
+    # Deprecating slope method
+    if slope_method is not None:
+        warnings.warn(
+            "'slope_method' is deprecated, use 'surface_fit' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        surface_fit = slope_method  # override
+        slope_method = None
```
rhugonnet: Right now this function is non-public (starts with `_`), so we wouldn't need to repeat the deprecation check.
But I think you mentioned that it'd be good to also have access to fit coefficients? If we do this, we might need to modify this function into an intermediate that can also return the coefficients directly. (and make it public)

rhugonnet: Maybe better to tackle this in a separate PR later though

xdem/terrain.py
```diff
@@ -1790,11 +2277,23 @@ def hillshade(
...
+
+    # Deprecating slope method
+    if method is not None:
+        warnings.warn(
+            "'method' is deprecated, use 'surface_fit' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        surface_fit = method  # override
+        method = None
```
rhugonnet, line 2584: To centralize, we could keep the new deprecation check only in `get_terrain_attribute`, and mention both argument names in the warning message there (`method` or `slope_method`, depending on the function level, as you noticed; it was not a good idea to have 2 names to begin with, good the change fixes that! :sweat_smile: ).

xdem/dem.py
rhugonnet: In case you're not familiar: Even if it's called "pre-commit", you can also run it a posteriori and commit the changes due to linting as a separate commit (it's what I do; the drawback is that it makes individual commit harder to revert, but this is rarely useful for me, and for viewing changes at the scale of a PR it's perfectly fine to commit the linting separately).

trchudley: Thanks! Yes, you're right, I've automated `black` on save in my VS Code environment - apologies!

I did actually run `pre-commit` and received a long error. I'm not familiar with it so mentally put it into my 'sort out later' pile. I've pasted it below in case you're familiar with it:
```
          [242/308] Compiling C++ object numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_umath_string_ufuncs.cpp.o
          FAILED: [code=1] numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_umath_string_ufuncs.cpp.o
          c++ -Inumpy/_core/_multiarray_umath.cpython-313-darwin.so.p -Inumpy/_core -I../numpy/_core -Inumpy/_core/include -I../numpy/_core/include -I../numpy/_core/src/common -I../numpy/_core/src/multiarray -I../numpy/_core/src/npymath -I../numpy/_core/src/umath -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Accelerate.framework/Headers -I/Users/tom/miniforge3/envs/xdem-dev/include/python3.13 -I/private/var/folders/h7/6y8bw0210l5fp4pyvbtkx54r0000gn/T/pip-install-bdb54gar/numpy_317c9d632fd647c380983e4a1d2627e6/.mesonpy-qxtx8gwp/meson_cpu -fvisibility=hidden -fvisibility-inlines-hidden -fcolor-diagnostics -DNDEBUG -Wall -Winvalid-pch -std=c++17 -O3 -ftrapping-math -DNPY_HAVE_CLANG_FPSTRICT -DNPY_HAVE_NEON_VFPV4 -DNPY_HAVE_NEON_FP16 -DNPY_HAVE_NEON -DNPY_HAVE_ASIMD -F/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Accelerate.framework -DACCELERATE_NEW_LAPACK -DHAVE_CBLAS '-DBLAS_SYMBOL_SUFFIX=$NEWLAPACK' -DNPY_INTERNAL_BUILD -DHAVE_NPY_CONFIG_H -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE=1 -D_LARGEFILE64_SOURCE=1 -D__STDC_VERSION__=0 -fno-exceptions -fno-rtti -MD -MQ numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_umath_string_ufuncs.cpp.o -MF numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_umath_string_ufuncs.cpp.o.d -o numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_umath_string_ufuncs.cpp.o -c ../numpy/_core/src/umath/string_ufuncs.cpp
          In file included from ../numpy/_core/src/umath/string_ufuncs.cpp:20:
          ../numpy/_core/src/umath/string_fastsearch.h:132:5: error: no type named 'ptrdiff_t' in namespace 'std'; did you mean simply 'ptrdiff_t'?
          
          [snip]
          
          7 errors generated.
          [243/308] Compiling C++ object numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_umath_special_integer_comparisons.cpp.o
          [244/308] Compiling C object numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_umath_umathmodule.c.o
          [245/308] Compiling C object numpy/_core/_simd.cpython-313-darwin.so.p/src__simd__simd.c.o
          [246/308] Compiling C object numpy/linalg/lapack_lite.cpython-313-darwin.so.p/lapack_lite_python_xerbla.c.o
          [247/308] Compiling C object numpy/linalg/lapack_lite.cpython-313-darwin.so.p/lapack_litemodule.c.o
          [248/308] Compiling C++ object numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_npysort_timsort.cpp.o
          [249/308] Compiling C object numpy/linalg/_umath_linalg.cpython-313-darwin.so.p/lapack_lite_python_xerbla.c.o
          [250/308] Compiling C object numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_umath_ufunc_object.c.o
          [251/308] Compiling C++ object numpy/_core/_multiarray_umath.cpython-313-darwin.so.p/src_umath_stringdtype_ufuncs.cpp.o
          [252/308] Compiling C++ object numpy/linalg/_umath_linalg.cpython-313-darwin.so.p/umath_linalg.cpp.o
          [253/308] Compiling C++ object numpy/fft/_pocketfft_umath.cpython-313-darwin.so.p/_pocketfft_umath.cpp.o
          ninja: build stopped: subcommand failed.
          [end of output]
      
      note: This error originates from a subprocess, and is likely not a problem with pip.
    error: metadata-generation-failed
    
    × Encountered error while generating package metadata.
    ╰─> See above for output.
```

xdem/terrain.py
trchudley: Thought so, but yes, hopefully checks will have caught things by this point!

xdem/terrain.py
trchudley: Thanks - yes, this is something I was thinking about too. Ultimately, referring to $p$ and $q$ or $g$ and $h$ are really referring to $\frac{\delta z}{\delta x}$ and $\frac{\delta z}{\delta y}$. The Minár paper refer to them in shorthand as $z_x$ and $z_y$ (and so on - e.g. $z_{xy}$ for $\frac{\delta^2 z}{\delta x \delta y}$). So perhaps, to be 'paper agnostic', we could refer to the as such in the variable names - i.e. `z_x`, `z_y`, `z2_xx`, `z_yy`, `z_xy`? This would be more 'explicit' in the code so long as it explained.

xdem/terrain.py
trchudley: Agreed. I was generous with the checks to help me keep track of things, knowing we could remove later.

xdem/terrain.py
trchudley: Sounds like a good idea, will do that!

xdem/dem.py
rhugonnet: Weird error, probably a conflict with NumPy or Python version.

The easiest way to resolve it should be to re-install the dev environment from scratch:
```
mamba env create -f dev-environment.yml
conda activate xdem-dev
pip install pre-commit
pre-commit run --all
```
You might get a NumPy plugin error with Mypy. That's fine, but hopefully everything else should run.

rhugonnet: Otherwise you can specify in `black` that the line length is 120 for xDEM (only parameter we have for this, then we use defaults; all is parametrized in the pre-commit config file).

xdem/terrain.py
rhugonnet: Agree, `z_x`, `z_y`, etc is much better to link to the Minár paper!

xdem/dem.py
trchudley: I did try this again with the #783 update but it still didn't like it - just switched to a 120 line length for now!

xdem/terrain.py
trchudley: Have coalesced on one check within `get_terrain_attribute`, as per below.

xdem/terrain.py
trchudley: Done - `_make_attribute_from_coefs` now converts method-specific `idx` to `z_x_idx` and so on. The calculations are a lot cleaner now - great idea!

xdem/terrain.py
trchudley: As above, now there's only one check in `get_terrain_attribute`.

xdem/terrain.py
trchudley: Done!

xdem/dem.py
rhugonnet: OK, not a problem. I can checkout your fork/branch once the PR is finalized, and run the linting in just a few minutes (otherwise CI won't pass). :wink:

rhugonnet:
@trchudley: Amazing! As you want for tests/docs, we could set-up a call in about 2 weeks (until then I'm swamped with proposal writing :sweat_smile:). I'll look more into the changes then!

trchudley:
Have updated docs, including updating docstrings in `terrain.py` and updates to the following doc pages that refer to either curvatures or elevation derivatives:

 - `terrain.md`
 - `api.md`
 - `citation.md`
 - `plot_terrain_attributes.py`
 - `plot_slope_methods.py`

I left some of the more detailed analysis in `plot_slope_methods.py` focussed on Horn and ZT. 

I think that should be every page necessary...

rhugonnet:
@trchudley Amazing, thanks!
I've setup the documentation to render on hidden branch here so that we can check the build + output: STILL BUILDING, WILL UPDATE LATER

It's faster than you linking your fork to RTD, you can update that branch by pushing on my fork:
```
# Setting up the remote
git remote add romain https://github.com/rhugonnet/xdem.git
git fetch romain
# Once setup you only need to do this
git checkout romain/geomorpho_docs
git rebase geomorphometry  # Rebase with your own branch
git push romain HEAD:geomorpho_docs
```
But if you want to do the setup of your `geomorphometry` branch directly, here's how: https://github.com/GlacioHack/xdem/wiki/Build-the-documentation-for-a-given-fork-and-branch

I'll get started on tests!
P.S.: I also see there are some breaking changes on other parts of the doc, I'll do that first!

rhugonnet:
Alright I've merged + fixed some small issues in other modules (mostly the uncertainty stuff that relied on max curvature), the doc is now building fully from scratch, see the link above!

tests/test_terrain.py
```diff
@@ -268,6 +372,186 @@ def test_curvatures(self, name: str) -> None:
...
+        if name == "planform_curvature":
+            assert np.allclose(curv_g, curv_d, equal_nan=True)
+        # For tangential, profile and flowline, only the divider is different with a multiplier greater than 1,
+        # so the absolute geometric curvature should always have lower value
+        elif name in ["tangential_curvature", "profile_curvature", "flowline_curvature"]:
+            ind_valid = np.logical_and(np.isfinite(curv_g), np.isfinite(curv_d))
+            assert np.all(np.abs(curv_d[ind_valid]) >= np.abs(curv_g[ind_valid]))
+        # For max/min, no particular relation
+        else:
+            return
```
rhugonnet, line 409: @trchudley Anything we could test here for the relationship between maximal/maximum and minimal/minimum?
For now I only test the relationship of other curvature definitions (geometric vs directional) as above.

tests/test_terrain.py
```diff
@@ -268,6 +372,186 @@ def test_curvatures(self, name: str) -> None:
...
+            if conc == "convex":
+                assert pytest.approx(fl_curv_max) == 0
+                assert pytest.approx(zt_curv_max) == 0
+                assert fl_curv_min < 0
+                assert zt_curv_min < 0
+            else:
+                assert pytest.approx(fl_curv_min) == 0
+                assert pytest.approx(zt_curv_min) == 0
+                assert fl_curv_max > 0
+                assert zt_curv_max > 0
```
rhugonnet, line 554: @trchudley Above, I've for now limited the tests to checking if values were zero/non-zero/positive/negative, based on a synthetic 5x5 DEM, as we expect them depending on convexity + slope orientation + type of curvature.
Do those make sense to you?
Note that for case 2/ (curvature in direction tangent to slope), both flowline and profile are zero.
While for case 3/ (curvature in direction of slope), only profile is non-zero.

Any idea on a simple synthetic DEM where we could check the expected curvature value, quantitatively?

tests/test_terrain.py
```diff
@@ -202,6 +211,92 @@ def test_attribute_functions_against_richdem(self, attribute: str, get_test_data
...
+            # The last curvatures are sensitive in flat terrain, so let's remove that
+            else:
+                ind = moderate_steep
+
+            # Remove terrain
+            diff_zt_fl[~ind] = np.nan
+
+            # The 50% lowest differences should be relatively small (less than 10% of magnitude, e.g. a couple degrees
+            # for slope)
+            assert np.nanpercentile(diff_zt_fl, 50) < 0.1 * magnitude
```
rhugonnet, line 269: @trchudley This test checks that Florinsky or ZevenbergThorne give generally similar values (especially when masking areas where there's high sensitivity of the algorithms).
I also checked this visually, and it looks great! :smile:

xdem/terrain.py
```diff
@@ -109,24 +286,35 @@ def _divider_method_coef(res: float, coef: str) -> float:
...
-        "zt_g": res,
-        "zt_h": res,
+        "zt_a": 4 * res**4,
+        "zt_b": 4 * res**3,
+        "zt_c": 4 * res**3,
+        "zt_d": res**2,  # Divided by 2 compared to Zevenberg to match z_xx definition
+        "zt_e": res**2,  # Divided by 2 compared to Zevenberg to match z_xx definition
+        "zt_f": 4 * res**2,  # Times 2 what is reported in ZevenbergThorne because later formula multiplies by 2
+        "zt_g": 2 * res,
+        "zt_h": 2 * res,
```
rhugonnet, line 296: @trchudley I've removed the fractions in the ZT kernels for clearer reading, and divided by 2 where necessary to make ZT/Minar formulas consistent.

xdem/terrain.py
```diff
@@ -342,71 +629,337 @@ def _make_attribute_from_coefs(
...
+
+        if curv_method_id in [0, 1]:
+
+            # Geometric planform curvature following Sobolevsky, 1932
+            # Planform derivation is the same in a geometric and directional derivative context
+            # (see Minár et al. 2020 following Jenčo, 1992)
+            # plancurv = - (z_xx * z_y**2 - 2 * z_xy * z_x * z_y + z_yy * z_x**2) / sqrt((z_x**2 + z_y**2)**3)
+
+            plancurv = np.where(
+                C[z_x_idx] ** 2 + C[z_y_idx] ** 2 < 10e-15,
```
rhugonnet, line 758: @trchudley A mysterious one that took me a bit of time to figure out! Planform and Flowline curvatures being very sensitive on flat surfaces, floating point precision started to become problematic.
For instance, for a flat DEM made only of ones `dem=np.ones((5,5), dtype=np.float32)`, where we expect 0, the Planform and Flowline would yield non-zero values (in reasonable orders of magnitude, like 0.5-2), despite all coefficients being near zero (`10e-20`)! 
This was much worse with Florinsky (due to the higher degree of derivatives) than with ZT.

I solved this by adding the `<10e-15` condition here instead of `==0`... This ideal scenario now passes in the tests.
This should hopefully suffice.

tests/test_terrain.py
```diff
@@ -153,6 +158,10 @@ def test_attribute_functions_against_richdem(self, attribute: str, get_test_data
         attr_richdem_rst = gu.Raster(get_test_data_path(os.path.join("richdem", f"{attribute}.tif")), load_data=True)
         attr_richdem = gu.raster.get_array_and_mask(attr_richdem_rst)[0].squeeze()
 
+        # TODO: Profile curvature is opposite sign as of RichDEM, check if this is warranted
+        if attribute == "profile_curvature":
+            attr_richdem = -attr_richdem
```
rhugonnet, line 164: @trchudley After homogenizing the X/Y directions and scaling factors of coefficients, this is the only difference left with RichDEM. All other outputs are consistent.
We should double-check if this is a difference due to inconsistent sign in Minar vs ZT, or an error in RichDEM.

xdem/terrain.py
rhugonnet: Maybe we should be careful of the order of magnitude of the check, as the denominator is squared, so it can go fast...

tests/test_terrain.py
rhugonnet: I just checked Minar/ZT, and both have a negative sign for profile curvature.
From the old code, it looks like I had actually removed that negative sign when implementing ZT several years ago to be consistent with RichDEM. So I think we can consider that this is a RichDEM inconsistency! :smile:

tests/test_terrain.py
trchudley: A simple relationship between the two is that the max* should always be greater than the min* - the maximal/minimal are effectively the eigenvectors of the slope (think first/second principal stress - indeed, Minár refers to the as the _principal normal curvatures_). Is this the sort of thing you were thinking of?

tests/test_terrain.py
trchudley: Hmm, a couple other shapes that are mathematically simple but might give robust and understandable curvatures could be

1. A 5x5 saddle-shaped surface like this:

```
# grid
n = 5
x = np.linspace(-1, 1, n)
y = np.linspace(-1, 1, n)
X, Y = np.meshgrid(x, y)

# hyperbolic paraboloid
Z = X**2 - Y**2
```

This should have opposing max and min curvatures, so might be useful for testing this.

2. A simple slope in the X direction with a Gaussian ridge or trough in the y direction

```
a = 0.6
b = 1.0
sigma = 1
# ridge
Z = a * X + b * np.exp(-(Y/sigma)**2)
# trough
Z = a * X - b * np.exp(-(Y/sigma)**2)
```
These should (?) have opposing tangential and planform curvatures. I think the profile and flowline curvatures should be zero, at least for the trough format?

tests/test_terrain.py
trchudley: Excellent news!

xdem/terrain.py
trchudley: Thanks!

xdem/terrain.py
trchudley: It doesn't surprise me that planform and flowline curvatures are sensitive to flat surfaces - these are the curvatures of a projected contour line or flow line, so they should begin to break down as the slope approaches 'true' flatness. If anything, I'm slightly confused they're returning reasonable magnitudes rather than just breaking completely... Regardless, if you're happy then I'm happy here.

tests/test_terrain.py
trchudley: Fantastic! Glad to hear.

tests/test_terrain.py
rhugonnet: Ah yes I can add this too!
I was thinking more of between maximal/maximum and minimal/minimum, but I'm not sure there is a clear relationship there really?

trchudley: Ah, right! No, I don't think there is - they are analogous but without a strict mathematical relationship, as far as I can interpret.

tests/test_terrain.py
rhugonnet: Indeed, all of this works! Added to the tests.

trchudley:
Thanks @rhugonnet, fantastic! I don't think any of the code here is taken directly from `pdemtools` so don't worry about that - if anything, I will have to go back to `pdemtools` and update it in light of what I've learned here! Maybe instead there might be scope to thank Ian Evans and Josef Minár for some of the inputs they've had - perhaps that might be appropriate for the acknowledgements section of the paper.

P.S. I realise that I never added a [multidirectional hillshade](https://pubs.usgs.gov/of/1992/of92-422/of92-422.pdf) option, but that's so simple we can leave that for another day!

examples/advanced/plot_slope_methods.py
```diff
@@ -2,12 +2,27 @@
 Slope and aspect methods
 ========================
 
-Terrain slope and aspect can be estimated using different methods.
+Within `xdem`, slope is always calculated as
+
+$$
+\alpha = arctan \\left( \\sqrt{ \\left( \frac{\\partial z}{\\partial x} \right)^2 + \\left( \frac{\\partial z}{\\partial y} \right)^2 } \right),
```
belletva: problem with the markdown in the doc, the equation has a problem

examples/advanced/plot_slope_methods.py
```diff
@@ -2,12 +2,27 @@
...
+$$
+
+Where $z$ is elevation, and $x$ and $y$ are the Cartesian co-ordinates.
+
+Aspect is always calculated as
+
+$$
+\theta = \\left( -\tan^{-1}\\left( \frac{ \\partial z / \\partial x }{ \\partial z / \\partial y } \right) - \\pi \right) \bmod (2\\pi).
```
belletva: same here

belletva: image

examples/advanced/plot_slope_methods.py
```diff
@@ -2,12 +2,27 @@
...
+$$
+
+However, calculating terrain attributes - not only slope and aspect but also curvatures - requires estimating the elevation derivatives of the surface (e.g. $\frac{\\partial z}{\\partial x}$). `xdem` offers three different ways to calculate elevation derivatives, which can result in slightly different results.
```
belletva: problem equation

examples/advanced/plot_slope_methods.py
```diff
@@ -43,31 +58,39 @@ def plot_attribute(attribute, cmap, label=None, vlim=None):
 # Slope with method of Horn (1981) (GDAL default), based on a refined
```
belletva: this text is not commented in the code, it is outside of the code

examples/advanced/plot_slope_methods.py
```diff
@@ -43,31 +58,39 @@ def plot_attribute(attribute, cmap, label=None, vlim=None):
...
-# We compute the difference between the slopes computed with each method.
+# Slope with method of Florinsky (2009).
+
+slope_florinsky = xdem.terrain.slope(dem, surface_fit="Florinsky")
+
+plot_attribute(slope_florinsky, "Reds", "Slope of Florinsky (2009) (°)")
+
+
+# %%
+# We can compute the difference between the slopes computed with each method.
```
belletva: there is only the difference between horn and zevenberg, the difference between horn and florinsky is missing as well as the difference between zevenberg and florinsky

doc/source/terrain.md
```diff
@@ -170,11 +193,29 @@ and take values of either the center ($0$), west or south ($-$), or east or nort
...
+It is most often described in degrees, where a flat surface is 0° and a vertical cliff is 90°.
+No tilt direction is stored in the slope map; a 45° tilt westward is identical to a 45° tilt eastward.
+
+The slope $\alpha$ is defined as
+
+$$
+\arctan \left( \sqrt{ z_x^2 + z_y^2 } \right),
+$$
+
+and the surface derivatives can be computed either by the method of [Horn (1981)](http://dx.doi.org/10.1109/PROC.1981.11918), [Zevenbergen and Thorne (1987)](http://dx.doi.org/10.1002/esp.3290120107), or [Florinsky (2009)](https://doi.org/10.1080/13658810802527499) (default).
```
belletva: I prefer adding a sentence such as: By default, `"Florinsky"` is used . 
It is clearer

doc/source/terrain.md
```diff
@@ -189,15 +230,13 @@ The aspect describes the orientation of strongest slope.
...
-[Zevenbergen and Thorne (1987)](http://dx.doi.org/10.1002/esp.3290120107):
+The aspect $\theta$ is defined as
 
 $$
-\theta = arctan(\frac{p}{q}),
+\theta = -\arctan\left( \frac{-z_x}{z_y} \right) \bmod (2\pi).
 $$
 
-with $p$ and $q$ defined in the slope section.
+Like with slope, the surface derivatives can be calculated following the methods of of [Horn (1981)](http://dx.doi.org/10.1109/PROC.1981.11918), [Zevenbergen and Thorne (1987)](http://dx.doi.org/10.1002/esp.3290120107), or [Florinsky (2009)](https://doi.org/10.1080/13658810802527499) (default).
```
belletva: I prefer adding a sentence such as: By default, "Florinsky" is used .

doc/source/terrain.md
```diff
@@ -218,11 +257,10 @@ The hillshade is a slope map, shaded by the aspect of the slope.
 With a westerly azimuth (a simulated sun coming from the west), all eastern slopes are slightly darker.
 This mode of shading the slopes often generates a map that is much more easily interpreted than the slope.
 
-The hillshade $hs$ is directly based on the slope $\alpha$ and aspect $\theta$, and thus also varies between the method of [Horn (1981)](http://dx.doi.org/10.1109/PROC.1981.11918) (default) and that of [Zevenbergen and Thorne (1987)](http://dx.doi.org/10.1002/esp.3290120107), and
-is often scaled between 1 and 255:
+The hillshade $hs$ is directly based on the slope $\alpha$ and aspect $\theta$, and thus also varies between the methods of [Horn (1981)](http://dx.doi.org/10.1109/PROC.1981.11918), [Zevenbergen and Thorne (1987)](http://dx.doi.org/10.1002/esp.3290120107), and [Florinsky (2009)](https://doi.org/10.1080/13658810802527499) (default). It is often scaled between 1 and 255:
```
belletva: I prefer adding a sentence such as: By default, "Florinsky" is used .

doc/source/terrain.md
```diff
@@ -235,80 +273,187 @@ hillshade = dem.hillshade()
...
+
+There are two parallel systems of defining curvatures: either _geometric_ (curvatures can be defined by the radius of a circle), or _directional derivative_ (curvatures can be understood as directional derivatives of the elevation field). For more information on this, [Minár et al. (2020)](https://doi.org/10.1016/j.earscirev.2020.103414) provides a comprehensive review. The choice of system be be set in xDEM via the `curv_method` parameter. This defaults to the `"geometric"` method, which should be suitable for most users, although `"directional"` is also available for those interested.
 
-{func}`xdem.DEM.curvature`
+All curvatures require $z_{xx}$, $z_{xy}$, and/or $z_{yy}$ partial derivatives to calculate: as a result, only [Zevenbergen and Thorne (1987)](http://dx.doi.org/10.1002/esp.3290120107), and [Florinsky (2009)](https://doi.org/10.1080/13658810802527499) (default) surfacr fit methods can be used.
```
belletva: I prefer adding a sentence such as: By default, "Florinsky" is used .

doc/source/terrain.md
belletva: Add also the code: surface_fit="Florinsky" or method="Florinky"

examples/advanced/plot_slope_methods.py
trchudley: Thanks for catching. I think this is a working version -- commenting here for now and will return and fix when I'm back in work next week.

$$
\alpha = \arctan\!\left(
  \sqrt{
    \left(\frac{\partial z}{\partial x}\right)^{2}
    +
    \left(\frac{\partial z}{\partial y}\right)^{2}
  }
\right),
$$

examples/advanced/plot_slope_methods.py
trchudley: As above

$$
\theta =
\left( -\tan^{-1} \left(
    \frac{\partial z/\partial x}{\partial z/\partial y}
  \right) - \pi
\right) \bmod (2\pi).
$$

examples/advanced/plot_slope_methods.py
trchudley: And again, to fix next week
$\frac{\partial z}{\partial x}$

examples/advanced/plot_slope_methods.py
trchudley: Not quite understanding this, sorry - can you explain your expected output?

examples/basic/plot_terrain_attributes.py
```diff
@@ -8,7 +8,7 @@
 For more information, see the :ref:`terrain-attributes` feature page.
 
 **References:** `Horn (1981) <https://ieeexplore.ieee.org/document/1456186>`_ (slope, aspect, hillshade),
```
belletva, line 10: I think it could be a good idea to add a sentence such as : "As a reminder the slope, aspect, hillshade are computed using the Horn method, ..."
In addition to references: ""

examples/advanced/plot_slope_methods.py
trchudley: Made a deliberate choice here not to complicate things by attempting to compute all differences (https://github.com/GlacioHack/xdem/pull/777#issuecomment-3521504423), and kept the original horn/zevenberg difference as an example rather than a full explanation. Could always go further if it is considered necessary.

doc/source/terrain.md
trchudley: Thanks, will edit when I return to this next week!

examples/advanced/plot_slope_methods.py
belletva: Sorry it was not very clear : this sentence is a bit long "Slope with method of Horn (1981) (GDAL default), based on a refined approximation of the gradient (page 18, bottom left, and pages 20-21)."
I propose to replace with:
```
 # Slope with method of Horn (1981)
 # .. note::  (GDAL default), based on a refined approximation of the gradient (page 18, bottom left, and pages 20-21).
  ````

File: examples/advanced/plot_slope_methods.py (hunk 17 above)
Comment (belletva) [reply examples/advanced/plot_slope_methods.py]: OK perfect! In this case, maybe changing the sentence: "We can compute the difference between the slopes computed with Horn and Zevenberg methods"

File: xdem/dem.py [hunk 23]
```diff
@@ -385,60 +385,143 @@ def to_vcrs(
     @copy_doc(terrain, remove_dem_res_params=True)
     def slope(
         self,
-        method: Literal["Horn", "ZevenbergThorne"] = "Horn",
+        method: Literal["Horn", "ZevenbergThorne"] = None,
+        surface_fit: Literal["Horn", "ZevenbergThorne", "Florinsky"] = "Florinsky",
```
belletva, line 389: Ok I understand now ! I didn't understand that method is depreciated now! Maybe add a note in the doc?

doc/source/terrain.md
rhugonnet: Also typo "surfacr fit"!

examples/advanced/plot_slope_methods.py
rhugonnet: Ah it's in the gallery example! I thought I had broken those when I edited them in the main doc page "Terrain" :sweat_smile:

@trchudley FYI: On the main doc page, there were 2 equations for aspects (I think you hadn't fully removed the old one), so I removed one + I simplified all equations to use `z_x` and `z_y` defined at the top directly (instead of rewriting the partial derivatives). I also replaced `tan-1` by `arctan` for clarity (`tan-1` can be misinterpreted as 1/tan).

For the gallery example: We could simply point to the main doc page to keep it short, and not re-write the equations altogether :wink:

xdem/dem.py
trchudley: Great - all relevant docstrings now read:
```
    :param method: Deprecated; use `surface_fit` instead.
    :param surface_fit: Surface fit method to use for slope: "Horn", "ZevenbergThorne" or "Florinsky".
```

rhugonnet:
@trchudley A big chunk of work finalized, thanks a ton! :partying_face: :tada: 

This change will be included in the next release (xDEM 0.2; along with point cloud support), so soon accessible without being on the dev branch. 
It will be followed by a 0.3 (Xarray accessor) in the next months.

FYI: The equations for the slope example still weren't rendering (as sphinx-gallery examples are in RST). I tried to adjust them, but it was capricious, and we would have had to add a new Sphinx math extension. So I simply removed them and point towards the main page, I think it's easier :wink:

PR #781, opened 2025-09-11
Robustify initial shift for N&K single method.

### LINKED ISSUE #774: Robustify initial_shift
# Context
The initial shift option is not robust.
The goal of this ticket is to implement the necessary changes based on the context we have created.
(These issues may be due to the change from dem_coregistration to coregister3d.)

[pb_shift_initial.zip](https://github.com/user-attachments/files/21913039/pb_shift_initial.zip)

## Explanation of the context

- ref = DEM_2009_ref.tif, which I shifted by 2 pixels in X and 3.5 in Y (with 20 m pixels) for the test.
    - a coreg without initial shift → ok
    - initial shift set to the expected value (-2, -3.5 pixels) → misaligned output, and the shift value is inconsistent: (-120.00000046220806, -210.00000554783298)
    - initial shift set to the opposite value (2, 3.5 pixels) → I get a misaligned output, and the shift value is inconsistent: (39.96, 69.95)

# Code changes

- if we move the reference in one direction here https://github.com/GlacioHack/xdem/blob/main/xdem/dem.py#L552, we then need to apply the opposite shift here: https://github.com/GlacioHack/xdem/blob/main/xdem/dem.py#L581
- the aligned DEM is aligned on the shifted reference: https://github.com/GlacioHack/xdem/blob/main/xdem/dem.py#L558
So there’s an issue: the fitting should be done on the shifted reference, and the application should be done on the reference + the shift.
- to be consistent with the retrieved shifts, we should use a “-” here https://github.com/GlacioHack/xdem/blob/main/xdem/dem.py#L552
. (In my case, if I get a shift of (-40, -70), and I already know in advance that I will have this offset, I want to input (-2, -3.5), not the opposite.)

# Unit tests

We need to make the existing unit tests more robust here: https://github.com/GlacioHack/xdem/blob/main/tests/test_dem.py#L403, using the context described in this ticket.

# Documentation

A new paragraph should be added to the documentation about the initial shift, ideally with a diagram: https://github.com/GlacioHack/xdem/blob/main/doc/source/coregistration.md?plain=1#L58

marinebcht:
Resolves #774 

This PR fixes errors introduced by the initial shit value in coregistration_3D function for N&K single method. 

## Contexte 
Issue #774 highligths the actual outputs: the `DEM.coregister_3d()` gives wrong values of the output shifts, when the initial shift given is != (0, 0), as well as a wrong output coregistred dem.  

With an actual  dem to be aligned shifted to (-2.0, -3.5) pixels :  

```
Input shift = (0.0, 0.0) pixels => Output shift = (-2.0, -3.5) pixels  
Input shift = (-2.0, -3.5) pixels s => Output shift = (-6.0, -10.5) pixels 
Input shift = (1.999, 3.499) pixels => Output shift = (2.0, 3.5) pixels  
Input shift = (-1.0, -1.0) pixels => Output shift = (-4.0, -5.5) pixels 
```

 ## Resolution

Indeed, the shift is applied twice in the same direction before the fit_and_apply [here](https://github.com/GlacioHack/xdem/blob/b7b0fdd6e6fc8dbcdc67202a83ae08943faa331d/xdem/dem.py#L569) and after [here](https://github.com/GlacioHack/xdem/blob/b7b0fdd6e6fc8dbcdc67202a83ae08943faa331d/xdem/dem.py#L592).

As the initial_shift is distance to add to the shifted dem to match the reference,  we need to 
it from reference dem to match shifted dem. 
Which gives: `reference_elev  = reference_elev.translate(-shift_x, -shift_y)` and not ` reference_elev  = reference_elev.translate(shift_x, shift_y)`

After this, we have now : 
```
Input shift = (0.0, 0.0) pixels => Output shift = (-2.0, -3.5) pixels  
Input shift = (-2.0, -3.5) pixels s => Output shift =  (-2.0, -3.5) pixels 
Input shift = (1.999, 3.499) pixels => Output shift = (-2.0, -3.5) pixels  
Input shift = (-1.0, -1.0) pixels => Output shift =  (-2.0, -3.5) pixels 
```

## Optimisation

As the initial shift is only a peculiarity of affine coregistration methods, the part of the code dealing with it have been move to `NuthKaab` initialisation, inherit from `AffineCoreg`.

So now, we need to do:

```
xdem.coreg.NuthKaab(initial_shift=(10, 10))
dem_aligned = dem_tba.coregister_3d(dem_ref, coreg_method=coreg_method)
```

This "new" information were added to the metadata in inputs/affine/initial_shift which gives for exemple :  

```
c = xdem.coreg.NuthKaab(initial_shift=(10, 10))
print (c.meta)
```
{'inputs': {'random': {'subsample': 500000.0}, 'fitorbin': {'fit_or_bin': 'bin_and_fit', 'fit_func': <function _nuth_kaab_fit_func at 0x7f979bc3b6a0>, 'fit_optimizer': <function curve_fit at 0x7f97a8f39300>, 'bin_sizes': 72, 'bin_statistic': <function nanmedian at 0x7f97c78ab130>}, 'iterative': {'max_iterations': 10, 'tolerance': 0.001}, 'specific': {}, 'affine': {'apply_vshift': True, *'initial_shift': (10, 10)*}}, 'outputs': {}}

```
print (c.info())
```
Generic coregistration information 
  Method:       NuthKaab 
  ....
  Affine
    Vertical shift activated:                     True
    *Estimated initial shift:                      (10, 10)*
  ....

### Initial_shift values

For NuthKaab : 
- Initial_shift = None if not instantiated 
- Initial_shift = (x, y) if instantiated by (x, y)
- Error if different : "Argument `initial_shift` must be a list or tuple of exactly two numerical values."

For others methods : 
- Initial_shift = None

## Tests

The unit tests focus on the initial shifts have been updated to verify : 
- consistency of the initial shift and error raised in case
- presence of the inital shift in metadata if given for `AffineCoreg` and `AffineCoreg` in `CoregPipeline`
- output shift after  `DEM.coregister_3d()` for several input shift : None, (0,0), (50, 10), (10, 50) and (10, 50, 20)
- output dem after  `DEM.coregister_3d()` 
- if fit_and_apply with/without initial shift parameters for NuhKaab method do the same
- no intial shifts in case of an CoregPipeline

## Open subjects

- Application with others methods through `matrix` + Add an `initial_transform` parameter as well ?  => https://github.com/GlacioHack/xdem/issues/784 
- Case of a CoregPipeline  like `coreg_method = xdem.coreg.NuthKaab(initial_shift=(10, 5)) + xdem.coreg.VerticalShift()`

`dem_ref` in `dem_tba.coregister_3d(dem_ref, coreg_method=coreg_method)` is changed from DEM type to NDArray type in [_preprocess_coreg_fit_raster_raster](https://github.com/marinebcht/xdem/blob/c88b9617aad811f2ac07b2f41b9d159403edcb6f/xdem/coreg/base.py#L2787) then the [reference_elev = reference_elev.translate(-shift_x, -shift_y)](https://github.com/marinebcht/xdem/blob/c88b9617aad811f2ac07b2f41b9d159403edcb6f/xdem/coreg/base.py#L2103) can't work... 

  => issue (?)

### DIFF
--- a/tests/test_dem.py
+++ b/tests/test_dem.py
@@ -165,7 +165,7 @@ class TestDEM:
         dem = DEM.from_array(data=np.ones((5, 5)), transform=transform, crs=CRS("EPSG:4326"), nodata=None)
 
         mask_dem = dem > 1
-        assert isinstance(mask_dem, gu.Mask)
+        assert isinstance(mask_dem, gu.Raster) and np.dtype(mask_dem.dtype) == np.bool_
 
     def test_copy(self) -> None:
         """
--- a/xdem/coreg/base.py
+++ b/xdem/coreg/base.py
@@ -49,7 +49,7 @@ import scipy.ndimage
 import scipy.optimize
 from geoutils.interface.gridding import _grid_pointcloud
 from geoutils.interface.interpolate import _interp_points
-from geoutils.raster import Mask, RasterType, raster
+from geoutils.raster import Raster, RasterType, raster
 from geoutils.raster._geotransformations import _resampling_method_from_str
 from geoutils.raster.array import get_array_and_mask
 from geoutils.raster.georeferencing import _cast_pixel_interpretation, _coords
@@ -129,7 +129,7 @@ dict_key_to_str = {
 def _preprocess_coreg_fit_raster_raster(
     reference_dem: NDArrayf | MArrayf | RasterType,
     dem_to_be_aligned: NDArrayf | MArrayf | RasterType,
-    inlier_mask: NDArrayb | Mask | None = None,
+    inlier_mask: NDArrayb | Raster | None = None,
     transform: rio.transform.Affine | None = None,
     crs: rio.crs.CRS | None = None,
     area_or_point: Literal["Area", "Point"] | None = None,
@@ -206,7 +206,7 @@ def _preprocess_coreg_fit_raster_raster(
 
     # Make sure that the mask has an expected format.
     if inlier_mask is not None:
-        if isinstance(inlier_mask, Mask):
+        if isinstance(inlier_mask, Raster):
             inlier_mask = inlier_mask.data.filled(False).squeeze()
         else:
             inlier_mask = np.asarray(inlier_mask).squeeze()
@@ -234,7 +234,7 @@ def _preprocess_coreg_fit_raster_raster(
 def _preprocess_coreg_fit_raster_point(
     raster_elev: NDArrayf | MArrayf | RasterType,
     point_elev: gpd.GeoDataFrame,
-    inlier_mask: NDArrayb | Mask | None = None,
+    inlier_mask: NDArrayb | Raster | None = None,
     transform: rio.transform.Affine | None = None,
     crs: rio.crs.CRS | None = None,
     area_or_point: Literal["Area", "Point"] | None = None,
@@ -262,7 +262,7 @@ def _preprocess_coreg_fit_raster_point(
 
     # Make sure that the mask has an expected format.
     if inlier_mask is not None:
-        if isinstance(inlier_mask, Mask):
+        if isinstance(inlier_mask, Raster):
             inlier_mask = inlier_mask.data.filled(False).squeeze()
         else:
             inlier_mask = np.asarray(inlier_mask).squeeze()
@@ -294,7 +294,7 @@ def _preprocess_coreg_fit_point_point(
 def _preprocess_coreg_fit(
     reference_elev: NDArrayf | MArrayf | RasterType | gpd.GeoDataFrame,
     to_be_aligned_elev: NDArrayf | MArrayf | RasterType | gpd.GeoDataFrame,
-    inlier_mask: NDArrayb | Mask | None = None,
+    inlier_mask: NDArrayb | Raster | None = None,
     transform: rio.transform.Affine | None = None,
     crs: rio.crs.CRS | None = None,
     area_or_point: Literal["Area", "Point"] | None = None,
@@ -520,7 +520,7 @@ def _get_subsample_on_valid_mask(params_random: InRandomDict, valid_mask: NDArra
     """
     Get mask of values to subsample on valid mask (works for both 1D or 2D arrays).
 
-    :param valid_mask: Mask of valid values (inlier and not nodata).
+    :param valid_mask: Raster of valid values (inlier and not nodata).
     """
 
     # This should never happen
@@ -1977,7 +1977,7 @@ class Coreg:
         """
         Get mask of values to subsample on valid mask.
 
-        :param valid_mask: Mask of valid values (inlier and not nodata).
+        :param valid_mask: Raster of valid values (inlier and not nodata).
         """
 
         # Get random parameters
@@ -2049,7 +2049,7 @@ class Coreg:
         self: CoregType,
         reference_elev: NDArrayf | MArrayf | RasterType | gpd.GeoDataFrame,
         to_be_aligned_elev: NDArrayf | MArrayf | RasterType | gpd.GeoDataFrame,
-        inlier_mask: NDArrayb | Mask | None = None,
+        inlier_mask: NDArrayb | Raster | None = None,
         bias_vars: dict[str, NDArrayf | MArrayf | RasterType] | None = None,
         weights: NDArrayf | None = None,
         subsample: float | int | None = None,
@@ -2065,7 +2065,7 @@ class Coreg:
 
         :param reference_elev: Reference elevation, either a DEM or an elevation point cloud.
         :param to_be_aligned_elev: To-be-aligned elevation, either a DEM or an elevation point cloud.
-        :param inlier_mask: Mask or boolean array of areas to include (inliers=True).
+        :param inlier_mask: Raster or boolean array of areas to include (inliers=True).
         :param bias_vars: Auxiliary variables for certain bias correction classes, as raster or arrays.
         :param weights: Array of weights for the coregistration.
         :param subsample: Subsample the input to increase performance. <1 is parsed as a fraction. >1 is a pixel count.
@@ -2272,7 +2272,7 @@ class Coreg:
         self,
         reference_elev: NDArrayf | MArrayf | RasterType | gpd.GeoDataFrame,
         to_be_aligned_elev: MArrayf,
-        inlier_mask: NDArrayb | Mask | None = None,
+        inlier_mask: NDArrayb | Raster | None = None,
         bias_vars: dict[str, NDArrayf | MArrayf | RasterType] | None = None,
         weights: NDArrayf | None = None,
         subsample: float | int | None = None,
@@ -2292,7 +2292,7 @@ class Coreg:
         self,
         reference_elev: NDArrayf | MArrayf | RasterType | gpd.GeoDataFrame,
         to_be_aligned_elev: NDArrayf,
-        inlier_mask: NDArrayb | Mask | None = None,
+        inlier_mask: NDArrayb | Raster | None = None,
         bias_vars: dict[str, NDArrayf | MArrayf | RasterType] | None = None,
         weights: NDArrayf | None = None,
         subsample: float | int | None = None,
@@ -2312,7 +2312,7 @@ class Coreg:
         self,
         reference_elev: NDArrayf | MArrayf | RasterType | gpd.GeoDataFrame,
         to_be_aligned_elev: RasterType | gpd.GeoDataFrame,
-        inlier_mask: NDArrayb | Mask | None = None,
+        inlier_mask: NDArrayb | Raster | None = None,
         bias_vars: dict[str, NDArrayf | MArrayf | RasterType] | None = None,
         weights: NDArrayf | None = None,
         subsample: float | int | None = None,
@@ -2331,7 +2331,7 @@ class Coreg:
         self,
         reference_elev: NDArrayf | MArrayf | RasterType | gpd.GeoDataFrame,
         to_be_aligned_elev: NDArrayf | MArrayf | RasterType | gpd.GeoDataFrame,
-        inlier_mask: NDArrayb | Mask | None = None,
+        inlier_mask: NDArrayb | Raster | None = None,
         bias_vars: dict[str, NDArrayf | MArrayf | RasterType] | None = None,
         weights: NDArrayf | None = None,
         subsample: float | int | None = None,
@@ -2349,7 +2349,7 @@ class Coreg:
 
         :param reference_elev: Reference elevation, either a DEM or an elevation point cloud.
         :param to_be_aligned_elev: To-be-aligned elevation, either a DEM or an elevation point cloud.
-        :param inlier_mask: Mask or boolean array of areas to include (inliers=True).
+        :param inlier_mask: Raster or boolean array of areas to include (inliers=True).
         :param bias_vars: Auxiliary variables for certain bias correction classes, as raster or arrays.
         :param weights: Array of weights for the coregistration.
         :param subsample: Subsample the input to increase performance. <1 is parsed as a fraction. >1 is a pixel count.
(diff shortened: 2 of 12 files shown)


xdem/coreg/base.py
```diff
@@ -1677,6 +1678,8 @@ class OutSpecificDict(TypedDict, total=False):
 class InAffineDict(TypedDict, total=False):
     """Keys and types of inputs associated with affine methods."""
 
+    # Estimated initial shift
+    initial_shift: tuple[float, float, float]
```
adebardo: Isn't it tuple[float, float] ?

xdem/dem.py
```diff
@@ -1,4 +1,3 @@
-# Copyright (c) 2025 xDEM developers
```
adebardo, line 1: We cannot delete le xDEM developers copyrights

xdem/coreg/base.py
marinebouchet: oui, j'avais corrigé :)

xdem/dem.py
marinebouchet: oh !

doc/source/coregistration.md
```diff
@@ -551,6 +551,7 @@ For both **inputs** and **outputs**, four consistent categories of metadata are
 
 - An input `only_translation` to define if a coregistration should solve only for translations instead of a full rigid transformation (translations and rotations),
 - An input `standardize` to define if the input data should be standardized to the unit sphere before coregistration (to improve numerical convergence),
+- An input `initial_shift` that define the estimated initial x and y shifts in georeferenced units, applied before fit step.
```
belletva: that defines

tests/test_coreg/test_affine.py
```diff
@@ -596,3 +596,28 @@ def test_nuthkaab_no_vertical_shift(self) -> None:
         # Assert horizontal shifts are the same
         matrix2[2, 3] = matrix1[2, 3]
         assert np.array_equal(matrix1, matrix2)
+
+    def test_nuthkaab_initial_shift(self) -> None:
+        """
+        Test that the initial_shift don't impact fit_and_apply process.
```
belletva: Maybe we need to precise that the fit_and_apply is for the nuth and kaab coregistration ?

tests/test_dem.py
```diff
@@ -411,59 +407,78 @@ def test_coregister_3d(coreg_method, initial_shift, expected_pipeline_types) ->
...
-                xdem.coreg.AffineCoreg() + xdem.coreg.VerticalShift(),
-                ValueError,
-                r".*two numerical values.*",
-                [2, 3, 5],
-            ),
+            pytest.param(None, None, False, id="NuthKaab method: No initial shift"),
+            pytest.param([0, 0], None, False, id="NuthKaab method: [0, 0] initial shift"),
+            pytest.param([50, 10], None, False, id="NuthKaab method: [5, 10] initial shift"),
+            pytest.param([10, 50], None, False, id="NuthKaab + VerticalShift methods: [10, 5] initial shift"),
+            pytest.param(["2", 2], r".*two numerical values.*", False, id='NuthKaab method: ["2", 2] initial shift'),
```
belletva, line 428: double quote

tests/test_coreg/test_affine.py
```diff
@@ -596,3 +596,28 @@ def test_nuthkaab_no_vertical_shift(self) -> None:
...
+        """
+
+        # Use entire DEMs here (to compare to original values from older package versions)
+        ref, tba = load_examples(crop=False)[0:2]
+
+        # Get the coregistration method and expected shifts from the inputs
+        inlier_mask = ~self.outlines.create_mask(ref)
+
+        c = coreg.NuthKaab(initial_shift=(0, 0), subsample=50000)
```
belletva: parenthesis for initial_shift?

tests/test_dem.py
```diff
@@ -411,59 +407,78 @@ def test_coregister_3d(coreg_method, initial_shift, expected_pipeline_types) ->
...
-                xdem.coreg.AffineCoreg() + xdem.coreg.VerticalShift(), ValueError, r".*two numerical values.*", ["2", 2]
-            ),
-            pytest.param(
-                xdem.coreg.AffineCoreg() + xdem.coreg.VerticalShift(),
-                ValueError,
-                r".*two numerical values.*",
-                [2, 3, 5],
-            ),
+            pytest.param(None, None, False, id="NuthKaab method: No initial shift"),
+            pytest.param([0, 0], None, False, id="NuthKaab method: [0, 0] initial shift"),
```
belletva: square brackets for initial_shift?

tests/test_dem.py
```diff
@@ -371,36 +371,32 @@ def test_terrain_attributes_wrappers(self, terrain_attribute: str) -> None:
...
     @pytest.mark.parametrize(  # type: ignore
-        "coreg_method, initial_shift, expected_pipeline_types",
+        "coreg_method, expected_pipeline_types",
         [
-            pytest.param(xdem.coreg.Deramp(), None, [xdem.coreg.Deramp], id="Custom method: Deramp"),
+            pytest.param(xdem.coreg.Deramp(), [xdem.coreg.Deramp], id="Custom method: Deramp"),
             pytest.param(
-                xdem.coreg.NuthKaab() + xdem.coreg.VerticalShift(),
-                [10, 5],
+                xdem.coreg.NuthKaab(initial_shift=[10, 5]) + xdem.coreg.VerticalShift(),
```
belletva: square brackets for initial_shift?

xdem/coreg/base.py
```diff
@@ -2092,6 +2095,18 @@ def fit(
         if self._meta["inputs"]["random"]["subsample"] != 1:
             self._meta["inputs"]["random"]["random_state"] = random_state
 
+        # Apply the shift to the source dem if given
+        initial_shift_apply = False
+        if (
+            isinstance(reference_elev, gu.Raster)
+            and "initial_shift" in self._meta["inputs"]["affine"]
+            and self._meta["inputs"]["affine"]["initial_shift"]
```
rhugonnet: The `translate` function also works if the input is not a raster, but a point cloud (and we need to! otherwise the functionality would be inconsistent depending on input type). 

Maybe we can simplify to only `if "initial_shift" in self._meta["inputs"]["affine"]:`?

xdem/coreg/base.py
```diff
@@ -2131,6 +2146,15 @@ def fit(
             **kwargs,
         )
 
+        # Unapply the shift to the source dem if apply before
+        if initial_shift_apply and "outputs" in self.meta and "affine" in self.meta["outputs"]:
+            if "shift_x" in self.meta["outputs"]["affine"]:
```
rhugonnet, line 2155: These cases don't cover affine methods that don't store `shift_x` and `shift_y`, but only the affine `matrix` (so LZD, ICP, CPD, etc).

One way to have it fully consistent across all affine methods would be to write it this way:
```python
matrix = self.to_matrix()
matrix[0, 3] += shift_x
matrix[1, 3] += shift_y
self.meta["outputs"]["affine"]["matrix"] = matrix
```

But the problem is that "shift_x" and "shift_y" won't be updated accordingly (when they exist)... Maybe we should define a new `Affine.set_matrix` that automatically derives `shift_x`, `shift_y`, etc... in the `.meta` (and we could even add `rot_x`, `rot_y`, `rot_z` this way!).

tests/test_dem.py
```diff
@@ -411,59 +407,80 @@ def test_coregister_3d(coreg_method, initial_shift, expected_pipeline_types) ->
...
-                [2, 3, 5],
+                (2, 3, 5, 4), r".*three numerical values.*", False, id="NuthKaab method: [2, 3, 5] initial shift"
             ),
+            pytest.param((50, 10), None, True, id="NuthKaab method: [5, 10] initial shift + VerticalShift"),
         ],
     )
-    def test_exceptions_initial_shift(
-        self, coreg_method, error, expected_match, test_shift_tuple
-    ) -> None:  # type: ignore
+    def test_nuthkaab_initial_shift(shift, expected_error, pipeline) -> None:  # type: ignore
```
rhugonnet: I think we can test every affine method here, in addition to Nuth and Kääb!

tests/test_coreg/test_affine.py
```diff
@@ -596,3 +596,28 @@ def test_nuthkaab_no_vertical_shift(self) -> None:
         # Assert horizontal shifts are the same
         matrix2[2, 3] = matrix1[2, 3]
         assert np.array_equal(matrix1, matrix2)
+
+    def test_nuthkaab_initial_shift(self) -> None:
```
rhugonnet, line 600: Same comment here, we can test this works well for every affine method, so just Nuth and Kääb!

tests/test_dem.py
belletva: Maybe for this PR we can do it only for N&K and create a new issue for all the other coregistration methods ? :)

tests/test_coreg/test_affine.py
belletva: Maybe for this PR we can do it only for N&K and create a new issue for all the other coregistration methods ? :)

rhugonnet: For the tests, it's only adding `@pytest.mark.parametrize("coreg_method", [all_coreg_methods])` everywhere, and for the code itself within `fit()`, it's what I wrote above.
Probably less than 10 lines of code overall, so I think it's shorter to do it here? 

Happy to provide more help on a new `set_matrix()` as I described above, if you want. Or take over the PR.

marinebcht: Currently, initial_shift param only exists in xdem.coreg.NuthKaab class. Let's finish this PR with the N&K single method use before :) After, I will use set_matrix() that looks perfect !

xdem/coreg/base.py
marinebcht: Subject to deal in https://github.com/GlacioHack/xdem/issues/784

tests/test_dem.py
marinebcht: Subject to deal in https://github.com/GlacioHack/xdem/issues/784

tests/test_coreg/test_affine.py
marinebcht: Subject to deal in https://github.com/GlacioHack/xdem/issues/784

tests/test_coreg/test_base.py
```diff
@@ -60,6 +60,8 @@ def assert_coreg_meta_equal(input1: Any, input2: Any) -> bool:
     # If input is a dictionary, we recursively call this function to check equality of all its sub-keys
     elif isinstance(input1, dict):
         return all(assert_coreg_meta_equal(input1[k], input2[k]) for k in input1.keys())
+    elif input2 == input1:
```
rhugonnet: Why is this needed? Initial shift is a tuple, it's not already tested in the second line?

xdem/coreg/affine.py
```diff
@@ -2391,6 +2395,8 @@ def __init__(
         :param bin_statistic: Statistic of central tendency (e.g., mean) to apply during the binning.
         :param subsample: Subsample the input for speed-up. <1 is parsed as a fraction. >1 is a pixel count.
         :param vertical_shift: Whether to apply the vertical shift or not (default is True).
+        :param initial_shift: Tuple containing x, y and z shifts (in georeferenced units).
+        These shifts are applied before the fit() part.
```
rhugonnet: Need to add a tabulation here, or the line won't appear as part of the previous `:param` :wink:

xdem/coreg/base.py
```diff
@@ -1677,6 +1678,8 @@ class OutSpecificDict(TypedDict, total=False):
 class InAffineDict(TypedDict, total=False):
     """Keys and types of inputs associated with affine methods."""
 
+    # Estimated initial shift
+    initial_shift: tuple[float, float, float] | None
```
rhugonnet, line 1682: Do we need to allow `tuple[float, float]` here as well? (Or did I miss the conversion somewhere?)

xdem/coreg/base.py
```diff
@@ -2092,6 +2100,14 @@ def fit(
         if self._meta["inputs"]["random"]["subsample"] != 1:
             self._meta["inputs"]["random"]["random_state"] = random_state
 
+        # Apply the shift to the source dem if given
+        initial_shift_apply = False
+        if self._meta["inputs"]["affine"].get("initial_shift") is not None:
+            shift_x = self._meta["inputs"]["affine"]["initial_shift"][0]  # type: ignore
+            shift_y = self._meta["inputs"]["affine"]["initial_shift"][1]  # type: ignore
+            reference_elev = reference_elev.translate(-shift_x, -shift_y)  # type: ignore
+            initial_shift_apply = True
```
rhugonnet, line 2112: This applies X/Y but no Z shift? 
Not sure if the idea is to also apply Z shift, or allow only a `tuple[float, float]`. Both work for me!

tests/test_coreg/test_base.py
marinebcht: With the removal of the initial_shift key when equals to None, I can delete this case

Issue #782, opened 2025-09-14
Remove Mypy plugin for NumPy that is deprecated


rhugonnet:
Mirrors https://github.com/GlacioHack/geoutils/issues/728

PR #783, opened 2025-09-14
Fix setup of MyPy temporarily and add `pre-commit` to dev-environment

rhugonnet:
This PR fixes the setup of Mypy which was breaking locally due to the NumPy plugin (and soon in CI), by pinning NumPy 2.0.X.
This PR also adds `pre-commit` to `dev-environment.yml` to have it already setup after a dev install.

We'll have to remove the NumPy plugin in time due to deprecation, see #782.
But this involves updating to NumPy 2.3.X, which triggers many more static typing checks, so let's do this in a PR updating most of the static typing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,7 +72,7 @@ repos:
                         --disable-error-code=no-any-return
 
                   ]
-                  additional_dependencies: [tokenize-rt, numpy==2]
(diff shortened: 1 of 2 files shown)


rhugonnet:
@trchudley This small fix should allow you to install+run `pre-commit` locally from the `dev-environment.yml` file without any issue :wink:

Issue #522, opened 2024-05-06
Unpin `coverage<7` in CI once `coveragepy-lcov` fixed


rhugonnet:
In our build tests, see https://github.com/chaychoong/coveragepy-lcov/issues/5

PR #791, opened 2025-10-01
Unpin `coverage` version in CI for compatibility with Numba and LCOV coverage reporting

### LINKED ISSUE #790: Latest Numba fails during tests because of optional dependency with Coverage
Failure in CI:
```E AttributeError: module 'coverage' has no attribute 'types'```

See issue in Numba here: https://github.com/numba/numba/issues/10239

I think this is because we have `coverage<7` enforced in the tests. I'll try to unpin it.

rhugonnet:
This PR fixes dependencies between Coverage, Numba in relation to the generation of covergae in the LCOV format.

We had pinned `coverage` to 7.0 max because the plugin `coverage-lcov` didn't work anymore: https://github.com/GlacioHack/xdem/issues/522
But Numba now needs more recent `coverage` versions, see #790.

It turns out that `pytest-cov` now supports the LCOV format natively through the addition of the feature in `coverage`: https://github.com/pytest-dev/pytest-cov/pull/536

So it's an easy fix.


rhugonnet:
Curiously, the ddem values have changed ever so slightly (0.001), so I had to update `test_examples`. No obvious culprit...

This was also affecting the stats tests in `test_topo`: We might want to remove those, to having to update values across different modules when a coregistration output slightly changes. Opening an issue.

rhugonnet:
Moving forward: The discrepancy that started appearing for the ddem is only for Python 3.11 and above...

rhugonnet:
Opened #796 due to second CI issue above.
Will solve it in a separate #797.

PR #803, opened 2025-10-13
Blockwise multiprocessing bug and intersection in CLI

adebardo:
# Context
The purpose of this PR is to propose a set of initial fixes to the CLI following Valentine’s work and to fix the blockwise multiprocessing bug.

# Code

## Blockwise bug
Incorrect function call: fixed and unit tests added.
image

## None for stats
It was decided that if the stats field was set to `None` in the config, then the statistics would not be computed.  
It was also proposed to set unnecessary information to `None` during display. However, this double handling caused various conflicts, which have now been resolved.

## Addition of intersection
The goal is to add the computation of the intersection between two DEMs so that they have the same size and projection: new function created — `_prepare_datas_for_coreg`.

### DIFF
--- a/tests/test_coreg/test_blockwise.py
+++ b/tests/test_coreg/test_blockwise.py
@@ -130,7 +130,46 @@ class TestBlockwiseCoreg:
 
         assert actual == expected
 
-    @pytest.mark.parametrize("block_size", [500, 985, 1332], ids=["2d_shifts", "1d_shifts_x", "monotile"])
+    def test_apply_z_correction_parameter(self, step, blockwise_coreg):
+        """
+        Test function to verify the application a vertical shift correction.
+        """
+
+        data = np.ones((500, 500), dtype="uint8")
+        transform = (30.0, 0.0, 478000.0, 0.0, -30.0, 3108140.0)
+        raster = Raster.from_array(data, transform, 32645)
+
+        coeff_x = (0.0, 0.0, 0.0)
+        coeff_y = (0.0, 0.0, 0.0)
+        coeff_z = (0.0, 0.0, 2.0)
+
+        dem_no_correction = blockwise_coreg._wrapper_apply_epc(
+            tba_dem_tile=raster,
+            coeff_x=coeff_x,
+            coeff_y=coeff_y,
+            coeff_z=coeff_z,
+            apply_z_correction=False,
+        )
+
+        assert np.allclose(dem_no_correction.data, raster.data, equal_nan=True)
+
+        dem_with_correction = blockwise_coreg._wrapper_apply_epc(
+            tba_dem_tile=raster,
+            coeff_x=coeff_x,
+            coeff_y=coeff_y,
+            coeff_z=coeff_z,
+            apply_z_correction=True,
+        )
+
+        expected = raster.data + 2.0
+
+        assert np.allclose(
+            dem_with_correction.data,
+            expected,
+            equal_nan=True,
+        )
+
+    @pytest.mark.parametrize("block_size", [500, 985, 1332])
     def test_blockwise_coreg_pipeline(self, step, example_data, tmp_path, block_size):
         """Test end-to-end blockwise coregistration and validate output."""
         ref, tba, mask = example_data
@@ -143,13 +182,14 @@ class TestBlockwiseCoreg:
         aligned = xdem.DEM(tmp_path / "aligned_dem.tif")
 
         # Ground truth comparison with full image coregistration
-        nuth_kaab = xdem.coreg.NuthKaab()
-        expected = nuth_kaab.fit_and_apply(ref, tba, mask)
+        expected = step.fit_and_apply(ref, tba, mask)
 
-        valid = (expected.data.data != expected.nodata) & (aligned.data.data != aligned.nodata)
-        assert np.allclose(expected.data.data[valid], aligned.data.data[valid], atol=20)
+        diff = np.abs(expected - aligned)
+        ind_valid = np.isfinite(diff.data.data)
+        # 90% of the aligned data differs by less than 2m
+        assert np.nanpercentile(diff.data.data[ind_valid], 90) < 2
 
-    @pytest.mark.parametrize("block_size", [500, 985, 1332], ids=["2d_shifts", "1d_shifts_x", "monotile"])
+    @pytest.mark.parametrize("block_size", [500])
     def test_blockwise_coreg_pipeline_with_multiprocessing(self, step, example_data, tmp_path, block_size):
         """Test end-to-end blockwise coregistration in multiprocessing and validate output."""
         ref, tba, mask = example_data
@@ -164,12 +204,12 @@ class TestBlockwiseCoreg:
         aligned = xdem.DEM(tmp_path / "aligned_dem.tif")
 
         # Ground truth comparison with full image coregistration
-        nuth_kaab = xdem.coreg.NuthKaab()
-        expected = nuth_kaab.fit_and_apply(ref, tba, mask)
+        expected = step.fit_and_apply(ref, tba, mask)
 
-        valid = (expected.data.data != expected.nodata) & (aligned.data.data != aligned.nodata)
-        # TODO: careful with 20 value
-        assert np.allclose(expected.data.data[valid], aligned.data.data[valid], atol=20)
+        diff = np.abs(expected - aligned)
+        ind_valid = np.isfinite(diff.data.data)
+        # 90% of the aligned data differs by less than 2m
+        assert np.nanpercentile(diff.data.data[ind_valid], 90) < 2
 
     def test_ransac_on_horizontal_tiles(self, blockwise_coreg) -> None:
         """Test case where RANSAC works on horizontal tiles."""
--- a/tests/test_workflows/test_accuracy.py
+++ b/tests/test_workflows/test_accuracy.py
@@ -275,9 +275,3 @@ def test_mask_init(tmp_path, get_accuracy_inputs_config):
     inlier_mask = ~mask.create_mask(dem)
     assert workflows.inlier_mask == inlier_mask
     assert Path(tmp_path / "plots").joinpath("masked_elev_map.png").exists()
-
-
-def test_prepare_datas_for_coreg():
-    """
-    Test prepare_datas_for_coreg function
-    """


xdem/workflows/accuracy.py
```diff
@@ -140,33 +138,40 @@ def _compute_coregistration(self) -> RasterType:
...
+            if sampling_source == "reference_elev":
+                self.to_be_aligned_elev = self.to_be_aligned_elev.reproject(self.reference_elev, silent=True)
+            elif sampling_source == "to_be_aligned_elev":
+                self.reference_elev = self.reference_elev.reproject(self.to_be_aligned_elev, silent=True)
+        else:
+            self.to_be_aligned_elev = self.to_be_aligned_elev.reproject(crs=crs_utm, res=(1, 1))
+            self.reference_elev = self.reference_elev.reproject(crs=crs_utm, res=(1, 1))
+
+        # Intersection
+        logging.info("Computing reprojection")
```
belletva: modify the log by "Computing intersection"

xdem/workflows/accuracy.py
```diff
@@ -140,33 +138,40 @@ def _compute_coregistration(self) -> RasterType:
...
-            src, target = self.to_be_aligned_elev, self.reference_elev
-            name = "to_be_aligned_elev"
-            reprojected = src.reproject(target, silent=True)
-            self.to_be_aligned_elev = reprojected
-        elif sampling == "to_be_aligned_elev":
-            src, target = self.reference_elev, self.to_be_aligned_elev
-            name = "reference_elevation"
-            reprojected = src.reproject(target, silent=True)
-            self.reference_elev = reprojected
+        if crs_utm is None:
```
belletva: Add a log "Computing reprojection"

xdem/workflows/accuracy.py
```diff
@@ -140,33 +138,40 @@ def _compute_coregistration(self) -> RasterType:
...
+        else:
+            self.to_be_aligned_elev = self.to_be_aligned_elev.reproject(crs=crs_utm, res=(1, 1))
+            self.reference_elev = self.reference_elev.reproject(crs=crs_utm, res=(1, 1))
+
+        # Intersection
+        logging.info("Computing reprojection")
+        coord_intersection = self.reference_elev.intersection(self.to_be_aligned_elev)
+        if sampling_source == "reference_elev":
+            self.reference_elev = self.reference_elev.crop(coord_intersection)
+            self.generate_plot(self.to_be_aligned_elev, "crop_reference_elev_map")
```
belletva: I suggest to rename in "cropped_reference_elev_map"

xdem/workflows/accuracy.py
```diff
@@ -140,33 +138,40 @@ def _compute_coregistration(self) -> RasterType:
...
+        else:
+            self.to_be_aligned_elev = self.to_be_aligned_elev.crop(coord_intersection)
+            self.generate_plot(self.to_be_aligned_elev, "crop_to_be_aligned_elev_map")
```
belletva: same here : "cropped_to_be_aligned_elev_map"

xdem/coreg/blockwise.py
belletva, line 1: All the files saved in "plots" have the suffix "map" except diff_elev_after_coreg.png and diff_elev_before_coreg.png. Add this suffix "map"

xdem/workflows/workflows.py
belletva, line 1: Use the same extent between the png images if there is an intersection (the largest one)

xdem/dem.py
```diff
@@ -267,6 +267,20 @@ def set_vcrs(
         self._vcrs = _vcrs_from_user_input(vcrs_input=new_vcrs)
         self._vcrs_grid = _grid_from_user_input(vcrs_input=new_vcrs)
 
+    def get_utm_zone_as_epsg_code(self) -> int | None:
+        """
+        Get the UTM zone as an EPSG code.
+        """
+
+        if self.crs.is_geographic:
+            lon_center = (self.bounds.left + self.bounds.right) / 2
```
marinebcht: I think It doesn't work with dem at antimeridian  
if [ 179 , -178 ] lon_center = 0.5 => utm_zone = 31 (not 1)

tests/test_coreg/test_blockwise.py
```diff
@@ -147,6 +147,27 @@ def test_blockwise_coreg_pipeline(self, step, example_data, tmp_path, block_size
...
+        blockwise_coreg.apply()
+
+        aligned = xdem.DEM(tmp_path / "aligned_dem.tif")
+
+        # Ground truth comparison with full image coregistration
+        nuth_kaab = xdem.coreg.NuthKaab()
+        expected = nuth_kaab.fit_and_apply(ref, tba, mask)
+
+        valid = (expected.data.data != expected.nodata) & (aligned.data.data != aligned.nodata)
+        assert np.allclose(expected.data.data[valid], aligned.data.data[valid], atol=20)
```
marinebcht: note for the futur : can be > 20 for block_size = 500

tests/test_dem.py
```diff
@@ -517,3 +517,17 @@ def test_estimate_uncertainty(self) -> None:
 
         assert isinstance(sig_h, gu.Raster)
         assert callable(corr_sig)
+
+    def test_utm_zone_as_epsg_code(self) -> None:
+        """
+        Test utm_zone_as_epsg_code function with random faked data
+        """
+        dem = DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
+        print(dem.crs.is_geographic)
```
marinebcht: little print to delete :)

xdem/workflows/workflows.py
adebardo: https://github.com/GlacioHack/geoutils/issues/755

xdem/workflows/accuracy.py
```diff
@@ -140,33 +138,41 @@ def _compute_coregistration(self) -> RasterType:
...
-            self.reference_elev = reprojected
+        # Reprojection
+        if sampling_source == "reference_elev":
+            crs_utm = self.reference_elev.get_utm_zone_as_epsg_code()
+        else:
+            crs_utm = self.to_be_aligned_elev.get_utm_zone_as_epsg_code()
+
+        logging.info("Computing reprojection")
+        if crs_utm is None:
+            if sampling_source == "reference_elev":
```
marinebcht: I don't understand how the crs_utm == None from input dem selected AND this one input can also lead reprojection ?

xdem/workflows/accuracy.py
```diff
@@ -140,33 +138,41 @@ def _compute_coregistration(self) -> RasterType:
...
+        logging.info("Computing intersection")
+        coord_intersection = self.reference_elev.intersection(self.to_be_aligned_elev)
+        if sampling_source == "reference_elev":
+            self.reference_elev = self.reference_elev.crop(coord_intersection)
+            self.generate_plot(self.to_be_aligned_elev, "cropped_reference_elev_map")
+        else:
+            self.to_be_aligned_elev = self.to_be_aligned_elev.crop(coord_intersection)
+            self.generate_plot(self.to_be_aligned_elev, "cropped_to_be_aligned_elev_map")
 
         if self.level > 1:
```
marinebcht, line 174: I don't find the doc about this ?

xdem/workflows/accuracy.py
```diff
@@ -264,7 +268,7 @@ def run(self) -> None:
             diff_pairs = [("", self.to_be_aligned_elev)]
 
         for label, dem in diff_pairs:
-            diff = dem - ref_elev
+            diff = dem.reproject(ref_elev) - ref_elev
```
marinebcht: in case of "after", there is a double reprojection  
dem = aligned_elev.reproject(ref_elev)).reproject(ref_elev) - ref_elev (l66+l271)

xdem/workflows/workflows.py
```diff
@@ -174,17 +174,29 @@ def load_dem(config_dem: Dict[str, Any] | None) -> tuple[DEM, Raster, str | None
 
     def remove_none(self, dico: Union[Dict[str, Any], List[Any]]) -> Union[Dict[str, Any], List[Any]]:
         """
-        Recursively remove all keys whose values are None from a dictionary.
+        Recursively remove all keys whose values are None from a dictionary,
```
marinebcht: Maybe remove empty keys too ?

dico = {"a":None, "b":None, "c": {"h": None}}
=> {'c': {}}

xdem/workflows/accuracy.py
marinebcht: same for "before" reproj(ref) x2 (in _prepare_datas_for_coreg + here)

rhugonnet: If useful:
To check if a reprojection is required or not, you can use `georeferenced_grid_equal`: https://geoutils.readthedocs.io/en/stable/gen_modules/geoutils.Raster.georeferenced_grid_equal.html#geoutils.Raster.georeferenced_grid_equal

xdem/workflows/accuracy.py
adebardo: https://xdem.readthedocs.io/en/latest/cli_accuracy.html 

image

adebardo: @marinebcht Is it enough ? :)

marinebcht: Yes ! Thanks

xdem/workflows/accuracy.py
adebardo: Corrected x)

xdem/workflows/accuracy.py
adebardo: It was due to the misunderstanding before :)

belletva (dismissed):
Good for me @rhugonnet @adehecq

tests/test_coreg/test_blockwise.py
```diff
@@ -147,6 +147,28 @@ def test_blockwise_coreg_pipeline(self, step, example_data, tmp_path, block_size
         valid = (expected.data.data != expected.nodata) & (aligned.data.data != aligned.nodata)
         assert np.allclose(expected.data.data[valid], aligned.data.data[valid], atol=20)
 
+    @pytest.mark.parametrize("block_size", [500, 985, 1332], ids=["2d_shifts", "1d_shifts_x", "monotile"])
```
rhugonnet: I don't think the "ids" correspond to anything in the tests, do they? The example data is fixed, so the shifts are always the same?

tests/test_coreg/test_blockwise.py
```diff
@@ -147,6 +147,28 @@ def test_blockwise_coreg_pipeline(self, step, example_data, tmp_path, block_size
...
+        )
+        blockwise_coreg = xdem.coreg.BlockwiseCoreg(step=step, mp_config=config_mc, block_size_fit=block_size)
+        blockwise_coreg.fit(ref, tba, mask)
+        blockwise_coreg.apply()
+
+        aligned = xdem.DEM(tmp_path / "aligned_dem.tif")
+
+        # Ground truth comparison with full image coregistration
+        nuth_kaab = xdem.coreg.NuthKaab()
+        expected = nuth_kaab.fit_and_apply(ref, tba, mask)
```
rhugonnet: This should use "step" instead, not necessarily NuthKaab?

rhugonnet: It will ensure full consistency. Actually, `step` currently uses `NuthKaab(vertical_shift=False)`, but not the NuthKaab redefined here.
This might explain why you have such drastic differences below for the `np.allclose()`

tests/test_coreg/test_blockwise.py
```diff
@@ -147,6 +147,28 @@ def test_blockwise_coreg_pipeline(self, step, example_data, tmp_path, block_size
...
+
+        valid = (expected.data.data != expected.nodata) & (aligned.data.data != aligned.nodata)
+        # TODO: careful with 20 value
+        assert np.allclose(expected.data.data[valid], aligned.data.data[valid], atol=20)
```
rhugonnet: A diff of 20 is enormous indeed.
Some pixels will have large differences because they are near extreme slopes (~70 degrees), and a tiny shift causes a big diff.
This happens a lot in the `terrain` module, when derivatives (especially curvatures) are very sensitive to little changes. To write a robust quantitative test, it's better to remove `np.allclose` and use only a difference check on a percentile of the data with a much finer tolerance, like this:
```
diff = np.abs(expected - aligned)
ind_valid = np.isfinite(diff)
# 90% of the aligned data differs by less than 10e-2
assert np.nanpercentile(diff[ind_valid], 90) < 10e-2
```
Hopefully this will work with a small tolerance :wink:

tests/test_workflows/test_accuracy.py
```diff
@@ -286,3 +275,9 @@ def test_mask_init(tmp_path, get_accuracy_inputs_config):
     inlier_mask = ~mask.create_mask(dem)
     assert workflows.inlier_mask == inlier_mask
     assert Path(tmp_path / "plots").joinpath("masked_elev_map.png").exists()
+
+
+def test_prepare_datas_for_coreg():
+    """
+    Test prepare_datas_for_coreg function
+    """
```
rhugonnet: Remove or finalize this test?

xdem/coreg/blockwise.py
```diff
@@ -295,12 +295,13 @@ def _ransac(
...
         return a, b, c
 
+    @staticmethod
     def _wrapper_apply_epc(
-        self,
         tba_dem_tile: RasterType,
         coeff_x: tuple[float, float, float],
         coeff_y: tuple[float, float, float],
         coeff_z: tuple[float, float, float],
+        apply_z_correction: bool = False,
```
rhugonnet, line 304: Add a test for this new parameter in blockwise?

tests/test_coreg/test_blockwise.py
rhugonnet: Same further above (not changed in this PR).

belletva (approved):

PR #812, opened 2025-10-15
Adapt xDEM with latest GeoUtils

marinebcht:
This PR updates xDEM to the latest GeoUtils changes (to test on CI and merge only once a new GeoUtils is released!).

Changes : 
- Import of subsample_array from geoutils.raster to geoutils.stats.sampling
- Updates Vector.create_mask(raster= to ref= et xres= to res=)
- Updates Raster.interp_points() to use as_array=True in tests, to avoid returning a point cloud,
- Updates the call to _grid() to return only the first output, as there are now 2.

EDIT from @rhugonnet: Added the last fixes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ pyproj>=3.4,<4
 rasterio>=1.3,<2
 scipy==1.*
 tqdm
-geoutils==0.1.17
+geoutils==0.2.0
 affine
 pandas
 pyogrio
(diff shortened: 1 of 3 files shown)


rhugonnet:
No worries, done! All passing locally :slightly_smiling_face: 
There were updates to `interp_points()` and `grid()` that changed the type/shape of the outputs.

We can't merge this PR yet, we need to wait until we want to release in GeoUtils. But we can use it in the meantime to work synchronously on both packages.

PR #815, opened 2025-10-23
Downsampling DEM CLI

adebardo:
# Context

The purpose of this PR is to provide simple, interim solutions for CLI users to reduce memory usage while waiting for more substantial future improvements.

We propose one first solution:
- The ability to load the input DEM in a downsampled version using a simple input parameter (`downsample`), including units tests and documentation for users. 

**Additional changes:** minor renaming of `info` to `topo`, and a fix in the `-h` option with the complete command display.

# Extra

We use the partial function to load function, it will be easier later for overload those function

### DIFF
--- a/doc/source/cli_accuracy.md
+++ b/doc/source/cli_accuracy.md
@@ -48,9 +48,12 @@ xdem accuracy --display_template_config
                "path_to_mask", "Path to mask associated to the elevation", "str", "", "No"
                "from_vcrs", "Original vcrs", "str, int", None, "No"
                "to_vcrs", "Destination vcrs", "str, int", None, "No"
-               "downsample", "Downsampling elevation", "int", 1, "No"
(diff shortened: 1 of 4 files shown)


adebardo:
@belletva @rhugonnet the a proposition suggested by @adehecq  to process some large datas the easy way

xdem/workflows/workflows.py
```diff
@@ -71,6 +71,7 @@ def __init__(self, user_config: str | Dict[str, Any]) -> None:
 
         self.outputs_folder = Path(self.config["outputs"]["path"])
         self.outputs_folder.mkdir(parents=True, exist_ok=True)
+        logging.info(f"Outputs will be save at {self.outputs_folder}")
```
rhugonnet: typo: "saved"

PR #819, opened 2025-11-03
Lack shape consistency on fit and apply 

### LINKED ISSUE #756: Lack shape consistency check in xdem.coreg fit_and_apply
There is no test in `fit_and_apply` to check that the DEM to coregister and the inlier_mask have the same shape. if that happens, this leads to an error downstream that is difficult to understand.
Here is a MWE (an opportunity for me to test the new `icrop` 😄) :
```
import geoutils as gu
import numpy as np
import xdem

reference_dem = xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
dem_to_be_aligned = xdem.DEM(xdem.examples.get_path("longyearbyen_tba_dem"))
glacier_outlines = gu.Vector(xdem.examples.get_path("longyearbyen_glacier_outlines"))

# We create a stable ground mask (not glacierized) to mark "inlier data".
inlier_mask = ~glacier_outlines.create_mask(reference_dem)

# crop the mask by a few pixels on each side
nrows, ncols = inlier_mask.shape
inlier_mask2 = inlier_mask.icrop((0, 0, ncols-2, nrows-2))
print(inlier_mask2.shape)
print(inlier_mask.shape)

# Try running coregistration
nuth_kaab = xdem.coreg.NuthKaab()
aligned_dem = nuth_kaab.fit_and_apply(reference_dem, dem_to_be_aligned, inlier_mask2)
```
returns the following error:

> File /usr/local/Caskroom/miniconda/base/envs/xdem/lib/python3.12/site-packages/xdem/coreg/base.py:268, in _preprocess_coreg_fit_raster_raster(reference_dem, dem_to_be_aligned, inlier_mask, transform, crs, area_or_point)
>     265     raise ValueError("'dem_to_be_aligned' had only NaNs")
>     267 # Isolate all invalid values
> --> 268 invalid_mask = np.logical_or.reduce((~inlier_mask, ref_mask, tba_mask))
>     270 if np.all(invalid_mask):
>     271     raise ValueError("All values of the inlier mask are NaNs in either 'reference_dem' or 'dem_to_be_aligned'.")
> 
> ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.

marinebcht:
Resolves #756 

# Context

The problem appears when we need to do a fit_and_apply() with a different size mask of the elevation inputs.

```
  File "....xdem/xdem/coreg/base.py", line 227, in _preprocess_coreg_fit_raster_raster
    invalid_mask = np.logical_or.reduce((~inlier_mask, ref_mask, tba_mask))
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.
```

As a reminder, the `fit_and_apply(reference_dem, dem_to_be_aligned)` works like this : 

| ref \\ tba | DEM | array | PC |
|------------|-----|-------| -------|
| DEM | OK | OK (1/2\*) with `transform/crs` information | OK |
| array | OK | OK (3\*) with `transform/crs` information | OK (1\*) with `transform/crs` information  |
| PC | OK | OK (1\*) with `transform/crs` information | KO |

(1\*)  error in `_preprocess_coreg_apply`  if not transform/crs

(2\*) /!\\ `warning f"'{name}' of type {type(dem)} overrides the given 'transform'")` dans `_preprocess_coreg_fit_raster_raster` car `If any input is a Raster, use its transform if 'transform is None'.`

(3\*) error in `_preprocess_coreg_fit` if not transform/crs

# Resolution 

When the mask is not the same size as elevation data, here are the processes : 

## mask is a raster

| ref \\ tba | DEM | array | PC |
|------------|-----|-------| -------|
| DEM | reproject with ref (a) |  reproject with ref/transform/crs (a) | reproject with ref  (b) |
| array | reproject with tba (a) |  reproject with ref/transform/crs (a) | reproject with array/transform/crs (b) |
| PC | reproject with tba (b) | reproject with array/transform/crs  (b)| not possible |

(a) _preprocess_coreg_fit_raster_raster
(b) _preprocess_coreg_fit_raster_point

## mask is an array 

(E2) = No solution found to reproject mask array    
=> Raises error "Input mask array can't be a different size array as input elevation."

# Tests 

The reference results (dem + shift values) are computed with a cropped mask, reprojected with reference_dem. 

```
dem_aligned_crop = nuth_kaab_crop.fit_and_apply(reference_dem, to_be_aligned_dem, inlier_mask_crop_proj, random_state=42)
```

All `nuth_kaab_crop.fit_and_apply` with the different configs listed in the previous table + the cropped mask `inlier_mask_crop_proj` need to match the reference results.

### DIFF
--- a/dev-environment.yml
+++ b/dev-environment.yml
@@ -30,14 +30,17 @@ dependencies:
   # Test dependencies
   - pytest
   - pytest-xdist
+  - pytest-instafail
+  - pytest-socket
+  - pytest-cov
+  - coveralls
   - flake8
   - pylint
   - pre-commit
 
   # Doc dependencies
   - sphinx
-  - pydata-sphinx-theme
-  - sphinx-book-theme>=1.0
+  - sphinx-book-theme
   - sphinxcontrib-programoutput
   - sphinx-design
   - sphinx-autodoc-typehints


rhugonnet:
@marinebcht Amazing description and summary, I grasped everything immediately. :wink:

And great idea for the extra consistency tests! (crop/uncropped and pointcloud/raster)

Here are possible explanations:
- **For uncropped/crop:** In the first case (`inlier_mask` is cropped), the random subsample coordinates are computed using random indices of the N*M array. While, in the second case, they are computed using indices of your cropped array (N - 100 x M - 100). So even with `random_state=42`, the code is selecting different subsamples.
If you pin `subsample=1` to use all samples in both cases, then there should be no more randomness, and your test should yield the exact same result for both cases! (hopefully)

- **For pointcloud/raster input and order:** Same as above but with points pre-selected differently du to the various types of the datasets. With `subsample=1` you should get the same result hopefully. If not, we have things to fix!

For the PR changes:
(E1): If the mask is a raster, we can run `mask.reproject(reference_dem, resampling="nearest")` using the `_reproject()` that works with transform/crs input. But we should raise an error if the mask doesn't intersect the original raster: `mask.get_projected_footprint(crs).intersects(dem.get_projected_footprint())`, also here using the equivalent with transform/crs input.
(E2): OK for me.

xdem/coreg/base.py
```diff
@@ -313,6 +314,33 @@ def _preprocess_coreg_fit(
     ):
         raise ValueError("Input elevation data should be a raster, an array or a geodataframe.")
 
+    # If inlier_mask has not the same shape of input dems, reproject it into an into dem # todo
```
rhugonnet: This code only applies when the inputs are rasters, so should be moved to `_preprocess_coreg_fit_raster_raster` called directly below.

rhugonnet:
Quick answers:
- **Inlier mask out of bounds:** I think our only option is to raise an error: if the user is passing an `inlier_mask` that does not intersect the DEMs, then all other values are by definition `False` (outside the mask), so we can't run the coregistration further.
- **Reprojection:** No, we **can** reproject, this is what I meant above. Here is the way: `ref_rst = Raster.from_array(data=ref_dem, transform=transform, crs=crs); inlier_mask = inlier_mask.reproject(ref_rst)`,
- **New tests differences:** Then another difference is introduced somewhere, we need to dig more... We could start by checking the output of `preprocess_fit` is indeed the same in your 2 cases. There actually aren't many tests for preprocessing functions right now. Amélie had started some here but we never got to merge them: https://github.com/GlacioHack/xdem/pull/525

dev-environment.yml
```diff
@@ -30,17 +30,14 @@ dependencies:
   # Test dependencies
   - pytest
   - pytest-xdist
-  - pytest-instafail
-  - pytest-socket
-  - pytest-cov
-  - coveralls
```
rhugonnet, line 36: This file needs to be restored to match `main` (wrong merge?).
All these dependencies mirror the ones in `setup.cfg`, for those using a conda install.

tests/test_coreg/test_base.py
```diff
@@ -438,6 +438,95 @@ def test_fit_and_apply__pipeline(self) -> None:
...
+            ("array", "raster", True, "raster", "passes", ""),
+            ("raster", "array", True, "raster", "passes", ""),
+            ("array", "array", True, "raster", "passes", ""),
+            ("pc", "raster", False, "raster", "passes", ""),
+            ("raster", "pc", False, "raster", "passes", ""),
+            ("pc", "array", True, "array", "error", "Input mask array"),
+            ("array", "pc", True, "array", "error", "Input mask array"),
+        ],
+    )  # type: ignore
+    def test_fit_and_apply__cropped_mask(self, combination: tuple[str, str, str, str, str, str]) -> None:
```
rhugonnet, line 456: Amazing series of tests, thanks a lot :wink:

rhugonnet (approved):
Perfect! See the small issue with the `dev-env` file.
We can add the new tests in a separate PR following your issue.

PR #826, opened 2025-11-25
Raise new error when `vcrs` user input has no extension, and fix HTTP error catching during download attempt

### LINKED ISSUE #824: 垂直参考设置为EGM2008时有问题
dem.set_vcrs("EGM2008") 
/home/ sd/anaconda3/envs/xdem-dev/share/proj下明明存在us_nga_egm08_25.tif，但就是报出urllib.error.HTTPError: HTTP Error 404: Not Found

rhugonnet:
Fixes the catching of HTTP error when attempting to download non-existing grids, which changed from `http.client` to `urllib`. Adds a test to verify this works correctly.
Adds a new check verifying that a user-provided grid has an extension (.tif/.json/.pol), otherwise fails directly with clear instructions on possible inputs.

Resolves #824

### DIFF
--- a/tests/test_dem.py
+++ b/tests/test_dem.py
@@ -238,7 +238,7 @@ class TestDEM:
 
         # -- Test 2: we check with grids --
         # Most grids aren't going to be downloaded, so this warning can be raised
-        warnings.filterwarnings("ignore", category=UserWarning, message="Grid * not found in *")
+        warnings.filterwarnings("ignore", category=UserWarning, message="Grid .*")
 
         dem.set_vcrs(new_vcrs="us_nga_egm96_15.tif")
         assert dem.vcrs_name == "unknown using geoidgrids=us_nga_egm96_15.tif"


tests/test_workflows/test_schemas.py
```diff
@@ -212,12 +212,12 @@ def test_valid_to_vcrs(get_topo_inputs_config, pipeline_topo, prefix, vcrs):
     [
         pytest.param(
             "wrong",
-            UserWarning,
+            ValueError,
```
rhugonnet: @adebardo With the update of this PR (we now properly catch `HTTPError` and other wrong string input that went undetected before), I couldn't easily adjust the error in this test.
I'm not sure I understood the logic: The `pyproj.CRS` EPSG error goes through normally, but the others don't and instead add something to logging? 
Now that error raising is more robust and clear directly in `vcrs`, should we simply raise the error every time? :thinking:

rhugonnet: I'm referring to the try/except clauses here: https://github.com/GlacioHack/xdem/blob/782a92e0ff21bcdd85c58fae50a3087e31b59927/xdem/workflows/schemas.py#L47

rhugonnet: @adebardo Tagging you again in case you missed it, need your input :wink:

rhugonnet: Alright, I've resolved this using the `caplog` fixture to capture logging, and updating the test to check a logging error is raised instead.

Issue #632, opened 2024-11-02
Include vcrs in DEM.info


adehecq:
Just some thought while reading the new doc. I don't think the VCRS is displayed when calling `DEM.info()`. It could be interesting to add it.

rhugonnet:
Joins #370.

rhugonnet:
@marinebcht The CCRS will be removed eventually (we'll override the CRS instead, now that Rasterio/Proj support 3D CRS directly). 
So the best is to display only CRS + VCRS.

rhugonnet:
@marinebcht I'm not sure I understood the problem above. Everything seems to already be consistent with what we support in xDEM. We just need to merge `ccrs` with `crs` eventually as noted in #370.

If I understood correctly:

**For the name:** "Coordinate system" is short for "Coordinate reference system (CRS)". We can use the full version if you like.

**For the display:** Showing the EPSG in `info()` is not a good idea, as many Vertical CRS don't have an EPSG, so it wouldn't work in a general case. Right now we print the Vertical CRS in WKT2 format after being assimilated by PROJ (whether from a WKT, PROJ grid, or EPSG), which is the most generic and covers all cases.

If that wasn't your question, I'll need more details!

PR #830, opened 2025-12-08
Include vcrs in DEM.info 

marinebcht:
Resolves https://github.com/GlacioHack/xdem/issues/632 

Now, the DEM.info() prints the VCRS information, after the Coordinates System line : 

Results dem.info()

image

image

image

### Resolution

To do this, `Raster.infos()` was overwrited and a dynamic insertion was done on its results.
1. `info_str = super.info(stats=stats, verbose=False)`
2. Coordinates System line found
3. replace by  Coordinates System: [ crs.name ] if 3D ; [ crs , vrcs ]

#### Tests

Added test_info() with some checks (all inputs possibles) like the addition on the line and the line itself (when VRCS is None and when initialized).

/!\ need to add 3D vcrs to xdem-data and unskip the corresponding test

### DIFF
--- a/xdem/dem.py
+++ b/xdem/dem.py
@@ -189,7 +189,7 @@ class DEM(Raster):  # type: ignore
 
         # Change crs values if not 3D
         if len(CRS(self.crs).axis_info) > 2:
-            new_crs = self.crs
+            new_crs = [CRS(self.crs).name]
         else:
             new_crs = [self.crs.to_string() if self.crs is not None else None, str(self.vcrs)]
 
--- a/tests/test_dem.py
+++ b/tests/test_dem.py
@@ -369,10 +369,9 @@ class TestDEM:
 
         assert dem_class_attr.raster_equal(terrain_module_attr)
 
-    def test_info(self) -> None:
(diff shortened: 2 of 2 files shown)


xdem/dem.py
```diff
@@ -166,6 +166,41 @@ def __init__(
...
+
+        :returns: Summary string or None.
+        """
+
+        # Get super.infos()
+        raster_info = super().info(stats=stats, verbose=False)  # type: ignore
+        raster_info_split = raster_info.split("\n")
+
+        # VCRS info
+        vcrs_str = f"VCRS:                 {self.vcrs}"
```
belletva: I propose "3D CRS" instead of "VCRS"?

rhugonnet: Maybe "Vertical CRS"?. The `vcrs` in xDEM is a vertical CRS, not a 3D CRS (A 3D CRS is the combination of a 2D CRS + a vertical CRS).
In the future `DEM.crs` might be allowed to be 3D (for now it's only `ccrs` which is internal). See #370.

marinebcht:
Results dem.info()

image

image

image

PR #860, opened 2026-01-05
Documentation update before `0.2`

### LINKED ISSUE #869: Add `output` argument to CLI to easily set dynamic output folder
Having the output folder being defined only in the YAML file makes it a bit difficult for users to apply the workflows to multiple DEMs (they'd need to write code to dynamically change the YAML; For those doing everything in bash especially, it's quite difficult).

It also makes it hard to generate reports in specific location for CI/documentation (with the same setup), as I'm doing in #860 and #852.

I think we should have an `--output` argument that override the output folder path in the YAML, so that it's easy to make the variable dynamic.
We might think of doing the same for the primary elevation path (for `topo`) or two elevation paths (for `accuracy`), with two more arguments?

This way we provide the best of both world! 😄

### LINKED ISSUE #870: Embedded report in documentation fails due to ReadTheDocs security
The workflows I added in #852 works locally, but fails on ReadTheDocs, because PDF/HTML files that don't exist in a `_static` folder BEFORE runtime are disregarded for security reasons.

Adding a solution in #860.

### LINKED ISSUE #871: Add automated check for report consistency in CI
Due to #870, we can't show the report generated during runtime. Hence, the only solution is to have a report pre-saved in `_static`.

We don't want to have to check if it has changed manually everytime we make a change, so we need to add a CI job that does this for us.

Working on it in #860.

### LINKED ISSUE #872: CLI not printing `Coreg` dictionaries as string
For instance, "least_squares" is a function in the dictionary, and so renders badly in the report (void entry for the metadata "fit_optimizer").

I'm fixing quickly in #860, but maybe we want to use the equivalent of `Coreg.info()` in the report with full description names available here: https://github.com/GlacioHack/xdem/blob/c88a49ab46fa6f91e27b4844f4a20e529a0cb741/xdem/coreg/base.py#L86 instead of the technical metadata names of `Coreg`?

rhugonnet:
This PR updates the documentation, or aspects related to the documentation, before releasing xDEM 0.2.

Doc before the recent updates (to compare to CLI not updated yet): https://xdem-rhugonnet.readthedocs.io/en/small_terrain_doc_fixes/global_cli_information.html
**Doc now: https://xdem-rhugonnet.readthedocs.io/en/doc_before_0.2/cli.html**

In particular, this PR:
- Updates the README, 
- Adds release notes for 0.2,
- Streamlines various sections and corrects small errors/fixes.

Additionally, this PR fixes the rendering of the CLI report in the documentation introduced in #852 (which worked locally but failed on ReadTheDocs due to security reasons, see #870).

For this, this PR:
- Uploads the PDF/HTML reports "manually" to a folder `doc/source/_workflows/` (which is the only possible option for this, see #870), the PDF file is used to render in the documentation while the HTML (devoids of its subplots) is only used for CI purposes (see next point),
- Adds a CI job that automatically checks that the report displayed in the CLI documentation pages are up to date. If not, the job fails and gives instruction to re-generate the reports in a few lines. For checking if the report has changed, it runs each CLI workflow (currently `xdem accuracy` and `xdem topo`) for every PR commit, and computes the diff of the output HTML files (Computing a PDF diff would be asking for trouble; The HTML diff only includes the text, but this is a pretty good indicator of change already; We could extend to include the plot files at some point if wanted but it might become oversensitive).
- The above workflow requires 1/ Skipping some runtime metadata displayed in the report (computing time, date; which are removed directly in the `bash` script of the CI job), 2/ Pinning the `random_state` of the coregistration in `xdem accuracy` to ensure the report is reproducible, and 3/ Adding a `--output` CLI argument to dynamically override the output folder and be able to compare the 2 outputs (injecting a different output in the YAML is not practical).

Resolves #869 
Resolves #872
Resolves #871 
Resolves #870 

## TO-DO

- [x] (Big one) Embedded PDF for CLI workflows appear on Sphinx build locally but not on Readthedocs...
- [x] Update README,
- [x] Fix Sphinx warnings and unwanted std_err,
- [x] Add release notes for 0.2,
- [x] Update copyright date to 2026,
- [x] Florinsky missing from citations for methods?
- [x] Add introduction to detailed configuration parameters for "Topo" workflow (mirroring "Accuracy" one).

### DIFF
--- a/doc/source/config.md
+++ b/doc/source/config.md
@@ -80,6 +80,12 @@ def my_xdem_function():
   ...
 ```
 
+Some functions are already profiled automatically when the Profiler is enabled, with a memory consumption report each 0.05 seconds.
+Those are:
+- DEM loading through {class}`~xdem.DEM`,
+- All terrain attributes such as {func}`~xdem.DEM.slope`,
+- Co-registration through {func}`~xdem.coreg.Coreg.fit_and_apply` and {func}`~xdem.DEM.coregister_3d`.
+
 Finally, in any other script, the profiler can be activated and the output directory defined:
 
 ```{code-cell} ipython3
@@ -88,15 +94,13 @@ from geoutils.profiler import Profiler
 Profiler.enable(save_graphs=True, save_raw_data=True)
 
 # ...
-# Code calling decorated functions
+# Code calling decorated functions, such as terrain attribute
+import xdem
+dem = xdem.DEM(xdem.examples.get_path("longyearbyen_ref_dem"))
+slope = dem.slope()
 # ...
 
+# Generate the output summary
 my_output_directory="./profile_output/"
 Profiler.generate_summary(my_output_directory)
 ```
-
-Some functions are already profiled automatically when the Profiler is enabled, with a memory consumption report each 0.05 seconds.
-Those are:
-- DEM loading through {class}`~xdem.DEM`,
-- All terrain attributes such as {func}`~xdem.DEM.slope`,
-- Co-registration through {func}`~xdem.Coreg.fit_and_apply` and {func}`~xdem.DEM.coregister_3d`.
--- a/.gitignore
+++ b/.gitignore
@@ -159,6 +159,9 @@ examples/advanced/aligned_dem.tif
 doc/source/_workflows/outputs_accuracy/
 doc/source/_workflows/outputs_topo/
 
+# Should be cleaned during Sphinx execution but could be left in case of error
+doc/source/profile_output
+
 # PDF to keep to display in documentation
 !doc/source/_workflows/outputs_accuracy/report.pdf
 !doc/source/_workflows/outputs_topo/report.pdf


doc/source/cli_accuracy.md
belletva, line 1: I think it is important to just add which values are excluded in the line: "The **path to a mask of unstable surfaces**, to exclude terrain during the analysis **(0 values in the mask are excluded)**".

doc/source/cli_accuracy.md
```diff
@@ -80,33 +80,12 @@ Now that we have this configuration file, we run the workflow.
...
-shutil.copytree(src, dst)
-```
-
 Finally, a report is created (both in HTML and PDF formats) in the output directory.
 
 We can visualize the report of our workflow above:
 
 ```{raw} html
-<iframe src="_workflows/outputs_accuracy/report.html" width="100%" height="800"></iframe>
+<iframe src="_static/outputs_accuracy/report.pdf" width="100%" height="800"></iframe>
```
belletva, line 88: really nice

doc/source/cli_topo.md
```diff
@@ -116,10 +95,18 @@ The `topo` workflow is described by the following chart:
 (params-topo)=
 ### Configuration parameters
 
+The parameters to pass to the `topo` workflow are divided into four categories:
+- The `inputs` define file opening and pre-processing, including **one required path to elevation data**, but also optional masking, CRS, nodata over-riding, and downsampling factor,
+- The `outputs` define file writing and report generation, with various **levels** of detail for the produced outputs,
+- The `terrain attributes` define steps for coregistration, directly **interfacing with the {ref}`terrain` module** of xDEM,
+- The `statistics` define steps for computing statistics before/after coregistration, directly **interfacing with the [Statistics](https://geoutils.readthedocs.io/en/stable/stats.html) module** of GeoUtils.
```
belletva, line 102: Linked to this issue: https://github.com/GlacioHack/geoutils/issues/800, not sure that is useful to compute the 15 statistics by default.

doc/source/config.md
```diff
@@ -67,20 +67,36 @@ These changes will then apply to all your operations in xDEM, such as coregistra
...
+
+my_output_directory="./profile_output/"
+Profiler.generate_summary(my_output_directory)
+```
+
+Some functions are already profiled automatically when the Profiler is enabled, with a memory consumption report each 0.05 seconds.
+Those are:
+- DEM loading through {class}`~xdem.DEM`,
+- All terrain attributes such as {class}`~xdem.DEM.slope`,
+- Co-registration through {function}`~xdem.Coreg.fit_and_apply` and {function}`~xdem.DEM.coregister_3d`.
```
belletva: Add an illustration ?

README.md
```diff
@@ -14,20 +14,21 @@
 [![Checked with mypy](http://www.mypy-lang.org/static/mypy_badge.svg)](http://mypy-lang.org/)
 [![Imports: isort](https://img.shields.io/badge/%20imports-isort-%231674b1?style=flat&labelColor=ef8336)](https://pycqa.github.io/isort/)
 
-**xDEM** is an open source project to develop a core Python package for the analysis of digital elevation models (DEMs).
+**xDEM** is an open source project to develop a core Python package for the analysis of digital elevation models (DEMs) and elevation point clouds.
```
belletva: and elevation point clouds (EPC).

doc/source/quick_start.md
```diff
@@ -116,24 +116,27 @@ os.remove("dh_error.tif")
...
-    path_to_elev: "xdem/examples/data/Longyearbyen/data/DEM_2009_ref.tif"
-  to_be_aligned_elev:
-    path_to_elev: "xdem/examples/data/Longyearbyen/data/DEM_1990.tif"
-    path_to_mask: "xdem/examples/data/Longyearbyen/data/glacier_mask/CryoClim_GAO_SJ_1990.shp"
+```{caution}
+The command-line interface for workflows is experimental, and its API might change rapidly.
 ```
 
-And run the workflow, which will generate several outputs including an HTML report (by default in an `outputs` folder of your current directory).
+The configuration file should contain, at minima, the path to elevation files but aims to be modular to tune the use of xDEM's features (see details in the **{ref}`cli` reference page**):
```
belletva: Precision that in this case it is the configuration file and the command line for the accuracy workflow?

doc/source/release_notes.md
```diff
@@ -1,7 +1,36 @@
+(release-notes)=
 # Release notes
 
 Below, the release notes for all minor versions and our roadmap to a first major version.
 
+## 0.2.0
+
+xDEM version 0.2 is the **second minor release**, notably marked by the merge in efforts with [demcompare](https://github.com/CNES/demcompare) to build a unified tool!
```
belletva: Perhaps it should be noted that demcompare is no longer maintained?

doc/source/release_notes.md
```diff
@@ -1,7 +1,36 @@
...
+It is the result of months of work to consolidate several features, in particular support for **elevation point clouds**. Parallel work on scalability with Dask and Multiprocessing
```
belletva: (EPC)

doc/source/release_notes.md
belletva: in about_xdem.html : replace the note "xDEM is merging efforts with CNES’s [demcompare](https://github.com/CNES/demcompare) to combine the best of both tools into one!" by "CNES has stopped developing its [demcompare](https://github.com/CNES/demcompare) tool and is now contributing to xDEM."

doc/source/cli_accuracy.md
rhugonnet: Looking at the code in `accuracy.py`, right now the "masks" are only supported for vector files (.SHP, .GPKG), and excludes the terrain inside the polygons. See here:
https://github.com/GlacioHack/xdem/blob/c88a49ab46fa6f91e27b4844f4a20e529a0cb741/xdem/workflows/workflows.py#L183

We could have a helper function that supports more inputs (inclusion/exclusion shapefile and inclusive/exclusive raster masks), a bit in the same direction as the helper function of Amaury's old `dem_coregistration`:
https://github.com/GlacioHack/xdem/blob/01e09cde8e99a354b4628aae7447a1f5bb60a5e4/xdem/coreg/workflows.py#L42
(but without the filtering, just simply masking terrain)

I thought we had other code doing this...

rhugonnet: Found it!
We have an old function doing the same thing in `spatialstats`: https://github.com/GlacioHack/xdem/blob/c88a49ab46fa6f91e27b4844f4a20e529a0cb741/xdem/spatialstats.py#L652

@adebardo @belletva I think we should consolidate into a single function modular-enough to be used everywhere? :slightly_smiling_face:
I'll open an issue!

rhugonnet: Opened #874

doc/source/cli_topo.md
rhugonnet: Yes, I opened it exactly when I was working on this section :wink:

README.md
rhugonnet: Fixed

doc/source/release_notes.md
rhugonnet: Good point, added for both

doc/source/quick_start.md
rhugonnet: Added!

doc/source/config.md
rhugonnet: Just looked into it, and it's pretty hard to do now, for the same reason as the HTML report: We can't render an HTML created **on disk** during runtime in the documentation, for security reasons: #870. (The plots for profiling are saved as HTML)

But for profiling, there's an easier solution than for the CLI (because it's a single plot, not a full report): We could modify functions of the `Profiler` class to be able to return the plots directly (and not only save to disk). This way the figures would be accessible interactively (Python API/Documentation using `Profiler.plot()`) while maintaining the save on disk (for the `generate_summary()` function).

@marinebcht @belletva Thoughts?

PR #863, opened 2026-01-06
Update management of optional dependencies

### LINKED ISSUE #755: Separate standard and expert dependencies (pip install xdem[expert])
# Context:

Currently, installing `xdem` installs all dependencies, including those only required for advanced or expert-level use. This can unnecessarily increase the installation time and complexity for users who only need basic functionality or prefer a lightweight setup.

* `pip install xdem` → Minimal installation (light mode)
* `pip install xdem[expert]` → Full installation with additional expert dependencies

# Tasks:

* Identify the dependencies required for basic (light) usage.

  * [ ] List the "expert" dependencies

  * [ ] Modify the `setup.py` file to include `extras_require`:

    Example:

    ```python
    extras_require={
        "expert": [
            "matplotlib",
            "scikit-learn",
            "rasterio",
            # other expert-level dependencies
        ]
    }
    ```

  * [ ] Update the documentation

  * [ ] Add a test in the CI to ensure both installation modes work correctly.

### LINKED ISSUE #795: Drop optional dependency to `pytransform3d` in favor of new SciPy class?
We only use `pytransform3d` to check rigid 3D matrices.

This can be done fairly easily with just NumPy, and is now added in SciPy 1.16 for consistent future support: "A new class, scipy.spatial.transform.RigidTransform, provides functionality to convert between different representations of rigid transforms in 3-D space"

We could replace towards this moving forward.

### LINKED ISSUE #866: CLI dependencies (cerberus, weasyprint, pyyaml) are not optional
Need to move them to optional as we discussed, doing it in #863

### LINKED ISSUE #867: Wrap `tqdm` as an optional dependency
We shouldn't impose a "progress-bar" software package on anyone, there is a way to wrap it so that it's available if desired, otherwise not.
Will do this in #863.

rhugonnet:
Mirrors https://github.com/GlacioHack/geoutils/pull/810
Resolves #795
Resolves #866
Resolves #867 
Resolves #755 

This PR updates the management of optional dependencies following good practices (see details here:  https://github.com/GlacioHack/xdem/wiki/Good-practices-for-managing-optional-dependencies).

In particular, this PR:
- Removes the dependency on `pytransform3d` (was only used to manipulate a couple aspects of rotation matrices that we already did manually within registration functions, so I added NumPy logic there),
- Moves `matplotlib`, `cerberus`, `weasyprint`, `pyyaml` and `tqdm` to optional dependencies.

For `tqdm`, this involves creating a helper a function, which fails silently when the dependency does not exist (no-one cares if there's no progress bar for a given feature). While all other lead to failure, which is now checked in the CI. :smile:

### DIFF
--- a/tests/test_spatialstats.py
+++ b/tests/test_spatialstats.py
@@ -456,8 +456,9 @@ class TestBinning:
                 dvalues=self.diff.get_nanarray(), stable_mask=self.outlines, list_var=[self.slope.get_nanarray()]
             )
 
-    @pytest.mark.skipif(find_spec("matplotlib") is not None,
-                        reason="Only runs if matplotlib is missing.")  # type: ignore
+    @pytest.mark.skipif(
+        find_spec("matplotlib") is not None, reason="Only runs if matplotlib is missing."
+    )  # type: ignore
     def test_plot_binning__missing_dep(self) -> None:
         """Check that proper import error is raised when matplotlib is missing"""
 
@@ -978,8 +979,9 @@ class TestVariogram:
                 dvalues=diff_on_stable_arr, stable_mask=self.outlines, list_models=["Gau", "Sph"], random_state=42
             )
 
-    @pytest.mark.skipif(find_spec("matplotlib") is not None,
-                        reason="Only runs if matplotlib is missing.")  # type: ignore
+    @pytest.mark.skipif(
+        find_spec("matplotlib") is not None, reason="Only runs if matplotlib is missing."
+    )  # type: ignore
     def test_plot_variogram__missing_dep(self) -> None:
         """Check that proper import error is raised when matplotlib is missing"""
 


xdem/cli.py
```diff
@@ -18,31 +18,33 @@
...
+    lib_gobject_name = ctypes.util.find_library("gobject-2.0")
+    lib_pango_name = ctypes.util.find_library("pango-1.0")
+    if lib_gobject_name and lib_pango_name:
+        from weasyprint import HTML
+        _has_libgobject = True
+    else:
+        _has_libgobject = False
+    _has_weasyprint = _has_libgobject
+except ImportError:
+    _has_weasyprint = False
```
rhugonnet, line 40: @adebardo @belletva I had to adapt the code above. Doing it, I noticed that, when running without the checks for `weasyprint` on `lib_gobject/pango` written above (basically just doing `try/except` for `import weasyprint`), the CI fails for Mac (but not on Linux).

So, right now, by using the code above, we are later not producing a report on Mac. And we essentially "fail without raising error" then.
I don't think that's good? :sweat_smile:
Should we open an issue on this? Any ideas or experience from other projects?

I'll tag the other spot where we skip the report generation below.

xdem/cli.py
```diff
@@ -112,7 +114,12 @@ def main() -> None:
     else:
         raise ValueError(f"{args.command} doesn't exist, valid command are 'accuracy', 'topo'")
 
-    if args.config and _has_libgobject:
+    if args.config:
+        if not _has_weasyprint:
+            msg = ("Optional dependency 'weasyprint' required. "
```
rhugonnet: Here we skip if the dependencies are not correctly found, which is the case on Mac.

PR #893, opened 2026-01-21
Update with GeoUtils accessor

rhugonnet:
To potentially merge with #656


rhugonnet:
I've fixed inconsistencies but, as we need to wait for a GeoUtils release anyway, I merge in #656

PR #906, opened 2026-02-04
CLI improvement

### LINKED ISSUE #841: CLI : plot intersection
# Context

Currently, in the report, DEMs are displayed in their entirety. We believe it is more relevant to present only the intersecting areas. This ticket depends on ticket #840 .

# Code

It is therefore necessary to ensure that the PNGs are saved at the appropriate stage of the processing, preferably during the data preparation step.  

# Unit Tests

Add tests.

# Documentation

Update the documentation.

marinebcht:
Resolves #841 #922 

### Default method in the doc vs method in the `accuracy_config.yaml` 

Changed NuthKaab to LZD

### Preprocessed datasat presentation 

Save the projected/cropped DEM after data preparation if activated :
- preprocessed_reference_elev_map.png if `sampling_grid = to_be_aligned_elev`
- preprocessed_to_be_aligned_elev_map.png if `sampling_grid = reference_elev`

Show them in the report, in the section after inputs metadata information.

image

### Turned statistics table 

Before in HTML and PDF: image 

After in HTML and PDF: 
image

### Elevation difference plot with the same scale 

image

### Mask management 

Masked data were not plot and correctely handle 

image 
image

+ in the stats now:

image

## TOPO

Remove empty terrain attributes case : before/After

image image

### Dynamic unit

`(m)` is now taken from the CRS info (here: `metre`)

### Stats : 
Format data in the stats  :
- if "count": int
- if "percentage": '{:.2f}'.format(val) + "%"
- if abs(val) > 10E4 or abs(val) < 10E-4: np.format_float_scientific(val, precision=3)
- else '{:.3f}'.format(val)

+ Remove std/standarddeviation duplicate in each template and alias std gives "Standart deviation" too 

See before for accuracy and topo : 

image

# Reduce default stats 

image


xdem/workflows/accuracy.py
```diff
@@ -195,21 +213,26 @@ def _prepare_datas(self) -> None:
         # Intersection
         logging.info("Computing intersection")
         coord_intersection = self.reference_elev.intersection(self.to_be_aligned_elev)
+
         if sampling_source == "reference_elev":
-            self.reference_elev = self.reference_elev.crop(coord_intersection)
+            self.to_be_aligned_elev = self.to_be_aligned_elev.crop(coord_intersection)
```
marinebcht, line 206: to verify @adebardo

doc/source/_workflows/accuracy_config.yaml
```diff
@@ -8,8 +8,8 @@ outputs:
   path: "outputs_accuracy"
 coregistration:
   step_one:
-    method: "LZD"
-    extra_information: {"subsample": 10000}
+    method: NuthKaab
+    extra_information: null
```
rhugonnet: Why replace by `NuthKaab`? I used `LZD` here on purpose because it's the best-performing method.

doc/source/cli.md
```diff
@@ -37,6 +37,8 @@ Also, this template can directly be saved in a YAML file like this:
 xdem workflow_name --template-config template_config.yaml
 ```
 
+Optionally, a output path to save the template can be set directly in the command-line using `--output`.
```
rhugonnet: Isn't that already the effect of `--template-config` followed by the filepath, as done right above? If yes, need to remove this

xdem/workflows/accuracy.py
```diff
@@ -74,24 +74,35 @@ def __init__(self, config_dem: str | Dict[str, Any], output: str | None = None)
...
+        Load data
 
+        :return vmin, vmax: to plot elevation data with the same scale
+        """
         self.to_be_aligned_elev, tba_mask, tba_path_mask = self.load_dem(self.config["inputs"]["to_be_aligned_elev"])
         self.reference_elev, ref_mask, ref_mask_path = self.load_dem(self.config["inputs"].get("reference_elev", None))
         if self.reference_elev is None:
             self.reference_elev = self._get_reference_elevation()
+
+        vmin = min(self.reference_elev.get_stats("min"), self.to_be_aligned_elev.get_stats("min"))
```
rhugonnet: It's not a good idea to use "min" and "max" from the DEMs, many DEMs have outliers (clouds, stereo artefacts), which are 1000+ meters.
If we use `vmin` and `vmax` as min/max in this case, we basically plot nothing (a raster full of a single color). And it shouldn't be the responsibility of the user to filter their DEM perfectly ahead, especially in the CLI.

This is why I changed to median +/- 3 NMAD for the color-scale in #852 (this should always cover 95%-99% of values, depending on the normality of the distribution), and I think we should keep this.

xdem/workflows/accuracy.py
```diff
@@ -321,38 +344,46 @@ def run(self) -> None:
...
-            diff_pairs = [("", self.to_be_aligned_elev)]
-
-        for label, dem in diff_pairs:
-            diff = dem - ref_elev
-            stats_keys = ["min", "max", "nmad", "median"]
-            stats = diff.get_stats(stats_keys)
-
-            if label == "before":
-                self.diff_before, self.stats_before = diff, stats
-                vmin, vmax = -(stats["median"] + 3 * stats["nmad"]), stats["median"] + 3 * stats["nmad"]
```
rhugonnet, line 338: I would keep those robust vmin/vmax as argued above.

xdem/workflows/accuracy.py
rhugonnet: Noticing now this is only for the DEM, not the elevation difference.

It still applies, an outlier cloud would mess things up otherwise. But maybe better to use `np.nanpercentile(5)` and `np.nanpercentile(95)`, because the distribution can be quite asymmetrical for elevation.

xdem/workflows/accuracy.py
rhugonnet: Actually nevermind those are used below, but should also use a robust equivalent for the DEM. Commented above.

xdem/cli.py
```diff
@@ -106,8 +106,7 @@ def main() -> None:
         epilog="examples:\n"
         "  xdem accuracy --config config.yaml\n"
         "  xdem accuracy --config config.yaml --output myoutputfolder\n"
-        "  xdem accuracy --template-config\n"
-        "  xdem accuracy --template-config template_config.yaml",
+        "  xdem accuracy --template-config --output template_config.yaml",
```
rhugonnet: So --output is added now? I didn't see anything about it in the PR description, just checking if intentional.

doc/source/cli.md
marinebcht: My bad, I got confused

doc/source/_workflows/accuracy_config.yaml
marinebcht: Ohhhh ! It was our idea to homogenize with the --template-config 
We'll talk about it tomorrow @belletva @adebardo ^^

xdem/workflows/accuracy.py
marinebcht: Oh perfect yes I understand, it's safer

xdem/workflows/accuracy.py
```diff
@@ -74,24 +74,35 @@ def __init__(self, config_dem: str | Dict[str, Any], output: str | None = None)
...
         self.reference_elev, ref_mask, ref_mask_path = self.load_dem(self.config["inputs"].get("reference_elev", None))
         if self.reference_elev is None:
             self.reference_elev = self._get_reference_elevation()
+
+        vmin = min(self.reference_elev.get_stats("min"), self.to_be_aligned_elev.get_stats("min"))
+        vmax = max(self.reference_elev.get_stats("max"), self.to_be_aligned_elev.get_stats("max"))
+
         self.generate_plot(
             self.reference_elev,
             title="Reference DEM",
```
belletva: reference elev

xdem/workflows/accuracy.py
```diff
@@ -74,24 +74,35 @@ def __init__(self, config_dem: str | Dict[str, Any], output: str | None = None)
...
             filename="reference_elev_map",
+            vmin=vmin,
+            vmax=vmax,
             cmap="terrain",
             cbar_title="Elevation (m)",
         )
         self.generate_plot(
             self.to_be_aligned_elev,
             title="To-be-aligned DEM",
```
belletva: to-be-aligned elev

xdem/workflows/accuracy.py
```diff
@@ -195,21 +213,26 @@ def _prepare_datas(self) -> None:
...
         coord_intersection = self.reference_elev.intersection(self.to_be_aligned_elev)
+
         if sampling_source == "reference_elev":
-            self.reference_elev = self.reference_elev.crop(coord_intersection)
+            self.to_be_aligned_elev = self.to_be_aligned_elev.crop(coord_intersection)
             self.generate_plot(
                 self.to_be_aligned_elev,
-                title="Cropped reference DEM",
-                filename="cropped_reference_elev_map",
+                title="Preprocessed to-be-aligned DEM",
```
belletva: DEM --> elev

xdem/workflows/accuracy.py
```diff
@@ -195,21 +213,26 @@ def _prepare_datas(self) -> None:
...
             )
         else:
-            self.to_be_aligned_elev = self.to_be_aligned_elev.crop(coord_intersection)
+            self.reference_elev = self.reference_elev.crop(coord_intersection)
             self.generate_plot(
-                self.to_be_aligned_elev,
-                title="Cropped to-be-aligned DEM",
-                filename="cropped_to_be_aligned_elev_map",
+                self.reference_elev,
+                title="Preprocessed reference DEM",
```
belletva: DEM --> elev

doc/source/_workflows/accuracy_config.yaml
belletva: Template config should be homogenized with LZD

xdem/workflows/accuracy.py
belletva, line 1: Replace all DEM by elev to get homogeneity with file's names.

xdem/workflows/accuracy.py
```diff
@@ -321,38 +344,46 @@ def run(self) -> None:
...
-                self.diff = diff
-                vmin, vmax = -(stats["median"] + 3 * stats["nmad"]), (stats["median"] + 3 * stats["nmad"])
+        stats_keys = ["min", "max", "nmad", "median"]
 
+        def generate_plot_diff(label: str, diff: RasterType, vmin: float, vmax: float) -> None:
             suffix = f"_elev_{label}_coreg_map" if label else "_elev"
             self.generate_plot(
                 diff,
-                title=f"Difference\n{label} coregistration",
+                title=f"Difference {label} coregistration",
```
belletva: Elevation difference

rhugonnet:
@marinebcht For displaying the masks: Perfect! It's a shame that the "terrain" colormap has white in it, because we can confuse nodata or the edge of the colormap. We could maybe cut the colorbar, or choose another one... 
What do you think?

rhugonnet:
As you prefer, both work well I think! :slightly_smiling_face: 
(gist_earth + white, or terrain + black)

rhugonnet:
Perfect, thanks! :smile:  

I don't have much to re-review.
My only two comments: 
- It would be nice to get a RTD link (following the Wiki) to see the new doc rendered, easier to read + get a sense of the changes + potentially help spot typos/issues.
- Should we define default statistics? The full list is a bit long, we already discussed this... Maybe simply Min/Max/Mean/Median/STD/NMAD and Valid/Total/Percent counts is already plenty.

PR #911, opened 2026-02-12
Add every affine translation coregistration method in blockwise

### LINKED ISSUE #910: Test evry coregistration method for blockwise
# Context

The purpose of this ticket is to certify the possibility of using all co-registration methods in blockwise.

1. Test affine methods
2. Update the code to accept all of them
3. Find out about non-affine methods
4. Take the necessary steps

adebardo:
Resolves #910 

# Context

The purpose of PR is to implement all coregistration methods for blockwise. In previous PRs, we only worked with NuthKaab.

# Code 

We now cover all **Affine** methods. To do this, we had to handle the case where only a vertical shift is returned. For this, we use a default value of 0. 

Question : Should we keep 0, or use np.nan in cases where the method does not return any shifts in the output dictionary? (lines 204, 205, 206) 

For non-affine methods that I do not think are compatible with our vision of apply, I return an error to the user. But perhaps there are improvements to blockwise that could be considered, and a ticket could be opened.

# Tests
I have added parameters to pass all coregistration methods and verify that we are getting the desired output.

### DIFF
--- a/doc/source/coregistration.md
+++ b/doc/source/coregistration.md
@@ -577,6 +577,7 @@ for {class}`xdem.coreg.DirectionalBias`, an input `angle` to define the angle at
 ```{caution}
 The {class}`~xdem.coreg.BlockwiseCoreg` feature is still experimental. Currently, it is only tested for the affine
 coregistration methods and for translation only. Rotation shifts will be soon implemented, but not yet available.
+Please set `only_translation=True` when initializing the coregistration method ICP(), LZD() or CPD().
 ```
 
 A {class}`~xdem.coreg.BlockwiseCoreg` splits a coregistration across different spatial blocks of an elevation dataset, running that
--- a/tests/test_coreg/test_blockwise.py
+++ b/tests/test_coreg/test_blockwise.py
@@ -226,6 +226,33 @@ class TestBlockwiseCoreg:
         ):
             _ = xdem.coreg.BlockwiseCoreg(step=step_coreg, mp_config=config_mc, block_size_fit=block_size)
 
+    @pytest.mark.parametrize(
+        "step_coreg",
+        [
+            pytest.param(xdem.coreg.CPD(only_translation=False), id="CPD"),
+            pytest.param(xdem.coreg.ICP(only_translation=False), id="ICP"),
+            pytest.param(xdem.coreg.LZD(only_translation=False), id="LZD"),
+            pytest.param(xdem.coreg.CPD(), id="CPD"),
+            pytest.param(xdem.coreg.ICP(), id="ICP"),
+            pytest.param(xdem.coreg.LZD(), id="LZD"),
+        ],
+    )
+    @pytest.mark.parametrize("block_size", [32])
+    def test_blockwise_only_translation_coreg_pipeline(self, step_coreg, tmp_path, block_size):
+        """
+        Test end-to-end blockwise coregistration for non-affine steps and validate output.
+        """
+
+        config_mc = MultiprocConfig(
+            chunk_size=block_size, outfile=tmp_path / "test.tif", cluster=ClusterGenerator("multi", nb_workers=4)
+        )
+        with pytest.raises(
+            ValueError,
+            match="The provided coregistration method is configured to only estimate translation. "
+            "Consider setting 'only_translation' to True to allow for more complex transformations.",
+        ):
+            _ = xdem.coreg.BlockwiseCoreg(step=step_coreg, mp_config=config_mc, block_size_fit=block_size)
+
     @pytest.mark.parametrize("block_size", [32])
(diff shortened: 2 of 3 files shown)


doc/source/coregistration.md
```diff
@@ -575,7 +575,7 @@ for {class}`xdem.coreg.DirectionalBias`, an input `angle` to define the angle at
 ### The {class}`~xdem.coreg.BlockwiseCoreg` object
 
 ```{caution}
-The {class}`~xdem.coreg.BlockwiseCoreg` feature is still experimental. Currently, it is only tested for the Nuth and Kaab method.
+The {class}`~xdem.coreg.BlockwiseCoreg` feature is still experimental. Currently, it is only tested for the affine method.
```
belletva: affine coregistration methods

tests/test_coreg/test_blockwise.py
```diff
@@ -179,20 +191,40 @@ def test_blockwise_coreg_pipeline(self, step, example_data, tmp_path, block_size
...
+    @pytest.mark.parametrize("block_size", [32])
+    def test_blockwise_not_affine_coreg_pipeline(self, step_coreg, tmp_path, block_size):
+        """
+        Test end-to-end blockwise coregistration for non-affine steps and validate output.
+        """
+
+        config_mc = MultiprocConfig(
+            chunk_size=block_size, outfile=tmp_path / "test.tif", cluster=ClusterGenerator("multi", nb_workers=4)
+        )
+        with pytest.raises(ValueError, match="The blockwise coregistration only supports affine coregistration steps."):
```
belletva: affine coregistration methods

xdem/coreg/blockwise.py
```diff
@@ -81,6 +81,8 @@ def __init__(
             raise ValueError(
                 "The 'step' argument must be an instantiated Coreg subclass. " "Hint: write e.g. ICP() instead of ICP"
             )
+        if not step.is_affine:
```
belletva, line 84: affine coregistration methods

xdem/coreg/blockwise.py
```diff
@@ -81,6 +81,8 @@ def __init__(
...
+            raise ValueError("The blockwise coregistration only supports affine coregistration steps.")
```
belletva: affine coregistration methods

rhugonnet:
Great @adebardo!

> Question : Should we keep 0, or use np.nan in cases where the method does not return any shifts in the output dictionary? (lines 204, 205, 206)

I think we really need to use `np.nan`, otherwise the aggregation method (2d plan RANSAC, or other: could be an average of all) is **biased** by the zero values.

**Another major comment:**
Other methods than Nuth and Kääb perform a rotation as well (while right now `BlockwiseCoreg` only uses the `shift_x/shift_y/shift_z` outputs. We need to adjust the `apply()` of blockwise to work on this generic case. We can discuss in more details at the next meeting. :wink:

doc/source/coregistration.md
```diff
@@ -575,7 +575,8 @@ for {class}`xdem.coreg.DirectionalBias`, an input `angle` to define the angle at
 ### The {class}`~xdem.coreg.BlockwiseCoreg` object
 
 ```{caution}
-The {class}`~xdem.coreg.BlockwiseCoreg` feature is still experimental. Currently, it is only tested for the Nuth and Kaab method.
+The {class}`~xdem.coreg.BlockwiseCoreg` feature is still experimental. Currently, it is only tested for the affine
+coregistration methods and for translation only. Rotation shifts will be soon implemented, but not yet available.
```
belletva, line 579: with this PR and this code: 
```
mp_config = MultiprocConfig(chunk_size=1000, cluster=ClusterGenerator("multi", nb_workers=2))
coreg_method = xdem.coreg.ICP()
tile_size = 1000
blockwise = xdem.coreg.BlockwiseCoreg(
coreg_method,
block_size_fit=tile_size,
mp_config=mp_config)
blockwise.fit(reference_dem, dem_to_be_aligned)
aligned_dem_blockwise = blockwise.apply()
print(coreg_method.meta)
```
I got:
```
{'inputs': {'random': {'subsample': 500000.0}, 'fitorbin': {'fit_minimizer': <function least_squares at 0x14ac21df5e10>, 'fit_loss_func': 'linear'}, 'iterative': {'max_iterations': 20, 'tolerance': 0.01}, 'specific': {'icp_method': 'point-to-plane', 'icp_picky': True}, 'affine': {'only_translation': False, 'standardize': True}}, 'outputs': {}}
```

`'only_translation'` should be at `True `

I got the same for `LZD()` and `CPD()`

rhugonnet: Maybe we just raise a warning if the user passes a `Coreg` method that contains `only_translation=False` or is a ` BiasCorr` method? (But those will be solved soon no, @adebardo? In that case no need to spend too much time on this)

PR #767, opened 2025-08-04
Command line interface V0

adebardo:

PR #948, opened 2026-04-09
Enhancements CLI

### LINKED ISSUE #854: Enhancements for workflows and CLI interface
While editing the documentation (and slightly the workflows) in #852, I noted some things we can improve:
- The `topo` workflow expects a `reference_elev` subsection between `inputs` and `path_to_elev`, which is unexpected (and not really documented) because the workflow works on a single elevation input. We can probably remove, except if we intend to add support for multiple elevation data in the future?
- Having `to_vcrs` as input in `topo`/`accuracy` seems unnecessary: The `sampling_grid` argument already specifies which grid/CRS should be used when comparing both (reference or to-be-aligned). The vertical CRS being part of the CRS, it's transformed automatically during that step, so it never needs to be transformed during input reading with `to_vcrs`. The user should only be able to set or override the vertical CRS of the input data (`set_vcrs`).
- The `accuracy` workflow detailed description has two sub-tabs for `reference/tba_elev` within the tab `inputs`, but the description are exactly the same word-for-word for each. We could simplify by having those as a single page to avoid sub-tabs.

Some thoughts of global parameters that could be useful for users to access:
- The `resampling` during reprojection, or during point-grid interpolation (bilinear, cubic, etc). We could use the [global GeoUtils parameters](https://geoutils.readthedocs.io/en/stable/config.html) for this.
- The `nodata_propagation` behaviour during reprojection/interpolation.

And thoughts on the HTML reports:
- The plots sometimes occupy 50% or 100% of a page, making the fontsize vary a lot (readable to unreadable...), depending on the figure width. Should we simply make double-width figures (twice 50%) directly into Matplotlib (vertical subplots), instead of adding 2 plots to the HTML, to avoid having varying-size fonts across a report?

### LINKED ISSUE #864: Add direct tests for CLI
Working on #863, I notice that we only have Python tests for Workflows right now, no direct CLI tests.

We should inspire ourselves from the CLI tests in GeoUtils here: https://github.com/GlacioHack/geoutils/blob/main/tests/test_geoviewer.py and add a `test_cli.py` test suite.

Among others, this would allow to check that the users do get the right error message when optional dependencies are missing (what I'm working on), and that the CLI does raise standard errors (what I mentioned in #858).

### LINKED ISSUE #921: Adapt CLI to run topo for several inputs
Add several input informaton for several dem in the topo config file and adapt de workflow. 

Note: adapt `generate_plot` too

### LINKED ISSUE #949: Fractal roughness error in topo workflow
Hello,

I noticed a small bug in the topo workflow.

If the .yaml file is configured with these settings
```
terrain_attributes:
-fractal_roughness
```

The error:
`KeyError: 'fractal_roughness'`

The correction:

In `topo.py` `generate_terrain_attributes_png` `attribute_params`, the key `fractal_dimension` (line 151) is incorrect, it should be named `fractal_roughness`.

marinebcht:
### 1. Inputs config in topo workflow

Resolves #854 (point 1)
Resolves #921  

Now, we can run several Topo Workflows for several DEMs :

```
inputs:
  - path_to_elev: "path_dem_1"
    [other_param: xxx]
  - path_to_elev: "path_dem_2"
    [other_param: xxx]
```

Corresponding tree of outputs for level 1:

```
output_path (from outputs/path in the yml)
  ├─ dem_[X]  (if several inputs)
  │   ├─ tables ...
  │   ├─ plots ...
  │   ├─ report.html
  │   ├─ report.pdf
  └─ used_config.yaml

```

If only one dem needed to be study, the input section can look like this :

```
inputs:
    path_to_elev: "path_dem_1"
    [other_param: xxx]
```

And the outputs tables/plots(/raster)  are now at the root directory (from outputs/path in the yml) as before.

### 2.A VCRS parameters simplification 

Resolves #854 (point 2)

Currently, input elev have two parameters :
image

As VCRS becomes part of the CRS, the DEM can have already a VCRS. So, we can simplify to only one optional "force_vcrs" :
- if the actual CRS elev has no vcrs: set it 
- else: replace it

```
inputs:
    path_to_elev: "path_dem_1"
    force_vcrs: XXXX
```

The possibility to offer this possibility for the CRS: CRS are mostly present and correct, unlike VCRS, which can be absent from the metadata or erroneous => NO add

### 2.B Attributes computaion in case of geographic CRS 

For the topo workflow, as not projected elev can lead to wrong attributes outputs, we propose the following keys : 

a.  no reproject key, or `{reproject: None}`, or `{reproject: {crs: None}}` => reproject the elev to the corresponting CRS UTM (.get_metric_crs())

```
[reproject:]
   [crs:]
```

b. `{reproject: {crs: None}}` => no reprojection => warning `As the CRS is a geographic CRS, the following surface fit attributes might be wrong. Please use a projected CRS or let it empty to reproject in default projected CRS.`

```
reproject:
   crs: False
```

c.  If we wan't to customize the projection : 
    - if CRS is projected => projection before atributes computation (and only!) 
          - CRS projected covering the area => OK
          - Need to raises an error (GDAL EXCEPTION = > issues https://github.com/GlacioHack/xdem/issues/955)
   - if CRS is not projected => warning `As the CRS is a geographic CRS  and the "reproject/crs" either, the following surface fit attributes might be wrong.`

```
reproject:
    crs: xxx
```

In two words :
- the reprojection is use ONLY if the initial proj is geographic and if the attributes need as input projection
- the default behaviour = reproject when needed to the corresponding UTM
- no reproject (leading to possible errors) if `{reproject: {crs: None}}`
- personal choice projection `{reproject: {crs: [CRS]}}`

If level > 1 and a reprojection is done, "elev_reprojected.tif" is saved in rasters
/!\ NEED TO USE THE NEW GEOUTIL WITH XARRAY AND VCRS IN CRS to complete validation https://github.com/GlacioHack/xdem/issues/956

### 3. Add direct CLI tests (arguments + workflow) in test_cli.py 

Resolves #864 (point 1)

- test_invalid_parameters
- test_raises_help (xdem -h, xdem --help, xdem)
- test_missing_param_after_workflow (xdem topo)
- test topo and accuracy workflow
- test default config (print and file)
- test config file errors (does not exist, not .yml or .yaml, error in it)

### 4. Fixes other CLI problems/little change in it

In topo workflows:
- if attributes : None, no error anymore
- attributes are presented in columns of 3 if greater than 6
- in the config, if extra parameters needed (dict version), the   `extra_information` key is not more needed
- the attribtues computation is now call once (before: once in png generation function and one in raster function)
- `fractal_roughness` attibutes was misscall `fractal_dimension` Resolve https://github.com/GlacioHack/xdem/issues/949

- 
In accuracy : 
- remove `output_grid` key ( https://github.com/GlacioHack/xdem/pull/948#issuecomment-4369509160)
  =>  output_grid = reference_elev as default behaviour
  => new map aligned_dem - tba when level > 1 

image

For both workflows :
-  `create_html` and `generate_pdf` functions  moved to [workflow] as we call it for each input in topo, same for `creation_output_dir` function and `_load_data` takes an input dict.
- the `_validate_crs` call for previous `set_vcrs` and `to_vcrs`, renamed `_validate_vcrs` for `force_vcrs` and `_validate_crs` validated  `CRS.from_user_input(value)` (@ topo key `reproject/crs`)
- Add outputs/generate_pdf (boolean) to deactivate this part (error in the cluster)

(*) Added `test_attributes_default_call` in `test_terrain.py` to check if default parameters in `get_terrain_attribute()` and in `[attribute]()` are the same (and lead to same results for each terrain attributes).

### 5. Documentation

- [x] Change cli_topo for (1) 
- [x] Change cli_topo + cli_accuracy for (2.A) and (generate_pdf in 4)
- [x] Change cli_topo for (2.B)
- [x]  Remove double sub sections for reference/tba_elev in cli_accuracy Resolves #854 (point 3)

### 6. Propositions for the now/future

- [ ] Waiting VCRS in CRS possibility => complete reprojection (CRS & VCRS) https://github.com/GlacioHack/xdem/issues/956
- [ ] Test when mask has differents vcrs/crs than input elev https://github.com/GlacioHack/xdem/issues/956 + https://github.com/GlacioHack/xdem/issues/954 
- [ ] -> Test/simplify _prepare_datas in accuracy workflow https://github.com/GlacioHack/xdem/issues/956
- [ ] Do coreg test in test_accuracy.py (skipped here)
- [ ] Optimize terrain attributes lists an functions  `[attributes]()` (terrain.py/topo workflow)
- [ ] Raise error when  `reproject/crs` doest not cover the input dem https://github.com/GlacioHack/xdem/issues/955 
- [ ] See if dep opt  errors are catch https://github.com/GlacioHack/xdem/issues/864 => https://github.com/GlacioHack/xdem/issues/955
- [ ] Add resample and nodata_propagation as global param https://github.com/GlacioHack/xdem/issues/854 => https://github.com/GlacioHack/geoutils/issues/884


marinebcht:
This PR is almost ready (question point 2 + its doc) :) 
@belletva @rhugonnet @adehecq 

Maybe another proposition for the point 2 (with all the dev simplification to do): 
image

Furthermore, the `set/from_vcrs` does not appear to persist after "data preparation", but this will be automatically fixed when xDem will be merged to use the brand new Geoutils and its VCRS-part-of-the-CRS functionality :partying_face:

marinebcht:
Output example for two inputs @ topo workflow : 

image

doc/source/cli_topo.md
```diff
@@ -134,8 +134,11 @@ Please refer to {ref}`data-example` to have more information.
 
 :::{code-block} yaml
 inputs:
-  reference_elev:
-    path_to_elev: "path_to/reference_elev.tif"
+  - path_to_elev: "path_to/one_elev.tif"
+    force_source_nodata: -32768
+    from_vcrs: null
+    to_vcrs: null
+  - path_to_elev: "path_to/another_elev.tif"
```
belletva, line 141: first_elev.tif and second_elev.tif

doc/source/cli_topo.md
```diff
@@ -180,8 +183,8 @@ If a mask is provided, the statistics are also computed inside the mask.
 List or set of dictionaries for extra information.
 
 :::{note}
-- If no block is specified, slope, aspect, and curvature attributes are calculated by default.
-- If a block is specified but no information is provided, then no attributes will be calculated.
+- If the block is specified, slope, aspect, and curvature attributes are calculated by default.
+- If the block is specified but no information is provided, then no attributes will be calculated.
```
belletva, line 187: quite difficult to understand

doc/source/cli_topo.md
```diff
@@ -134,8 +134,11 @@ Please refer to {ref}`data-example` to have more information.
 
 :::{code-block} yaml
 inputs:
-  reference_elev:
-    path_to_elev: "path_to/reference_elev.tif"
+  - path_to_elev: "path_to/one_elev.tif"
+    force_source_nodata: -32768
+    from_vcrs: null
```
belletva, line 139: add path_to_mask

doc/source/cli_topo.md
```diff
@@ -237,41 +239,42 @@ outputs:
...
-  ├─ report.html
-  ├─ report.pdf
+path_to/outputs
+  ├─ dem_[X]  (if several inputs)
+  │   ├─ tables
+  │   │   ├─ elev_stats.csv
+  │   │   └─ elev_with_mask_stats.csv (if mask_elev is given in input)
+  │   ├─ plots
+  │   │   ├─ elev_map.png
+  │   │   ├─ masked_elev_map.png (if mask_elev is given in input)
```
belletva, line 249: add [ ] for optional files

doc/source/cli_topo.md
```diff
@@ -237,41 +239,42 @@ outputs:
...
-  │   ├─ slope.tif
-  │   ├─ terrain_ruggedness_index.tif
-  │   └─ ...
-  ├─ report.html
-  ├─ report.pdf
+path_to/outputs
+  ├─ dem_[X]  (if several inputs)
+  │   ├─ tables
+  │   │   ├─ elev_stats.csv
+  │   │   └─ elev_with_mask_stats.csv (if mask_elev is given in input)
```
belletva, line 263: add [ ] for optional files

rhugonnet:
Thanks a lot @marinebcht! This is amazing (especially the direct CLI tests :stuck_out_tongue_winking_eye:)

I can't review fully today, but my quick reaction on `vcrs` management:

This is not specific to vertical CRS, but CRS in general, and there's two questions mixed in here:

1/ Do we provide support for edge cases where a user has a badly defined input, with potentially no CRS/VCRS written in the metadata, or not? Currently we don't, except for the VCRS. This is not an issue in Python, where the user can set it manually beforehand, but I can see how this could be limiting in the CLI.

2/ Do we allow users to reproject through xDEM for workflows on single-source data like `Topo`? (`Accuracy` is relative change, so not an issue there). This is actually quite relevant, as some terrain attributes are nonsensical in some common coordinate systems (geographic), and need to be projected to local UTM/other beforehand. (We could even have a default that reprojects/derives/reprojects back for the geographic case)

I think the way forward could be simply:

1/ We always accept optional inputs `force_crs` and `force_vcrs` that override the DEM `crs/vcrs`, in `inputs` (or we can keep `from_vcrs/crs` naming too),
2/ We add an optional processing step `reprojection` where the user can provide a `crs`/`vcrs`, or even a string `"utm"` (which would wrap `.get_metric_crs()`), so that it's clear in the workflow that this step is done. I don't think it works in `outputs` because this happens ahead in the processing chain.

In this context, we won't need `to_vcrs` anymore (only `from/force_crs/vcrs`), as this is covered in the `reprojection` step for the whole 3D CRS.

marinebcht:
Hello @rhugonnet :) Thank you for all of your comments ! 

Several questions/proposition : 
- Is that pretty common to can CRS error / missing CRS ? I thought that was just the case on purpose (like sensor grid) 😥 
In this context of wrong metadata : if we have a dem with (V)RS A, and the user give an (V)CRS B, it means that we have to trust hit and force elev to (V)CRS B with a warning ?
- I don't quite understand this sentence "We could even have a default that reprojects/derives/reprojects back for the geographic case" ? 🤔
- In the topo workflow, I feel like the output reprojection will be more accurate in new section (same for all inputs =  one purpose/process by .yml) than in each inputs

rhugonnet:
@marinebcht Thanks! :slightly_smiling_face: 

Some quick answers:

- **CRS missing:** It is quite uncommon, especially nowadays, but I've seen it (usually messed up by the user through some software, often the nodata too). Vertical CRS is the most frequent one.
- Reprojecting back: If the CRS is geographic, most attributes fail (unsupported), or give erroneous units. We could have a default behaviour that detects (if EPSG=4326 or same family), projects the DEM to the local utm (`get_metric_crs`), derives the attributes (right units and value), then projects back to the original EPSG (simple resampling, values/units stay valid).
- Last point: Not sure I understand. You mean the new "reprojection" section I propose would be clearer? (I liked the idea of a new section, but I think "outputs" that was proposed doesn't work in terms of processing order and logic of other CLI steps)

marinebcht:
@belletva @rhugonnet @adehecq 
ready to review (modulo doc point 2.B) :) 

Added:
+ complete new prop for point 2 after discussions with @belletva and @rhugonnet (can be modify!) (/!\ need to check how to handle a CRS in to_crs if it is not covering the elev zone)
+ small fixes point 4 (with @ould-a issue #949)

doc/source/cli_topo.md
```diff
@@ -180,8 +185,8 @@ If a mask is provided, the statistics are also computed inside the mask.
 List or set of dictionaries for extra information.
 
 :::{note}
-- If no block is specified, slope, aspect, and curvature attributes are calculated by default.
-- If a block is specified but no information is provided, then no attributes will be calculated.
+- If no block is specified: slope, aspect, and curvature attributes are calculated by default.
+- If the block is specified but but empty, no attributes will be calculated.
```
belletva: twice "but"

doc/source/cli_topo.md
```diff
@@ -216,16 +220,16 @@ Its use is entirely the responsibility of the user.
 
 Outputs information. Operates by levels:
 
-1. Level 1 → aligned elevation only
-2. Level 2 → more detailed output
+1. Level 1 → save reports and stats and plots used in it
```
belletva: save reports (HTML and PDF formats) in parent directory, stats (CSV formats)  in `tables` directory and plots (PNG formats) in `plots` directory

doc/source/cli_topo.md
```diff
@@ -216,16 +220,16 @@ Its use is entirely the responsibility of the user.
...
+2. Level 2 → save terrain attributes in addition (TIF format)
```
belletva: save raster terrain attributes (TIFF formats) in `rasters` directory

doc/source/cli_topo.md
```diff
@@ -237,41 +241,42 @@ outputs:
...
-  │   ├─ masked_elev_map.png (if mask_elev is given in input)
-  │   └─ terrain_attributes_map.png
-  ├─ rasters
-  ├─ report.html
-  ├─ report.pdf
+path_to/outputs
+  ├─ [dem_X]  (if several inputs)
+  │   ├─ tables
+  │   │   ├─ elev_stats.csv
+  │   │   └─ elev_with_mask_stats.csv (if mask_elev is given in input)
```
belletva: path_to_mask no instead of mask_elev?

doc/source/cli_topo.md
```diff
@@ -237,41 +241,42 @@ outputs:
...
-  │   ├─ elev_stats.csv
-  │   └─ elev_with_mask_stats.csv
-  ├─ plots
-  │   ├─ elev_map.png
-  │   ├─ masked_elev_map.png (if mask_elev is given in input)
-  │   └─ terrain_attributes_map.png
-  ├─ rasters
-  ├─ report.html
-  ├─ report.pdf
+path_to/outputs
```
belletva, line 265: caution: `rasters` folder is still created in level 1 and is empty

xdem/workflows/topo.py
```diff
@@ -186,62 +156,155 @@ def generate_terrain_attributes_png(self) -> None:
...
+                if CRS.from_user_input(self.config["reproject"]["to_crs"]).is_geographic:
+                    warnings.warn(
+                        'As the input dem is not in a projected CRS and the "reproject/to_crs" either,'
+                        "the following surface fit attributes might be wrong.",
+                        UserWarning,
+                    )
+                self.dem = self.dem.reproject(crs=self.config["reproject"]["to_crs"])
+
+        dem_proj_utm = self.dem
+        logging.info("Computing reprojection in needed")
```
belletva: tipo

doc/source/cli_topo.md
marinebcht: it is removed at the end if empty => I will add a "if" at the folder creation,
good catch

xdem/workflows/topo.py
```diff
@@ -186,62 +156,155 @@ def generate_terrain_attributes_png(self) -> None:
...
+            self.dico_to_show[k].append(("Elevation information", dem_informations))
+
+            # Statistics
+            list_metrics = self.config["statistics"]
+            if list_metrics is not None:
+                stats_dem = self.dem.get_stats(list_metrics)
+                stats_dem = {_ALIAS.get(k, k): v for k, v in stats_dem.items()}
+                self.save_stat_as_csv(stats_dem, "stats_elev")
+                self.dico_to_show[k].append(("Statistics", self.floats_process(stats_dem)))
+                logging.info(f"Computing metrics on reference elevation: {list_metrics}")
```
belletva: delete "reference"

xdem/workflows/topo.py
```diff
@@ -186,62 +156,155 @@ def generate_terrain_attributes_png(self) -> None:
...
-        t1 = time.time()
-        self.elapsed = t1 - t0
+        # Generate terrain attributes png
+        self.generate_terrain_attributes_png(attributes)
 
-        self.create_html(self.dico_to_show)
+        # Generate terrain attributes tif
+        if export_tif:
+            for k, attr in enumerate(self.list_attributes):
+                logging.info(f"Saving {attr} as a raster file ({attr}.tif)")
```
belletva: `logging.info(f"Saving {attr} as a raster file (rasters/{attr}.tif)")`

doc/source/cli_accuracy.md
```diff
@@ -297,50 +296,60 @@ outputs:
...
   ├─ tables
-  │   └─ aligned_elev_stats.csv
-  ├─ plots
-  │   ├─ diff_elev_after_coreg_map.png
-  │   ├─ diff_elev_before_coreg_map.png
-  │   ├─ diff_elev_before_after_hist.png
-  │   ├─ reference_elev_map.png
-  │   ├─ masked_elev_map.png (if mask_elev is given in input)
-  │   └─ to_be_aligned_elev_map.png
+  │   ├─ aligned_elev_stats.csv
```
belletva: [ ] (if coregistration)

doc/source/cli_accuracy.md
```diff
@@ -297,50 +296,60 @@ outputs:
...
+  │   ├─ [diff_elev_before_coreg_stats.csv] (if compute_coreg)
```
belletva: what is compute_coreg ? it's coregistration: process: True

doc/source/cli_accuracy.md
```diff
@@ -297,50 +296,60 @@ outputs:
...
   └─ used_config.yaml
 :::
 
-Tree of outputs for level 2 (including coregistration step):
+Tree of outputs for level 2:
 
 :::{code-block} text
 - root
   ├─ tables
   │   ├─ aligned_elev_stats.csv
```
belletva: if coregistration

doc/source/cli_accuracy.md
```diff
@@ -297,50 +296,60 @@ outputs:
...
   │   ├─ reference_elev_stats.csv
   │   └─ to_be_aligned_elev_stats.csv
   ├─ plots
-  │   ├─ diff_elev_after_coreg_map.png
-  │   ├─ diff_elev_before_coreg_map.png
-  │   ├─ diff_elev_before_after_hist.png
-  │   ├─ reference_elev_map.png
-  │   ├─ masked_elev_map.png (if mask_elev is given in input)
-  │   └─ to_be_aligned_elev_map.png
+  │   ├─ inputs.png
```
belletva, line 335: what is inputs.png?

doc/source/cli_accuracy.md
```diff
@@ -297,50 +296,60 @@ outputs:
...
+  │   ├─ [diff_elev_after_coreg_map.png] (if compute_coreg)
+  │   ├─ [diff_elev_before_after_hist.png] (if compute_coreg)
+  │   ├─ [diff_elev_without_coreg_map.png] (if no compute_coreg)
+  │   └─ [elev_diff_histo.png] (if compute_coreg))
   ├─ rasters
-  │   ├─ aligned_elev.tif
-  │   ├─ diff_elev_after_coreg.tif
-  │   ├─ diff_elev_before_coreg.tif
-  │   └─ to_be_aligned_elev_reprojected.tif
+  │   ├─ reference_elev_reprojected.tif
```
belletva: always computed?

doc/source/cli_accuracy.md
```diff
@@ -297,50 +296,60 @@ outputs:
...
+  │   ├─ inputs.png
+  │   ├─ [masked_elev_map.png] (if `path_to_mask` is given in input)
+  │   ├─ [preprocessed_to_be_aligned_elev_map.png or preprocessed_reference_elev_map.png] (if sampling_grid)
+  │   ├─ [diff_elev_before_coreg_map.png] (if compute_coreg)
+  │   ├─ [diff_elev_after_coreg_map.png] (if compute_coreg)
+  │   ├─ [diff_elev_before_after_hist.png] (if compute_coreg)
+  │   ├─ [diff_elev_without_coreg_map.png] (if no compute_coreg)
+  │   └─ [elev_diff_histo.png] (if compute_coreg))
   ├─ rasters
   │   └─ aligned_elev.tif
```
belletva: if coregistration

doc/source/cli_accuracy.md
```diff
@@ -297,50 +296,60 @@ outputs:
...
+  │   ├─ aligned_elev_stats.csv
+  │   ├─ [diff_elev_before_coreg_stats.csv] (if compute_coreg)
+  │   ├─ [diff_elev_after_coreg_stats.csv] (if compute_coreg)
+  │   ├─ reference_elev_stats.csv
+  │   └─ to_be_aligned_elev_stats.csv
+- ├─ plots
+  │   ├─ inputs.png
+  │   ├─ [masked_elev_map.png] (if `path_to_mask` is given in input)
+  │   ├─ [preprocessed_to_be_aligned_elev_map.png or preprocessed_reference_elev_map.png] (if sampling_grid)
+  │   ├─ [diff_elev_before_coreg_map.png] (if compute_coreg)
```
belletva: we should keep the diff even if the coregistration is not computed

doc/source/cli_accuracy.md
```diff
@@ -297,50 +296,60 @@ outputs:
...
-  │   ├─ diff_elev_after_coreg_map.png
-  │   ├─ diff_elev_before_coreg_map.png
-  │   ├─ diff_elev_before_after_hist.png
-  │   ├─ reference_elev_map.png
-  │   ├─ masked_elev_map.png (if mask_elev is given in input)
-  │   └─ to_be_aligned_elev_map.png
+  │   ├─ inputs.png
+  │   ├─ [masked_elev_map.png] (if `path_to_mask` is given in input)
+  │   ├─ [preprocessed_to_be_aligned_elev_map.png or preprocessed_reference_elev_map.png] (if sampling_grid)
+  │   ├─ [diff_elev_before_coreg_map.png] (if compute_coreg)
```
belletva: same here we should keep the diff even if the coregistration is not computed

doc/source/cli_accuracy.md
belletva, line 154: it needs an explanation in the doc for the difference between `sampling_grid` and `output_grid`

doc/source/cli_accuracy.md
belletva: if grid_sampling

doc/source/cli_accuracy.md
```diff
@@ -129,13 +129,12 @@ Inputs information, split between reference and to-be-aligned elevation data.
 | `path_to_elev`        | Path to reference elevation              | str        |         | Yes      |
 | `force_source_nodata` | No data elevation                        | int        |         | No       |
 | `path_to_mask`        | Path to mask associated to the elevation | str        |         | No       |
-| `from_vcrs`           | Original vcrs                            | int, str   | `null`  | No       |
-| `to_vcrs`             | Destination vcrs                         | int, str   | `null`  | No       |
+| `force_vcrs`          | Vertical CRS of the elevation            | int, str   |         | No       |
```
rhugonnet: No need for "of the elevation" here I think, just "Vertical CRS"

doc/source/cli_accuracy.md
```diff
@@ -129,13 +129,12 @@ Inputs information, split between reference and to-be-aligned elevation data.
...
 | `downsample`          | Downsampling elevation factor >= 1       | int, float | 1       | No       |
 :::
 
 :::{note}
-For transforming between vertical CRS with ``from_vcrs``/``to_vcrs`` please refer to {ref}`vertical-ref`.
+To set the vertical CRS with ``force_vcrs``, please refer to {ref}`vertical-ref`.
```
rhugonnet: Maybe an additional note: "Overrides a vertical CRS that might exist in the metadata".

marinebcht:
Hello, I took some time to understand the use of the param `output_grid` in `outputs`.
Here's what I saw, depending on its value, the value of `sampling_grid`  and if the coreg is activated or not.

https://docs.google.com/spreadsheets/d/109xGCD0ILPQTy0bbZuMFch2qr7C-ZdZHpkOkSOmBe-8/edit?usp=sharing

When `output_grid` when equals to `to_be_aligned_elev` gives `diff_before` and diff_elev null. The only purpose can be to compute the difference between `to_be_aligned_elev` before and after coreg (if coreg!) => to see the error of an elev relative to the reference it should coregister.

To me, the main accuracy metric is really the difference between `to_be_aligned_elev` (before/after/like that) and the `reference_elev`.  So, maybe we can "fix" `output_grid` to `reference_elev` (so remove the key from the config file) and add a new key dedicated to compute the diff before/after `to_be_aligned_elev` (if coreg) (if useful ?). 

Corresponding PR: https://github.com/GlacioHack/xdem/pull/767

belletva (dismissed):
For me this PR is OK. We need to open a new issue concerning the `output_grid` and `sampling_grid`.

rhugonnet:
Good for me too. 

On `output_grid`: I don't understand the explanation on the purpose, sorry :sweat_smile:. 
If it's useless, we scrap it!

marinebcht:
Hello @belletva 
New titles for the plots in the accuracy workflow with coreg and level = 2 : 
[report_acc_11_05.pdf](https://github.com/user-attachments/files/27587107/report_acc_11_05.pdf)

Add the (visual) doc for the reprojection parameters in the topo workflow : 
image

marinebcht:
Little note after  https://github.com/GlacioHack/xdem/pull/948/commits/f9c6d7cc0377b8d8e7d422d9ce23309a1db34379
When we use multi process to compute several attributes, ex:
`mp_config = MultiprocConfig(chunk_size=50, outfile="tmp_mp_output.tif")`
The outputs files are  "tmp_mp_output_[attributes].tif"

marinebcht:
Final note : 
- added "run: conda install -c conda-forge weasyprint" [here](https://github.com/marinebcht/xdem/blob/35035575b7ee03fcb6e9b25e689e493b59978483/.github/workflows/python-tests.yml#L75)  to have the dependencies like pango (no matter the os) in the environment 
- update `matplotlib.cm.get_cmap(cmap)` to `matplotlib.pyplot.get_cmap(cmap)` [here](https://github.com/marinebcht/xdem/blob/35035575b7ee03fcb6e9b25e689e493b59978483/xdem/spatialstats.py#L3559), seems to be depecriated

belletva (approved):
Good for me 👍 Good job!

PR #970, opened 2026-06-17
Add possibility to have a raster path_to_mask in Workflows config file

### LINKED ISSUE #954: Raster masks not supported in workflows
I noticed a small bug in the workflows.

Here's the context :
In the Accuracy workflow, if a raster mask is provided in the configuration file, as shown here 
```
inputs:
  reference_elev:
    path_to_elev: "dems/longyearbyen/longyearbyen_ref_dem.tif"
    path_to_mask: "dems/longyearbyen/inlier_mask.tif"
  to_be_aligned_elev:
    path_to_elev: "dems/longyearbyen/longyearbyen_tba_dem.tif"
```
The following error appears:
`pyogrio.errors.DataSourceError: 'dems/longyearbyen/inlier_mask.tif' not recognized as being in a supported file format.`

Raster masks don't work, only vector files do. 
In `workflow.py` `load_dem()` the current logic is to convert a `gu.Vector` into a `gu.Raster` with `create_mask()`.

I think we should also add support for `gu.Raster` mask as input.

marinebcht:
Resolves #954 

Workflow accuracy and topo now accept raster path_to_mask, no matter its projection/footprint : 

```
inputs:
  reference_elev:
    path_to_elev: "longyearbyen_ref_dem"
    path_to_mask: "path/mask.tif"
  to_be_aligned_elev:
    path_to_elev: "longyearbyen_tba_dem"
  sampling_grid: reference_elev
```

```
inputs:
  - path_to_elev: "longyearbyen_ref_dem"
    path_to_mask: "path/mask.tif"
```

#### Nan values in the mask 

Nan value are translated by False : see [here](https://github.com/GlacioHack/xdem/pull/970#issuecomment-4790729141). For example, in the glacier mask case: NaN (no data before or after reprojection) values are not conidarated as Glacier. 

#### Errors Messages

`masknotexist.tif` : `ValueError: User configuration invalid for 'inputs': [{'reference_elev': [{'path_to_mask': ['Path does not exist: tmp/masknotexist.tif']}]}]`

`mask.txt` : `ValueError: You provided a 'path_to_mask' value that is not recognised as a mask. Vector error: 'tmp/mask.txt' not recognized as being in a supported file format.; It might help to specify the correct driver explicitly by prefixing the file path with '<DRIVER>:', e.g. 'CSV:path'.. Raster error: 'tmp/ddd.txt' not recognized as being in a supported file format..`

#### Operation 

This mask is : 
1. read as Raster
2. translated as boolean (0 is False, !=0 is True)
3. filled with False
4. reprojected in the corresponding dem projection


marinebcht:
@belletva raster path_to_mask must it necessarily be a boolean or are int+ rasters accepted?

marinebcht:
Some thoughts : 

- A) shapefile -> create_mask(dem) -> raster with true : pink, false: black

image

- B) from this raster, inlier mask used mask in terrain/coreg become true : black, pink: white -> black is the values needed to be masked

image


To use raster as mask we need to provide a tiff as the input of (A) :
- True = glacier = to be masked
- False = to keep, whitch is prettry logical 

Is the case where we have a mask smaller than the dem : the projection need to fit it gives nan values. 
Nan values need to be considerate as values to keep (false) = there is no glaciers in it. 

------------------------

If mask not boolean => mask.astype(bool) (0 values = True, != 0 set False
If mask with nan => mask.data.filled(False)

marinebcht:
Hi @rhugonnet, can you confirm the process presented above to translate a raster in a actual mask ? :) Thanks

rhugonnet:
@marinebcht Seems good to me.
@adehecq Do you see other cases?

The only thing that comes to mind right now is that the reprojection step should probably be done BEFORE masking if the mask is derived from a continuous-value raster, such as elevation (e.g., elevation > 500 m and < 1000 m), but I don't think this applies to a user input for Workflows (where the mask is a file already on disk). 
If the user wants a different resampling of their raster being used as mask, they need to do this themselves beforehand anyway.

Another question for the API and to facilitate workflow input is: Do we only accept a single type of mask file from users (0 + NaN  = False, other = True), or do we also allow a tag "invert_mask" to use the binary opposite of the mask (whether it is from a raster or vector)?

marinebcht:
@belletva @rhugonnet ready to review so :)

tests/test_workflows/test_workflows.py
```diff
@@ -340,3 +335,57 @@ def test_load_dem_alias():
...
+    config_dem = dict()
+    config_dem["path_to_elev"] = "longyearbyen_ref_dem"
+    output_dem, inlier_mask, mask_path = Workflows.load_dem(config_dem)
+    assert mask_path is None
+    assert inlier_mask is None
+
+    # path_to_mask = shapefile path
+    config_dem["path_to_mask"] = xdem.examples.get_path("longyearbyen_glacier_outlines")
+    dem, inlier_mask, mask_path = Workflows.load_dem(config_dem)
+    assert inlier_mask == ~gu.Vector(mask_path).create_mask(inlier_mask)
```
rhugonnet: This equality works? I would have expected it to need `raster_equal` or `all()`

rhugonnet: Same as done below actually

xdem/workflows/workflows.py
```diff
@@ -388,13 +390,26 @@ def load_dem(config_dem: Dict[str, Any] | None) -> tuple[DEM, Raster, str | None
...
+                    inlier_mask = ~mask.create_mask(dem)
+
+                except pyogrio.errors.DataSourceError:
+                    try:
+                        inlier_mask = gu.Raster(mask_path).astype(bool)
+                        inlier_mask.data = inlier_mask.data.filled(False)
+
+                        inlier_mask = inlier_mask.reproject(dem, silent=True)
+                    except rasterio.errors.RasterioIOError:
+                        raise ValueError("You provided a 'path_to_mask' value that is not recognised as a mask.")
```
rhugonnet: We should be more specific on the error message, and maybe collect original errors to return them through this one, something like:
```python

# Treat the mask according to its type (Vector or Raster)
try:
    mask = gu.Vector(mask_path)
    inlier_mask = ~mask.create_mask(dem)

except pyogrio.errors.DataSourceError as vector_error:
    try:
        inlier_mask = gu.Raster(mask_path).astype(bool)
        inlier_mask.data = inlier_mask.data.filled(False)
        inlier_mask = inlier_mask.reproject(dem, silent=True)

    except rasterio.errors.RasterioIOError as raster_error:
        errors = ExceptionGroup(
            f"Could not read {mask_path!r} as either a vector or raster mask.",
            [vector_error, raster_error],
        )
        raise ValueError(
            "You provided a 'path_to_mask' value that is not recognised as a mask."
        ) from errors
```

rhugonnet (dismissed):
Perfect, a few comments above.
I wondered: do we also need to update the documentation to specify how this behaves?

xdem/workflows/workflows.py
marinebcht: Thanks for introducing me to ExceptionGroup; I’ve just discovered it.
But what is the added value, as it is not the dependancies ?
PS1: Non-existent files also first raise an error during the schema validation step.
PS2: rasterio gives the reason above our "ValueError"

doc/source/cli_accuracy.md
```diff
@@ -134,8 +134,13 @@ Inputs information, split between reference and to-be-aligned elevation data.
 :::
 
 :::{note}
+
```
marinebcht: remove empty line

doc/source/cli_accuracy.md
```diff
@@ -134,8 +134,13 @@ Inputs information, split between reference and to-be-aligned elevation data.
...
+The `path_to_mask` parameter accepts both vector and raster masks.
+For the second case, pixels with a value of 0 are considered part of the mask, while any non-zero and nodata values are considered valid.
+
 To set the vertical CRS or to override one that might exist in the metadata with ``force_vcrs``,
 please refer to {ref}`vertical-ref`.
+
```
marinebcht, line 142: The renduring  

Image

xdem/workflows/workflows.py
rhugonnet: The added value is that the errors are traced all the way back, instead of being interrupted by our own `ValueError` in xDEM.

Concretely: In case of an error, instead of just printing "ValueError in xdem.xxx: You provided a .... that is not recognized as a mask" and stopping there, it also prints "from RasterioIOError in rasterio.xxx: message". Most often, the "message" from Rasterio will be "GDAL could not recognize filename path_to_mask as a valid input...".
So the user knows all the underlying causes, and it's easier to debug.

Not sure I understood your point about "not in the dependencies": if related to `ExceptionGroup`, it is available in base Python! :smile:

marinebcht: Thanks you 

For the "not in the dependencies" : I saw that `ExceptionGroup` is built-in for Python 3.11+ and do not require an import, but for Python 3.10 we need a `pip install` @rhugonnet

rhugonnet: Ah I didn't notice this. In this case we can pass down errors ourselves and link to only one:
```python
 raise ValueError(
            f"You provided a 'path_to_mask' value that is not recognised as a mask. "
            f"Vector error: {vector_error}. "
            f"Raster error: {raster_error}."
        ) from raster_error
 ```

belletva (approved):

PR #971, opened 2026-06-18
Validation of the coreg method inputs

### LINKED ISSUE #761: Typing mistake
# Context

We have highlighted that typing errors in APIs are not very clear. 

`aligned_dem = tba_dem.coregister_3d(ref_dem, xdem.coreg.NuthKaab(max_iterations="6"))`

Image

@rhugonnet indicates that this relies on the typing work of our classes.

marinebcht:
Resolves https://github.com/GlacioHack/xdem/issues/761

The input parameters for the coregistration functions were not tested, and the potential errors were not very clear (see https://github.com/GlacioHack/xdem/issues/761).

To avoid this, I introduced a new check after the construction of the CoregDict dictionary to verify that it contains the correct types : `validate_typed_dict([the_dict_to_check, the_corresponding_type)`. This function raises an error when one of the key differs to the expected key type and can be use for any dictionary/type. 

Examples of errors : 

```
ValueError: Argument `icp_method`must be a typing.Literal['point-to-point', 'point-to-plane'], got `point-to-dplane` of type <class 'str'>)`

ValueError: Argument `tolerance`must be a <class 'float'>, got `aaa` of type <class 'str'>)
```

/!\ Some of the parameters are checked before the construction on the dict (`CoregMethod __init__()`) : 

`TypeError: Argument `fit_optimizer` must be a function (callable), got <class 'str'>.`

#### Types Changes

- initial_shift: tuple[float, float, float] | None to tuple[float **| int**, float **| int**, float **| int**] | None 
- cpd_weight: float to cpd_weight: float **| int**
- bias_var_names: list[str] -> bias_var_names: list[str] | None
- bin_sizes: int | dict[str, int | Iterable[float]] = > int | dict[str, int | Iterable[float] | list | NDArrayf | tuple] to fit examples from test_bias (ex: "elevation": np.arange(200, 500, 10))
- 
fit_loss_func can be : 
- callable for DHMinize method (_dh_minimize_fit function)
- string fo ICP method (_icp_fit function)
=>     fit_loss_func: Callable[[NDArrayf], np.floating[Any]] **| str**
=> Raises error if string in DHMinize init

#### Delete old checks

In NuthKaab:

- fit_optimizer 
```
# Before : TypeError: Argument `fit_optimizer` must be a function (callable), got <class 'str'>.
# After : TypeError: Argument 'fit_optimizer' invalid, must be a function (callable), got 'a' (str)
```

- bin_sizes
```
# Before : TypeError: Argument `bin_sizes` must be an integer, or a dictionary of integers or iterables, got <class 'str'>.
# After : TypeError: Argument 'bin_sizes' invalid, must be a int or dict[str, int or Iterable or list or ndarray or tuple], got 'a' (str)
```

- bin_statistic
```
# Before : TypeError: Argument `bin_statistic` must be a function (callable), got <class 'str'>.
# After : TypeError: Argument 'bin_statistic' invalid, must be a function (callable), got 'a' (str)
```

- initial_shift 
```
# Before : ValueError: Argument `initial_shift` must be a tuple of exactly two or three numerical values.
# After : TypeError: Argument 'initial_shift' invalid, must be a (float or int, float or int) or (float or int, float or int, float or int) or NoneType, got (0, 1, 5, 6) (tuple)
```

#### Notes 

Error when type is Literal : 
```
TypeError: Argument 'icp_method' invalid, must be a one of these values ('point-to-point', 'point-to-plane'), got 'aaa' (str)
```


marinebcht:
@rhugonnet just to be sure : 
As few parameters are checked before the call to `super()` in each subclass, can you validate the idea of ​​checking each parameter against its type using the CoregDict dictionary ? Little bit lost about the duplicate tests/readability :)

rhugonnet:
@marinebcht Great to finally solve it. I like your solution. And you're right, I think we can then remove the checks in the subclasses to avoid duplication. The `validate_typed_dict` should be able to take care of everything.

Small remark: Maybe we can add an override for the "Literal" case to not print directly its type (I'm sure many users will not know what "Literal" is), to instead print something like `Argument "icp_method" must be a one of ['point-to-point', 'point-to-plane']` directly?

marinebcht:
@rhugonnet  Speaking of your example, the param in the ICP method (method) and its name in the CoregType (icp_method) is not the same ... So the error does not correspond 😔

rhugonnet:
> @rhugonnet Speaking of your example, the param in the ICP method (method) and its name in the CoregType (icp_method) is not the same ... So the error does not correspond 😔

We can have an exception for "Specific" parameters, so that the name prefix (icp_, lzd_, etc) is removed before raising the error; Or we can make them the same as the CoregDict in the `__init__` of methods.

belletva (dismissed):
Good for me

PR #983, opened 2026-08-04
Don't force reprojection when it is possible in the CLI accuracy workflow

### LINKED ISSUE #909: Don't force reprojection for all coreg method in the CLI accuracy workflow
Following https://github.com/GlacioHack/xdem/pull/759#issuecomment-3827256889 and https://github.com/GlacioHack/xdem/pull/873 `sampling_grid` modif 
ICP and CPD work even if datasets are not reprojected 

- [ ] ICP method is not allow yet, check why
- [ ] allow `sampling_grid: None` when it is possible and test it
- [ ] verify for all method the behaviour of the coreg when dem1 and dem2 don't intersect each other

marinebcht:
[Work In Progress]

Resolves #909 

`sampling_grid = None` can be possible:
- coreg with ICP and CPD (work even if datasets are not reprojected)
     - ICP/CPD methods only: OK 
     - Other method(s) only: KO
     - ICP/CPD + other method(s): KO
- no coreg : KO

Tests ne no resampling with : 
- overlapping dems: dem vs smaller dem
- no overlapping dems ?

/!\ Openned https://github.com/GlacioHack/geoutils/issues/900

### DIFF
--- a/tests/test_workflows/test_accuracy.py
+++ b/tests/test_workflows/test_accuracy.py
@@ -230,24 +230,25 @@ def test_run_without_coreg(get_accuracy_inputs_test, tmp_path, level, generated_
 @pytest.mark.parametrize(
     "methods, raises_error",
     [
+        (["CPD", None, None], False),
+        (["ICP", None, None], False),
+        (["ICP", "CPD", None], False),
         (["NuthKaab", None, None], True),
         (["DhMinimize", None, None], True),
         (["VerticalShift", None, None], True),
         (["DirectionalBias", None, None], True),
         (["TerrainBias", None, None], True),
         (["LZD", None, None], True),
-        (["CPD", None, None], False),
-        (["ICP", None, None], False),
-        # ([None, "LZD", "ICP"], False),
         (["CPD", "VerticalShift", None], True),
     ],
 )
 def test_sampling_grid_is_None(get_accuracy_inputs_test, tmp_path, methods, raises_error):
     """
-    Test the behavior when sampling_grid = None with the different regulation methods.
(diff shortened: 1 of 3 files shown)


marinebcht:
Hey ! I found out that in even in ICP/CPD, there is a always a reprojection between ref and tba here 
https://github.com/GlacioHack/xdem/blob/7d849049b1f1075f7bd024e1dd56c716922352d9/xdem/coreg/base.py#L156 (in the _preprocess_coreg_fit_raster_raster function). Am I correct ? @rhugonnet (Checked https://github.com/GlacioHack/xdem/pull/759 in case) Perhaps we should start there before working on the workflow extention @belletva

belletva:
> Hey ! I found out that in even in ICP/CPD, there is a always a reprojection between ref and tba here
> 
> https://github.com/GlacioHack/xdem/blob/7d849049b1f1075f7bd024e1dd56c716922352d9/xdem/coreg/base.py#L156
> 
> (in the _preprocess_coreg_fit_raster_raster function). Am I correct ? @rhugonnet (Checked #759 in case) Perhaps we should start there before working on the workflow extention @belletva

good catch!

rhugonnet:
It is not forced in #759, with edits adding a `reproj_same_grid` argument:

https://github.com/rhugonnet/xdem/blob/5c9d8c3211db3024425135904c57fe11317f4245/xdem/coreg/base.py#L189

marinebcht:
Oki top @rhugonnet read your PR https://github.com/GlacioHack/xdem/pull/759 a little too quickly ! :)
I think I need to wait that this PR is finished @belletva

Issue #986, opened 2026-08-06
Blockwise with a coregistration pipeline


marinebcht:
Blockwise can normaly be done with CoregPipeline but :

```
import xdem

ref_dem = xdem.DEM(xdem.examples.get_path_test("longyearbyen_ref_dem"))
tba_dem = xdem.DEM(xdem.examples.get_path_test("longyearbyen_tba_dem"))

blockwise = xdem.coreg.BlockwiseCoreg(xdem.coreg.NuthKaab() + xdem.coreg.NuthKaab()  ,
                                      block_size_fit=500,
                                      block_size_apply=100,
                                      parent_path="")

```

gives :

```
Traceback (most recent call last):
  File "/home/mbouchet/Documents/xDem_project/new_xdem/xdem/bug/859_bw_info.py", line 6, in <module>
    blockwise = xdem.coreg.BlockwiseCoreg(xdem.coreg.NuthKaab() + xdem.coreg.NuthKaab()  ,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/new_xdem/xdem/xdem/coreg/blockwise.py", line 84, in __init__
    if not step.is_affine:
           ^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/new_xdem/xdem/xdem/coreg/base.py", line 2028, in is_affine
    self.to_matrix()
  File "/home/mbouchet/Documents/xDem_project/new_xdem/xdem/xdem/coreg/base.py", line 3189, in to_matrix
    return self._to_matrix_func()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/new_xdem/xdem/xdem/coreg/base.py", line 3196, in _to_matrix_func
    new_matrix = coreg.to_matrix()
                 ^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/new_xdem/xdem/xdem/coreg/affine.py", line 1840, in to_matrix
    return self._to_matrix_func()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mbouchet/Documents/xDem_project/new_xdem/xdem/xdem/coreg/affine.py", line 2538, in _to_matrix_func
[...]

PR #988, opened 2026-08-06
Blockwise don't work with a coregistration pipeline

marinebcht:
The documentation suggested that Blockwise can be used with a coregistration pipeline, but it does not (see https://github.com/GlacioHack/xdem/issues/986). 

So I removed all mentions of `CoregPipeline` in `blockwise.py` (doc + type).


belletva (approved):
Good for me

PR #992, opened 2026-08-17
Initial shift in case of CoregPipeline

### LINKED ISSUE #801: Initial shift in case of CoregPipeline
After merging this https://github.com/GlacioHack/xdem/pull/781 (initial shift for N&K) and moving all operations needed in [base.py](https://github.com/GlacioHack/xdem/blob/1c4ee60d6088ce15dedcbc553b9d5d90aba18d94/xdem/coreg/base.py#L2048), a new problem emerges in case of CoregPipeline. 

Indeed, we can use [reference_elev = reference_elev.translate(-shift_x, -shift_y)](https://github.com/marinebcht/xdem/blob/c88b9617aad811f2ac07b2f41b9d159403edcb6f/xdem/coreg/base.py#L210) as reference_elev was cast from DEM to array in [_preprocess_coreg_fit_raster_raster](https://github.com/marinebcht/xdem/blob/c88b9617aad811f2ac07b2f41b9d159403edcb6f/xdem/coreg/base.py#L278).

@rhugonnet I am pointing you to the new ticket; if you could explain it to me in more detail the trick I need to use, that would be great :)
Thank you

marinebcht:
Resolves https://github.com/GlacioHack/xdem/issues/801

This PR proposes to put at the same level every `Coreg` in a `CoregPipeline` (removing nested `CoregPipeline`) and the use the `initial_shift` parameter in it.

### Behavior 

With the actual `CorePipeline` `__init__`, nested CoregPipeline(s) can be accepted.
ex : 
```
coreg.CoregPipeline(
                [
                    coreg.NuthKaab(initial_shift=is1),
                    coreg.CoregPipeline([coreg.NuthKaab(initial_shift=is2), coreg.NuthKaab(initial_shift=is3)])
                ]
            )
```
But they does not pass the `fit()` function. So, we now put all the coreg in it in series like this: 
```
coreg.CoregPipeline(
                [
                    coreg.NuthKaab(initial_shift=is1),
                    coreg.NuthKaab(initial_shift=is2),
                    coreg.NuthKaab(initial_shift=is3)])
                ]
            )
```
#### Dev info

This change was done in the `CorePipeline` `__init__` (recursive function).

### Behavior 

- Accepted : `initial_shift` in the first method of the Pipeline
- Not accepted : `initial_shift` in all other method after => delete it

Warning message : **No initial shift can be defined in a coregistration pipeline other than for the first step. Overidding to initial_shift=None for step number [i]. Remove initial shift parameters outside of the first step to silence this warning.**

#### Dev info

Currently, removing all the initial shift was done in: 

- `Coreg.__add__` : to manage `coreg1 + coreg2`
- `CoregPipeline.__add__` : to manage `coreg1 + coreg2 + coreg3` (= `coregPipeline1 + coreg3`) and `CoregPileline(list_of_coreg)`

The fact to keep only the first initial shift if exists in the FIRST coreg was done just after the first point.

#### Preprocessing 

Had to remove the preprocess in the `CoregPipeline.fit()` [here](https://github.com/GlacioHack/xdem/blob/58180c65925f0fff5d7456bce1a206a8b45a9438/xdem/coreg/base.py#L3006)  and only use the existing preprocess in the `Coreg.fit()` to have the same result between `coreg1 + x` and `coreg1` in case of an `initial_shift`
image

### DIFF
--- a/doc/source/coregistration.md
+++ b/doc/source/coregistration.md
@@ -551,7 +551,7 @@ For both **inputs** and **outputs**, four consistent categories of metadata are 
 
 - An input `only_translation` to define if a coregistration should solve only for translations instead of a full rigid transformation (translations and rotations),
 - An input `standardize` to define if the input data should be standardized to the unit sphere before coregistration (to improve numerical convergence),
-- An input `initial_shift` that defines the estimated initial x and y shifts in georeferenced units, applied before fit step.
+- An input `initial_shift` that defines the estimated initial x and y shifts in georeferenced units, applied before fit step (when used within a CoregPipeline, only the initial offset specified in the first coregistration is applied, others are ignored),
 - An output `matrix` that stores the estimated affine matrix,
 - An output `centroid` that stores the centroid coordinates with which to apply the affine transformation,
 - Outputs `shift_x`, `shift_y` and `shift_z` that store the easting, northing and vertical offsets, respectively.
--- a/tests/test_dem.py
+++ b/tests/test_dem.py
@@ -539,29 +539,6 @@ class TestDEM:
             # Output DEM
             assert dem_aligned.raster_equal(manually_aligned, warn_failure_reason=True)
 
-    @staticmethod
-    @pytest.mark.parametrize(
-        "pipeline",
-        [
-            xdem.coreg.NuthKaab() + xdem.coreg.VerticalShift(),
-            xdem.coreg.NuthKaab(initial_shift=(10, 5)) + xdem.coreg.VerticalShift(),
-            xdem.coreg.VerticalShift() + xdem.coreg.NuthKaab(initial_shift=(10, 5)),
-            xdem.coreg.NuthKaab(initial_shift=(10, 5))
-            + xdem.coreg.VerticalShift()
-            + xdem.coreg.NuthKaab(initial_shift=(10, 5)),
-            xdem.coreg.VerticalShift()
-            + xdem.coreg.NuthKaab(initial_shift=(10, 5))
-            + xdem.coreg.NuthKaab(initial_shift=(10, 5)),
-        ],
-    )
-    def test_nuthkaab_coregpipeline(pipeline: Any) -> None:
-        """
-        Test initial shift cancellation in coreg pipeline method
-        """
-
-        for method in pipeline.pipeline:
-            assert "initial_shift" not in method.meta["inputs"]["affine"]
-
     def test_estimate_uncertainty(self) -> None:
 
         # Import optional skgstat or skip test
--- a/xdem/coreg/base.py
+++ b/xdem/coreg/base.py
@@ -2010,11 +2010,6 @@ class Coreg:
         if not isinstance(other, Coreg):
             raise ValueError(f"Incompatible add type: {type(other)}. Expected 'Coreg' subclass")
 
-        # Cancel possible initial shift(s) in coreg(s) in "other"
-        if "affine" in other.meta["inputs"] and "initial_shift" in other.meta["inputs"]["affine"]:
-            warnings.warn(message="No initial shift can be xxx", category=UserWarning)
-            del other.meta["inputs"]["affine"]["initial_shift"]
-
         return CoregPipeline([self, other])
 
     @property
@@ -2892,7 +2887,26 @@ class CoregPipeline(Coreg):
 
         :param: Processing steps to run in the sequence they are given.
         """
-        self.pipeline = pipeline
+
+        def delete_is(pipeline: CoregPipeline, init: bool = False):
+            """Delete initial shift according to its place in the pipeline"""
+            for i, step in enumerate(pipeline):
+                if i == 0 and not init:
+                    continue
+
+                if not isinstance(step, CoregPipeline):
+                    if "affine" in step.meta["inputs"] and "initial_shift" in step.meta["inputs"]["affine"]:
+                        warnings.warn(
+                            message="No initial shift can be initialized in a coregistration pipeline other "
+                            "than for the first element.",
+                            category=UserWarning,
+                        )
+                        del step.meta["inputs"]["affine"]["initial_shift"]
+                else:
+                    delete_is(step, init=True)
+            return pipeline
+
+        self.pipeline = delete_is(pipeline)
 
         super().__init__()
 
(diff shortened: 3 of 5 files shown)


xdem/coreg/base.py
```diff
@@ -2309,9 +2308,14 @@ def fit(
...
             shift_y = self._meta["inputs"]["affine"]["initial_shift"][1]  # type: ignore
+
             # shift_z is currently always equal to zero
-            reference_elev = reference_elev.translate(-shift_x, -shift_y)  # type: ignore
-            initial_shift_apply = True
+            if isinstance(reference_elev, gu.Raster):
+                reference_elev = reference_elev.translate(-shift_x, -shift_y)  # type: ignore
+                initial_shift_apply = True
+            else:
+                transform = _translate(transform, xoff=-shift_x, yoff=-shift_y)
```
marinebcht, line 2312: Case if ref_dem is an array

xdem/coreg/base.py
```diff
@@ -3045,10 +3038,10 @@ def fit(
             # Step apply: one output for a geodataframe, two outputs for array/transform
             # We only run this step if it's not the last, otherwise it is unused!
             if i != (len(self.pipeline) - 1):
-                if isinstance(tba_dem_mod, gpd.GeoDataFrame):
+                if isinstance(tba_dem_mod, (Raster, gpd.GeoDataFrame, PointCloud)):
```
marinebcht, line 3062: fit the apply returns management

xdem/coreg/blockwise.py
```diff
@@ -199,10 +199,10 @@ def fit(
             shift_y = coreg.meta["outputs"]["affine"].get("shift_y", np.nan)
             shift_z = coreg.meta["outputs"]["affine"].get("shift_z", np.nan)
 
-            x, y = (
+            x, y = to_be_aligned_elev.transform * (
```
marinebcht, line 202: CI problem: see https://github.com/GlacioHack/xdem/pull/994

tests/test_dem.py
```diff
@@ -539,29 +539,6 @@ def test_nuthkaab_initial_shift(shift: Any, expected_message: Any) -> None:
             # Output DEM
             assert dem_aligned.raster_equal(manually_aligned, warn_failure_reason=True)
 
-    @staticmethod
-    @pytest.mark.parametrize(
```
marinebcht, line 543: move the "tests" in test_affine with the others

rhugonnet: I created a new `pipeline.py` module in #759. I'm going to have fun merging 😅

xdem/coreg/base.py
```diff
@@ -3177,9 +3170,14 @@ def __add__(self, other: list[Coreg] | Coreg | CoregPipeline) -> CoregPipeline:
 
         pipelines = self.pipeline + other
 
-        # Cancel possible initial shift(s) in CoregPipeline case
-        for method in pipelines:
+        # Cancel possible initial shift(s) in coreg(s) in "other"
+        for method in pipelines[1:]:
             if "affine" in method.meta["inputs"] and "initial_shift" in method.meta["inputs"]["affine"]:
+                warnings.warn(
+                    message="No initial shift can be initialized in a coregistration pipeline other "
```
rhugonnet: I think we need to more precise on what is done: We override. And repeat for every step:
"No initial shift can be defined in a coregistration pipeline other than for the first step. Overridding to `initial_shift=None` for step number {X}. Remove initial shift parameters outside of the first step to silence this warning."

rhugonnet (commented):
Great, thanks! As I commented separately before, I still think we need to move the logic to `CoregPipeline.__init__`, because this is the function that actually chains the step, and its class description is public.

So someone could do `CoregPipeline([NuthKaab(), NuthKaab(initial_shift=(2, 2, 2))])` and it would not raise an error/warning, yet be correctly documented code. This would be a problem.

marinebcht:
@rhugonnet Was little confused about the behaviour of nested CoregPipeline like NuthKaab(initial_shift=(2, 2, 0)) + CoregPipeline([NuthKaab(initial_shift=(1, 1, 0)), NuthKaab()]) but I worked on it ! ... But I don't know if it's supposed to work ^^ Is that a case we have/we need ? The initialization works (even if it is not in the signature function).

rhugonnet:
It's not a new case, it's the same as before but passed directly through `__init__` instead of `__add__`.

This is how `__add__` works, see here:
https://github.com/GlacioHack/xdem/blob/7bf95b197f6f45dcdfa9876220b5ba97b7f451d3/xdem/coreg/base.py#L2019

Which is why we need to cover the behaviour directly in `__init__` instead of `__add__`

tests/test_dem.py
marinebcht: oh no ^^''

marinebcht:
Yes Yes no problem about this :)  @rhugonnet 
I was wondering how "nested" `CorePipeline` are handled. With the actual `CorePipeline` `__init__`, they can be accepted.
ex : 
```
coreg.CoregPipeline(
                [
                    coreg.NuthKaab(initial_shift=is1),
                    coreg.CoregPipeline([coreg.NuthKaab(initial_shift=is2), coreg.NuthKaab(initial_shift=is3)])
                ]
            )
```
But I checked, it does not pass the `fit()` function. So, we can simply: 
- refuse `coregPipeline` object in the init inuput `list(Coreg)`
- accept them and put all the coregs in series like 
```
coreg.CoregPipeline(
                [
                    coreg.NuthKaab(initial_shift=is1),
                    coreg.NuthKaab(initial_shift=is2),
                    coreg.NuthKaab(initial_shift=is3)])
                ]
            )
```
it is the same, no ?

rhugonnet:
Yes it is the same, good idea to account for this clearly!

marinebcht:
@rhugonnet @belletva done with the new subject :)

rhugonnet (dismissed):
Looks good! 😄

doc/source/coregistration.md
```diff
@@ -551,7 +551,7 @@ For both **inputs** and **outputs**, four consistent categories of metadata are
 
 - An input `only_translation` to define if a coregistration should solve only for translations instead of a full rigid transformation (translations and rotations),
 - An input `standardize` to define if the input data should be standardized to the unit sphere before coregistration (to improve numerical convergence),
-- An input `initial_shift` that defines the estimated initial x and y shifts in georeferenced units, applied before fit step.
+- An input `initial_shift` that defines the estimated initial x and y shifts in georeferenced units, applied before fit step (when used within a CoregPipeline, only the initial offset specified in the first coregistration is applied, others are ignored),
```
marinebcht, line 554: @belletva @rhugonnet here a prop to have an explanation of the is behavior in the CoregPipeline (if it is not overfit)

belletva (approved):
Good for me

PR #993, opened 2026-08-17
Avoid forking a multi-threaded parent in variogram sampling

### LINKED ISSUE #507: Multiprocessing new multi-threading `DeprecationWarning` for multiple forks in `sample_empirical_variogram`
Calling the function with `n_jobs>1` raises:
```DeprecationWarning: This process (pid=10362) is multi-threaded, use of fork() may lead to deadlocks in the child.```

Maybe using `multiprocessing.get_context` as in the KH9 repo would solve the issue.

manduinca:
Fixes #507

`sample_empirical_variogram` built its pool with `mp.Pool(...)`, which uses the default start method. On Linux that is `fork`, and forking a parent with live threads is what raises the `DeprecationWarning` in the issue. This uses `forkserver` where it exists and `spawn` otherwise, as suggested in the issue, so workers start from a process without threads.

Two things you should weigh, since they are not free:

- **Start-up cost.** `maxtasksperchild=1` recycles a worker after every variogram, so the method matters: `forkserver` forks from a pre-imported process and stays cheap, while `spawn` re-imports xdem each time. That is why the choice prefers `forkserver` rather than going straight to `spawn`.
- **Behaviour change for callers.** With `fork`, calling `sample_empirical_variogram(n_jobs>1)` from a plain script works; with `forkserver`/`spawn` the caller needs the `if __name__ == "__main__":` guard. Python 3.14 already changed the default on Linux, so this is where things are heading anyway, but it will affect scripts written against older behaviour.

On the test: it holds a thread open while sampling and turns the fork `DeprecationWarning` into an error, which reproduces the issue's conditions. I could not watch it fail locally — macOS already defaults to `spawn`, so the warning never appears here — but it should cover the regression on the Linux CI.

### DIFF
--- a/tests/test_spatialstats.py
+++ b/tests/test_spatialstats.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+import threading
 import warnings
 from importlib.util import find_spec
 from typing import Any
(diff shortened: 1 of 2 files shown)


rhugonnet:
Thanks @manduinca, this warning will stop polluting the outputs.

A note: The `sample_empirical_variogram` function will remain active for a while for backwards-comp, but will be replaced during the revamp of the uncertainty module, see https://github.com/GlacioHack/xdem/issues/947

manduinca:
Thanks @rhugonnet, good to know about the revamp.

One thing worth flagging: the 15 failures the CI reports here are not from this branch. They are all in `tests/test_coreg/test_blockwise.py` with `TypeError: can't multiply sequence by non-int of type 'Affine'`, and I get the same 14 failures running that file on `main` at 58180c6 with nothing of mine applied. `blockwise.py` does not touch the variogram sampling this PR changes.

The `doc` job is red too, which I have not looked into.

rhugonnet:
@manduinca Thanks, indeed it is a separate issue that @marinebcht is fixing in #994. Then we should be able to merge here as well :wink:

rhugonnet (approved):

PR #994, opened 2026-08-18
Fix failure to show help with `xdem` CLI command and `Affine` class multiplication in blockwise

marinebcht:
I noticed that simply calling `xdem` causes an error instead of launching the help utility:

```
Traceback (most recent call last):
  File "/home/mbouchet/Documents/xDem_project/new_xdem/xdem/venv/bin/xdem", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/mbouchet/Documents/xDem_project/new_xdem/xdem/xdem/cli.py", line 129, in main
    if args.output and not args.config:
       ^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'output'
```
#### Correction

`output` belongs to the workflows and can't be use without so args.output does not exists.
I removed this line => if `--output` is called without topo/accuracy, it behaves like any  other `--xxx` provided option. 

#### Test

Indeed, the test verifying its proper functioning was insufficient, as the testing method did not fully simulate the call via the terminal in case of the parameter : https://github.com/marinebcht/xdem/blob/12ad5fbe16696c960a8b19d4966c5a6551cf4f2d/tests/test_workflows/test_cli.py#L45 

So I used subprocess module : https://github.com/marinebcht/xdem/blob/12ad5fbe16696c960a8b19d4966c5a6551cf4f2d/tests/test_workflows/test_cli.py#L41

## NOTES DEV

I took the opportunity to correct another error in the CI : https://github.com/marinebcht/xdem/blob/936c4341f6968ce3dc39812f673521535ccd662a/xdem/coreg/blockwise.py#L202

- `something` * `Affine` => KO see https://github.com/GlacioHack/xdem/actions/runs/32118791212/job/95654249902
- `Affine` * `something` => OK

### DIFF
--- a/tests/test_workflows/test_cli.py
+++ b/tests/test_workflows/test_cli.py
@@ -39,7 +39,11 @@ pytest.importorskip("cerberus")
 
 
 def test_xdem():
-    subprocess.run(["xdem"])
+    """Check if running simply 'xdem' in CLI does indeed show the help, without failure."""
+    result = subprocess.run(["xdem"], capture_output=True, text=True)
+
+    assert result.returncode == 0
+    assert "usage:" in result.stdout.lower()
 
 
 @pytest.mark.parametrize("help_arg", [[], ["-h"], ["--help"]])
--- a/tests/test_workflows/test_cli.py
+++ b/tests/test_workflows/test_cli.py
@@ -23,6 +23,7 @@ test for CLI class
 
 import logging
 import os
+import subprocess
 
 # mypy: disable-error-code=no-untyped-def
 from pathlib import Path
@@ -35,6 +36,10 @@ from xdem.workflows.schemas import COMPLETE_CONFIG_ACCURACY, COMPLETE_CONFIG_TOP
 
 pytestmark = pytest.mark.filterwarnings("ignore::UserWarning")
 pytest.importorskip("cerberus")
+
+
+def test_xdem():
+    subprocess.run(["xdem"])
 
 
 @pytest.mark.parametrize("help_arg", [[], ["-h"], ["--help"]])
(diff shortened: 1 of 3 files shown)


tests/test_workflows/test_cli.py
```diff
@@ -37,6 +38,10 @@
 pytest.importorskip("cerberus")
 
 
+def test_xdem():
+    subprocess.run(["xdem"])
```
rhugonnet: Just calling `subprocess.run()` captures/raises an error in case of issue? I've never used it in tests before, and the rest uses `cli.main()` I think.

rhugonnet: Anyway to `grep` the `stdout` to check it is the help that comes up, and that `stderr` is empty?

rhugonnet: Here's a more rigorous version: captures output as text, checks the process did not fail (returncode == 0) and check the content of the output.

```python

def test_xdem():
    result = subprocess.run(["xdem"], capture_output=True, text=True)

    assert result.returncode == 0
    assert "usage:" in result.stdout.lower()
```

rhugonnet: Done

rhugonnet:
Good thing I improved the test, the behaviour was still broken without args to `xdem` in CLI :wink:.

OK, with these fixes merged, we should have all CI tests running. I'll be merging other PRs, so I wanted to get this out the way!

xdem/cli.py
```diff
@@ -126,9 +127,6 @@ def main(arg_list: list[str] | None = None) -> None:
     logging.getLogger("fontTools").setLevel(logging.WARNING)
     logging.getLogger("fontTools").propagate = False
 
-    if args.output and not args.config:
```
rhugonnet, line 129: We need to keep this error, as the two arguments need each other.

rhugonnet: Solving this

tests/test_workflows/test_cli.py
marinebcht: Oh nice, thanks for the improvement !

PR #1005, opened 2026-09-13
Fix small CI/doc issues

rhugonnet:
Fixing some small CI issues that are breaking because of GeoUtils updates (and CI runs on its main branch right now).

Most are warning/tiny fixes.

The most important are:
- Since interpolation at points was corrected near outer edge, the `test_example` that runs Nuth and Kääb changes slightly. @marinebcht Don't forget to delete your test data folder to reset this for local Pytest runs.
- Since we don't separate 2D CRS and vertical CRS anymore, we have some issues during `reproject()`, see #1006. I temporarily update the doc to pass by circumventing the issue. Will fix this in a separate PR.


rhugonnet:
FYI: Will merge this once https://github.com/GlacioHack/geoutils/pull/938 is done

rhugonnet:
GeoUtils main is changing again anyway, so other things would break, easier to include these changes into #759. Closing this.

